PR #474, opened 2024-02-01
feat(cli): CATALYST-246 create-catalyst login, create env, scaffold project

matthewvolk:
## What/Why?

- Adds initial commands for the `create-catalyst` CLI: `create` and `init`. 
- `create` is the default command, and invoked by running `[npm|pnpm|yarn] create @bigcommerce/catalyst@latest`. The command accepts a number of configurable options that can be passed as flags the CLI command (this is mostly useful for running the CLI in a CI environment). A number of the configurable options have some default values if they are left empty. If no options are passed as flags, the user is prompted to enter values for each option via terminal input. The end result is a Catalyst project, created in a local directory, with an (optionally) configured environment variable file, with dependencies installed, and if the environment variable file is configured, the `create` command also runs GraphQL codegen for GraphQL types + lints the project to ensure the project can build without errors.
- `init` is a utility command that can be used to swap the store that the Catalyst project is connected to. Running `[npx|pnpx|yarn dlx] @bigcommerce/create-catalyst init` prompts the user for authentication credentials to a BigCommerce store, and when successfully authenticated, overwrites the current local `.env.local` file with new environment variables that belong to the newly authenticated store.

## Testing
Includes Unit and Integration tests

### DIFF
--- a/packages/create-catalyst/.swcrc
+++ b/packages/create-catalyst/.swcrc
@@ -1,0 +1,17 @@
+{
+  "$schema": "https://json.schemastore.org/swcrc",
+  "sourceMaps": true,
+  "jsc": {
+    "parser": {
+      "syntax": "typescript",
+      "decorators": true,
+      "dynamicImport": true
+    },
+    "transform": {
+      "legacyDecorator": true,
+      "decoratorMetadata": true
+    },
+    "baseUrl": "./"
+  },
+  "minify": false
+}
--- a/packages/create-catalyst/jest.config.cjs
+++ b/packages/create-catalyst/jest.config.cjs
@@ -1,0 +1,12 @@
+module.exports = {
+  extensionsToTreatAsEsm: ['.ts'],
+  transform: {
+    '^.+\\.(t|j)s?$': '@swc/jest',
+  },
+  transformIgnorePatterns: [],
+  // moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
+  moduleNameMapper: {
+    '^(\\.{1,2}/.*)\\.js$': '$1',
+  },
+  testEnvironment: 'node',
+};
--- a/packages/create-catalyst/package.json
+++ b/packages/create-catalyst/package.json
@@ -7,17 +7,44 @@
     "dist"
   ],
   "scripts": {
+    "dev": "tsup --watch",
     "typecheck": "tsc --noEmit",
     "lint": "eslint . --max-warnings 0",
-    "test": "jest"
+    "test": "jest",
+    "build": "tsup"
   },
   "engines": {
     "node": ">=18.16"
   },
+  "dependencies": {
+    "@inquirer/prompts": "^3.3.0",
+    "chalk": "^5.3.0",
+    "commander": "^11.1.0",
+    "fs-extra": "^11.2.0",
+    "giget": "^1.2.1",
+    "lodash.kebabcase": "^4.1.1",
+    "lodash.merge": "^4.6.2",
+    "lodash.set": "^4.3.2",
+    "lodash.unset": "^4.5.2",
+    "msw": "^2.1.7",
+    "nypm": "^0.3.6",
+    "ora": "^8.0.1",
+    "zod": "^3.22.4",
+    "zod-validation-error": "^3.0.2"
+  },
   "devDependencies": {
     "@bigcommerce/eslint-config": "^2.7.0",
     "@bigcommerce/eslint-config-catalyst": "workspace:^",
+    "@swc/core": "^1.3.107",
+    "@swc/jest": "^0.2.34",
+    "@types/fs-extra": "^11.0.4",
+    "@types/jest": "^29.5.11",
+    "@types/lodash.kebabcase": "^4.1.9",
+    "@types/lodash.merge": "^4.6.9",
+    "@types/lodash.set": "^4.3.9",
+    "@types/lodash.unset": "^4.5.9",
     "@types/node": "^18.17.12",
+    "@types/prompts": "^2.4.9",
     "eslint": "^8.55.0",
     "jest": "^29.7.0",
     "prettier": "^3.1.1",
--- a/packages/create-catalyst/package.json
+++ b/packages/create-catalyst/package.json
@@ -30,7 +30,7 @@
     "nypm": "^0.3.6",
     "ora": "^8.0.1",
     "zod": "^3.22.4",
-    "zod-validation-error": "^3.0.2"
+    "zod-validation-error": "^3.0.0"
   },
   "devDependencies": {
     "@bigcommerce/eslint-config": "^2.7.0",
--- a/packages/create-catalyst/src/commands/init.ts
+++ b/packages/create-catalyst/src/commands/init.ts
@@ -1,0 +1,115 @@
+import { input, select } from '@inquirer/prompts';
+import chalk from 'chalk';
+import * as z from 'zod';
+
+import { Https } from '../utils/https.js';
+import { login } from '../utils/login.js';
+import { parse } from '../utils/parse.js';
+import { writeEnv } from '../utils/write-env.js';
+
+export const init = async (opts: unknown) => {
+  const BigCommerceHostnameSchema = z.string().min(1);
+  const SampleDataApiUrlSchema = z.string().url();
+  const StoreHashSchema = z.string().optional();
+  const AccessTokenSchema = z.string().optional();
+
+  const OptionsSchema = z.object({
+    bigCommerceHostname: BigCommerceHostnameSchema,
+    sampleDataApiUrl: SampleDataApiUrlSchema,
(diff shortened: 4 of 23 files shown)


.github/workflows/cli.yml
```diff
@@ -0,0 +1,42 @@
+name: create-catalyst
```
chanceaclark, line 1:
```suggestion
name: Create Catalyst CLI
```

.github/workflows/cli.yml
```diff
@@ -0,0 +1,42 @@
...
+jobs:
+  lint-typecheck:
+    name: Run Tests
+
+    runs-on: ubuntu-latest
+
+    env:
+      TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
+      TURBO_TEAM: ${{ vars.TURBO_TEAM }}
+      BIGCOMMERCE_STORE_HASH: ${{ secrets.BIGCOMMERCE_STORE_HASH }}
```
chanceaclark, line 20: Just added it, but make sure to add this https://github.com/bigcommerce/catalyst/blob/main/.github/workflows/basic.yml#L20

packages/create-catalyst/src/index.ts
```diff
@@ -0,0 +1,44 @@
...
+
+const program = new Command();
+
+// CATALYST_DIR
+// BIGCOMMERCE_API_URL
+// STORE_HASH
+// ACCESS_TOKEN
+// BIGCOMMERCE_AUTH_URL
+// BIGCOMMERCE_CLI_CLIENT_ID
+// PACKAGE_MANAGER
```
chanceaclark, line 22: Can these be removed?

.github/workflows/cli.yml
matthewvolk: Thank you!

packages/create-catalyst/package.json
```diff
@@ -1,23 +1,50 @@
 {
-  "name": "@bigcommerce/create-catalyst",
-  "version": "0.1.0",
+  "name": "create-catalyst-storefront",
+  "version": "0.3.0",
   "type": "module",
```
matthewvolk, line 4: Will revert after testing

matthewvolk:
@bigcommerce/team-catalyst PR description updated + branch rebased. This one is ready for a final review 👍

jorgemoya (approved):
A lot to unpack but looks good to me!

packages/create-catalyst/jest.config.cjs
```diff
@@ -0,0 +1,12 @@
...
+  transform: {
+    '^.+\\.(t|j)s?$': '@swc/jest',
+  },
+  transformIgnorePatterns: [],
+  // moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
+  moduleNameMapper: {
+    '^(\\.{1,2}/.*)\\.js$': '$1',
+  },
+  testEnvironment: 'node',
+};
```
chanceaclark, line 12: All you need is this (see `tsconfig.json` comments):
```suggestion
module.exports = {
  transform: {
    '^.+\\.(t|j)s?$': '@swc/jest',
  },
};
```

packages/create-catalyst/tsconfig.json
chanceaclark, line 1: Try changing the tsconfig to this along with the jest config changes.

```json
{
  "$schema": "https://json.schemastore.org/tsconfig",
  "display": "Node 18",
  "compilerOptions": {
    "strict": true,
    "target": "es6",
    "module": "ESNext",
    "esModuleInterop": true,
    "moduleResolution": "Bundler",
    "resolveJsonModule": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": true,
  }
}
```

Additionally, let's remove disable the `dts` option and enable `sourcemap` in `tsup.config.ts`. We don't need types exported with the CLI and sourcemaps will be helpful for debugging.

packages/create-catalyst/.swcrc
chanceaclark, line 1: This should be okay to remove with the jest + tsconfig changes.

packages/create-catalyst/src/utils/write-env.ts
```diff
@@ -0,0 +1,33 @@
...
+   */
+  outputFileSync(
+    join(projectDir, '.env.local'),
+    [
+      `AUTH_SECRET=${randomBytes(32).toString('hex')}`,
+      `ENABLE_ADMIN_ROUTE=true`,
+      `BIGCOMMERCE_STORE_HASH=${storeHash}`,
+      `BIGCOMMERCE_CHANNEL_ID=${channelId}`,
+      `BIGCOMMERCE_ACCESS_TOKEN=${accessToken}`,
+      `BIGCOMMERCE_CUSTOMER_IMPERSONATION_TOKEN=${customerImpersonationToken}`,
```
chanceaclark, line 30: Should we also add this?
```
CLIENT_LOGGER="false"
```

packages/create-catalyst/src/utils/spinner.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { oraPromise } from 'ora';
+
+import { spinner } from './spinner.js';
```
chanceaclark, line 3: With the tsconfig changes, you can remove the `.js` extension.

packages/create-catalyst/src/utils/project-config.ts
```diff
@@ -0,0 +1,33 @@
...
+import kebabCase from 'lodash.kebabcase';
+import { join } from 'path';
+
+export const projectConfig = async (opts: { projectDir?: string; projectName?: string }) => {
+  let projectName = kebabCase(opts.projectName);
+  let projectDir = opts.projectDir && projectName ? join(opts.projectDir, projectName) : undefined;
+
+  if (projectDir && projectName) {
+    return { projectDir, projectName };
+  }
```
chanceaclark, line 12: 🍹 Instead of using `let`'s, we can just do something like this just to clean these files up.
```suggestion
if (opts.projectName && opts.projectDir) {
  return {
    projectDir: join(opts.projectDir, projectName),
    projectName: kebabCase(opts.projectName),
  };
}
```

packages/create-catalyst/src/utils/https.ts
```diff
@@ -0,0 +1,274 @@
...
+    sampleDataApiUrl,
+    storeHash,
+    accessToken,
+  }: {
+    bigCommerceApiUrl?: string;
+    bigCommerceAuthUrl?: string;
+    sampleDataApiUrl?: string;
+    storeHash?: string;
+    accessToken?: string;
+  }) {
```
chanceaclark, line 48: 🍹 Would love some descriptive interfaces for these overloaded args.

packages/create-catalyst/src/commands/init.ts
```diff
@@ -0,0 +1,115 @@
...
+  let customerImpersonationToken = '';
+
+  if (!canCreateChannel) {
+    const channelSortOrder = ['catalyst', 'next', 'bigcommerce'];
+
+    const existingChannel = await select({
+      message: 'Which channel would you like to use?',
+      choices: activeChannels
+        .sort((a, b) => channelSortOrder.indexOf(a.platform) - channelSortOrder.indexOf(b.platform))
+        .map((ch) => ({
```
chanceaclark, line 72: 🍹 :nit:
```suggestion
        .map((channel) => ({
```

chanceaclark (commented):
I know you already merged this but left a few comments for cleanup whenever you get around to it!

packages/create-catalyst/jest.config.cjs
matthewvolk: All of this looks good! However, I do need to keep the `moduleNameMapper` because this is a `"type": "module"` package, and the `.js` extensions on imports are not found if the mapper is not present

matthewvolk: Oops 🤦‍♂️ Had to restart ESLint, and now the changes are working. Removing moduleNameMapper! Thank you!

packages/create-catalyst/src/utils/project-config.ts
matthewvolk: This one is a little tricky, and the more I think about it, I think there's a lot of refactoring that I'm gonna need to do around each option. In this case:
* `projectDir` and `projectName` can be passed as CLI flags
* `projectName` is optional, and if not passed as a flag it would be equal to `undefined` in this function.
* If `projectName` is undefined, we need to **not** return, because we need to prompt the user to enter a project name
* If `projectName` is defined but invalid (such as entering weird characters like `*`), I need to throw an error (just noticed this isn't currently accounted for)
* It's possible for `projectName` to be a space character, in which case I need to throw an error (this isn't accounted for either)
* `projectDir` always defaults to `process.cwd()`, but if the user enters a path that does not exist, I need to throw

I think I can reuse the `validate` method lower down in this file to help with a lot of this. I'll rework it in a separate PR 👍

PR #480, opened 2024-02-03
Support 301 Redirects and Raw Web Pages

bookernath:
## What/Why?
Adds support for 301 redirects created in the BigCommerce platform, so these will work automatically on Catalyst storefronts.

Also add an optimized route for supporting Raw Web Pages, which previously were broken as a complete HTML document attempted to render inside a template.

## Notes and future improvement opportunities
- Although Next.js defaults to HTTP 308, I have selected HTTP 301 to ensure we are correctly understood by as many crawlers as possible for SEO purposes, as this feature is intended for maintaining SEO value for URLs when a store has been replatformed from a different system.
- This adds more data to the `route()` lookup return type which in turn increases the storage used by KV
- A future refactor could simplify the KV storage to use the GQL global ids (`id`) instead of (`entityId`) which would give us a simpler way to store entities generically across all possible route types as well as redirects
- I intentionally updated names to standardize on `route()` syntax and remove references to `customUrls` so we can move closer to ubiquitous language between API and codebase


bookernath:
Easier to start from scratch with this one, I'll make new PRs.

PR #498, opened 2024-02-06
chore(core): rename file names to match kebab case

jorgemoya:
## What/Why?

- Remove override to match file name rule.
- Rename files to match kebab case.
- Update imports.

## Testing
Locally.


apps/core/app/(default)/product/[slug]/_components/Breadcrumbs.tsx
```diff
@@ -1,4 +1,4 @@
-import { getProduct } from '~/client/queries/getProduct';
```
Comment (bc-alexsaiannyi) [inline apps/core/app/(default)/product/[slug]/_components/Breadcrumbs.tsx:1]: @jorgemoya is it Ok that components like `Breadcrumbs`, `Gallery` & `Reviews` are still in PascalCase  comparing to  `related-products` as an example?

Comment (jorgemoya) [reply apps/core/app/(default)/product/[slug]/_components/Breadcrumbs.tsx:1]: Seemed to be a a bug with github, locally all files were kebab case. I forced a push to fix it.

deini (approved):

PR #506, opened 2024-02-07
Skip getCustomUrlNode() entirely for home page

bookernath:
## What/Why?
The `route()` lookup is unnecessary on the home page, so we can improve performance by skipping it entirely.

## Testing
Confirmed in logs that I only see a single KV request where there used to be two.


bookernath:
Closing this since it'll be easier to re-write it than do the gnarly rebase.

PR #515, opened 2024-02-07
fix(core): Add type guards to middleware, properly handle empty cache

christensenep:
## What/Why?
Adds type guards for the RouteCache and StorefrontStatusCache values returned from the KV, and properly handle the case when the cache is empty. (In main at the moment an error is thrown when the cache is empty, although otherwise everything does function properly and the cache is hydrated/page rendered appropriately).

## Testing
Tested locally with both the vercel KV store and dev adapter, with hydrated and empty caches.

### DIFF
--- a/apps/core/middlewares/with-custom-urls.ts
+++ b/apps/core/middlewares/with-custom-urls.ts
@@ -1,5 +1,6 @@
 import { cookies } from 'next/headers';
 import { NextRequest, NextResponse } from 'next/server';
+import { z } from 'zod';
 
 import { getSessionCustomerId } from '~/auth';
 import { StorefrontStatusType } from '~/client/generated/graphql';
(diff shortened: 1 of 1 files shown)


apps/core/middlewares/with-custom-urls.ts
```diff
@@ -32,16 +32,44 @@ const createRewriteUrl = (path: string, request: NextRequest) => {
   return url;
 };
 
+const isRouteCache = (
```
chanceaclark, line 35: Thoughts on using `zod` instead of these custom type-guards?

christensenep: my thoughts are that I will use `zod` instead of these custom type-guards

christensenep:
♻️

chanceaclark (approved):

PR #518, opened 2024-02-07
chore(core): Remove v2_ prefix from kv keys

christensenep:
## What/Why?
Now that https://github.com/bigcommerce/catalyst/pull/476 has been merged to main, we can remove the `v2_` prefix, as all  branches will now be using the new format of the cache values. (Can probably hold off a day or so to merge this just to play nice with existing branches)

## Testing
Verified the KV store is still getting/setting properly, both on initial hydration and when revalidating.

### DIFF
--- a/apps/core/app/api/revalidate/route/route.ts
+++ b/apps/core/app/api/revalidate/route/route.ts
@@ -19,7 +19,7 @@ const handler = async (request: NextRequest) => {
   const expiryTime = Date.now() + 1000 * 60 * 30; // 30 minutes;
 
   try {
-    await kv.set(`v2_${pathname}`, { node, expiryTime });
+    await kv.set(pathname, { node, expiryTime });
   } catch (error) {
     // eslint-disable-next-line no-console
     console.error(error);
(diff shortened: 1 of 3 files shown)


apps/core/app/api/revalidate/route/route.ts
```diff
@@ -19,7 +19,7 @@ const handler = async (request: NextRequest) => {
   const expiryTime = Date.now() + 1000 * 60 * 30; // 30 minutes;
 
   try {
-    await kv.set(`v2_${pathname}`, { node, expiryTime });
+    await kv.set(`${pathname}`, { node, expiryTime });
```
deini, line 22: We can remove the template literal here.

apps/core/middlewares/with-custom-urls.ts
```diff
@@ -92,7 +92,7 @@ const setKvRoute = async (request: NextRequest, node: Node) => {
   try {
     const expiryTime = Date.now() + 1000 * 60 * 30; // 30 minutes;
 
-    await kv.set(`v2_${request.nextUrl.pathname}`, { node, expiryTime });
+    await kv.set(`${request.nextUrl.pathname}`, { node, expiryTime });
```
deini, line 95: Same here

apps/core/app/api/revalidate/route/route.ts
christensenep: Good catch

deini (approved):

PR #531, opened 2024-02-08
fix(core): Target catalyst latest deployment for lighthouse scores

avattipalli:
## What/Why?
Targeting the correct deployment to generate lighthouse scores so that we have the accurate data.

## Testing
Correct scores are now posted and that can be verified on this PR.


deini (approved):
Thanks!

jorgemoya (approved):

PR #532, opened 2024-02-08
chore: make the CI more stable on deployments

chanceaclark:
## What/Why?

- Adds a little more stability to the functional and lighthouse tests, relying on the environment rather than the url.
- Splits the functional test suites into two separate workflows so they can be triggered independently and will show up as separate checks.
- Adds the ability to bypass Vercel Deployment Protection for playwright tests
- Removes an unused Counter test as we fixed that in #492 

## Testing

Unfortunately, we still have an issue where the workflows will still be triggered by other deployments which might cause the checks below to have skipped runs. This can't be fixed without something like this https://github.com/orgs/community/discussions/33210 put into place on GitHubs end. That being said, the CI checks below might show skipped, so here's some links to preview the latest runs on this branch:
- 💡 [Lighthouse Action](https://github.com/bigcommerce/catalyst/actions/workflows/lighthouse.yml?query=branch%3Astronger-ci++) 
- 🖥️  [Playwright UI](https://github.com/bigcommerce/catalyst/actions/workflows/playwright-ui.yml?query=branch%3Astronger-ci) 
- 👓  [Playwright Visual Regression](https://github.com/bigcommerce/catalyst/actions/workflows/playwright-visual-regression.yml?query=branch%3Astronger-ci)


.github/workflows/playwright-ui.yml
```diff
@@ -0,0 +1,47 @@
+name: Playwright UI Tests
+
+on:
+  deployment_status:
+    states: ['success']
+
+jobs:
+  ui-tests:
+    name: Playwright UI Tests
```
avattipalli, line 9: 🍹 I feel specifying name is redundant here as we have only one workflow which is named the same and it looks `Playwright UI Tests / Playwright UI Tests` on the checks in a PR.

.github/workflows/playwright-visual-regression.yml
```diff
@@ -0,0 +1,47 @@
+name: Playwright Visual Regression Tests
+
+on:
+  deployment_status:
+    states: ['success']
+
+jobs:
+  visual-regression-tests:
+    name: Visual Regression Tests
```
avattipalli, line 9: 🍹 Same as the above comment. Feels redundant.

avattipalli (approved):
🥇

PR #541, opened 2024-02-09
fix(core): open quick-add drawer via search param

jorgemoya:
## What/Why?
Fix issue with quick add drawer not working when switching from `static` to `dynamic` route.

## How

- Change `Quick add` button to a link that appends a search param `showQuickAdd`.
- Move `ProductSheet` to layout.
- Handle state of `ProductSheet` via this new query param.

## Related fixes

- Fix issue with pricing font size
- Add missing focused styles to close sheet button

## Testing
Locally


apps/core/app/(default)/layout.tsx
```diff
@@ -11,6 +12,9 @@ export default function DefaultLayout({ children }: PropsWithChildren) {
       <main className="flex-1 px-6 2xl:container sm:px-10 lg:px-12 2xl:mx-auto 2xl:px-0">
         {children}
       </main>
+      <Suspense fallback={null}>
+        <ProductSheet />
+      </Suspense>
```
jorgemoya, line 17: Wrapping in suspense because of this: https://nextjs.org/docs/app/api-reference/functions/use-search-params#static-rendering

apps/core/components/product-sheet/index.tsx
```diff
@@ -1,33 +1,52 @@
...
+  const pathname = usePathname();
+  const searchParams = useSearchParams();
+  const productId = searchParams.get('showQuickAdd');
 
-export const ProductSheet = ({ children, title }: PropsWithChildren<{ title: string }>) => {
   const [open, setOpen] = useState(false);
 
+  useEffect(() => {
+    setOpen(!!productId);
+  }, [productId]);
```
deini, line 26: We can derive `open` by doing something like:

```
const open = !!productId;
```

No need for this effect and state.

deini (approved):

apps/core/components/product-sheet/index.tsx
jorgemoya: Unfortunately there is a hydration issue with this approach. I believe it could be related to something to do with the searchParams when rendering, or an issue with the state directly in the radix components. Will merge as is for now and we can investigate other solutions at a later point.

PR #550, opened 2024-02-13
fix: Update tests to match product changes

avattipalli:
## What/Why?
Looks like we had some product changes that require test updates. This change should take care of that.

## Testing
Tested against latest build.

Screenshot 2024-02-13 at 2 38 49 PM

### DIFF
--- a/packages/functional/tests/ui/core/components/RadioGroup.spec.ts
+++ b/packages/functional/tests/ui/core/components/RadioGroup.spec.ts
@@ -12,21 +12,21 @@ test('Changing selection on radio button options should update query parameters'
 
   await expect(page).toHaveURL('shop-all/');
 
-  await page.getByRole('button', { name: 'Quick add' }).first().click();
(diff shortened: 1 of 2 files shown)


packages/functional/tests/ui/core/components/RadioGroup.spec.ts
```diff
@@ -12,21 +12,21 @@ test('Changing selection on radio button options should update query parameters'
 
   await expect(page).toHaveURL('shop-all/');
 
-  await page.getByRole('button', { name: 'Quick add' }).first().click();
+  await page.getByRole('link', { name: 'Quick add' }).first().click();
```
avattipalli, line 15: @jorgemoya @chanceaclark Does this look correct (change from link -> button)?

Screenshot 2024-02-13 at 2 56 19 PM
?

jorgemoya: Yes

jorgemoya (approved):

PR #551, opened 2024-02-14
fix(core): Remove product carousel looping to fix layout shift

christensenep:
## What/Why?
Removing the looping behavior on the product carousels, as it is causing layout shifts. This seems non-critical enough functionality that I think we should consider removing looping until we can address this.


jorgemoya:
Do you have a recording of what you mean by layout shifts? 😮

matthewvolk (approved):
Tested locally, LGTM 👍

https://github.com/bigcommerce/catalyst/assets/28374851/af177774-c35e-4f62-90f5-ecee4a1de916

jorgemoya (approved):

PR #554, opened 2024-02-14
fix(core): show correct product card info

jorgemoya:
## What/Why?

- Correct case for titles in product carousels
- Change "Best selling" to "Popular products"
- Swap Featured and Popular in home page
- Allow reviews to be hidden in product cards
- Hide elements (review, compare) in other pages (pdp, not-found).
- Fix spacing
- Matches designs 100%

## Testing
Locally


chanceaclark (approved):

packages/functional/playwright.config.js
```diff
@@ -19,6 +19,7 @@ export default defineConfig({
       'x-vercel-protection-bypass': process.env.VERCEL_PROTECTION_BYPASS,
       'x-vercel-set-bypass-cookie': process.env.CI ? 'true' : 'false',
     },
+    actionTimeout: 1000
```
chanceaclark, line 22: Can we avoid setting timeouts? This is usually a smell that points to something in our test or the storefront that we need to look into.

avattipalli: This is required for the test which is verifying the cyclical behavior of Carousel. As there are multiple clicks that is refreshing the carousel items being displayed this was necessary. 

I'm not sure why this is a bad practise? This isn't an explicit wait, right? We are waiting for the page/DOM loads to finish on actions we think will take time.

chanceaclark: It shouldn't be required for this test to work. [The docs](https://playwright.dev/docs/test-timeouts#advanced-low-level-timeouts) for this option even say that you should be looking elsewhere if the test is flaky:
>If you happen to be in this section because your test are flaky, it is very likely that you should be looking for the solution elsewhere.

It should work without this option:
Screenshot 2024-02-14 at 16 26 30


The `actionTimeout` option should be for something like a button that is asynchronously waiting for data to be fetched and is showing a loading state before clicking (which I think is actually being accounted for [according to the docs](https://playwright.dev/docs/input#mouse-click)), but timeouts in general should rarely be used as it's most likely a cause of something else.

avattipalli: I was legit seeing the test fail locally once in a few runs but it seems to be working fine 🤷 . Remove it.

PR #557, opened 2024-02-14
feat(cli): minimum node.js version check

matthewvolk:
## What/Why?
Exits the Node.js process for the `create-catalyst` CLI if the user's Node version is not at least 18.

## Testing

https://github.com/bigcommerce/catalyst/assets/28374851/d908e36b-01c3-47a2-9d2e-c112f51b5911

### DIFF
--- a/packages/create-catalyst/src/index.ts
+++ b/packages/create-catalyst/src/index.ts
@@ -2,11 +2,26 @@
 
 import chalk from 'chalk';
 import { Command, Option } from 'commander';
+import { satisfies } from 'semver';
 
 import PACKAGE_INFO from '../package.json';
 
 import { create } from './commands/create.js';
 import { init } from './commands/init.js';
+
+if (!satisfies(process.version, PACKAGE_INFO.engines.node)) {
+  console.error(
+    chalk.red(
+      `\nYou are using Node.js ${process.version}. Catalyst requires ${PACKAGE_INFO.engines.node}. Please upgrade your Node.js version to continue.\n`,
+    ),
+  );
+  console.log(
+    chalk.yellow(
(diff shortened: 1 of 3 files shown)


packages/create-catalyst/src/index.ts
```diff
@@ -8,6 +8,18 @@ import PACKAGE_INFO from '../package.json';
 import { create } from './commands/create.js';
 import { init } from './commands/init.js';
 
+const currentNodeVersion = process.versions.node;
+const minimumNodeVersion = 18;
+
+if (parseInt(process.versions.node.split('.')[0], 10) < minimumNodeVersion) {
+  console.error(
+    chalk.red(
+      `\nYou are running Node.js v${currentNodeVersion}. Catalyst requires Node.js v${minimumNodeVersion}.0.0 or higher.\n`,
```
matthewvolk, line 17: Open to wording suggestions

packages/create-catalyst/src/index.ts
```diff
@@ -8,6 +8,18 @@ import PACKAGE_INFO from '../package.json';
 import { create } from './commands/create.js';
 import { init } from './commands/init.js';
 
+const currentNodeVersion = process.versions.node;
+const minimumNodeVersion = 18;
+
+if (parseInt(process.versions.node.split('.')[0], 10) < minimumNodeVersion) {
```
chanceaclark, line 14: 🍹 I'm down to include `semver` for this:
```
semver.satisfies(process.versions.node, package.engines.node)
```

chanceaclark (approved):

packages/create-catalyst/src/index.ts
bookernath: Maybe include a next step, like "If you use nvm, you can run ..."

PR #559, opened 2024-02-14
feat(build): add changesets and initial changeset config

matthewvolk:
## What/Why?
Adds initial configuration for [`changesets`](https://github.com/changesets/changesets). This PR only includes configuration, and will not result in any versioning or publishing of packages until the following steps are completed:
1. Install the [changesets bot](https://github.com/apps/changeset-bot) in the Catalyst repository
2. Add the [GitHub action](https://github.com/changesets/action?tab=readme-ov-file#with-publishing) to automate releases
3. Add action vars for `${{ secrets.GITHUB_TOKEN }}` (which allows the Changesets GitHub action to open Pull Requests for new releases and `${{ secrets.NPM_TOKEN }}` (which allows the Changesets GitHub action to publish on behalf of the BigCommerce NPM org.

## Testing
N/A

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -1,0 +1,18 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@3.0.0/schema.json",
+  "changelog": [
+    "@changesets/changelog-github",
+    { "repo": "bigcommerce/catalyst" }
+  ],
+  "commit": false,
+  "fixed": [["@bigcommerce/docs", "@bigcommerce/components"]],
+  "linked": [],
+  "access": "public",
+  "privatePackages": {
+    "version": true,
+    "tag": true
+  },
+  "baseBranch": "main",
+  "updateInternalDependencies": "patch",
+  "ignore": []
+}
--- a/package.json
+++ b/package.json
@@ -11,6 +11,8 @@
     "typecheck": "turbo typecheck"
   },
   "devDependencies": {
+    "@changesets/changelog-github": "^0.5.0",
+    "@changesets/cli": "^2.27.1",
     "@inquirer/prompts": "^3.3.0",
     "dotenv-cli": "^7.3.0",
     "turbo": "^1.11.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8,6 +8,12 @@ importers:
 
   .:
     devDependencies:
+      '@changesets/changelog-github':
+        specifier: ^0.5.0
+        version: 0.5.0
+      '@changesets/cli':
+        specifier: ^2.27.1
+        version: 2.27.1
       '@inquirer/prompts':
         specifier: ^3.3.0
         version: 3.3.0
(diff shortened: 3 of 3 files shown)


.changeset/README.md
matthewvolk, line 1: Should we keep this file for the links to documentation?

.changeset/config.json
```diff
@@ -0,0 +1,18 @@
+{
+  "$schema": "https://unpkg.com/@changesets/config@3.0.0/schema.json",
+  "changelog": [
+    "@changesets/changelog-github",
+    { "repo": "bigcommerce/catalyst" }
+  ],
+  "commit": false,
```
matthewvolk, line 7: Assumption: By default, Changesets should not write the commit messages for version bumps. We instead want the author of the PR that results in a release to leave a descriptive commit message for the work done in their PR.

.changeset/config.json
```diff
@@ -0,0 +1,18 @@
...
+  "fixed": [],
+  "linked": [],
```
matthewvolk, line 9: Docs: [`fixed`](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#fixed-array-of-arrays-of-package-names)
Docs: [`linked`](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#fixed-array-of-arrays-of-package-names)

I don't believe that we have a need for these two properties, so I'm leaving them blank. The only idea I had is if we wanted to tie the version of Storybook to the current version of the component library, but I don't think that every new version of the component library would always result in a new version of Storybook.

.changeset/config.json
```diff
@@ -0,0 +1,18 @@
...
+  "access": "public",
+  "privatePackages": {
+    "version": true,
+    "tag": true
+  },
```
matthewvolk, line 14: More information: https://github.com/changesets/changesets/blob/main/docs/versioning-apps.md

TL;DR: This option tells Changesets that while package.json files with `"private": true` indeed should not be published to NPM, that we still want to version, tag, and generate changelogs for them. This is useful for apps such as `apps/core` and `packages/components`, which won't be published to NPM, but will still need semver and changelogs. 

Unfortunately, this is all or nothing, so this behavior will apply to any package.json with `"private": true`, which includes:
- `apps/docs`
- `packages/functional`

.changeset/config.json
```diff
@@ -0,0 +1,18 @@
...
+  "baseBranch": "main",
+  "updateInternalDependencies": "patch",
+  "ignore": []
+}
```
chanceaclark, line 18: What happens if we change it to something like this?
```suggestion
{
  "$schema": "https://unpkg.com/@changesets/config@2.0.0/schema.json",
  "changelog": "@changesets/cli/changelog",
  "commit": false,
  "fixed": [],
  "linked": [],
  "access": "public",
  "baseBranch": "main",
  "updateInternalDependencies": "patch",
  "ignore": ["@bigcommerce/docs", "@bigcommerce/components", "functional"]
}
```

.changeset/README.md
chanceaclark: remove it

.changeset/config.json
matthewvolk: The only reason I was hesitant on the `ignore` property is because of [this warning](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#ignore-array-of-packages):
> THIS FEATURE IS DESIGNED FOR TEMPORARY USE TO ALLOW CHANGES TO BE MERGED WITHOUT PUBLISHING THEM - If you want to stop a package from being published at all, set private: true in its package.json.

I think we want to keep the `privatePackages` prop for tag generation of core, right?

Also, the changelog being generated by `@changesets/changelog-github` is preferred because of the links to PR's 👍

chanceaclark: >Also, the changelog being generated by @changesets/changelog-github is preferred because of the links to PR's 👍

Okay cool

>I think we want to keep the privatePackages prop for tag generation of core, right?

Most likely we'll remove `private` on core, unless I'm missing something.

matthewvolk:
@bigcommerce/team-catalyst Hey all 👋 

Converting this PR to draft, Chance and I had a good chat and we think it'd be beneficial to review some process-oriented discussion topics around how we want to manage changesets for different pieces of Catalyst. 

For example, we could think of a few questions related to `apps/core`:
- how often do we release new tags for core? 
- should tags be released differently for core pre-native hosting?

I'm thinking about scheduling a meeting (1hr) where we all sit down and read the [changesets documentation](https://github.com/changesets/changesets) together for 30 minutes, with a focus on some of the questions above. Then we'll reserve the last 30 minutes for some discussion.

Once we have a good idea of the answers to these questions, I'll re-open this PR for a final review.

matthewvolk:
👋 Marking this PR ready for review. 

An update on our decisions:
* We should be releasing `apps/core` frequently, and should not feel shy to do so
* https://0ver.org/ is most closely aligned with how we should be thinking about the versioning strategy, specifically only for `apps/core`. This allows us to release frequently and allows us to not have to worry about what a breaking change is (because at the moment, the definition of what is breaking is ambiguous)
* Packages published to NPM, such as `packages/client` and `packages/eslint-configs-catalyst` will follow traditional semver, as "breaking changes" in these packages are more obvious

There are some smaller points that are worth mentioning, but I'll add those as comments in the specific lines in this PR to which they apply.

🚀

.changeset/config.json
matthewvolk: We're still somewhat undecided on this point; the ultimate decision comes down to whether or not we choose to version + tag Storybook `apps/docs`, for example.

.changeset/config.json
matthewvolk: ✅

.changeset/config.json
```diff
@@ -0,0 +1,18 @@
...
+  "fixed": [],
+  "linked": [],
+  "access": "public",
+  "privatePackages": {
+    "version": true,
+    "tag": true
+  },
+  "baseBranch": "main",
+  "updateInternalDependencies": "patch",
+  "ignore": []
```
matthewvolk, line 17: ⚠️ There is a chance we might want to use this field for `apps/docs` and `packages/functional` so that we can keep them out of NPM, and also not have to worry about versioning them.

.changeset/config.json
matthewvolk: See PR comment below. Marking this as resolved based on our last meeting 👍

deini (approved):

.changeset/config.json
matthewvolk: Since `ignore` is meant to be used as a [temporary solution](https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#ignore-array-of-packages), we'll just leave `packages/functional` with the `private` property in `package.json` set to `true`.

PR #569, opened 2024-02-15
refactor(component): carousel

jorgemoya:
## What/Why?
Refactor component to improve:
- State management in parent element
- Use context for everything else
- Add keys functionality from Shadcn
- Fix issue with arrow indicators on non loop carousels
- Prevent shifting issue (still testing).

Should look & behave virtually the same.

## Testing
Locally

### DIFF
--- a/apps/core/components/product-card-carousel/index.tsx
+++ b/apps/core/components/product-card-carousel/index.tsx
@@ -1,9 +1,9 @@
 import {
   Carousel,
   CarouselContent,
+  CarouselItem,
   CarouselNextIndicator,
   CarouselPreviousIndicator,
-  CarouselSlide,
 } from '@bigcommerce/components/carousel';
 import { useId } from 'react';
 
@@ -45,7 +45,7 @@ export const ProductCardCarousel = ({
   }, []);
 
   return (
-    <Carousel aria-labelledby="title" className="mb-14">
+    <Carousel aria-labelledby="title" className="mb-14" opts={{ loop: true }}>
       <div className="flex items-center justify-between">
         <h2 className="text-3xl font-black lg:text-4xl" id="title">
           {title}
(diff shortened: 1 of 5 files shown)


packages/components/src/components/carousel/carousel.tsx
```diff
@@ -8,134 +8,211 @@ import {
...
-
-    useImperativeHandle<ForwardedRef, ForwardedRef>(ref, () => mutableRef.current, []);
-
-    const refCallback = useCallback(
-      (current: ForwardedRef) => {
-        emblaRef(current);
-        mutableRef.current = current;
-      },
-      [emblaRef],
-    );
```
jorgemoya, line 56: @matthewvolk Do you remember why we needed this?

matthewvolk: Yup! That was added based on an assumption that the ref was required for Makeswift registration. Now that we know the Content component won't be registered in Makeswift, we should definitely be safe to remove this. The `emblaRef` might still be required by Embla, but we should at least be able to remove the imperativeHandle and refCallback.

apps/core/components/product-card-carousel/index.tsx
```diff
@@ -45,7 +45,7 @@ export const ProductCardCarousel = ({
   }, []);
 
   return (
-    <Carousel aria-labelledby="title" className="mb-14">
+    <Carousel aria-labelledby="title" className="mb-14" opts={{ loop: true }}>
```
deini, line 48: Are we changing this back to `true`?

matthewvolk: I had the same thought, but I was thinking loop was disabled for the slideshow only because of CLS. Was the product carousel also causing CLS @jorgemoya?

jorgemoya: It was but I don't see any artifacts now when running locally or on the build.

matthewvolk (approved):
👏 Much simpler and more readable.

apps/core/components/product-card-carousel/index.tsx
christensenep: Late here, but chiming in to confirm it seems to have vanished 🤷 The CLS *itself* has shifted!

PR #570, opened 2024-02-16
feat(cli): use oauth device grant by default for cli auth

matthewvolk:
## What/Why?
Default to the OAuth Device Authorization Grant flow for authentication inside the `create-catalyst` CLI. 

(Additional PR description left as comments in files in this PR)

## Testing
CI

### DIFF
--- a/packages/create-catalyst/src/commands/create.ts
+++ b/packages/create-catalyst/src/commands/create.ts
@@ -71,8 +71,6 @@ export const create = async (options: CreateCommandOptions) => {
       canCreateChannel = false;
     } else if (channels.length >= limits.total_including_inactive) {
       canCreateChannel = false;
-    } else {
-      console.log(`${limits.active - activeChannels.length} active channels remaining.`);
     }
 
     if (!canCreateChannel) {
--- a/packages/create-catalyst/src/utils/https.ts
+++ b/packages/create-catalyst/src/utils/https.ts
@@ -30,7 +30,7 @@ interface HttpsConfig {
 }
 
 export class Https {
-  DEVICE_OAUTH_CLIENT_ID = 'o094w49fnipwgbo0h9772pc2q6pw1p7';
+  DEVICE_OAUTH_CLIENT_ID = 'acse0vvawm9r1n0evag4b8e1ea1fo90';
 
   bigCommerceApiUrl: string;
   bigCommerceAuthUrl: string;
@@ -118,10 +118,7 @@ export class Https {
     });
 
     if (response.status !== 200) {
-      console.error(
-        chalk.red(`\nPOST /device/token failed: ${response.status} ${response.statusText}\n`),
-      );
-      process.exit(1);
+      throw new Error('Device code not yet verified');
     }
 
     const DeviceCodeSuccessSchema = z.object({
--- a/packages/create-catalyst/src/utils/login.ts
+++ b/packages/create-catalyst/src/utils/login.ts
@@ -1,12 +1,10 @@
 /* eslint-disable no-await-in-loop */
 
-import { confirm, input } from '@inquirer/prompts';
+import { confirm } from '@inquirer/prompts';
 import chalk from 'chalk';
 
 import { Https } from './https';
 import { spinner } from './spinner';
-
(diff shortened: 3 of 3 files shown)


packages/create-catalyst/src/commands/create.ts
```diff
@@ -71,8 +71,6 @@ export const create = async (options: CreateCommandOptions) => {
       canCreateChannel = false;
     } else if (channels.length >= limits.total_including_inactive) {
       canCreateChannel = false;
-    } else {
-      console.log(`${limits.active - activeChannels.length} active channels remaining.`);
     }
```
matthewvolk, line 74: As-is, this log is leading to more confusion than anticipated. I will rework this + the logic above as part of CATALYST-313

packages/create-catalyst/src/utils/login.ts
```diff
@@ -6,7 +6,7 @@ import chalk from 'chalk';
 import { Https } from './https';
 import { spinner } from './spinner';
 
-const DEVICE_OAUTH_ENABLED = process.env.DEVICE_OAUTH_ENABLED ?? false;
+const DEVICE_OAUTH_ENABLED = process.env.DEVICE_OAUTH_ENABLED ?? true;
```
matthewvolk, line 9: I'm conflicted as to whether we should leave this as a configurable variable. 

One reason we would be interested in keeping it is in the event of any issues with our device flow API; users would be able to set this var to `false` (`DEVICE_OAUTH_ENABLED=false pnpm create @bigcommerce/catalyst@latest`) and workaround the issue by providing store hash and access token as prompt input. The benefit to prompt input is that we can log which scopes the access token needs, so the user doesn't need to look at other documentation to figure that out. 

However, one could argue that the environment variable is redundant, because the CLI currently accepts flags for `--store-hash` and --access-token` already. That's true, but the problem with the flags is that they don't provide the same documentation that the logs I mentioned above do. We _could_ consider adding scope requirements in the `--help` menu that ships with `commander`... I'm open to either option here. I can build a quick POC to show what this would look like.

But ultimately, this variable isn't documented anywhere; so unless we either document it, tell the user, or the user looks through the source code and finds this var, they probably won't know to use it.

packages/create-catalyst/src/commands/create.ts
bookernath: 🍹  It would be cleaner if this was its own commit, but given that you're the main contributor at the moment it's probably fine.

packages/create-catalyst/src/utils/login.ts
bookernath: I think you get rid of it, accept the args as an alternate format of providing the needed information, and improve your _error message_ when the token has the wrong scopes instead of trying to pre-empt that (because to your point, you don't really have the opportunity anymore).

chanceaclark: ^ what booker said

packages/create-catalyst/src/utils/https.ts
```diff
@@ -118,10 +118,7 @@ export class Https {
     });
 
     if (response.status !== 200) {
-      console.error(
-        chalk.red(`\nPOST /device/token failed: ${response.status} ${response.statusText}\n`),
-      );
-      process.exit(1);
+      throw new Error('Device code not yet verified');
```
chanceaclark, line 121: 🍹 If I ever ran into this error code, I would be confused why the command wasn't working. Could we elaborate here on why or suggest a path towards debugging it (even if it means submitting a bug report)?

P.S. this is a sign that we also need a debug command.

chanceaclark (approved):

packages/create-catalyst/src/utils/login.ts
matthewvolk: Nice! Great point about checking scopes, I can leverage GQL for that. Adding now! Thank y'all 🙏

packages/create-catalyst/src/utils/https.ts
matthewvolk: This is a great point, ultimately this error is caught silently by the `poll` method in `login.ts`. It's basically a way to early return and retry. Maybe I should replace it with `return;`, since it's not technically an error? It's moreso just "The user hasn't logged into their store and entered their code yet".

PR #571, opened 2024-02-16
feat(core): add default revalidate target

jorgemoya:
## What/Why?
Allow an env variable to set a default revalidate target.

## Testing
Locally.

### DIFF
--- a/apps/core/client/queries/get-route.ts
+++ b/apps/core/client/queries/get-route.ts
@@ -1,5 +1,6 @@
 import { client } from '..';
 import { graphql } from '../generated';
+import { revalidate } from '../revalidate-target';
 
 export const GET_ROUTE_QUERY = /* GraphQL */ `
   query getRoute($path: String!) {
@@ -38,6 +39,7 @@ export const getRoute = async (path: string) => {
   const response = await client.fetch({
     document: query,
     variables: { path },
+    fetchOptions: { next: { revalidate } },
   });
 
   return response.data.site.route;
--- a/apps/core/client/queries/get-store-status.ts
+++ b/apps/core/client/queries/get-store-status.ts
@@ -1,5 +1,6 @@
 import { client } from '..';
 import { graphql } from '../generated';
+import { revalidate } from '../revalidate-target';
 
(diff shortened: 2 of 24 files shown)


apps/core/client/queries/get-route.ts
```diff
@@ -28,6 +29,7 @@ export const getRoute = async (path: string) => {
   const response = await client.fetch({
     document: query,
     variables: { path },
+    fetchOptions: { next: { revalidate } },
```
jorgemoya, line 32: Maybe we don't want this here?

apps/core/client/queries/get-store-status.ts
```diff
@@ -13,7 +14,7 @@ export const GET_STORE_STATUS_QUERY = /* GraphQL */ `
 
 export const getStoreStatus = async () => {
   const query = graphql(GET_STORE_STATUS_QUERY);
-  const { data } = await client.fetch({ document: query });
+  const { data } = await client.fetch({ document: query, fetchOptions: { next: { revalidate } } });
```
jorgemoya, line 17: Should we set a sensible limit here, like 3-5 mins?

apps/core/client/revalidate-target.ts
```diff
@@ -0,0 +1,3 @@
+export const revalidate = process.env.NEXT_PUBLIC_DEFAULT_REVALIDATE_TARGET
+  ? Number(process.env.NEXT_PUBLIC_DEFAULT_REVALIDATE_TARGET)
+  : undefined;
```
matthewvolk, line 3: If we default to 3600 in `.env.example`, should we also default to 3600 here?

```suggestion
  : 3600;
```

matthewvolk (commented):
This LGTM! 👍 Before I approve, can you add this default environment variable to https://github.com/bigcommerce/catalyst/blob/main/packages/create-catalyst/src/utils/write-env.ts as well?

apps/core/client/revalidate-target.ts
jorgemoya: Good catch, I think that makes sense.

matthewvolk (approved):
👍👍

PR #572, opened 2024-02-16
refactor(cli): refactor option parsing, create new channel

matthewvolk:
## What/Why?
This PR fixes two issues:
1. Before this PR, CLI flag (option) parsing missed edge cases, such as `--project-name` potentially conflicting with another directory name in the target project directory.
2. Before this PR, users capable of creating a channel on their store were not offered the option to use an existing channel; it was incorrectly assumed that if a user _could_ create a new channel, they always would. Now, they have the option to create a new channel or use an existing one. 

## Testing
CI


packages/create-catalyst/README.md
```diff
@@ -13,11 +13,3 @@ npm create catalyst-storefront@latest
...
 pnpm create catalyst-storefront@latest
 ```
-
-# To Do
-
-- [ ] Channel Site Routes
-- [ ] Check Active Storeront Limit
-- [ ] Prompt Sample Data API for New Channels
-- [ ] Yarn module resolution bug
-- [ ] Write root `.vscode/settings.json`
```
matthewvolk, line 23: These are now accounted for in the backlog, removing from `README`.

packages/create-catalyst/src/commands/create.ts
```diff
@@ -1,37 +1,94 @@
...
+  const sampleDataApiUrl = parse(options.sampleDataApiUrl, URLSchema);
+  const bigcommerceApiUrl = parse(`https://api.${options.bigcommerceHostname}`, URLSchema);
+  const bigcommerceAuthUrl = parse(`https://login.${options.bigcommerceHostname}`, URLSchema);
+
+  let projectName;
+  let projectDir;
+  let storeHash = options.storeHash;
+  let accessToken = options.accessToken;
+  let channelId;
+  let customerImpersonationToken = options.customerImpersonationToken;
```
matthewvolk, line 35: These values _may_ be parsed from options, but typically need additional validation, formatting, etc. which is denoted below by the `if (options.property)` statements below

packages/create-catalyst/src/commands/create.ts
```diff
@@ -1,37 +1,94 @@
...
-    bigCommerceAuthUrl,
-    options.storeHash,
-    options.accessToken,
-  );
+  const { packageManager, ghRef } = options;
+
+  const URLSchema = z.string().url();
+  const sampleDataApiUrl = parse(options.sampleDataApiUrl, URLSchema);
+  const bigcommerceApiUrl = parse(`https://api.${options.bigcommerceHostname}`, URLSchema);
+  const bigcommerceAuthUrl = parse(`https://login.${options.bigcommerceHostname}`, URLSchema);
```
matthewvolk, line 28: These values are parsed from options, and never need to be modified again (hence, the const declaration).

packages/create-catalyst/src/utils/check-storefront-limit.ts
```diff
@@ -0,0 +1,4 @@
+import { Https } from './https';
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export const checkStorefrontLimit = (bc: Https) => true;
```
matthewvolk, line 4: Placeholder stub - will be completed as part of CATALYST-313

packages/create-catalyst/src/utils/login.ts
```diff
@@ -34,8 +34,12 @@ export const login = async (
     return { storeHash, accessToken };
   }
 
-  const shouldLogin = await confirm({
+  const shouldLogin = await select({
```
matthewvolk, line 37: Switched for consistency with other yes/no prompts.

matthewvolk (commented):
While the diff looks large, there aren't any material changes to how the CLI works. This PR mostly just fixes a bunch of validations errors that previous versions of the CLI were letting slip through.

packages/create-catalyst/src/utils/check-storefront-limit.ts
matthewvolk: PR for this is up: #573

deini (approved):

PR #573, opened 2024-02-16
refactor(cli): more reliable check for storefront limits

matthewvolk:
⚠️ Depends on #572 

## What/Why?
As mentioned in https://github.com/bigcommerce/catalyst/pull/572#discussion_r1492990903, this PR adds more reliable checks for active storefront limit, allowing the user the option to create a new channel if the store is allowed to (or use an existing channel), but forcing them to use an existing channel if the store cannot have any more channels created.

## Testing
CI

### DIFF
--- a/packages/create-catalyst/src/commands/create.ts
+++ b/packages/create-catalyst/src/commands/create.ts
@@ -83,8 +83,8 @@ export const create = async (options: CreateCommandOptions) => {
     accessToken = credentials.accessToken;
   }
 
-  if (!projectName) throw new Error('Someting went wrong, projectName is not defined');
-  if (!projectDir) throw new Error('Someting went wrong, projectDir is not defined');
+  if (!projectName) throw new Error('Something went wrong, projectName is not defined');
+  if (!projectDir) throw new Error('Something went wrong, projectDir is not defined');
 
   if (!storeHash || !accessToken) {
     console.log(`\nCreating '${projectName}' at '${projectDir}'\n`);
@@ -109,8 +109,10 @@ export const create = async (options: CreateCommandOptions) => {
 
   if (!channelId || !customerImpersonationToken) {
     const bc = new Https({ bigCommerceApiUrl: bigcommerceApiUrl, storeHash, accessToken });
+    const availableChannels = await bc.channels('?available=true&type=storefront');
+    const storeInfo = await bc.storeInformation();
 
-    const canCreateChannel = checkStorefrontLimit(bc);
+    const canCreateChannel = checkStorefrontLimit(availableChannels, storeInfo);
 
     let shouldCreateChannel;
 
@@ -148,13 +150,11 @@ export const create = async (options: CreateCommandOptions) => {
     }
 
     if (!shouldCreateChannel) {
-      const { data: channels } = await bc.channels('?available=true&type=storefront');
-
       const channelSortOrder = ['catalyst', 'next', 'bigcommerce'];
 
       const existingChannel = await select({
         message: 'Which channel would you like to use?',
-        choices: channels
+        choices: availableChannels.data
           .sort(
             (a, b) => channelSortOrder.indexOf(a.platform) - channelSortOrder.indexOf(b.platform),
           )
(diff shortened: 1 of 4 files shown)


packages/create-catalyst/src/utils/check-storefront-limit.ts
```diff
@@ -0,0 +1,30 @@
+import chalk from 'chalk';
+
+import { Https } from './https';
+
+export const checkStorefrontLimit = async (bc: Https) => {
```
matthewvolk, line 5: Going to see if I can refactor this a bit to include unit tests with this PR. I'm considering either:
* Passing `bigcommerceApiUrl`, `storeHash`, and `accessToken` as params, creating the client inside this function, and then using `msw` to mock responses
* Or, passing `availableChannels` and `features` into this method, using the types for these responses from the `zod` objects in `https.ts` (so they're always in sync). The benefit here is that we can reuse `availableChannels` in the "use existing channel" prompt in `create.ts`.

matthewvolk: ✅

packages/create-catalyst/src/commands/create.ts
```diff
@@ -1,37 +1,95 @@
...
+
+  if (!options.storeHash || !options.accessToken) {
+    const credentials = await login(bigcommerceAuthUrl);
+
+    storeHash = credentials.storeHash;
+    accessToken = credentials.accessToken;
+  }
+
+  if (!projectName) throw new Error('Someting went wrong, projectName is not defined');
+  if (!projectDir) throw new Error('Someting went wrong, projectDir is not defined');
```
bookernath, line 87: Spelling on these two lines - `Someting` --> `Something`

packages/create-catalyst/src/commands/create.ts
```diff
@@ -49,36 +107,54 @@ export const create = async (options: CreateCommandOptions) => {
     process.exit(0);
   }
 
-  const bc = new Https({ bigCommerceApiUrl, storeHash, accessToken });
+  if (!channelId || !customerImpersonationToken) {
```
bookernath, line 110: Idea for later, we could actually read the channel ID from the token

packages/create-catalyst/src/commands/create.ts
```diff
@@ -100,29 +176,16 @@ export const create = async (options: CreateCommandOptions) => {
...
-      } = await sampleDataApi.createChannel(newChannelName);
-
-      channelId = createdChannelId;
-      customerImpersonationToken = storefrontApiToken;
     }
   }
 
+  if (!channelId) throw new Error('Someting went wrong, channelId is not defined');
+  if (!customerImpersonationToken)
+    throw new Error('Someting went wrong, customerImpersonationToken is not defined');
```
bookernath, line 184: Spelling, `Someting` -> `Something` on two lines

packages/create-catalyst/src/utils/check-storefront-limit.ts
```diff
@@ -0,0 +1,33 @@
...
+export const checkStorefrontLimit = (
+  channels: BigCommerceChannelsV3Response,
+  storeInfo: BigCommerceStoreInfo,
+) => {
+  const { data: availableChannels } = channels;
+  const { features } = storeInfo;
+
+  if (availableChannels.length >= features.storefront_limits.total_including_inactive) {
+    console.error(
+      chalk.red('\nYou have reached the maximum number of storefronts allowed for your plan.\n'),
```
bookernath, line 14: What's my next step?

(Purchase additional storefront seats in BC control panel or contact BC)

packages/create-catalyst/src/utils/check-storefront-limit.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    return false;
+  }
+
+  const activeChannels = availableChannels.filter((channel) => channel.status === 'active');
+
+  if (activeChannels.length >= features.storefront_limits.active) {
+    console.error(
+      chalk.red(
+        '\nYou have reached the maximum number of active storefronts allowed for your plan.\n',
```
bookernath, line 25: What's my next step?

(Purchase additional storefront seats in BC control panel or de-activate a channel)

Could even provide a link to channel manager.

bookernath (approved):
Functionality LGTM, made a few suggestions on verbiage.

PR #574, opened 2024-02-17
Add custom image loader for BigCommerce CDN

bookernath:
## What/Why?
Adds a custom image loader to use images directly from the BigCommerce CDN

This allows for more optimal image loading and is a more generic solution that should work across hosts, and prevent wasteful double-processing of images by stacking CDNs.

## Testing
Ensured images are coming from BC CDN where relevant, and other URLs are not interfered with:
image
image
image


bookernath:
This makes lighthouse worse because some of the CDN optimizations aren't working correctly. Looking into that...

bookernath:
Closing in favor of https://github.com/bigcommerce/catalyst/pull/701

PR #578, opened 2024-02-19
feat(core): adds prefetch and prefetchKind to Link

deini:
## What/Why?
Adds the following props to our  custom `Link` component

```ts
interface PrefetchOptions {
  prefetch?: 'hover' | 'viewport' | 'none';
  prefetchKind?: 'auto' | 'full';
}
```

Defaults are `hover` and `auto` for now.

These props give more flexibility to Next.js' `Link` component. They do deviate from their `Link` api. Their `prefetch` prop IMO is hard to follow (eg: `false` and `undefined` doing different things). Adding `hover` was going to make this prop even weirder, went the explicit way.

### DIFF
--- a/apps/core/components/link/index.tsx
+++ b/apps/core/components/link/index.tsx
@@ -1,13 +1,46 @@
+'use client';
+
 // eslint-disable-next-line @typescript-eslint/no-restricted-imports
 import NextLink from 'next/link';
-import { ComponentPropsWithRef, ElementRef, forwardRef } from 'react';
+import { useRouter } from 'next/navigation';
+import { ComponentPropsWithRef, ElementRef, forwardRef, useReducer } from 'react';
 
 import { cn } from '~/lib/utils';
 
-type Props = ComponentPropsWithRef<typeof NextLink>;
+type NextLinkProps = Omit<ComponentPropsWithRef<typeof NextLink>, 'prefetch'>;
 
+interface PrefetchOptions {
+  prefetch?: 'hover' | 'viewport' | 'none';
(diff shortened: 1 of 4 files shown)


bookernath:
Love this approach!

christensenep:
Played around with this and it seems pretty nice. Delay is still noticable if you're in "click around and see how quickly the site responds" mode, but for a normal shopper I would imagine this is *nearly* as good as the more aggressive prefetching.

I tested and from what I can tell clicking partway through an inflight prefetch does *not* trigger another request, which is what I was hoping I would discover. 👍

apps/core/components/link/index.tsx
```diff
@@ -1,21 +1,56 @@
...
+
+interface PrefetchOptions {
+  prefetch?: 'hover' | 'viewport' | 'none';
+  prefetchKind?: 'auto' | 'full';
+}
+
+type Props = NextLinkProps & PrefetchOptions;
 
+/**
+ * This custom `Link` component extends Next.js's `Link` with additional prefetching controls, making
```
christensenep, line 20: It took me a bit to realize that `auto` and `full` were describing the difference between Next.js's `undefined` and `true` values, I think that could probably be better explained here. That said, this is probably fine for now as I think ultimately we're going to probably want to reference the catalyst.dev/docs here (as this kind of thing will certainly live there), which we should have linkable soon enough, so we can iterate there instead.

christensenep (approved):

chanceaclark (approved):

jorgemoya (approved):
So basically what we expected to happen (prefetch on hover)

PR #583, opened 2024-02-21
Support 301 redirects

bookernath:
## What/Why?
Adds a lookup for existing 301 redirects on a Route and redirects as appropriate.

I have also versioned the keys to v3 out of an abundance of caution, based on our previous change in https://github.com/bigcommerce/catalyst/pull/518/files

## Testing
Confirmed that both manual and entity-targeted redirects work on a real store, and that existing Route functionality is unaffected.

Also confirmed new v3_ keys are correctly writing to KV.

### DIFF
--- a/apps/core/app/api/revalidate/route/route.ts
+++ b/apps/core/app/api/revalidate/route/route.ts
@@ -3,6 +3,7 @@ import { z } from 'zod';
 
 import { getRoute } from '~/client/queries/get-route';
 import { kv } from '~/lib/kv';
+import { routeCacheKvKey } from '~/lib/kv/keys';
 
 import { withInternalAuth } from '../../internal-auth';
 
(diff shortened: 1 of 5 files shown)


bookernath:
Need to fix types/linting

bookernath:
Thanks @chanceaclark for helping me fix up the types. This is ready for review.

christensenep:
The way I did it in the past @bookernath  was to do a `v2_` prefix, and then a few days later remove it. It's not perfect, you could still have stale data if the routes have never been revalidated in that time, but don't see a strong reason *not* to do it.

bookernath:
@christensenep ♻️

matthewvolk (approved):
LGTM 👍 

I also tested the API route to invalidate locally, and looks like it's being handled properly


https://github.com/bigcommerce/catalyst/assets/28374851/3af4b91f-d79f-4966-aac1-406e3bf62460

Redirect config for store used in testing above:
Screenshot 2024-02-23 at 1 53 52 PM

PR #591, opened 2024-02-23
feat(core): Add CSP, minus upgrade-insecure-requests

christensenep:
## What/Why?
Adds a basic CSP to Catalyst. This omits `upgrade-insecure-requests` as that was breaking Quick Add in a local environment. Will add that back separately once fixed.

## Testing
Tested locally, Quick Add works without `upgrade-insecure-requests`.

### DIFF
--- a/apps/core/next.config.js
+++ b/apps/core/next.config.js
@@ -1,5 +1,11 @@
 // @ts-check
 
+const cspHeader = `
+  base-uri 'self';
+  form-action 'self';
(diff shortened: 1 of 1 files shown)


matthewvolk (approved):
LGTM 👍 

https://github.com/bigcommerce/catalyst/assets/28374851/fa17c636-0024-42f6-9057-fd0d9d10fbc5

PR #592, opened 2024-02-23
feat(core): update shipping estimations on cart

bc-alexsaiannyi:
## What/Why?
This PR extends functionality for working with shipping estimations. Now after adding shipping cost to Cart we can change/update it.

## Testing
locally

https://github.com/bigcommerce/catalyst/assets/67792608/f41e56a4-0f36-412c-bd39-5b40e96b1b2a


yurytut1993 (approved):
👍

PR #594, opened 2024-02-23
fix(core): send product options when updating cart item

jorgemoya:
## What/Why?
Currently, when a product has a required product modifier, I can't update the quantity without an error. This is because on the backend we require all modifiers to be sent back if we're updating a product, any missing modifiers will be deleted. Because we have a required modifiers it errors out, since it needs to be passed in.

With these changes we pass all product options that are returned from cart and we're able to update product succesfully.

## Testing
Locally.


bc-alexsaiannyi (approved):
looks good for me

PR #596, opened 2024-02-23
Support Raw page type properly

bookernath:
## What/Why?
Add proper support for Raw Web Pages, in which the user should be able to set 100% of the HTML content of the page, bypassing the framework or template.

Previously this was broken and trying to insert one HTML document inside of another; this fixes that.

I have opted to use the GQL global ID as the identifier on the API route, and I'd like to start moving towards GQL global IDs in general so this is intentional. I will look at moving the other objects over to this format in a later PR.

## Testing
Compare output on preview deployment vs prod

**Main (before this PR):**
This is a bootstrap HTML document weirdly embedded into the main one, polluting the styles of the page and doing weird stuff
image

**After this PR:**
HTML document is 100% what we get back from the API
image
image

## Follow-up items
Raw Web Pages support other content types like application/json, but they aren't exposed in the GQL API. Once they are, I can set the content-type from the API response to support this properly.

### DIFF
--- a/apps/core/client/queries/get-web-page.ts
+++ b/apps/core/client/queries/get-web-page.ts
@@ -8,17 +8,6 @@ export const GET_WEB_PAGE_QUERY = /* GraphQL */ `
     site {
       route(path: $path) {
         node {
-          ... on RawHtmlPage {
-            path
-            htmlBody
-            plainTextSummary(characterLimit: $characterLimit)
-            seo {
-              pageTitle
-              metaKeywords
-              metaDescription
-            }
-            ...WebPage
-          }
           ... on ContactPage {
             contactFields
             path
@@ -81,7 +70,6 @@ export const getWebPage = cache(async ({ path, characterLimit = 120 }: Options) 
   switch (webpage.__typename) {
     case 'ContactPage':
     case 'NormalPage':
-    case 'RawHtmlPage':
       return webpage;
 
     default:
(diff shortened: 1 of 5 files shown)


apps/core/client/queries/get-web-page.ts
```diff
@@ -8,17 +8,6 @@ export const GET_WEB_PAGE_QUERY = /* GraphQL */ `
...
         node {
-          ... on RawHtmlPage {
-            path
-            htmlBody
-            plainTextSummary(characterLimit: $characterLimit)
-            seo {
-              pageTitle
-              metaKeywords
-              metaDescription
-            }
```
jorgemoya, line 19: We don't need the rest of these values?

apps/core/middlewares/with-routes.ts
```diff
@@ -228,6 +229,13 @@ export const withRoutes: MiddlewareFactory = (next) => {
         return NextResponse.rewrite(url);
       }
 
+      case 'RawHtmlPage': {
+        // fast path for raw html pages
+        const url = createRewriteUrl(`/api/raw-page/${node.id}`, request);
```
jorgemoya, line 234: Curious to know what do you mean by fast path?

apps/core/client/queries/get-web-page.ts
bookernath: No, they will never be used.

apps/core/middlewares/with-routes.ts
bookernath: I mean that's it's an API that returns HTML as directly as possible without any further processing, which is why I'm using an API route (fastest way to get from middleware to HTML output)

PR #597, opened 2024-02-24
Use event.waitUntil to complete cache updates in the background

bookernath:
## What/Why?
Using [event.waitUntil](https://nextjs.org/docs/app/building-your-application/routing/middleware#waituntil-and-nextfetchevent) allows the KV write actions to happen after the middleware has released the request, improving performance in situations where a cold cache is encountered.

Thanks @christensenep for the idea.

## Testing
Confirmed on https://1millionproducts.store/ that performance is improved when cold URL cache is encountered and that KV still gets written as expected.


apps/core/middlewares/with-routes.ts
```diff
@@ -71,22 +71,30 @@ const getExistingRouteInfo = async (request: NextRequest) => {
...
+          method: 'POST',
+          headers: {
+            'x-internal-token': process.env.BIGCOMMERCE_CUSTOMER_IMPERSONATION_TOKEN ?? '',
+          },
+        }),
+      );
     }
 
     if (routeCache && routeCache.expiryTime < Date.now()) {
-      void fetch(new URL(`/api/revalidate/route`, request.url), {
```
christensenep, line 83: We should be able to remove the revalidate api routes entirely as well 🥳

jorgemoya (approved):

PR #599, opened 2024-02-26
fix(client): update types for shipping costs related methods

bc-alexsaiannyi:
## What/Why?
This PR updates types for shipping estimations methods.

## Testing
locally

### DIFF
--- a/packages/client/src/client.ts
+++ b/packages/client/src/client.ts
@@ -131,7 +131,7 @@ class Client<FetcherRequestInit extends RequestInit = RequestInit> {
       throw new Error(`Unable to get available Countries List: ${response.statusText}`);
     }
 
-    return response.json();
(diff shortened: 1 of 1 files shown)


yurytut1993 (approved):

packages/client/src/client.ts
```diff
@@ -131,7 +131,7 @@ class Client<FetcherRequestInit extends RequestInit = RequestInit> {
       throw new Error(`Unable to get available Countries List: ${response.statusText}`);
     }
 
-    return response.json();
+    return response.json() as Promise<unknown>;
```
chanceaclark, line 134: 🍹 Using `as` can slow down typescript compilation time. If you don't want to use `as` here, you might try an explicit return type instead. `async fetchAvailableCountries(): Promise<unknown> {}`

PR #622, opened 2024-03-06
Add ThemeUi color tokens to tailwinds config

onurstats:
## What/Why?

I think using color classes like `blue-400` is not the best approach for implementing design systems. Using Color Tokens form  [Storefront Ui Components](https://www.figma.com/community/file/1337451772126638767) make components more configurable. This way if developer tweak primary, secondary colors all components will obey the color without changing the all blue classes to yellow etc.

## Testing
It can be tested on storybook and with regression testing. 

Screenshot 2024-03-06 at 18 17 11
Screenshot 2024-03-06 at 18 17 50
Screenshot 2024-03-06 at 18 18 25


packages/components/tailwind.config.js
```diff
@@ -8,9 +8,17 @@ const config = {
...
-        secondary: '#3071EF',
+      primary: '#053FB0',
+      secondary: '#3071EF',
+      white: '#FFFFFF',
+      error: {
+        DEFAULT: '#AD0000',
+        secondary: '#C62828',
+      },
+      success: {
+        light: '#4FD055',
```
jorgemoya, line 19: I believe we don't need this, need to confirm if we can remove.

packages/components/src/components/message/message.tsx
```diff
@@ -7,8 +7,8 @@ import { cn } from '~/lib/utils';
 const messageVariants = cva('flex w-full gap-x-2.5 justify-start p-3 text-base', {
   variants: {
     variant: {
-      success: 'bg-green-300/20 [&>svg]:text-green-300',
-      error: 'bg-red-100/20 [&>svg]:text-red-100',
+      success: 'bg-success-light/[.15] [&>svg]:text-success',
+      error: 'bg-error/[.15] [&>svg]:text-error-secondary',
```
jorgemoya, line 11: Related to my other comment, this seems like a one off and we should probably remove.

jorgemoya (commented):
Hi @onurstats!

I think is a great improvement and definitely think it should be the path forward. From your work, I noticed some areas of improvement that we could potentially make.

Overall this looks awesome and a much better experience.

packages/components/tailwind.config.js
```diff
@@ -44,6 +45,9 @@ const config = {
       animation: {
         revealVertical: 'revealVertical 400ms forwards cubic-bezier(0, 1, 0.25, 1)',
       },
+      opacity: {
+        15: '.15',
+      },
```
jorgemoya, line 50: Do we need this? [Seems to be included in tailwind already](https://tailwindcss.com/docs/opacity).

packages/components/src/components/message/message.tsx
onurstats: oo sory will remove those lines. maybe I can include them on a separate pull regarding the message component. it does not uses standart color and it missing 3 variant. 
Screenshot 2024-03-07 at 01 18 28

jorgemoya: Let me confirm with design and I'll reach back, thanks!

jorgemoya: I confirmed we should stick to the changes below so we can get rid of `bg-success-light`.

```suggestion
      success: 'bg-success-secondary/15 [&>svg]:text-success',
      error: 'bg-error-secondary/15 [&>svg]:text-error',
```

packages/components/tailwind.config.js
onurstats: when I removed this. when I tried `bg-success/15` it did not register the 15% opacity at first. that's why I put those lines.

packages/components/tailwind.config.js
```diff
@@ -8,9 +8,16 @@ const config = {
...
+      secondary: '#3071EF',
+      white: '#FFFFFF',
+      error: {
+        DEFAULT: '#AD0000',
+        secondary: '#C62828',
+      },
+      success: {
+        DEFAULT: '#146622',
+        secondary: '#388E3C',
       },
```
onurstats, line 21: removed `succes light = green-300` color

packages/components/src/components/message/message.tsx
```diff
@@ -7,8 +7,8 @@ import { cn } from '~/lib/utils';
 const messageVariants = cva('flex w-full gap-x-2.5 justify-start p-3 text-base', {
   variants: {
     variant: {
-      success: 'bg-green-300/20 [&>svg]:text-green-300',
-      error: 'bg-red-100/20 [&>svg]:text-red-100',
+      success: 'bg-success/[.15] [&>svg]:text-success',
+      error: 'bg-error/[.15] [&>svg]:text-error-secondary',
```
onurstats, line 11: this will render darker green color on success variant

jorgemoya: Yes, that is fine, we'll update the Figma designs.

jorgemoya: Actually, not sure if you saw my comment but this is what we want from now on: https://github.com/bigcommerce/catalyst/pull/622#discussion_r1515244682

packages/components/src/components/message/message.tsx
onur-o7: done on commit `Change Message Component Background Colors Based on Design Comment`

packages/components/src/components/message/message.tsx
onur-o7: changed based on recomended code on `Change Message Component Background Colors Based on Design Comment` commit

apps/core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx
```diff
@@ -117,6 +117,7 @@ export default async function Brand({ params: { slug, locale }, searchParams }:
               prevLabel={tPagination('prev')}
               startCursor={startCursor}
             />
+
```
Comment (jorgemoya) [inline apps/core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx:120]: I don't think we want this empty line.

apps/core/app/[locale]/(default)/(faceted)/category/[slug]/page.tsx
```diff
@@ -125,6 +125,7 @@ export default async function Category({ params: { locale, slug }, searchParams
               prevLabel={tPagination('prev')}
               startCursor={startCursor}
             />
+
```
Comment (jorgemoya) [inline apps/core/app/[locale]/(default)/(faceted)/category/[slug]/page.tsx:128]: Same here.

packages/components/src/components/message/message.tsx
jorgemoya: Perfect, thanks!

apps/core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx
Comment (onurstats) [reply apps/core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx]: sorry for the trouble :D I  run pnpn run lint on core package

jorgemoya (approved):
Looks good to me! Thanks for doing this.

PR #629, opened 2024-03-08
chore(core): bump next-intl and use Link after fix

jorgemoya:
## What/Why?
Latest `next-intl` release fixes issue with Links.

## Testing
Locally.


deini (approved):
nice

PR #630, opened 2024-03-08
feat(component): add default 'Info' variant to Message component

jorgemoya:
## What/Why?
Screenshot 2024-03-08 at 11 14 27 AM

**Extra**: Additionally removed some unnecessary use cases of the Message component.

## Testing
Locally

### DIFF
--- a/packages/components/src/components/message/message.tsx
+++ b/packages/components/src/components/message/message.tsx
@@ -4,26 +4,24 @@ import { ComponentPropsWithRef, ElementRef, forwardRef } from 'react';
 
 import { cn } from '~/lib/utils';
 
-const messageVariants = cva(
-  'flex w-full gap-x-2.5 justify-start p-3 text-base bg-secondary/[.15] [&>svg]:text-primary',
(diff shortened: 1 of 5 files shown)


jorgemoya:
@avattipalli Updated the spec but might need help updating the snapshots.

packages/components/src/components/message/message.tsx
```diff
@@ -20,6 +23,7 @@ interface MessageProps extends ComponentPropsWithRef<'div'> {
 export const Message = forwardRef<ElementRef<'div'>, MessageProps>(
   ({ className, children, variant, ...props }, ref) => (
     <div className={cn(messageVariants({ variant, className }))} ref={ref} {...props}>
+      {!variant && <Info className="flex-none" />}
```
deini, line 26: What are your thoughts on adding a default `info` variant?

jorgemoya: Sure, that works!

deini (approved):

PR #628, opened 2024-03-07
feat(core): add mutations for reset password

bc-alexsaiannyi:
## What/Why?
This PR adds 2 mutations for reset password pages (Registered  customer flow):
- `submitResetPassword` is being used for sending a link on provided email,
-  `submitChangePassword` is aimed to updating existing password with new one.

## Testing
locally

PR #638, opened 2024-03-12
feat(core): reset password

bc-alexsaiannyi:
## What/Why?
This PR adds functionality to reset password for Customer and includes 2 steps:
- reset password email,
- change password

it's based on [PR](https://github.com/bigcommerce/catalyst/pull/628) with mutations.

## Testing
locally

### DIFF
--- a/apps/core/app/[locale]/(default)/login/_actions/submit-reset-password-form.ts
+++ b/apps/core/app/[locale]/(default)/login/_actions/submit-reset-password-form.ts
@@ -1,0 +1,44 @@
+'use server';
+
+import { z } from 'zod';
+
+import { ResetPasswordSchema, submitResetPassword } from '~/client/mutations/submit-reset-password';
+
+interface SubmitResetPasswordForm {
+  formData: FormData;
+  path: string;
+  reCaptchaToken: string;
+}
+
+export const submitResetPasswordForm = async ({
+  formData,
+  path,
+  reCaptchaToken,
+}: SubmitResetPasswordForm) => {
+  try {
+    const parsedData = ResetPasswordSchema.parse({
+      email: formData.get('email'),
+    });
+
+    const response = await submitResetPassword({
+      email: parsedData.email,
+      path,
+      reCaptchaToken,
+    });
+
+    if (response.customer.requestResetPassword.errors.length === 0) {
+      return { status: 'success', data: parsedData };
+    }
+
+    return {
+      status: 'error',
+      error: response.customer.requestResetPassword.errors.map((error) => error.message).join('\n'),
+    };
+  } catch (error: unknown) {
+    if (error instanceof Error || error instanceof z.ZodError) {
+      return { status: 'error', error: error.message };
+    }
+
+    return { status: 'error', error: 'Unknown error' };
+  }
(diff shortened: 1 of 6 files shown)


matthewvolk:
Needs changeset 🦋

apps/core/app/[locale]/(default)/login/_components/reset-password-form.tsx
```diff
@@ -0,0 +1,166 @@
...
+          <>
+            <span className="absolute z-10 flex h-full w-full items-center justify-center bg-gray-400">
+              <Spinner aria-hidden="true" className="animate-spin" />
+            </span>
+            <span className="sr-only">{t('spinnerText')}</span>
+          </>
+        )}
+        <span aria-hidden={pending}>{t('submitText')}</span>
+      </>
+    </Button>
```
Comment (BC-krasnoshapka) [inline apps/core/app/[locale]/(default)/login/_components/reset-password-form.tsx:56]: Just noticed the same component is used in #665 change password form.

Comment (bc-alexsaiannyi) [reply apps/core/app/[locale]/(default)/login/_components/reset-password-form.tsx:56]: Yeah. Actually, now we have around 6 forms with submit button which are different basically only  with their texts. I was thinking about creating abstraction with `text` params that will return `<SubmitButton>` & reuse it across all forms💅

jorgemoya (approved):
Looks good but we do need to know if we want to have a different path to reset passwords other than reusing `/login` 🤔

bc-alexsaiannyi:
> Looks good but we do need to know if we want to have a different path to reset passwords other than reusing `/login` 🤔

Yeah, it's a good point, Jorge. Now, when `RequestResetPasswordInput` can deal with path, I've [added](https://github.com/bigcommerce/catalyst/blob/1445935de6bd034a03d2d2b335ea0274bb6f8e0b/apps/core/app/%5Blocale%5D/(default)/login/_components/reset-password-form.tsx#L93) it to submit function also. I believe it should be done explicitly since if we skip this argument on calling submit fn we get next result - `https://your-store/login.php?c=XX&t=XXX`. In this way it inherits || reproduces current stencil behaviour, if we miss completely `path` in arguments, we get smth like `https://your-store?c=XX&t=XXX`.
 So to receive `https://your-store?login?c=XX&t=XXX` I'm providing  it as an argument:

`submitResetPasswordForm({ formData, reCaptchaToken, path: '/login' }); `

apps/core/app/[locale]/(default)/login/_actions/submit-reset-password-form.ts
```diff
@@ -0,0 +1,44 @@
...
+      path,
+      reCaptchaToken,
+    });
+
+    if (response.customer.requestResetPassword.errors.length === 0) {
+      return { status: 'success', data: parsedData };
+    }
+
+    return {
+      status: 'failed',
```
Comment (yurytut1993) [inline apps/core/app/[locale]/(default)/login/_actions/submit-reset-password-form.ts:34]: should we change `status` to `error` here?

Comment (bc-alexsaiannyi) [reply apps/core/app/[locale]/(default)/login/_actions/submit-reset-password-form.ts]: Hmm, I was sure that I did change it. Thanks for catching this.

yurytut1993 (approved):
👌

PR #639, opened 2024-03-12
fix(core): allow previous page navigation in pagination

jorgemoya:
## What/Why?
Pass in `before` variable from query params to navigate to previous pages in navigation. With `before` we have to use `last` to only fetch the limit desired. We can't pass in `first` and `last` in the same query.

https://github.com/bigcommerce/catalyst/assets/196129/f2abdeea-a741-47ba-abf0-a916188cbab8

## Testing
Locally.

### DIFF
--- a/apps/core/client/queries/get-product-search-results.ts
+++ b/apps/core/client/queries/get-product-search-results.ts
@@ -21,7 +21,9 @@ interface ProductSearch {
 const GET_PRODUCT_SEARCH_RESULTS_QUERY = /* GraphQL */ `
   query getProductSearchResults(
     $first: Int
+    $last: Int
     $after: String
+    $before: String
     $filters: SearchProductsFiltersInput!
     $sort: SearchProductsSortInput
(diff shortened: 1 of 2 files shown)


christensenep (approved):

deini (approved):

apps/core/client/queries/get-product-search-results.ts
```diff
@@ -173,7 +176,23 @@ export const getProductSearchResults = cache(
...
+            imageWidth,
+          }
+        : {
+            first: limit,
+            after,
+            filters,
+            sort,
+            imageHeight,
+            imageWidth,
+          },
```
jorgemoya, line 195: Any recommendations on how to do this? We can't mix `last` and `first` in the same query.

deini: We do this in multiple places in the repo, check `getBlogPosts` for eg.

PR #642, opened 2024-03-12
fix(core): Fix accessibility issues

christensenep:
## What/Why?
Fixes a few accessibility issues affecting our lighthouse score. Specifically, ensures that the footer links, mobile navigation dropdown, and social icons are appropriately inside `<li>` tags in their respect `<ul>` lists.

## Testing
Tested locally against both mobile and desktop Lighthouse tests.


deini (approved):

jorgemoya:
This was already fixed before but I think we lost it in the internationalization PR 😞 . Thanks for fixing it again!

PR #644, opened 2024-03-12
feat(core): use focus-visible instead of focus

christensenep:
## What/Why?
Use `focus-visible` instead of `focus`, so that we retain our focus styles for accessibility purposes, but do not style elements when focused via actions like click or touch events.

## Testing
Ensured no focus styling is visible when using click or touch events, but remains when navigating via keyboard.

### DIFF
--- a/.changeset/empty-icons-fold.md
+++ b/.changeset/empty-icons-fold.md
@@ -1,0 +1,7 @@
+---
+"@bigcommerce/components": patch
+"@bigcommerce/catalyst-core": patch
+"@bigcommerce/docs": patch
+---
+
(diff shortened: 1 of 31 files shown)


matthewvolk (approved):
🍹 Not sure if you meant to add `apps/core/components/header/index.tsx`

christensenep:
> 🍹 Not sure if you meant to add `apps/core/components/header/index.tsx`

Impressive catch, had modified but not saved the file, thanks.

jorgemoya:
Can you expand on why we don’t want focus to show on click?

jorgemoya (approved):

PR #645, opened 2024-03-13
fix(core): Add translation provider to No Search Results

christensenep:
## What/Why?
The No Search Results form was missing a translation provider. This fixes that.

## Testing
Verified that the "No Search Results" form now shows translated strings rather than e.g. `NotFound.searchProducts`

### DIFF
--- a/.changeset/breezy-spies-sleep.md
+++ b/.changeset/breezy-spies-sleep.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Add intl provider to No Search Results page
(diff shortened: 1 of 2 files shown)


BC-krasnoshapka:
@christensenep good catch. I think `SearchForm` component can be wrapped with provider only, without other markup. 

And BTW there is one more similar case https://github.com/bigcommerce/catalyst/blob/main/apps/core/app/%5Blocale%5D/(default)/(faceted)/search/page.tsx#L34 - do you mind fixing it as well?

christensenep:
Thanks for the catch @BC-krasnoshapka, done.

jorgemoya (approved):

PR #659, opened 2024-03-14
fix(core): use amount and discounts fields for cart summary

jorgemoya:
## What/Why?
Discount was not showing in cart for a marketing promotion. Noticed we were using the `discountedAmount` for each specific line item instead of the global discounted amount value. 

Screenshot 2024-03-14 at 2 59 56 PM

Screenshot 2024-03-14 at 6 55 10 PM

## Testing
Locally

### DIFF
--- a/.changeset/perfect-spies-cheat.md
+++ b/.changeset/perfect-spies-cheat.md
@@ -2,4 +2,4 @@
 "@bigcommerce/catalyst-core": patch
 ---
 
-Use total discount value for cart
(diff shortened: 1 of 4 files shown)


christensenep:
It looks like the Grand Total in your screenshot isn't taking into account the discounts, is that correct?

deini (approved):

jorgemoya:
> It looks like the Grand Total in your screenshot isn't taking into account the discounts, is that correct?

Fixed by using `amount` value from Cart.

PR #665, opened 2024-03-18
feat(core): add components for change password

bc-alexsaiannyi:
## What/Why?
This PR adds second step for changing password.

## Testing
locally

### DIFF
--- a/apps/core/app/[locale]/(default)/login/_components/change-password-form.tsx
+++ b/apps/core/app/[locale]/(default)/login/_components/change-password-form.tsx
@@ -1,0 +1,152 @@
+'use client';
+
+import { Button } from '@bigcommerce/components/button';
+import {
+  Field,
+  FieldControl,
+  FieldLabel,
+  FieldMessage,
+  Form,
+  FormSubmit,
+} from '@bigcommerce/components/form';
+import { Input } from '@bigcommerce/components/input';
+import { Message } from '@bigcommerce/components/message';
+import { Loader2 as Spinner } from 'lucide-react';
+import { useTranslations } from 'next-intl';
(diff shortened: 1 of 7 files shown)


apps/core/app/[locale]/(default)/login/_components/change-password-form.tsx
```diff
@@ -0,0 +1,165 @@
...
+        status: 'success',
+        message: t('successMessage'),
+      });
+
+      setTimeout(() => router.push('/login'), 2000);
+    }
+
+    if (submit.status === 'failed') {
+      setFormStatus({ status: 'error', message: submit.error ?? '' });
+    }
```
Comment (matthewvolk) [inline apps/core/app/[locale]/(default)/login/_components/change-password-form.tsx:98]: Should we just use the same `status` strings between the action and the component? What if `submitChangePasswordForm` returned `status: 'error'` when an error occurred? It could potentially simplify this 🙂

apps/core/client/mutations/submit-change-password.ts
```diff
@@ -4,8 +4,8 @@ import { client } from '..';
 import { graphql } from '../graphql';
 
 export const ChangePasswordSchema = z.object({
-  newPassword: z.string(),
-  confirmPassword: z.string(),
+  newPassword: z.string().min(7),
+  confirmPassword: z.string().min(7),
```
BC-krasnoshapka, line 8: Should min length be taken from password fields settings via GQL?

apps/core/app/[locale]/(default)/login/_components/change-password-form.tsx
Comment (bc-alexsaiannyi) [reply apps/core/app/[locale]/(default)/login/_components/change-password-form.tsx]: Good point, Matt. I've refactored it already

jorgemoya (approved):

PR #673, opened 2024-03-19
refactor(core): use gql.tada on get cart query

deini:
[Hide whitespace for this one
](https://github.com/bigcommerce/catalyst/pull/673/files?w=1)
## What/Why?
`getCart` now uses `gql.tada`.

`MoneyFields` fragment was only used in `getCart`.


jorgemoya (approved):
I feel like we should use MoneyFields on Product, but that is something we can change when needed.

PR #677, opened 2024-03-19
refactor(core): use gql.tada on get blog posts and category queries

deini:
## What/Why?
- Fixed a filename typo
- `getBlogPosts`,  and `getCategory` are now using `gql`
- Removed unused fields
- Inlined `PageDetails` fragment.


apps/core/client/queries/get-blog-posts.ts
```diff
@@ -31,15 +31,16 @@ const GET_BLOG_POSTS_QUERY = /* GraphQL */ `
           name
           posts(first: $first, after: $after, last: $last, before: $before, filters: $filters) {
             pageInfo {
-              ...PageDetails
+              hasNextPage
+              hasPreviousPage
+              startCursor
+              endCursor
```
jorgemoya, line 37: We don't want to use `PageDetails` fragment anymore?

jorgemoya (approved):

apps/core/client/queries/get-blog-posts.ts
deini: Trying to use the least amount of shared fragments for now. Until we get to colocation.

PR #678, opened 2024-03-19
refactor(core): use gql.tada on get product

deini:
## What/Why?
Move `getProduct` query to use `gql.tada`.

### DIFF
--- a/apps/core/client/fragments/prices.ts
+++ b/apps/core/client/fragments/prices.ts
@@ -1,0 +1,34 @@
+import { graphql } from '../graphql';
+
+export const PRICES_FRAGMENT = graphql(`
+  fragment Prices on Product {
+    prices {
+      basePrice {
+        currencyCode
+        value
+      }
+      price {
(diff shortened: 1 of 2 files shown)


apps/core/client/fragments/prices.ts
```diff
@@ -0,0 +1,34 @@
+import { graphql } from '../graphql';
+
+export const PRICES_FRAGMENT = graphql(`
```
deini, line 3: Moved this to a "global" fragment as it's currently being exported from `get-product` and being used in multiple places.

apps/core/client/fragments/prices.ts
```diff
@@ -0,0 +1,34 @@
...
+  fragment Prices on Product {
+    prices {
+      basePrice {
+        currencyCode
+        value
```
jorgemoya, line 8: Should we use `MoneyFields` fragment on all these fields?

jorgemoya (approved):

PR #697, opened 2024-03-21
feat(core): add customer & address form fields queries

yurytut1993:
## What/Why?
This pr adds

- `getAddressFormFields` query
- `getCustomerFormFields` query
- `FormFields` fragment used in those queries

## Testing
Locally

### DIFF
--- a/apps/core/client/queries/get-address-form-fields.ts
+++ b/apps/core/client/queries/get-address-form-fields.ts
@@ -1,0 +1,46 @@
+import { cache } from 'react';
+
+import { client } from '..';
+import { FORM_FIELDS_FRAGMENT } from '../fragments/form-fields';
+import { graphql, VariablesOf } from '../graphql';
+import { revalidate } from '../revalidate-target';
+
+const GET_ADDRESS_FORM_FIELDS = graphql(
+  `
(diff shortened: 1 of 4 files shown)


apps/core/client/queries/get-address-form-fields.ts
```diff
@@ -0,0 +1,45 @@
+import { cache } from 'react';
+
+import { client } from '..';
+import { FORM_FIELDS_FRAGMENT } from '../fragments/form-fields';
+import { FormFieldFiltersInput, FormFieldSortInput } from '../generated/graphql';
+import { graphql } from '../graphql';
+import { revalidate } from '../revalidate-target';
+
+interface AddressFormFields {
```
bc-alexsaiannyi, line 9: I think that after switching to gql.tada we can do it smth like:

```
import { graphql, VariablesOf } from '../graphql';

type Variables = VariablesOf<typeof GET_ADDRESS_FORM_FIELDS>;

interface AddressFormFields {
filters: Variables['Filters'];
sortBy: Variables['sortBy'];
}

```

yurytut1993: yeah, missed that update, good catch 👍

bc-alexsaiannyi (approved):
Looks good to me

PR #698, opened 2024-03-21
chore(core): bump gql.tada packages

deini:
## What/Why?
Bumping gql.tada pacakges as the new versions have improved performance.


matthewvolk (approved):
Reviewed changelogs, this LGTM 👍

PR #709, opened 2024-03-26
Remove remotePatterns from next.config.js

bookernath:
## What/Why?
Now that we've added a custom image loader in https://github.com/bigcommerce/catalyst/pull/701, we no longer need to specify these remotePatterns.

## Testing
https://catalyst-ksjmb4x1t-bigcommerce-platform.vercel.app/

### DIFF
--- a/apps/core/next.config.js
+++ b/apps/core/next.config.js
@@ -12,13 +12,6 @@ const cspHeader = `
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   reactStrictMode: true,
-  images: {
-    remotePatterns: [
-      {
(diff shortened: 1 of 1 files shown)


apps/core/next.config.js
```diff
@@ -11,14 +11,8 @@ const cspHeader = `
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
+  output: 'standalone',
```
deini, line 14: leftover change?

bookernath: 🙈

deini (approved):

PR #712, opened 2024-03-27
refactor(cli): change default core ref to latest release

matthewvolk:
## What/Why?
Currently, the CLI always clones the `main` branch of `catalyst-core` — this has recently caused issues in which the latest published CLI version is not compatible with changes recently merged to `main` for `catalyst-core`. As an example:

1. Assume `@bigcommerce/create-catalyst@A` expects `@bigcommerce/catalyst-core@main` to have an NPM script called `script-a`
2. A recent PR proposes removing `script-a` from `@bigcommerce/catalyst-core@main`. In this PR, `@bigcommerce/create-catalyst@main` is also modified so that it no longer relies on `script-a`
3. The PR is approved, which removes `script-a` from `@bigcommerce/catalyst-core@main`, and opens a `Version Packages` PR to release `@bigcommerce/create-catalyst@B`
4. Until `Version Packages` is merged, running `pnpm create @bigcommerce/catalyst@latest` runs `@bigcommerce/create-catalyst@A` which expects `@bigcommerce/catalyst-core@main` to have `script-a`, which it no longer does. This causes the CLI to fail. 

This PR uses the [GitHub Releases API](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#list-releases) to retrieve the latest `@bigcommerce/catalyst-core` release and uses it as the default `ghRef` for the remainder of the program. This guarantees that PR's introducing changes to the CLI that are dependent on changes made to core (or vice versa) are released together, so long as the PR includes a changeset for both packages. 

Legacy behavior can be achieved by running:
```sh
pnpm create @bigcommerce/catalyst@latest --gh-ref main
``` 

## Testing
Unit tests


packages/create-catalyst/jest.config.cjs
```diff
@@ -1,4 +1,5 @@
 module.exports = {
+  transformIgnorePatterns: [],
```
matthewvolk, line 2: Unit tests throw `unexpected token` errors for `import` keywords in CLI dependency source files if this line is not here 🤔 By default, [Jest does not transform files in `node_modules`](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#list-releases). 

This feels like a bandaid, open to suggestions for how to fix this differently

packages/create-catalyst/src/utils/get-latest-core-tag.ts
```diff
@@ -0,0 +1,19 @@
...
+export const getLatestCoreTag = async () => {
+  const latestReleasesUrl = 'https://api.github.com/repos/bigcommerce/catalyst/releases';
+
+  const response = await fetch(latestReleasesUrl);
+
+  const releases = z.array(z.object({ tag_name: z.string() })).parse(await response.json());
+
+  const latestCoreRelease = releases.find((release) =>
+    release.tag_name.startsWith('@bigcommerce/catalyst-core'),
+  );
```
christensenep, line 12: Is this relying on the releases being ordered from most recent to least recent? Does the API guarantee this behavior (a quick look at their docs didn't seem to say that it did, but I may be missing something)

packages/create-catalyst/jest.config.cjs
christensenep: This does strike me as awkward but I don't have a better solution.

packages/create-catalyst/src/utils/get-latest-core-tag.ts
matthewvolk: You're correct - unfortunately, I couldn't find anything to guarantee the order will always start with most recent 😞 

I'm open to suggestions for alternatives... I could modify the request to add a maximum `per_page` param and sort the response myself, but this still doesn't provide a 100% guarantee. I'm also trying to avoid any rate limits from GitHub ([60 requests per hour](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-unauthenticated-users)), otherwise I'd be open to paginating the response to grab every release and sort in-memory (cons: inefficient, and could cause issues as the list grows in the future)

packages/create-catalyst/src/index.ts
```diff
@@ -38,11 +39,16 @@ const createCommand = program
...
-  .option('--gh-ref <url>', 'Clone a specific ref from the bigcommerce/catalyst repository', 'main')
   .option('--store-hash <hash>', 'BigCommerce store hash')
   .option('--access-token <token>', 'BigCommerce access token')
   .option('--channel-id <id>', 'BigCommerce channel ID')
   .option('--customer-impersonation-token <token>', 'BigCommerce customer impersonation token')
+  .addOption(
+    new Option(
+      '--gh-ref <ref>',
+      'Clone a specific ref from the bigcommerce/catalyst repository',
+    ).default(getLatestCoreTag()),
```
deini, line 50: Won't this always fetch from gh even if `--gh-ref` is present? Thoughts on changing the default to only be the fn reference and only calling it when needed?

matthewvolk: Good catch!

jorgemoya (approved):

packages/create-catalyst/src/utils/get-latest-core-tag.ts
matthewvolk: Refactored this PR to pull the latest core tag by parsing the [raw content of the `apps/core/package.json` file](https://raw.githubusercontent.com/bigcommerce/catalyst/main/apps/core/package.json) on `main` 👍

packages/create-catalyst/src/utils/get-latest-core-tag.ts
```diff
@@ -0,0 +1,16 @@
+import * as z from 'zod';
+
+export const getLatestCoreTag = async () => {
+  const rawPackageJsonUrl =
+    'https://raw.githubusercontent.com/bigcommerce/catalyst/main/apps/core/package.json';
+
+  const response = await fetch(rawPackageJsonUrl);
+
+  try {
+    const { version } = z.object({ version: z.string() }).parse(JSON.parse(await response.text()));
```
chanceaclark, line 10:
```suggestion
    const { version } = z.object({ version: z.string() }).parse(await response.json());
```

matthewvolk: Good catch!

PR #714, opened 2024-03-27
feat: Generate lh scores for desktop and mobile

avattipalli:
## What/Why?
This change lets us take a look at the lighthouse scores for mobile and desktop configurations at once. This would help us from being blindsided to degradation on one or the other and also gives us option to track them continuosly. 

## Testing
This change's impact are posted as a comment on this PR.


chanceaclark (approved):
LGTM! If you push this fix up for the comment, no need to rereview from my end.

PR #715, opened 2024-03-28
Add support for checking out as a logged-in customer

bookernath:
## What/Why?
Adds the existing methodology for maintaining login state while checking out as a logged-in customer.

This approach uses the [Customer Login API](https://developer.bigcommerce.com/docs/start/authentication/customer-login), which requires that the Catalyst app have a Client ID and Client Secret from a Store API Account in order to sign login JWTs. This approach works for now, but we'd like to replace it in the future by supporting this checkout redirect directly in the GraphQL Storefront API - so this will be removed & replaced in the future, and that will remove the dependency on these additional credentials as well.

## Testing
Tested locally:
- Checking out as guest
- Checking out as customer with correct credentials
- Checking out as customer with no client ID/secret (falls back gracefully to guest checkout + logs error)

### DIFF
--- a/.changeset/neat-masks-travel.md
+++ b/.changeset/neat-masks-travel.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": minor
+---
+
+Add support for logged-in customer checkout using Customer Login API
--- a/apps/core/client/management/get-checkout-url.ts
+++ b/apps/core/client/management/get-checkout-url.ts
@@ -1,6 +1,12 @@
+import * as jose from 'jose';
 import { z } from 'zod';
 
 import { client } from '..';
+
+const clientSecret = process.env.BIGCOMMERCE_CLIENT_SECRET;
+const clientId = process.env.BIGCOMMERCE_CLIENT_ID;
+const channelId = process.env.BIGCOMMERCE_CHANNEL_ID ?? Number(process.env.BIGCOMMERCE_CHANNEL_ID);
+const storeHash = process.env.BIGCOMMERCE_STORE_HASH;
 
 const RedirectUrlsSchema = z.object({
   data: z.object({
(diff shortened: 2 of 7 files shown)


.changeset/neat-masks-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": minor
+---
+
+Add support for logged-in customer checkout using Customer Login API
```
chanceaclark, line 5: I think we need to add more information for users now that we are requiring the `BIGCOMMERCE_CLIENT_SECRET` and `BIGCOMMERCE_CLIENT_ID` to be required here. There will also need to be some changes to the CLI in order for new initializations to be using this.

apps/core/client/management/get-checkout-url.ts
```diff
@@ -10,13 +16,65 @@ const RedirectUrlsSchema = z.object({
...
+    .setIssuedAt(Date.now() / 1000 - 5)
+    .setExpirationTime('24h')
+    .sign(secret);
+
+  return `${origin}/login/token/${jwt}`;
+};
+
 // Url used to redirect the user to the checkout page
-export const getCheckoutUrl = async (cartId: string) => {
+export const getCheckoutUrl = async (cartId: string, customerId?: number) => {
```
chanceaclark, line 66: I think we should leave this method unchanged and move this customerLoginJwt logic closer to the server action.

apps/core/lib/checkout-action.ts
chanceaclark, line 1: This file should probably live closer to it's usage, i.e. `apps/core/app/[locale]/(default)/cart/_actions`

chanceaclark (changes requested):
Sorry to hit you with the RC, but I think there's some things that need to get updating in this PR in order for a smooth transition of this feature.

I also think we need to fix the functional test suite along with the change to account for the link -> button change.

bookernath:
Thanks @chanceaclark , these are great suggestions.

.changeset/neat-masks-travel.md
bookernath: I did add some new docs, but I can add the callout to the changelog also.

The CLI can't autoconfigure this for $reasons - merchants need to create an explicit API Account in the CP in order to get a client secret, the CLI has no ability to do that automatically.

apps/core/client/management/get-checkout-url.ts
bookernath: Yeah, I think baking the redirect URL into the page directly and then doing the JWT wrap at the last second (since these JWTs are very short-lived) would be a nice balance.

.changeset/neat-masks-travel.md
chanceaclark: That's making me think that maybe we don't do this change if it's required to include those environment variable. As much as this would make things a bit better from the storefront perspective, impeding the flow for creating a Catalyst app might be detrimental to devs.

apps/core/client/management/get-checkout-url.ts
CNanninga: That's what I did. You guys can feel free to reference this: https://github.com/CNanninga/bc-catalyst-contentful/blob/main/app/%5Blocale%5D/(default)/cart/_actions/redirect-to-checkout.ts

CNanninga (changes requested):

bookernath:
Closing in favor of updated approach using new Customer Access Token which will be merged soon.

PR #716, opened 2024-03-28
Prefetch the cart link immediately after add-to-cart

bookernath:
## What/Why?
Add opportunistic `prefetch` to the cart link a shopper sees immediately after add-to-cart, since they are very likely to click on it after adding to cart.

This is a low-traffic, high-value deployment of prefetch which significantly improves the experience.

## Testing
Added to cart on PDP, PLP, and compare page and compared load performance of the cart link to main.

Try it yourself on the preview deployment.

### DIFF
--- a/.changeset/fresh-lemons-lick.md
+++ b/.changeset/fresh-lemons-lick.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": minor
+---
+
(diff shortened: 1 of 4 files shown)


chanceaclark (approved):

jorgemoya (approved):
Great idea!

PR #717, opened 2024-03-28
feat(components): add tabs component

bc-alexsaiannyi:
## What/Why?
This PR adds tabs component.

## Testing
Locally

Tabs

### DIFF
--- a/packages/components/src/components/tabs/tabs.tsx
+++ b/packages/components/src/components/tabs/tabs.tsx
@@ -1,0 +1,46 @@
+import * as TabsPrimitive from '@radix-ui/react-tabs';
+import { ComponentPropsWithoutRef, ElementRef, forwardRef } from 'react';
+
+import { cn } from '~/lib/utils';
+
+const Tabs = TabsPrimitive.Root;
+
+const TabsList = forwardRef<
+  ElementRef<typeof TabsPrimitive.List>,
+  ComponentPropsWithoutRef<typeof TabsPrimitive.List>
+>(({ className, ...props }, ref) => (
+  <TabsPrimitive.List
+    className={cn('mb-8 flex list-none items-start overflow-x-auto text-base', className)}
+    ref={ref}
+    {...props}
+  />
+));
+
+TabsList.displayName = TabsPrimitive.List.displayName;
+
+const TabsTrigger = forwardRef<
+  ElementRef<typeof TabsPrimitive.Trigger>,
+  ComponentPropsWithoutRef<typeof TabsPrimitive.Trigger>
+>(({ className, ...props }, ref) => (
+  <TabsPrimitive.Trigger
+    className={cn(
+      'px-4 pb-2 data-[state=active]:border-b-4 data-[state=active]:border-primary data-[state=active]:text-primary',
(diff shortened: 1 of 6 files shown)


packages/components/src/components/tabs/tabs.tsx
```diff
@@ -0,0 +1,46 @@
+import * as TabsPrimitive from '@radix-ui/react-tabs';
+import * as React from 'react';
+
+import { cn } from '~/lib/utils';
+
+const Tabs = TabsPrimitive.Root;
+
+const TabsList = React.forwardRef<
+  React.ElementRef<typeof TabsPrimitive.List>,
```
jorgemoya, line 9: Lets follow convention from other components and import what we need from React instead of using React wildcard.

packages/components/src/components/tabs/tabs.tsx
```diff
@@ -0,0 +1,46 @@
...
+  React.ComponentPropsWithoutRef<typeof TabsPrimitive.List>
+>(({ className, ...props }, ref) => (
+  <TabsPrimitive.List
+    className={cn('mb-8 flex list-none items-start overflow-x-auto text-base', className)}
+    ref={ref}
+    {...props}
+  />
+));
+
+TabsList.displayName = 'TabsList';
```
jorgemoya, line 19: I prefer if we get the `displayName` from the radix primitive instead, like we do in other components.

```suggestion
TabsList.displayName = TabsPrimitive.List.displayName;
```

jorgemoya (commented):
Looking good, just a couple of suggestions.

packages/components/src/components/tabs/tabs.tsx
```diff
@@ -0,0 +1,46 @@
+import * as TabsPrimitive from '@radix-ui/react-tabs';
+import * as React from 'react';
+
+import { cn } from '~/lib/utils';
+
+const Tabs = TabsPrimitive.Root;
+
+const TabsList = React.forwardRef<
+  React.ElementRef<typeof TabsPrimitive.List>,
+  React.ComponentPropsWithoutRef<typeof TabsPrimitive.List>
```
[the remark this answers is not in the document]
jorgemoya: I guess because we wrap in forwardRef is not needed. 🤔

packages/components/src/components/tabs/tabs.tsx
bc-alexsaiannyi: agree and updated

packages/components/src/components/tabs/tabs.tsx
bc-alexsaiannyi: I've also changed for `TabsTrigger` & `TabsContent`.

packages/components/src/components/tabs/tabs.tsx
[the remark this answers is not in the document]
bc-alexsaiannyi: I'm thinking that you're right 
Screenshot 2024-03-29 at 17 16 19

jorgemoya (approved):

yurytut1993 (approved):

PR #719, opened 2024-04-01
feat(core): add tabs for customer account

bc-alexsaiannyi:
## What/Why?
This PR adds tabs for Customer's account page.

## Testing
Locally

### DIFF
--- a/apps/core/app/[locale]/(default)/account/[tab]/page.tsx
+++ b/apps/core/app/[locale]/(default)/account/[tab]/page.tsx
@@ -1,0 +1,55 @@
+import type { Metadata } from 'next';
+import { notFound } from 'next/navigation';
+import { getTranslations } from 'next-intl/server';
+
+import { LocaleType } from '~/i18n';
+
+import { TabType } from './layout';
+
+interface Props {
+  params: {
+    locale: LocaleType;
+    tab: TabType;
+  };
+}
+
+export async function generateMetadata({ params: { tab, locale } }: Props): Promise<Metadata> {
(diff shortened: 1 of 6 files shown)


apps/core/app/[locale]/(default)/account/[tab]/page.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+import { LocaleType } from '~/i18n';
+
+export type TabType =
+  | 'orders'
+  | 'messages'
+  | 'addresses'
+  | 'wishlists'
+  | 'recently-viewed'
+  | 'settings';
```
Comment (chanceaclark) [inline apps/core/app/[locale]/(default)/account/[tab]/page.tsx:13]: Maybe we can make this a single source of truth?
```suggestion

export const tabList = ['orders', 'messages', 'addresses', 'wishlists', 'recently-viewed', 'settings'] as const
export type TabType = keyof tabList;
```

jorgemoya:
I was playing with it on the demo site and it seems like there's a delay between when a tab is selected vs when the content renders. Wondering if we can do something about it so that it looks like an exact match.

apps/core/app/[locale]/(default)/account/[tab]/page.tsx
Comment (bc-alexsaiannyi) [reply apps/core/app/[locale]/(default)/account/[tab]/page.tsx]: Thanks, Chance. I've updated but with small changes. while trying to export `tabList` I got an error `... is not a valid Next.js entry export value` and since we use `tablist` as prop in `<AccountTabs/>` so I moved both of them to layout and export  only `TabType` across page & component. What do you think?

chanceaclark (approved):

bc-azvierieva (approved):
Nice job!

PR #722, opened 2024-04-02
fix(core): preselect first state when country is selected

jorgemoya:
## What/Why?
To match the expected behavior, this change will preselect first state (if country has states array) so that it visibly demonstrate state selection that is currently being preselected when form submits.

Screenshot 2024-04-02 at 1 31 57 PM

## Testing
Locally.

### DIFF
--- a/apps/core/app/[locale]/(default)/cart/_components/shipping-info.tsx
+++ b/apps/core/app/[locale]/(default)/cart/_components/shipping-info.tsx
@@ -106,6 +106,12 @@ export const ShippingInfo = ({
     }
   }, [formValues.country, t]);
 
+  useEffect(() => {
+    if (formValues.states && !formValues.state) {
+      setFormValues({ state: formValues.states[0]?.state || '' });
+    }
+  }, [formValues.state, formValues.states]);
+
   const onSubmit = async (formData: FormData) => {
     const { status } = await submitShippingInfo(formData, {
       checkoutId: checkout.entityId,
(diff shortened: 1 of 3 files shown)


packages/components/src/components/form/form.tsx
```diff
@@ -68,9 +68,7 @@ const FieldLabel = forwardRef<ElementRef<typeof Label>, FieldLabelProps>(
       {...props}
     >
       <span>{children}</span>
-      {isRequired && (
-        <span className="text-xs font-normal font-normal text-gray-500">Required</span>
-      )}
+      {isRequired && <span className="text-xs font-normal  text-gray-500">Required</span>}
```
deini, line 71: Extra whitespace

apps/core/app/[locale]/(default)/cart/_components/shipping-info.tsx
```diff
@@ -106,6 +106,12 @@ export const ShippingInfo = ({
     }
   }, [formValues.country, t]);
 
+  useEffect(() => {
```
Comment (deini) [inline apps/core/app/[locale]/(default)/cart/_components/shipping-info.tsx:109]: Thoughts on doing this on `fetchStates` where you set the states?

Comment (jorgemoya) [reply apps/core/app/[locale]/(default)/cart/_components/shipping-info.tsx:109]: I played with that idea. It is doable but that would mean taking out `fetchStates` from inside the effect, so that I can check if `formValues.state` has been prefilled.

deini (approved):

PR #724, opened 2024-04-02
feat(core): Support SSG when the cart is non-empty

christensenep:
## What/Why?
Move the cart badge icon (indicating the number of items in the cart) to a client component, allowing us to drop the requirement that the user must not have a cart yet in order to support static routes. Now, the static pages are generated with empty carts, and when the client `CartBadge` component is initially rendered, it fetches the cart quantity from a new API endpoint. Revalidation of data still works as it did previously.

## Testing
Tested locally and in a vercel preview build, verifying that the transition from empty cart -> populated cart works correctly, as well as subsequent additions to the cart.

### DIFF
--- a/apps/core/app/api/cart-quantity/route.ts
+++ b/apps/core/app/api/cart-quantity/route.ts
@@ -1,0 +1,18 @@
+import { cookies } from 'next/headers';
+import { NextResponse } from 'next/server';
+
+import { getCart } from '~/client/queries/get-cart';
+
+export const GET = async () => {
+  const cartId = cookies().get('cartId')?.value;
+
+  if (cartId) {
+    const cart = await getCart(cartId);
+
+    return NextResponse.json({ count: cart?.lineItems.totalQuantity ?? 0 });
+  }
+
+  return NextResponse.json({ count: 0 });
+};
+
+export const runtime = 'edge';
--- a/apps/core/components/header/cart-icon.tsx
+++ b/apps/core/components/header/cart-icon.tsx
@@ -1,0 +1,47 @@
+'use client';
+
+import { Badge } from '@bigcommerce/components/badge';
+import { ShoppingCart } from 'lucide-react';
+import { useEffect, useState } from 'react';
+import { z } from 'zod';
+
+export const CartQuantityResponseSchema = z.object({
+  count: z.number(),
+});
+
+interface CartIconProps {
+  count?: number;
+}
+
+export const CartIcon = ({ count }: CartIconProps) => {
+  const [fetchedCount, setFetchedCount] = useState<number | null>();
+  const computedCount = count ?? fetchedCount;
+
+  useEffect(() => {
+    async function fetchCartQuantity() {
+      const response = await fetch(`/api/cart-quantity/`);
+      const parsedData = CartQuantityResponseSchema.parse(await response.json());
+
+      setFetchedCount(parsedData.count);
+    }
+
+    // When a page is rendered statically via the 'force-static' route config option, cookies().get() always returns undefined,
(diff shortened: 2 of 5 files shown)


apps/core/components/header/cart.tsx
```diff
@@ -17,24 +18,19 @@ export const CartLink = ({ children }: { children: ReactNode }) => (
 
 export const Cart = async () => {
   const cartId = cookies().get('cartId')?.value;
+  let count: number | undefined;
 
-  if (!cartId) {
-    return (
-      <CartLink>
```
christensenep, line 23: I'm assuming there was no particular reason we needed to exclude 

`<span className="sr-only">Cart Items</span>` 

when there was no cart, but correct me if I'm wrong here.

apps/core/app/api/cart-quantity/route.ts
```diff
@@ -0,0 +1,19 @@
...
+  let count = 0;
+
+  if (cartId) {
+    const cart = await getCart(cartId);
+
+    count = cart?.lineItems.totalQuantity ?? 0;
+  }
+
+  return NextResponse.json({ count });
+};
```
chanceaclark, line 17: We try to avoid using `let`'s and mutations.
```suggestion
export const GET = async () => {
  const cartId = cookies().get('cartId')?.value;

  if (cartId) {
    const cart = await getCart(cartId);

    return NextResponse.json({ count: cart?.lineItems.totalQuantity ?? 0 });
  }

  return NextResponse.json({ count: 0 });
};
```

apps/core/components/header/cart-badge.tsx
```diff
@@ -0,0 +1,37 @@
...
+    if (count === undefined) {
+      void fetchCartQuantity();
+    }
+  }, [count]);
+
+  if (computedCount) {
+    return <Badge> {computedCount}</Badge>;
+  }
+
+  return null;
```
chanceaclark, line 36: I just thought of a better way after we chatted about this 😶‍🌫️ 
```suggestion
  const [fetchedCount, setFetchedCount] = useState<number | null>(count ?? null);

  useEffect(() => {
    async function fetchCartQuantity() {
      const response = await fetch(`/api/cart-quantity/`);
      const parsedData = CartQuantityResponseSchema.parse(await response.json());

      setFetchedCount(parsedData.count);
    }

    if (fetchedCount === null) {
      void fetchCartQuantity();
    }
  }, [fetchedCount]);

  if (!fetchedCount) {
    return null;
  }

  return <Badge> {fetchedCount}</Badge>;
```

Also, a :nit:, we try to `return null` before we return the component.

apps/core/components/header/cart.tsx
chanceaclark: For accessibility we wanted an empty cart to just say `"Link, cart"` and a cart with items, `"Link, Cart Items {count}"` for screen readers.

christensenep: Ah gotcha, I'll see about supporting that

apps/core/components/header/cart-icon.tsx
```diff
@@ -0,0 +1,44 @@
...
+      setFetchedCount(parsedData.count);
+    }
+
+    if (count === undefined) {
+      void fetchCartQuantity();
+    }
+  }, [count]);
+
+  if (!computedCount) {
+    return <ShoppingCart aria-label="cart" />;
```
chanceaclark, line 34:
```suggestion
    return <ShoppingCart aria-label="cart" />
```

christensenep: Prettier is actually adding this one here (but fixed the other one)

deini (approved):

PR #725, opened 2024-04-03
feat: use streaming + cache for cart [experimental]

chanceaclark:
## What/Why?
After doing some research from @christensenep's PR #724, we are not caching our cart request. We also discussed this a bit in #647. I am proposing to add this PR into main for a bit to see from Vercel's side what Data Cache implications are if we cached this request. We are mostly curious to see if our cache will be evicted more heavily now that we are caching user-specific requests.

## Testing
Doing some testing, most requests seemed to be sub-400ms with an item added to the cart. I want to measure for a week if that is the case in prod + identifying the Data Cache implications.


chanceaclark:
It's too slow still

PR #728, opened 2024-04-03
fix(core): Fix breadcrumb links on PDP

christensenep:
## What/Why?
The breadcrumb links on the PDP were all pointing to `#`. I've added the `path` field to the getProduct query so that we have the path of each category available to populate the link with.

## Testing
Verified that the breadcrumb links on the PDP now link to the correct page.

### DIFF
--- a/apps/core/app/[locale]/(default)/product/[slug]/_components/breadcrumbs.tsx
+++ b/apps/core/app/[locale]/(default)/product/[slug]/_components/breadcrumbs.tsx
@@ -17,7 +17,7 @@ export const BreadCrumbs = async ({ productId }: Props) => {
   return (
     <nav>
       <ul className="m-0 flex flex-wrap items-center p-0 md:container md:mx-auto ">
-        {category.breadcrumbs.map((breadcrumb, i, arr) => {
+        {category.breadcrumbs.map(({ name, path }, i, arr) => {
           const isLast = arr.length - 1 === i;
 
           return (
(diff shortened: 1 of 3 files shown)


chanceaclark (approved):

apps/core/app/[locale]/(default)/product/[slug]/_components/breadcrumbs.tsx
```diff
@@ -26,9 +26,9 @@ export const BreadCrumbs = async ({ productId }: Props) => {
                 'font-semibold': !isLast,
                 'font-extrabold': isLast,
               })}
-              key={breadcrumb.name}
+              key={name}
             >
-              <Link href="#">{breadcrumb.name}</Link>
+              <Link href={path ?? '#'}>{name}</Link>
```
Comment (christensenep) [inline apps/core/app/[locale]/(default)/product/[slug]/_components/breadcrumbs.tsx:31]: Not sure if we want to do something more complex here if `path` is null, I'm not sure how much of an edge case that is.

Comment (chanceaclark) [reply apps/core/app/[locale]/(default)/product/[slug]/_components/breadcrumbs.tsx:31]: Maybe it doesn't need to be a link if there is no path?

Comment (christensenep) [reply apps/core/app/[locale]/(default)/product/[slug]/_components/breadcrumbs.tsx:31]: I experimented with this, will also involve remove the `hover:text-primary` styling above, and I started to wonder if it's worth the complexity (I'm honestly not sure why a category would not have a path)

Comment (chanceaclark) [reply apps/core/app/[locale]/(default)/product/[slug]/_components/breadcrumbs.tsx:31]: >I'm honestly not sure why a category would not have a path

Good question, maybe that is something we need to look into at the API level 🤔

Comment (christensenep) [reply apps/core/app/[locale]/(default)/product/[slug]/_components/breadcrumbs.tsx:31]: Confirmed this is an extreme edge case, don't think we need to account for it any more gracefully than this.

PR #730, opened 2024-04-04
feat(core): create register customer page

yurytut1993:
## What/Why?
This pr adds Register Customer Page with Built In Fields
register-customer-form

## Testing
Locally

### DIFF
--- a/.changeset/nasty-news-retire.md
+++ b/.changeset/nasty-news-retire.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": minor
+---
+
+create register customer page
--- a/core/app/[locale]/(default)/login/page.tsx
+++ b/core/app/[locale]/(default)/login/page.tsx
@@ -96,15 +96,8 @@ export default async function Login({ params: { locale }, searchParams }: Props)
             <li>{t('CreateAccount.ordersTracking')}</li>
             <li>{t('CreateAccount.wishlists')}</li>
           </ul>
-          <Button asChild className="w-fit items-center px-8 py-2">
-            <Link
-              href={{
-                pathname: '/login',
-                query: { action: 'create_account' },
-              }}
-            >
-              {t('CreateAccount.createLink')}
-            </Link>
+          <Button asChild className="w-fit items-center px-8 py-2 hover:text-white">
+            <Link href="/login/register-customer">{t('CreateAccount.createLink')}</Link>
           </Button>
         </div>
       </div>
--- a/core/app/[locale]/(default)/login/register-customer/_components/register-customer-form/_actions/login.ts
+++ b/core/app/[locale]/(default)/login/register-customer/_components/register-customer-form/_actions/login.ts
@@ -1,0 +1,26 @@
+'use server';
+
+import { isRedirectError } from 'next/dist/client/components/redirect';
+
+import { signIn } from '~/auth';
+
+export const login = async (
+  email: FormDataEntryValue | null,
+  password: FormDataEntryValue | null,
+) => {
+  try {
+    return await signIn('credentials', {
+      email,
+      password,
+      redirectTo: '/account',
+    });
+  } catch (error: unknown) {
+    if (isRedirectError(error)) {
+      throw error;
+    }
+
+    return {
+      status: 'error',
+    };
+  }
+};
--- a/core/app/[locale]/(default)/login/register-customer/_components/register-customer-form/_actions/register-customer.ts
+++ b/core/app/[locale]/(default)/login/register-customer/_components/register-customer-form/_actions/register-customer.ts
@@ -1,0 +1,65 @@
+'use server';
+
+import {
+  Input,
+  registerCustomer as registerCustomerClient,
+} from '~/client/mutations/register-customer';
+
+interface RegisterCustomerForm {
+  formData: FormData;
+  reCaptchaToken?: string;
+}
+
+const isRegisterCustomerInput = (data: unknown): data is Input => {
+  if (typeof data === 'object' && data !== null && 'email' in data) {
+    return true;
+  }
+
+  return false;
+};
+
+export const registerCustomer = async ({ formData, reCaptchaToken }: RegisterCustomerForm) => {
+  formData.delete('customer-confirmPassword');
+
+  const parsedData = Array.from(formData.entries()).reduce<{
+    [key: string]: FormDataEntryValue | { [key: string]: FormDataEntryValue };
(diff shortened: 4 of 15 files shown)


packages/components/src/components/form/form.tsx
```diff
@@ -15,7 +15,8 @@ type ValidationPattern =
   | 'tooShort'
   | 'typeMismatch'
   | 'valid'
-  | 'valueMissing';
+  | 'valueMissing'
+  | ValidationFunction;
```
bc-alexsaiannyi, line 19: I think there can be an error in `ControlValidationPatterns` so basically you can update and use `ControlValidationPatterns` after:
` type ControlValidationPatterns = ValidationPattern | ValidationFunction;`
What do you think?

yurytut1993: Yes, changed after your updates

apps/core/app/[locale]/(default)/account/register-customer/page.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { NextIntlClientProvider } from 'next-intl';
+import { getMessages, getTranslations } from 'next-intl/server';
+
+import { auth } from '~/auth';
+import { FormFieldSortInput } from '~/client/generated/graphql';
+import { getCountries } from '~/client/management/get-countries';
+import { getAddressFormFields } from '~/client/queries/get-address-form-fields';
+import { getCustomerFormFields } from '~/client/queries/get-customer-form-fields';
+import { getReCaptchaSettings } from '~/client/queries/get-recaptcha-settings';
+import { getStoreCountry } from '~/client/queries/get-store-country';
```
Comment (jorgemoya) [inline apps/core/app/[locale]/(default)/account/register-customer/page.tsx:11]: Can we do this with a single query?

apps/core/components/register-customer-form/_actions/login.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const login = async (
+  email: FormDataEntryValue | null,
+  password: FormDataEntryValue | null,
+) => {
+  try {
+    const singin = await signIn('credentials', {
+      email,
+      password,
+      redirectTo: '/account',
```
jorgemoya, line 15: I believe we already redirect to account but I might be wrong.

jorgemoya (commented):
IMO `/register-customer-form` should be nested under a `_components` folder in the `register-customer` path, since it's not reused.

apps/core/app/[locale]/(default)/account/register-customer/page.tsx
Comment (yurytut1993) [reply apps/core/app/[locale]/(default)/account/register-customer/page.tsx:11]: Interesting idea, but `getAddressFormFields`, `getAddressFormFields` and `getReCaptchaSettings` need to be used in separate cases 🤔

apps/core/components/register-customer-form/_actions/login.ts
yurytut1993: Yes, we have, but `submitLoginForm` action used for form and has different arguments. And using it here requires huge refactoring and can add mess in currently existed action.

apps/core/app/[locale]/(default)/account/register-customer/page.tsx
Comment (jorgemoya) [reply apps/core/app/[locale]/(default)/account/register-customer/page.tsx:11]: You can try something like what @deini is doing with [fragment colocation](https://github.com/bigcommerce/catalyst/pull/766). What do you think?

apps/core/components/register-customer-form/_actions/login.ts
jorgemoya: This is okay then!

bc-alexsaiannyi:
> IMO `/register-customer-form` should be nested under a `_components` folder in the `register-customer` path, since it's not reused.

Hmm, It's interested point. I guess that in this case it makes sense to move `contactUs` form to relevant page `_components `and delete `core/components/forms` altogether. What do you think guys?🤔

yurytut1993:
> > IMO `/register-customer-form` should be nested under a `_components` folder in the `register-customer` path, since it's not reused.
> 
> Hmm, It's interested point. I guess that in this case it makes sense to move `contactUs` form to relevant page `_components `and delete `core/components/forms` altogether. What do you think guys?🤔

It will be consistent. And `blog-post-card` also should be replaced under `Blog` then

apps/core/components/register-customer-form/_actions/register-customer.ts
```diff
@@ -0,0 +1,65 @@
...
+      if (sections.includes('address')) {
+        acc.address[key] = value;
+      }
+
+      return acc;
+    },
+    { address: {} },
+  );
+
+  console.log(parsedData, 'parsedData');
```
bc-alexsaiannyi, line 43: don't forget to remove console

apps/core/components/register-customer-form/index.tsx
```diff
@@ -0,0 +1,410 @@
...
+    setCountryStatesPending(true);
+
+    const { data = [] } = await getShippingStates(country?.id || fallbackCountryId);
+
+    setCountryStatesPending(false);
+
+    setCountryStates(data);
+  };
+
+  const handlePicklistWihtTextValidation = (e: ChangeEvent<HTMLInputElement>) => {
```
bc-alexsaiannyi, line 211: here is a typo in method's name.

apps/core/app/[locale]/(default)/account/register-customer/page.tsx
Comment (yurytut1993) [reply apps/core/app/[locale]/(default)/account/register-customer/page.tsx]: @jorgemoya thats great idea, thanks. Refactored according to the new approach 👍

apps/core/app/[locale]/(default)/login/register-customer/page-data.ts
```diff
@@ -0,0 +1,105 @@
+import { cache } from 'react';
+
+import { getSessionCustomerId } from '~/auth';
+import { client } from '~/client';
+import { FORM_FIELDS_FRAGMENT } from '~/client/fragments/form-fields';
+import { graphql, VariablesOf } from '~/client/graphql';
+import { revalidate } from '~/client/revalidate-target';
+
+const RegisterCustomerQuery = graphql(
```
Comment (jorgemoya) [inline apps/core/app/[locale]/(default)/login/register-customer/page-data.ts:9]: I believe we can use fragment collocation better, similar to https://github.com/bigcommerce/catalyst/pull/848, but not sure if it should be part of this effort.

bc-alexsaiannyi (approved):
looks good to me

jorgemoya (approved):

bc-azvierieva (approved):
Nice job!

PR #731, opened 2024-04-04
feat(core): add tax total in checkout summary

jorgemoya:
## What/Why?
Add total tax from checkout.

Screenshot 2024-04-04 at 10 25 20 AM

## Testing
Locally.

### DIFF
--- a/apps/core/app/[locale]/(default)/cart/_components/checkout-summary.tsx
+++ b/apps/core/app/[locale]/(default)/cart/_components/checkout-summary.tsx
@@ -47,6 +47,11 @@ export const CheckoutSummary = async ({ cartId, locale }: { cartId: string; loca
         <span>-{currencyFormatter.format(checkout.cart?.discountedAmount.value || 0)}</span>
       </div>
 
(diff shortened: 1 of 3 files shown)


apps/core/app/[locale]/(default)/cart/_components/checkout-summary.tsx
```diff
@@ -47,6 +47,11 @@ export const CheckoutSummary = async ({ cartId, locale }: { cartId: string; loca
         <span>-{currencyFormatter.format(checkout.cart?.discountedAmount.value || 0)}</span>
       </div>
 
+      <div className="flex justify-between border-t border-t-gray-200 py-4">
+        <span className="font-semibold">{t('tax')}</span>
+        <span>{currencyFormatter.format(checkout.taxTotal?.value || 0)}</span>
```
Comment (christensenep) [inline apps/core/app/[locale]/(default)/cart/_components/checkout-summary.tsx:52]: 🍹 Curious why `taxTotal` can be missing here. The fallback of 0 seems reasonable but if it represents a concept of tax not existing at all I wonder if it's more appropriate to not show the tax field at all. (not a blocker at all for merging though imho)

christensenep (approved):

PR #735, opened 2024-04-05
chore: bump dependencies

deini:
## What/Why?
Bump most of the dependencies in prep for dependabot


apps/core/package.json
```diff
@@ -15,50 +15,50 @@
...
+    "@0no-co/graphqlsp": "^1.9.1",
     "@bigcommerce/eslint-config-catalyst": "workspace:^",
-    "@gql.tada/cli-utils": "^0.3.0",
+    "@gql.tada/cli-utils": "^0.3.3",
     "@tailwindcss/container-queries": "^0.1.1",
-    "@testing-library/react": "^14.1.2",
+    "@testing-library/react": "^14.2.2",
     "@types/lodash.debounce": "^4.0.7",
-    "@types/node": "^18.17.12",
+    "@types/node": "^20.12.4",
```
chanceaclark, line 48: Should we also change our `.nvmrc` and build pipeline node version too?

deini: No (at least not as part of this PR), my bad, didn't plan on updating the types. Good catch 👀

chanceaclark (approved):

jorgemoya (approved):

PR #757, opened 2024-04-12
fix: select component wouldn't scroll with long list

chanceaclark:
## What/Why?
Updates the behavior of the select component in order to scroll for longer lists.

## Testing

https://github.com/bigcommerce/catalyst/assets/10539418/c902de5b-829b-4635-a680-41505c35ee44

### DIFF
--- a/packages/components/src/components/select/select.tsx
+++ b/packages/components/src/components/select/select.tsx
@@ -59,19 +59,27 @@ type SelectContentType = typeof SelectPrimitive.Content;
 const SelectContent = forwardRef<
   ElementRef<SelectContentType>,
   ComponentPropsWithRef<SelectContentType>
->(({ children, className, ...props }, ref) => {
+>(({ children, className, position = 'popper', ...props }, ref) => {
   return (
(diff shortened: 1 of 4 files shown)


packages/components/src/components/select/select.tsx
```diff
@@ -59,19 +59,27 @@ type SelectContentType = typeof SelectPrimitive.Content;
...
           className,
         )}
         ref={ref}
       >
-        <SelectPrimitive.Viewport className="h-radix-select-content-available-height w-full min-w-[var(--radix-select-trigger-width)]">
+        <SelectPrimitive.Viewport
+          className={cn(
+            'w-full',
+            position === 'popper' &&
+              'h-[var(--radix-select-trigger-height)] w-full min-w-[var(--radix-select-trigger-width)]',
```
deini, line 80: do we need `w-full` here?

chanceaclark: Good catch

deini (approved):

PR #758, opened 2024-04-12
chore: remove old setup script

christensenep:
## What/Why?
Remove the old `scripts/setup.mjs` script, deprecated in favor of the Catalyst CLI.


matthewvolk (approved):
Looks good - 🍹 Should we add a line around `Setup > Step 3` to recommend running `pnpm create @bigcommerce/catalyst@latest init` to connect the monorepo to a store for local development?

PR #766, opened 2024-04-15
refactor(core): homepage fragment colocation

deini:
## What/Why?
In this PR, the primary objective is to optimize the homepage by consolidating multiple GraphQL queries into a single one. However, due to the reuse of shared components like product cards and pricing across various pages, it required adjustments to GraphQL requests throughout the codebase.


apps/core/client/queries/get-newest-products.ts
```diff
@@ -1,48 +0,0 @@
-import { removeEdgesAndNodes } from '@bigcommerce/catalyst-client';
```
deini, line 1: The homepage was the only one using this query.

apps/core/components/forbidden/index.tsx
```diff
@@ -1,33 +0,0 @@
-import { getFeaturedProducts } from '~/client/queries/get-featured-products';
```
deini, line 1: Looks like we weren't using this one. 🧹

jorgemoya (approved):

apps/core/app/[locale]/(default)/product/[slug]/page.tsx
```diff
@@ -68,12 +87,27 @@ export default async function Product({ params, searchParams }: ProductPageProps
...
+  const [product, { data }] = await Promise.all([
+    getProduct(productId, optionValueIds),
+
+    client.fetch({
+      document: ProductPageQuery,
+      variables: { entityId: productId, optionValueIds },
+      customerId,
+      fetchOptions: customerId ? { cache: 'no-store' } : { next: { revalidate } },
+    }),
+  ]);
```
Comment (chanceaclark) [inline apps/core/app/[locale]/(default)/product/[slug]/page.tsx:101]: 😬

chanceaclark (approved):

PR #768, opened 2024-04-15
feat(core): add get customer query

yurytut1993:
## What/Why?
This pr adds query to get Customer's account data

## Testing
Locally

### DIFF
--- a/apps/core/client/queries/get-customer.ts
+++ b/apps/core/client/queries/get-customer.ts
@@ -1,0 +1,60 @@
+import { cache } from 'react';
+
+import { client } from '..';
+import { graphql } from '../graphql';
+
+const GET_CUSTOMER_QUERY = graphql(`
+  query getCustomer {
+    customer {
+      entityId
+      company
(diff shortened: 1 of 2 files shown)


apps/core/client/queries/get-customer.ts
```diff
@@ -0,0 +1,61 @@
...
+        }
+      }
+    }
+  }
+`);
+
+export const getCustomer = cache(async (customerId: string) => {
+  const response = await client.fetch({
+    document: GET_CUSTOMER_QUERY,
+    fetchOptions: { next: { revalidate } },
```
bc-alexsaiannyi, line 50: don't we need to use here `cache: 'no-store' `option  for Customer?

bc-alexsaiannyi (approved):

apps/core/client/queries/get-customer.ts
yurytut1993: good catch 👍 thanks

PR #769, opened 2024-04-15
feat(core): Add tests for shipping estimates

avattipalli:
## What/Why?
Add UI tests for calculating shipping estimates on cart items.

## Testing
Newly added tests should run on this PR.

### DIFF
--- a/apps/core/app/[locale]/(default)/cart/_components/shipping-info.tsx
+++ b/apps/core/app/[locale]/(default)/cart/_components/shipping-info.tsx
@@ -139,11 +139,11 @@ export const ShippingInfo = ({
     >
       <>
         <Field className="relative space-y-2" name="country">
-          <FieldLabel>{t('country')}</FieldLabel>
+          <FieldLabel htmlFor="country">{t('country')}</FieldLabel>
           <FieldControl asChild>
             <Select
-              aria-label={t('countryAriaLabel')}
               autoComplete="country"
+              id="country"
               onValueChange={(value: string) => {
                 const countryId = value.split('-')[1];
 
@@ -171,12 +171,12 @@ export const ShippingInfo = ({
           </FieldControl>
         </Field>
         <Field className="relative space-y-2" name="state">
-          <FieldLabel>{t('state')}</FieldLabel>
+          <FieldLabel htmlFor="state">{t('state')}</FieldLabel>
           <FieldControl asChild>
             {formValues.states !== null ? (
               <Select
-                aria-label={t('stateAriaLabel')}
                 disabled={formValues.states.length === 0}
+                id="state"
                 onValueChange={(value) => setFormValues({ state: value })}
                 placeholder={t('statePlaceholder')}
                 value={formValues.state}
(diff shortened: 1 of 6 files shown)


apps/core/messages/en.json
```diff
@@ -128,13 +128,17 @@
     },
     "ShippingInfo": {
       "country": "Country",
-      "countryPlaceholder": "Select county",
```
avattipalli, line 131: Corrected a typo and added a few aria-label attributes.

apps/core/app/[locale]/(default)/cart/_components/shipping-info.tsx
```diff
@@ -217,6 +220,7 @@ export const ShippingInfo = ({
           <FieldLabel htmlFor="zip-field">{t('postcode')}</FieldLabel>
           <FieldControl asChild>
             <Input
+              aria-label={t('postcodeAriaLabel')}
```
Comment (chanceaclark) [inline apps/core/app/[locale]/(default)/cart/_components/shipping-info.tsx:223]: These should already have an accessible name attributed to them via the [`htmlFor` + `id`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/for) relationship. These should not be needed.

(Same with the input above)

apps/core/app/[locale]/(default)/cart/_components/shipping-info.tsx
```diff
@@ -142,6 +142,7 @@ export const ShippingInfo = ({
           <FieldLabel>{t('country')}</FieldLabel>
           <FieldControl asChild>
             <Select
+              aria-label={t('countryAriaLabel')}
```
Comment (chanceaclark) [inline apps/core/app/[locale]/(default)/cart/_components/shipping-info.tsx:145]: Looking above, it doesn't look like there is an `htmlFor` + `id` relationship. The best practice for accessibility is to use `aria-*` attributes as a last resort, specifically when there are no other native features to initially use.

That being said, I think we need to augment our `Select` component to accept an `id` prop and forward that attribute to the trigger. That way we can add an identifier to the field in order to attribute the `<label>` above. 

(Same with the select below)

chanceaclark (commented):
Tests look good, just some accessibility comments about the labels you added.

packages/functional/actions/cart-actions.ts
```diff
@@ -0,0 +1,26 @@
...
+  state: string,
+  city: string,
+  zip: string,
+): Promise<void> {
+  await page.getByRole('combobox', { name: 'Country' }).click();
+  await page.getByLabel(country).click();
+
+  await page
+    .getByRole('combobox', { name: 'State' })
+    .filter({ hasNotText: 'Select state' })
```
chanceaclark, line 15: I think we should be able to change it to this in order to not use `.filter`.
```suggestion
    .getByRole('combobox', { name: 'State/province' })
```

packages/functional/actions/cart-actions.ts
```diff
@@ -0,0 +1,26 @@
...
+    .click();
+  await page.getByLabel(state).click();
+
+  await page.getByLabel('City').fill(city);
+  await page.getByLabel('Postcode').fill(zip);
```
chanceaclark, line 20: 🍹 I would make sure that these are the right label:
```suggestion
  await page.getByLabel('Suburb/city').fill(city);
  await page.getByLabel('Zip/Postcode').fill(zip);
```

packages/functional/actions/cart-actions.ts
avattipalli: This was done mainly to avoid sporadic failures which were happening when playwright was quick to click on the dropdown before it loads the states based on selection of country. This was happening when testing against localhost and so, I tried to handle it for all scenarios like running it on CI and locally. I'll try and verify if it's working fine without the filter.

chanceaclark (approved):

PR #772, opened 2024-04-16
feat(core): Test coupon application at checkout

avattipalli:
## What/Why?
This change adds a new test to verify applying a coupon at checkout and consolidates all coupon related tests into its own test file.

## Testing
Tested against a deployed branch.

### DIFF
--- a/packages/functional/tests/ui/desktop/e2e/coupon.spec.ts
+++ b/packages/functional/tests/ui/desktop/e2e/coupon.spec.ts
@@ -1,0 +1,61 @@
+import { expect, test } from '@playwright/test';
+
+import { ProductActions } from '../../../../actions/product-actions';
+
+const sampleProduct = '[Sample] Able Brewing System';
+const couponCode = 'OFF25';
+
+test.beforeEach(async ({ page }) => {
+  await page.goto('/');
+  await page.getByLabel('Main').getByRole('link', { name: 'Kitchen' }).click();
+
+  await expect(page.getByRole('heading', { level: 1, name: 'Kitchen' })).toBeVisible();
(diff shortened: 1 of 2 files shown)


packages/functional/tests/ui/desktop/e2e/coupon.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+  await ProductActions.addProductToCart(page, sampleProduct);
+
+  await page.getByRole('link', { name: 'Cart Items 1' }).click();
+
+  await expect(page.getByRole('heading', { level: 1, name: 'Your cart' })).toBeVisible();
+
+  await page.getByRole('link', { name: 'Proceed to checkout' }).click();
+
+  await expect(page.getByRole('link', { name: 'Coupon/Gift Certificate' })).toBeVisible();
+  await expect(page.locator(totalAmount)).toContainText('$225.00');
```
avattipalli, line 80: Couldn't find a better way to assert text in a span.

chanceaclark: Does `page.getByText('$225.00')` not work?

avattipalli: It's at 3 different locations as it resolved to 3 elements and we might also want to know where we are looking at just so that it's at the right place so, I went with the locator.

chanceaclark: What about `page.getByText('Grand total $225.00')`?

chanceaclark (approved):

PR #782, opened 2024-04-17
feat(cli): copy vscode ts settings

chanceaclark:
## What/Why?
Creates the `.vscode/settings.json` file pointing to the Typescript SDK. This will help with new repos getting better TS support for gql-tada locally.

## Testing
Screenshot 2024-04-17 at 14 33 50

### DIFF
--- a/packages/create-catalyst/src/index.ts
+++ b/packages/create-catalyst/src/index.ts
@@ -65,6 +65,12 @@ const createCommand = program
       .default(getPackageManager())
       .hideHelp(),
   )
+  .addOption(
+    new Option('--code-editor <editor>', 'Your preferred code editor')
+      .choices(['vscode'])
+      .default('vscode')
+      .hideHelp(),
+  )
   .action((opts) => create(opts));
 
 export type CreateCommandOptions = Options<typeof createCommand>[0];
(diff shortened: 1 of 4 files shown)


packages/create-catalyst/src/index.ts
```diff
@@ -65,6 +65,12 @@ const createCommand = program
       .default(getPackageManager())
       .hideHelp(),
   )
+  .addOption(
+    new Option('--code-editor <editor>', 'Package manager to use for the project')
+      .choices(['vscode'])
```
chanceaclark, line 70: The thought here is that maybe we will support other editors TS servers.

packages/create-catalyst/src/index.ts
```diff
@@ -65,6 +65,12 @@ const createCommand = program
       .default(getPackageManager())
       .hideHelp(),
   )
+  .addOption(
+    new Option('--code-editor <editor>', 'Package manager to use for the project')
```
matthewvolk, line 69:
```suggestion
    new Option('--code-editor <editor>', 'The code editor you plan to use to work on this project')
```

Open to rewording, though!

matthewvolk (approved):
Very nice! Originally I was thinking we should prompt and ask if they want the settings file created, but I like this way more, as it's extensible to other editors - and, there's probably never a reason they wouldn't want editor-specific config files created.

packages/create-catalyst/src/index.ts
chanceaclark: Ah good catch!

jorgemoya (approved):

PR #790, opened 2024-04-18
chore(deps): bump most deps to latest

chanceaclark:
## What/Why?
Bumps most deps to the latest versions. I left out storybook since we are planning on removing it, and I left out eslint since it's a larger effort.

## Testing
Regenerated the types locally and linted + typechecked. CI should be the other proof.


apps/core/scripts/generate.cjs
```diff
@@ -42,11 +44,18 @@ const getEndpoint = () => {
 };
 
 const generate = async () => {
-  await generateSchema(getEndpoint(), {
+  await generateSchema({
+    input: getEndpoint(),
     headers: { Authorization: `Bearer ${getToken()}` },
+    output: join(__dirname, '../schema.graphql'),
```
chanceaclark, line 50: This API will use the current working directory if this is `undefined`. When being used in CI, it'll output the files at the root which will cause the next command to error out.

deini (approved):

PR #794, opened 2024-04-19
feat(core): add update customer form

yurytut1993:
## What/Why?
This PR adds Form for Settings tab to update Customer data like First Name, Last Name etc 

update-settings-form

## Testing
Locally


apps/core/app/[locale]/(default)/account/[tab]/page.tsx
```diff
@@ -37,10 +40,10 @@ const tabHeading = async (heading: string, locale: string) => {
 export default async function AccountTabPage({ params: { tab, locale }, searchParams }: Props) {
   switch (tab) {
     case 'orders':
-      return tabHeading(tab, locale);
+      return <TabHeading heading={tab} locale={locale} />;
```
Comment (jorgemoya) [inline apps/core/app/[locale]/(default)/account/[tab]/page.tsx:43]: Do we need to pass locale here? Couldn't get use `getLocale` inside the component instead?

Comment (yurytut1993) [reply apps/core/app/[locale]/(default)/account/[tab]/page.tsx]: @jorgemoya great idea! Refactored all over the accont

core/client/queries/get-customer.ts
```diff
@@ -1,5 +1,7 @@
 import { cache } from 'react';
```
jorgemoya, line 1: Is this being used?

core/client/queries/get-customer.ts
```diff
@@ -43,7 +45,9 @@ const GET_CUSTOMER_QUERY = graphql(`
   }
 `);
 
-export const getCustomer = cache(async (customerId: string) => {
+export const getCustomer = cache(async () => {
```
jorgemoya, line 48: Is this query being used? Can we delete?

core/app/[locale]/(default)/account/[tab]/page-data.ts
```diff
@@ -0,0 +1,111 @@
+import { cache } from 'react';
+
+import { getSessionCustomerId } from '~/auth';
+import { client } from '~/client';
+import { FORM_FIELDS_FRAGMENT } from '~/client/fragments/form-fields';
+import { graphql, VariablesOf } from '~/client/graphql';
+import { revalidate } from '~/client/revalidate-target';
+
+const CustomerSettingsQuery = graphql(
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/account/[tab]/page-data.ts:9]: This is kind of hard to explain but I feel `settings-content` and `addresses-content` should both have a gql fragment of the info they require, and this query should just be a recompilation of both fragments, and you simply pass data to both components and if typed correctly they should just have type of the fragment data.

Similar to: https://github.com/bigcommerce/catalyst/blob/main/core/app/%5Blocale%5D/(default)/login/page.tsx#L20

core/app/[locale]/(default)/account/[tab]/_components/update-settings-form/fields/text.tsx
```diff
@@ -0,0 +1,54 @@
...
+  defaultValue?: string;
+  isRequired?: boolean;
+  isValid?: boolean;
+  label: string;
+  name: string;
+  onChange: (e: ChangeEvent<HTMLInputElement>) => void;
+  type?: 'text' | 'email';
+}
+
+export const Text = ({
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/account/[tab]/_components/update-settings-form/fields/text.tsx:17]: This component could have it's own fragment as well.

core/app/[locale]/(default)/account/[tab]/page-data.ts
Comment (jorgemoya) [reply core/app/[locale]/(default)/account/[tab]/page-data.ts:9]: Sounds good to me

bc-alexsaiannyi (approved):
Looks 👍

jorgemoya (approved):

PR #798, opened 2024-04-22
chore: update tailwindFunctions to use correct utility name

chanceaclark:
## What/Why?
Remaps the `tailwindFunctions` class name utility from `cs` (old) to `cn` (new). Looks like it was missed in https://github.com/bigcommerce/catalyst/pull/457

## Testing
See the lint task


matthewvolk (approved):
Good catch!

deini (approved):

PR #806, opened 2024-04-22
feat(cli): add option to include functional test suite

chanceaclark:
## What/Why?
Adds the ability to include the functional test suite when creating a Catalyst repo from the CLI. We are about to migrate the tests over to the core app which this PR adds some preparation for it from the CLI perspective.

> [!WARNING]
> Just note that the functional test suite won't actually be included until we port it over. This is prep work for it actually being added.

## Testing
Screenshot 2024-04-23 at 11 27 26

### DIFF
--- a/packages/create-catalyst/src/utils/clone-catalyst.ts
+++ b/packages/create-catalyst/src/utils/clone-catalyst.ts
@@ -1,4 +1,5 @@
 import chalk from 'chalk';
+import { readFile, writeFile } from 'fs/promises';
 import { copySync, ensureDir, readJsonSync, removeSync, writeJsonSync } from 'fs-extra/esm';
 import { downloadTemplate } from 'giget';
 import merge from 'lodash.merge';
@@ -9,11 +10,13 @@ import { spinner } from './spinner';
 
 export const cloneCatalyst = async ({
   codeEditor,
+  includeFunctionalTests,
   projectDir,
   projectName,
   ghRef = 'main',
 }: {
   codeEditor: string;
+  includeFunctionalTests: boolean;
   projectDir: string;
   projectName: string;
   ghRef?: string;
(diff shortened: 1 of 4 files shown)


packages/create-catalyst/src/utils/clone-catalyst.ts
```diff
@@ -123,7 +132,24 @@ export const cloneCatalyst = async ({
 
   tsConfigJson.compilerOptions.paths['@bigcommerce/components/*'] = ['./components/ui/*'];
 
+  if (!includeFunctionalTests) {
+    tsConfigJson.include = tsConfigJson.include?.filter(
+      (include) => !['tests/**/*', 'playwright.config.ts'].includes(include),
```
chanceaclark, line 137: Waiting to chat as a team to figure out where the tests are going to live. `tests/**/*` is subject to change.

chanceaclark (commented):
I still need to see if there are any other TSConfig things that might need to changed, but that will have to wait till we have a small example of the tests running in `apps/core`.

jorgemoya (commented):
Looks right to me.

packages/create-catalyst/src/utils/clone-catalyst.ts
```diff
@@ -123,7 +132,24 @@ export const cloneCatalyst = async ({
...
+    );
+  }
+
   writeJsonSync(join(projectDir, 'tsconfig.json'), tsConfigJson, { spaces: 2 });
 
+  if (!includeFunctionalTests) {
+    const eslint = (await readFile(join(projectDir, '.eslintrc.cjs'), { encoding: 'utf-8' }))
+      .replace(/['"]\/playwright-report\/\*\*['"],?/, '')
+      .replace(/['"]\/test-results\/\*\*['"],?/, '');
```
chanceaclark, line 146: Using regex instead as it could be multi-lined. Pretty sure `eslint --fix` will take care of the formatting. If not we can always revisit it and do some cleanup here.

chanceaclark: Here were the regex test cases I used:
Screenshot 2024-04-23 at 11 28 52

deini (approved):

PR #812, opened 2024-04-24
refactor(functional): swap test runner to macos

chanceaclark:
## What/Why?
Since catalyst is OSS, we can start using the macos GitHub hosted runner without consuming our account credits. The biggest blocker for our visual regression tests was that we couldn't update the snapshots without running the test suite in a ubuntu docker container. This now allows us to update the snapshots locally.

## Testing
Ran locally twice, once to generate the snapshots and again to confirm the correct snapshot. Running in CI also confirms that the snapshots are valid as well.

Closes #807


chanceaclark:
Rebasing was easier by reseting the branch and redoing everything. ♻️ would be nice

deini (approved):

avattipalli (approved):

PR #816, opened 2024-04-25
feat(functional): Move functional tests to apps/core

avattipalli:
## What/Why?
As part of making functional tests optionally available to our users, we are moving our tests to `apps/core` so that they are now available through CLI setup.

## Testing
Tests should run as part of this PR. Verified it all locally.


avattipalli:
Successful test runs on CI:

[Playwright UI Tests](https://github.com/bigcommerce/catalyst/actions/runs/8835456260)
[Playwright Visual Regression Tests](https://github.com/bigcommerce/catalyst/actions/runs/8835449463)

apps/core/package.json
```diff
@@ -60,6 +60,9 @@
     "prettier": "^3.2.5",
     "prettier-plugin-tailwindcss": "^0.5.14",
     "tailwindcss": "^3.4.3",
-    "typescript": "^5.4.5"
+    "typescript": "^5.4.5",
+    "@faker-js/faker": "^8.4.1",
+    "@playwright/test": "^1.43.1",
+    "dotenv": "^16.4.5"
```
chanceaclark, line 66: These should to be sorted alphabetically.

apps/core/tests/playwright.config.js
chanceaclark, line 1: Could we actually move this to the root of `apps/core` as we may have a script in `package.json` to run these?

chanceaclark (commented):
Two minor comments, otherwise looks good. I'll pull this down later to test with the CLI before I give a ✅

apps/core/tests/playwright.config.js
avattipalli: Yeah, I was in the same dilemma on where to keep this. I'll do that.

apps/core/tests/.env.example
```diff
@@ -0,0 +1,2 @@
+# Environment variables for tests
+PLAYWRIGHT_TEST_BASE_URL='https://catalyst-demo.site'
```
chanceaclark, line 2: Now that I think about it, let's add remove this for now as we probably need to add [some documentation](https://github.com/bigcommerce/catalyst/blob/main/docs/environment-variables.md) around it.

apps/core/playwright.config.js
chanceaclark, line 1: 🍹 Convert this to a `.ts` file and add it in the `tsconfig.json` `includes` array.

apps/core/tests/.env.example
avattipalli: 👍 I have it documented in `docs/monorepo.md` but this might also be another placeholder to add it as it's an environment variable.

PR #830, opened 2024-04-29
refactor(core): cart page fragment colocation

deini:
## What/Why?
Continuing cart page / checkout fragment colocation.

Follow-up from https://github.com/bigcommerce/catalyst/pull/818


apps/core/app/[locale]/(default)/cart/_components/shipping-estimator/get-shipping-countries.ts
```diff
@@ -2,7 +2,8 @@ import { getCountries } from '~/client/management/get-countries';
 import { getShippingZones } from '~/client/management/get-shipping-zones';
 
 export const getShippingCountries = async () => {
-  const shippingZones = await getShippingZones();
+  const [shippingZones, allCountries] = await Promise.all([getShippingZones(), getCountries()]);
```
Comment (chanceaclark) [inline apps/core/app/[locale]/(default)/cart/_components/shipping-estimator/get-shipping-countries.ts:5]: Am I missing something, or should these not be invoked 🤔 
```suggestion
  const [shippingZones, allCountries] = await Promise.all([getShippingZones, getCountries]);
```

chanceaclark (approved):
I might be going crazy here, but other than my comment, looks good.

PR #831, opened 2024-04-29
chore: standardize actions

jorgemoya:
## What/Why?
Standardize actions by making sure actions have the same elementsd.

## Testing
Locally.

### DIFF
--- a/.changeset/nasty-crabs-do.md
+++ b/.changeset/nasty-crabs-do.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+chore: standardize actions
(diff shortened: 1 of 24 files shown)


apps/core/app/[locale]/(default)/cart/_queries/get-shipping-countries.ts
```diff
@@ -3,6 +3,7 @@ import { getShippingZones } from '~/client/management/get-shipping-zones';
```
Comment (jorgemoya) [inline apps/core/app/[locale]/(default)/cart/_queries/get-shipping-countries.ts:3]: I moved these files because they are not really `actions`, but mostly functions that call queries, not sure if this is the best approach.

chanceaclark (approved):

PR #839, opened 2024-05-01
chore(cli): remove cloning logic for components

jorgemoya:
## What/Why?
Since we no longer have `/packages/components`, we don't need to clone it.

## Testing
TBD

### DIFF
--- a/packages/create-catalyst/src/utils/clone-catalyst.ts
+++ b/packages/create-catalyst/src/utils/clone-catalyst.ts
@@ -65,9 +65,6 @@ export const cloneCatalyst = async ({
       }),
     );
 
-  delete packageJson.private;
-  delete packageJson.exports;
-  delete packageJson.sideEffects;
   delete packageJson.dependencies['@bigcommerce/catalyst-client'];
   delete packageJson.devDependencies['@bigcommerce/eslint-config-catalyst'];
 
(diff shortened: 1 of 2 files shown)


packages/create-catalyst/src/utils/clone-catalyst.ts
```diff
@@ -103,42 +78,29 @@ export const cloneCatalyst = async ({
...
-
   if (!includeFunctionalTests) {
+    const tsConfigJson = z
+      .object({
+        compilerOptions: z
+          .object({
+            declaration: z.boolean().optional(),
+            declarationMap: z.boolean().optional(),
+          })
+          .passthrough(),
```
chanceaclark, line 89: I think we should be able to remove this bit now (and the `delete` statements below).

packages/create-catalyst/src/utils/clone-catalyst.ts
```diff
@@ -103,42 +78,29 @@ export const cloneCatalyst = async ({
...
+        include: z.array(z.string()).optional(),
+      })
+      .passthrough()
+      .parse(merge({}, readJsonSync(join(projectDir, 'tsconfig.json'))));
```
chanceaclark, line 93: 🍹 I wonder if we still need to merge these now? Could we just read the file then parse it?

packages/create-catalyst/src/utils/clone-catalyst.ts
```diff
@@ -71,29 +56,19 @@ export const cloneCatalyst = async ({
...
-      ),
+      merge({}, readJsonSync(join(projectDir, 'package.json')), {
+        name: projectName,
+        description: '',
+      }),
     );
 
   delete packageJson.private;
   delete packageJson.exports;
   delete packageJson.sideEffects;
```
chanceaclark, line 70: I would audit these, in case we can remove more.

matthewvolk (approved):
🙌

packages/create-catalyst/src/utils/clone-catalyst.ts
jorgemoya: Makes sense!

packages/create-catalyst/src/utils/clone-catalyst.ts
jorgemoya: Do these even get added? Since I removed above, I think I can just remove these, no?

jorgemoya: What I don't get is why we added them and then deleted.

chanceaclark (approved):

PR #840, opened 2024-05-01
refactor(core): product details fragment colocation

deini:
## What/Why?
Progress on PDP fragment colocation, this PR now includes colocation on the product `<Details>` component.

### DIFF
--- a/apps/core/app/api/product/[id]/route.ts
+++ b/apps/core/app/api/product/[id]/route.ts
@@ -1,8 +1,33 @@
+/*
+ * This route handler is exclusively for use by the product sheet component.
+ */
 import { NextRequest, NextResponse } from 'next/server';
 
-import { getProduct } from '~/client/queries/get-product';
+import { client } from '~/client';
+import { graphql } from '~/client/graphql';
+import { ProductSheetContentFragment } from '~/components/product-sheet/fragment';
+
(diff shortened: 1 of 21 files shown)


apps/core/app/api/product/[id]/route.ts
```diff
@@ -1,8 +1,32 @@
...
+        }
+      }
+    }
+  `,
+  [ProductSheetContentFragment],
+);
 
 export const GET = async (request: NextRequest, { params }: { params: { id: string } }) => {
+  if (request.headers.get('x-catalyst-product-sheet') !== 'true') {
+    return NextResponse.json('Endpoint only available for product-sheet component.', {
```
Comment (chanceaclark) [inline apps/core/app/api/product/[id]/route.ts:25]: 🍹 Do you think we should normalize this error as the same as what @jorgemoya has been doing with server actions?, e.g. `{ status: 'error', message: '...', }`

chanceaclark (approved):

apps/core/app/api/product/[id]/route.ts
Comment (deini) [reply apps/core/app/api/product/[id]/route.ts]: Added in https://github.com/bigcommerce/catalyst/compare/bbe1ff9f93627d74585a2ac3d40ecc226ead6c1e..3d80524391e2c26916714c6ce4ca398bd61ba57a

PR #848, opened 2024-05-02
refactor(core): blogpost fragment colocation

jorgemoya:
## What/Why?
Fragment colocation in blog post page.

## Testing
Locally.


apps/core/app/[locale]/(default)/blog/[blogId]/_components/sharing-links.tsx
```diff
@@ -3,22 +3,32 @@
...
+      }
+      ...SEOFragment
+    }
+  `,
+  [SEOFragment],
+);
+
+interface Props {
+  blogPost: FragmentOf<typeof SharingLinksFragment>;
   vanityUrl?: string;
```
Comment (jorgemoya) [inline apps/core/app/[locale]/(default)/blog/[blogId]/_components/sharing-links.tsx:26]: I'm curious what is the best approach with this field. [Here](https://github.com/bigcommerce/catalyst/pull/848/files#diff-e2a5797679e0fca50a55870caa3f97d1bde21078f613ddf55a3722951627fa02R44) we return as the result of `getBlogPost` but it's not a field in a BlogPost fragment, it is a setting in Blog.

deini (approved):

Issue #849, opened 2024-05-02
suggestion for cache tags


nowlena:
Not an issue but a recommendation so that there are no "magic" strings.

Can we move to using a const for all cache tags?

This will allow for autocompletion and consistency when setting cache tags and when using `revalidateTag()` to purge them.

```ts
// define our tags
export const TAGS = {
  cart: 'cart',
  checkout: 'checkout',
} as const;

// use anytime we set a cache tag
const response = await client.fetch({
  document: GET_CART_QUERY,
  fetchOptions: {
    next: {
      tags: [TAGS.cart],
    },
  },
});

// use anytime we purge via a tag
revalidateTag(TAGS.cart);
```

Thanks all, appreciate all the work and effort here!

deini:
Completed by https://github.com/bigcommerce/catalyst/pull/850

PR #850, opened 2024-05-02
feat(core): Add cache tags as constant

nowlena:
## What/Why?
Add a "tags" constant instead of using raw strings for cache tags per issue #849


deini (commented):
Thank you for this 🙌

There is one small import order lint error can you please run
```
pnpm run lint -- --fix
```

deini (approved):
Thanks for your contribution!

PR #852, opened 2024-05-05
don't show featured products section if none exist (on not-found template)

electricenjindevops:
## What/Why?
The not found template doesn't account for stores that do not have any featured products, which is a dev error. 

If featured products do not exist, not-found template still shows the section and the h3 headline. this commit fixes that with a length check / fallback of return null.

The git file changes may show a big block replacement, but the only changes are the if statement above the section open and else condition below the section close.

## Testing

I tested by running two different Catalyst stores in dev & build, one on a store with no featured products, and one with featured products. I also had tested using a single store with featured products toggled on/off too.

### DIFF
--- a/.changeset/perfect-seas-breathe.md
+++ b/.changeset/perfect-seas-breathe.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Conditionally show featuredProducts on 404 page.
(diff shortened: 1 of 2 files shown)


electricenjindevops:
- William @ Electric Enjin

chanceaclark (approved):
LGTM! Thanks for the contribution!

jorgemoya (approved):

chanceaclark:
Lint task is not going to work as it's being ran on a fork. We don't share our secrets with forked branches. Confirmed it worked locally though.

Issue #853, opened 2024-05-05
Catalyst vs Stencil - Different behavior for Blog "isVisibleInNavigation" setting


electricenjindevops:
We really appreciate the help making Catalyst better. Every issue helps!

**Describe the bug**
The BC admin UI has an option to show or hide Blog in navigation, if you go by the setting 'isVisibleInNavigation' from the blog data. Catalyst uses this settings as enable/disable blog entirely. However, on Stencil Cornerstone, it only hides the blog from the navigation but still allows the /blog page to be seen, and allows individual articles to be seen.

This can cause confusion because of how the Stencil Cornerstone themes interpret the setting (navigation related only) versus how Catalyst interprets the setting (disable entirely). 

**To Reproduce**
1. Toggle Blog Visibility on/off in admin UI: /manage/content/blog
2. Look at Navigation on Stencil Cornerstone, then manually navigate to the /blog/ main page.
3. Look at Navigation on a Catalyst build, then manually navigate to the /blog/ main page.
4. Notice how default Cornerstone vs default Catalyst use the blog visibility setting differently.

**Expected behavior**
The property name is 'isVisibleInNavigation' and is used this way exclusively in Stencil Cornerstone. It was expected for it to be treated the same way in Catalyst. For example, if blog visibility is disabled then the expected behavior is for the navigation links to be removed, but for /blog/ itself to be enabled in Catalyst. Or alternatively, it could be expected that Catalyst is correct and it should be completely disabled in Stencil Cornerstone to match the Catalyst behavior.

**Additional context**
I do not know the philosophy behind Catalyst vs Stencil and normalization. 
If there is no goal to have exact matching behaviors between settings, then perhaps a tooltip could clarify this on BigCommerce UI so people understand the difference in default Catalyst vs Stencil stores, or could be included in dev documentation for Catalyst on its own which may be enough.

PR #854, opened 2024-05-06
fix(core): respect isVisibleInNavigation for blog pages

chanceaclark:
Fixes: #853

## What/Why?
Doesn't return 404 if the blog is disabled from navigation. This is to preserve functionality from Stencil -> Catalyst.

## Testing
Screenshot 2024-05-06 at 11 10 33

### DIFF
--- a/.changeset/early-bananas-wonder.md
+++ b/.changeset/early-bananas-wonder.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Respect isVisibleInNavigation for blog pages
--- a/apps/core/app/[locale]/(default)/blog/[blogId]/page-data.ts
+++ b/apps/core/app/[locale]/(default)/blog/[blogId]/page-data.ts
@@ -12,7 +12,6 @@ const PageQuery = graphql(
       site {
         content {
           blog {
-            isVisibleInNavigation
             post(entityId: $entityId) {
               author
               htmlBody
--- a/apps/core/app/[locale]/(default)/blog/[blogId]/page.tsx
+++ b/apps/core/app/[locale]/(default)/blog/[blogId]/page.tsx
@@ -40,9 +40,8 @@ export default async function BlogPostPage({ params: { blogId, locale } }: Props
 
   const data = await getPageData({ entityId: Number(blogId) });
   const blogPost = data?.content.blog?.post;
-  const isVisibleInNavigation = data?.content.blog?.isVisibleInNavigation;
 
-  if (!blogPost || !isVisibleInNavigation) {
+  if (!blogPost) {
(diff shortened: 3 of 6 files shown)


jorgemoya (approved):
So this prevent 404 but will we look into using `isVisibleInNavigation` in the footer link?

electricenjindevops:
Looks good on our side. From this change, our expectation would be that we could manually remove the blog structure for posts & the main index page if we wanted to complete disable it in a Catalyst build, with the setting now only controlling whether it appears in navigation links or not. 

Another thought we had is that we could do a length check on published blog posts to decide if the main blog page shows a default 404, or if it shows a no articles found display within the blog overview page - but this last part feels like it can be  a dev/client decision and not something that Catalyst would need to include by default.

deini (approved):

chanceaclark:
>Another thought we had is that we could do a length check on published blog posts to decide if the main blog page shows a default 404, or if it shows a no articles found display within the blog overview page - but this last part feels like it can be a dev/client decision and not something that Catalyst would need to include by default.

I was thinking about this as I was working through this fix. A big reason I didn't show an empty page is due to not having designs for an empty blog list so I omitted it for now. If we can work in a feature on BigCommerce's side to actually disable the blog entirely, I think we should implement an empty list UI.

electricenjindevops:
> I was thinking about this as I was working through this fix. A big reason I didn't show an empty page is due to not having designs for an empty blog list so I omitted it for now. If we can work in a feature on BigCommerce's side to actually disable the blog entirely, I think we should implement an empty list UI.

Yeah that makes sense. I think there are pros & cons for both ways. For example, if pages link to /blog/ and that is a full 404 page, now there is a SEO score "penalty" for linking to a missing page whereas the empty list UI version would not penalize for linking to a 404. However, the empty list UI could possibly lead to an SEO penalty for having a page with no useful content, causing the average quality score for backlinks to go down.

Obviously there is little reason to have either an empty blog page, or a 404 blog page, and this will most likely only ever affect a site on launch that hasn't published any blogs yet. Personally though I am a fan of empty list UI because that leaves the decision up to the developers and clients - at least until there is a setting to completely disable. 

I have seen cases where store themes/headless builds will go live without the content ready, with the client adding content later - in other blogging systems that have dev/staging/production publish statuses, it's also possible for a preview build to have no blogs, but a staging or production build to have published articles. The empty list UI fits this need better because when a full 404 is used, that decision may cause the blog to not get indexed initially / not re-indexed as quickly by search engines.

chanceaclark:
That's some great insight on SEO value for it. Overall it seems like it's a better merchant experience than the 404 page, so I'll go ahead and put a ticket in to get some designs for an empty blog page.

PR #859, opened 2024-05-06
refactor(core): brand page query relocation

jorgemoya:
## What/Why?
Following the pattern of fragment colocation, relocated query closer to page.

## Testing
Brand page still renders.


deini (approved):

apps/core/app/[locale]/(default)/(faceted)/brand/[slug]/static/page.tsx
```diff
@@ -1,10 +1,41 @@
...
+        }
+      }
+    }
+  }
+`);
+
+type BrandsPageQueryVariables = VariablesOf<typeof BrandsQuery>;
+
+const getBrands = cache(async (variables: BrandsPageQueryVariables = {}) => {
+  const response = await client.fetch({
```
Comment (jorgemoya) [inline apps/core/app/[locale]/(default)/(faceted)/brand/[slug]/static/page.tsx:30]: Do I need to cache this since it's only used during static gen once? 🤔

chanceaclark (approved):

PR #861, opened 2024-05-06
refactor(core): category page fragment colocation

jorgemoya:
## What/Why?

- Fragment collocation in category page
- Moved breadcrumbs component to global folder since it's reused in two pages

## Testing
Category page still works.


chanceaclark (approved):

apps/core/app/[locale]/(default)/(faceted)/category/[slug]/page-data.ts
```diff
@@ -0,0 +1,39 @@
...
+import { client } from '~/client';
+import { graphql, VariablesOf } from '~/client/graphql';
+import { revalidate } from '~/client/revalidate-target';
+import { BreadcrumbsFragment } from '~/components/breadcrumbs';
+
+import { CategoryTreeFragment } from './_components/sub-categories';
+
+const PageQuery = graphql(
+  `
+    query PageQuery($categoryId: Int!) {
```
Comment (deini) [inline apps/core/app/[locale]/(default)/(faceted)/category/[slug]/page-data.ts:13]: Can we add a more specific query name? If we use a generic name it will be harder to track with our gql logger.

deini (approved):

PR #862, opened 2024-05-06
feat(core): add sample rate for Vercel analytics and speed insights

chanceaclark:
## What/Why?
Adds an opinionated default on the sample rates for both Vercel Speed Insights and Vercel Analytics. Most of the time you don't need to capture 100%, especially on a heavily trafficked site.

For Vercel Analytics, there was a little lifting that was needed to be done. We don't necessarily want to send events 25% of the time, but rather 25% of the users get analytics tracked per session.

## Testing
Screenshot 2024-05-06 at 15 56 04
Screenshot 2024-05-06 at 15 56 16

### DIFF
--- a/.changeset/six-birds-accept.md
+++ b/.changeset/six-birds-accept.md
@@ -1,0 +1,9 @@
+---
+"@bigcommerce/catalyst-core": minor
+---
+
+Added a default sample rate of 25% for both Vercel Speed Insights and Vercel Analytics. This provides a good jumping off point for merchants to tweak their sample rates if they choose to.
+
+In order to tweak the sample rate, you can add the following environment variables to your Vercel project:
+- `NEXT_PUBLIC_ANALYTICS_SAMPLE_RATE` for Vercel Analytics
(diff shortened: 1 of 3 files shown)


.changeset/six-birds-accept.md
```diff
@@ -0,0 +1,9 @@
+---
+"@bigcommerce/catalyst-core": minor
+---
+
+Added a default sample rate of 25% for both Vercel Speed Insights and Vercel Analytics. This provides a good jumping off point for merchants to tweak their sample rates if they choose to.
+
+In order to tweak the sample rate, you can add the following environment variables to your Vercel project:
+- `ANALYTICS_SAMPLE_RATE` for Vercel Analytics
+- `SPEED_INSIGHTS_SAMPLE_RATE` for Vercel Speed Insights
```
jorgemoya, line 9: Shouldn't these be prefixed with `NEXT_PUBLIC` if used within a client component?

chanceaclark: 🤦‍♂️ yup

chanceaclark: 🤦‍♂️ yup

PR #869, opened 2024-05-07
test(core): update visual regression breadcrumbs test

chanceaclark:
## What/Why?
Updates the breadcrumbs test after @jorgemoya updated the styles in #861 

## Testing
See CI


deini (approved):

jorgemoya:
Thank you 🙏 . Worst case is I remembered last night I needed to fix these, but this morning I forgot.

PR #878, opened 2024-05-08
fix(core): fix accessibility issues on plp and pdp

christensenep:
## What/Why?
Small accessibility fixes to our PLP and PDP.

## Testing
Tested locally and with a preview vercel deployment.

### DIFF
--- a/core/components/product-card/compare.tsx
+++ b/core/components/product-card/compare.tsx
@@ -20,7 +20,7 @@ export const Compare = ({
   } | null;
   productName: string;
 }) => {
-  const checkboxId = useId();
+  const labelId = useId();
   const t = useTranslations('Product.ProductSheet');
   const [checkedState, setCheckedState] = useState(false);
   const { products, setProducts } = useCompareProductsContext();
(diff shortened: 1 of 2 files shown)


core/components/product-card/compare.tsx
```diff
@@ -21,6 +21,8 @@ export const Compare = ({
   productName: string;
 }) => {
   const checkboxId = useId();
+  const labelId = `${checkboxId}-label`;
```
chanceaclark, line 24: After looking at https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role, I think we can remove the `htmlFor` + `checkboxId` as we should be using the pattern you are proposing here.

deini (approved):

core/components/product-card/compare.tsx
chanceaclark: Let's just rename `checkboxId` to `labelId` instead as `useId` is guaranteed to be unique.

christensenep: Good catch thanks.

jorgemoya (approved):

PR #898, opened 2024-05-13
fix(core): adjust Pricing component localization in order to be used as client and server component

BC-krasnoshapka:
## What/Why?
Fixes error in quick search.

## Testing
Locally.
Pricing component works in various places like PDP, PLP, compare and quick search.

image


jorgemoya (approved):

avattipalli (approved):
Verified the change on feature branch.

PR #937, opened 2024-05-20
fix(core): show correct status and messaging for add to cart button

jorgemoya:
## What/Why?

- Create a shared component for the add to cart button.
- Check for product availability status to show correct text (unavailable, preorder).
- Use `isInStock` from `inventory` to check if a product/variant is available to be added to cart (when available).
- Show correct messaging in add to cart buttons when a product is not available/oos.
- Update translations object.

## Testing
Locally the button displays the correct messaging and disabled status.


avattipalli (commented):
Can we add a test to verify this?

jorgemoya:
> Can we add a test to verify this?

Good idea, looking into this.

core/components/add-to-cart-button/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  return (
+    <Button
+      className={className}
+      disabled={isProductDisabled}
+      loading={loading}
+      loadingText={t('processing')}
+      type="submit"
+    >
+      {showCartIcon && <ShoppingCart className="mr-2" />}
```
chanceaclark, line 51: Instead of having a boolean prop to control this, could we just use `children` instead?

jorgemoya: Great suggestion, made the change.

chanceaclark (approved):

PR #942, opened 2024-05-22
fix(core): make Select scrollable with popper positioning

jorgemoya:
## What/Why?
This fixes a side effect of #757, by adding a fixed heigh the Select as popper works as it should, allowing scrolling and properly rendering under the trigger.

Screenshot 2024-05-22 at 1 34 51 PM

## Testing
Locally.

### DIFF
--- a/.changeset/cuddly-dolphins-sparkle.md
+++ b/.changeset/cuddly-dolphins-sparkle.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Make select scrollable with popover functionality.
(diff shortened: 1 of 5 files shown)


bc-alexsaiannyi:
Hey @jorgemoya. It looks good but just curious if speed of scrolling is something that can be adjust? Doesn't it look too fast?


https://github.com/bigcommerce/catalyst/assets/67792608/a21d241c-6eb3-4f1e-8216-d9384cf0ac51

jorgemoya:
> Hey @jorgemoya. It looks good but just curious if speed of scrolling is something that can be adjust? Doesn't it look too fast?
> 
>  Screen.Recording.2024-05-23.at.14.11.26.mov

Good callout. I looked and it appears it's not configurable, so this will need to make do for now.

PR #943, opened 2024-05-23
Pass customer ID when requesting checkout URL

bookernath:
## What/Why?
Customer ID was not being passed to the request to get a checkout URL. As a result, there was a server error when checking out while logged in. This fixes that.

In the future, this request will return a customer-specific checkout URL, but for now it returns a guest checkout URL. This approach should automatically work in the future when the underlying URL is updated.

## Testing
Locally tested both logged-in checkout and guest checkout.


bookernath:
Tested against the preview URL also to confirm it works ✅

jorgemoya (approved):

PR #958, opened 2024-05-30
Create wishlist mutation

yurytut1993:
## What/Why?
This pr adds create wishlist mutation for customer account

## Testing
Locally

### DIFF
--- a/.changeset/heavy-otters-tell.md
+++ b/.changeset/heavy-otters-tell.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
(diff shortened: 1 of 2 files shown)


jorgemoya (commented):
Do we need to fetch all wishlist data on the mutate or could we simply get back `entityId` to make sure it was created?

yurytut1993:
> Do we need to fetch all wishlist data on the mutate or could we simply get back `entityId` to make sure it was created?

@jorgemoya we need to fetch it to add to the state and to display in wish lists list

jorgemoya (approved):

PR #976, opened 2024-06-03
feat(functional): Add test for shipping estimation recalculation

avattipalli:
## What/Why?
Add a test to verify that shipping information is persisted when cart items are modified but shipping estimates need to be recalculated.

## Testing
Verified the test locally. CI runs with the new test.

### DIFF
--- a/core/tests/ui/desktop/e2e/checkout.spec.ts
+++ b/core/tests/ui/desktop/e2e/checkout.spec.ts
@@ -70,9 +70,6 @@ test('Complete checkout as a logged in shopper', async ({ page }) => {
   await page.getByRole('heading', { level: 1, name: 'Your cart' }).click();
   await page.getByRole('button', { name: 'Proceed to checkout' }).click();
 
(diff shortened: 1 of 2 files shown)


core/tests/ui/desktop/e2e/checkout.spec.ts
```diff
@@ -70,9 +70,6 @@ test('Complete checkout as a logged in shopper', async ({ page }) => {
   await page.getByRole('heading', { level: 1, name: 'Your cart' }).click();
   await page.getByRole('button', { name: 'Proceed to checkout' }).click();
 
-  await page.getByLabel('Email').fill(faker.internet.email());
```
avattipalli, line 73: This test was failing because with a logged in customer, we are able to fill the email on the checkout page and so, the test was just waiting to find and fill the email. 

As the test is already failing, I am fixing it as part of this same PR as the change is small.

deini (approved):

jorgemoya (commented):
Thanks for adding this!

PR #979, opened 2024-06-06
fix(core): user is not redirected to Login page after requesting password reset link

bc-yevhenii-buliuk:
## What/Why?
This PR fixes the flow of redirecting users to the Login Page after changing the password

## Testing
locally

**before:**

https://github.com/bigcommerce/catalyst/assets/82589781/636fcd57-79ff-42a3-95b4-f7abbe559251

https://github.com/bigcommerce/catalyst/assets/82589781/8212c532-90e2-42bf-8e09-97735e668c06

**after:**

https://github.com/bigcommerce/catalyst/assets/82589781/f46f3a41-1a8f-4b0f-a907-2a0b384ac425

https://github.com/bigcommerce/catalyst/assets/82589781/475c7df3-ec6e-484d-9189-c67b742e3678

### DIFF
--- a/core/messages/en.json
+++ b/core/messages/en.json
@@ -259,7 +259,7 @@
     "Register": {
         "heading": "New account",
         "submit": "Create account",
-        "submitting": "Creating account...", 
+        "submitting": "Creating account...",
         "recaptchaText": "Pass ReCAPTCHA check",
         "successMessage": "Dear {firstName} {lastName}, your account was successfully created. Redirecting to account...",
         "stateProvincePrefix": "Choose state or province",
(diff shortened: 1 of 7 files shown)


bc-yevhenii-buliuk:
To redirect to the Login, I used the context that was created for the Account, so as not to create a separate context for the Login. Guys, do you think this approach is acceptable in this case? If so, should we move this context out of the Account directory and rename it to something more generic to use for both Account and Login? thanks!

chanceaclark (approved):

chanceaclark:
Some functional tests would be good with this PR as well 🙏

yurytut1993 (approved):

core/messages/en.json
```diff
@@ -286,7 +286,8 @@
       "confirmPasswordValidationMessage": "Entered passwords are mismatched. Please try again.",
       "newPasswordValidationMessage": "New password must be different from the current password or/and match confirm password.",
       "notEmptyMessage": "Field should not be empty",
-      "successMessage": "Password has been updated successfully!"
+      "successMessage": "Password has been updated successfully!",
+      "confirmChangePassword": "Your password has been successfully updated."
```
bc-alexsaiannyi, line 290: Do we need both translations?

bc-yevhenii-buliuk: @bc-alexsaiannyi I've double-checked this message in Figma and fixed it

bc-alexsaiannyi (approved):
lgtm

core/messages/en.json
bc-alexsaiannyi: thanks!

PR #980, opened 2024-06-06
fix(core): fix redirects causing middleware to break

chanceaclark:


core/middlewares/with-routes.ts
```diff
@@ -165,14 +162,16 @@ export const withRoutes: MiddlewareFactory = () => {
...
       return response;
     }
 
-    if (!locale) {
-      // Try to get locale detected by next-intl
-      locale = response.cookies.get('NEXT_LOCALE')?.value || '';
+    if (request.headers.has('x-action-redirect')) {
+      const redirectUrl = new URL(request.headers.get('x-action-redirect') ?? '', request.url);
+
+      return NextResponse.rewrite(new URL(`/${locale}${redirectUrl.pathname}`, request.url));
```
BC-krasnoshapka, line 174: @chanceaclark what's the case for this rewrite? I've just noticed that it handles prefix based locales strategy `domain.com/en` only. i18n supports subdomain based `en.domain.com` also. Should it be handles here?

chanceaclark: I'll add more info once I move this PR from Draft -> Open, but `next@14.2.0+` changed redirection behavior inside server actions causing things like login to break. I am not finished with this PR and this may not even be the fix for this issue, but if it is I still have other edge cases like external and absolute/relative redirects to test.

chanceaclark:
Closing this PR as I have created an issue with the Next.js team https://github.com/vercel/next.js/issues/66837

PR #994, opened 2024-06-10
feat(client): add channelId as an optional param

jorgemoya:
## What/Why?

- Add optional channelId param to client to allow fetching from multiple channels with the same client.
- Remove non channel-aware endpoint.
- Enforce channelId being used.

## Testing
TBD

### DIFF
--- a/packages/client/src/client.ts
+++ b/packages/client/src/client.ts
@@ -155,15 +155,7 @@ class Client<FetcherRequestInit extends RequestInit = RequestInit> {
   }
 
   private getEndpoint(channelId?: string) {
-    if (channelId && channelId !== '1') {
-      return `https://store-${this.config.storeHash}-${channelId}.${graphqlApiDomain}/graphql`;
-    }
-
-    if (!this.config.channelId || this.config.channelId === '1' || channelId === '1') {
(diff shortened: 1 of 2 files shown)


packages/client/src/client.ts
```diff
@@ -150,8 +154,12 @@ class Client<FetcherRequestInit extends RequestInit = RequestInit> {
     return response.json() as Promise<unknown>;
   }
 
-  private getEndpoint() {
-    if (!this.config.channelId || this.config.channelId === '1') {
+  private getEndpoint(channelId?: string) {
+    if (channelId && channelId !== '1') {
+      return `https://store-${this.config.storeHash}-${channelId}.${graphqlApiDomain}/graphql`;
```
chanceaclark, line 159: We've backfilled all/most default channels with the channel-specific canonical URL, so we should be able to simplify this to `https://store-${this.config.storeHash}-${channelId ?? this.config.channelId}.${graphqlApiDomain}/graphql` and enforce channel ID to be set.

jorgemoya: Good info, wasn't aware of this.

chanceaclark (approved):

PR #995, opened 2024-06-12
feat(core): make fetch channel aware per locale

jorgemoya:
## What/Why?
- Add channel <=> locale config example.
- Make `client.fetch` channel aware by fetching channel id assigned to locale.
- Fetch from default channel if no map is found. 

## Testing
TBD


core/channels.config.ts
```diff
@@ -0,0 +1,24 @@
...
+  fr: DEFAULT_CHANNEL_ID,
+  it: DEFAULT_CHANNEL_ID,
+  ja: DEFAULT_CHANNEL_ID,
+  nl: DEFAULT_CHANNEL_ID,
+  no: DEFAULT_CHANNEL_ID,
+  pl: DEFAULT_CHANNEL_ID,
+  'pt-BR': DEFAULT_CHANNEL_ID,
+  pt: DEFAULT_CHANNEL_ID,
+  sv: DEFAULT_CHANNEL_ID,
+};
```
BC-krasnoshapka, line 24: @jorgemoya does it make sense to combine this with locales from i18n into one config to minimize number of places where they're hardcoded?

jorgemoya: I have updated this object to just include the overrides, if any. I have made some type changes to make sure the keys match any of our configured locales. With these changes I don't think it makes sense to merge together, what do you think?

BC-krasnoshapka: Great idea!

chanceaclark (approved):

jorgemoya:
Not working as expected for static sites.

PR #1002, opened 2024-06-14
feat(core): cache all kv adapters with in-memory kv

deini:
## What/Why?
This PR adds an in-memory cache for all kv adapters. Should help reduce number of reads to remote KVs and improve performance.

### DIFF
--- a/core/lib/kv/index.ts
+++ b/core/lib/kv/index.ts
@@ -1,11 +1,15 @@
+import { MemoryKvAdapter } from './adapters/memory';
 import { KvAdapter, SetCommandOptions } from './types';
 
 interface Config {
   logger?: boolean;
 }
 
+const memoryKv = new MemoryKvAdapter();
+
 class KV<Adapter extends KvAdapter> implements KvAdapter {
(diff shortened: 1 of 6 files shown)


core/lib/kv/index.ts
```diff
@@ -20,24 +24,42 @@ class KV<Adapter extends KvAdapter> implements KvAdapter {
   }
 
   async get<Data>(key: string) {
-    const kv = await this.getKv();
-    const fullKey = `${this.namespace}_${key}`;
-
-    const value = await kv.get<Data>(fullKey);
```
christensenep, line 26: If I'm reading correctly this removes calls to the individual adapters' `get` methods entirely in favor of `mget`. Are we sure we want that? And if so, should we remove those methods?

chanceaclark (approved):

core/lib/kv/index.ts
deini: Good point, removed unused methods

jorgemoya (approved):

PR #1015, opened 2024-06-18
Add register account UI test and add MSW for mocking server-side API requests

bc-azvierieva:
## What/Why?
POC of adding Playwright tests with mocking GraphQL server-side requests. Mock Service Worker is used as a helper tool.

This is done for testing Catalyst functionality without actually sending data to backend. This might be helpful to check FE side functionality works fine in Catalyst and at the same time avoid creating and removing real test users.

- added `register.spec.ts` file with UI test of register account functionality 
- added `core/client/mocks` folder for storing all mocking related files in one place
- added `core/instrumentation.ts` hook that enables MSW server to intercept API requests we need
- as MSW is not yet working perfectly fine with new Next.js App router, a temporary workaround is added to `core/next.config.js` file to make it work fine; see more info in [issue](https://github.com/mswjs/msw/issues/1644)
- added `core/tests/test-data.ts` file to store test user we use across different tests to avoid code duplication
- added recaptcha bypass to make test pass in CI test run (kudos to Chance👏)

## Testing
Screenshot 2024-06-18 at 17 35 08

### DIFF
--- a/core/instrumentation.ts
+++ b/core/instrumentation.ts
@@ -1,0 +1,6 @@
+export async function register() {
+  if (process.env.NEXT_PUBLIC_API_MOCKING === 'enabled' && process.env.NEXT_RUNTIME === 'nodejs') {
+    const { server } = await import('~/client/mocks/server');
+    server.listen();
+  }
+}
--- a/core/tests/test-data.ts
+++ b/core/tests/test-data.ts
@@ -1,0 +1,13 @@
+import { faker } from '@faker-js/faker';
+
+export const testUser = {
+  emailAddress: process.env.TEST_ACCOUNT_EMAIL || '',
+  password: process.env.TEST_ACCOUNT_PASSWORD || '',
+  streetAddress: faker.location.streetAddress(),
+  firstName: faker.person.firstName(),
+  lastName: faker.person.lastName(),
+  phoneNumber: faker.phone.number(),
+  state: faker.location.state(),
+  city: faker.location.city(),
+  zipCode: '76286',
+};
--- a/core/app/[locale]/(default)/login/register-customer/page.tsx
+++ b/core/app/[locale]/(default)/login/register-customer/page.tsx
@@ -4,6 +4,7 @@ import { getMessages, getTranslations } from 'next-intl/server';
 
 import { auth } from '~/auth';
 import { LocaleType } from '~/i18n';
+import { bypassReCaptcha } from '~/lib/bypass-recaptcha';
 
 import { RegisterCustomerForm } from './_components/register-customer-form';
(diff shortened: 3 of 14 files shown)


core/instrumentation.ts
```diff
@@ -0,0 +1,6 @@
+export async function register() {
+  if (process.env.NEXT_PUBLIC_API_MOCKING === 'enabled' && process.env.NEXT_RUNTIME === 'nodejs') {
+    const { server } = await import('~/client/mocks/server');
+    server.listen();
+  }
+}
```
chanceaclark, line 6: So this doesn't actually work... Went to try this myself and it's not actually running the service worker as we are using the `edge` runtime. We'd also need the `NEXT_PUBLIC_API_MOCKING` environment variable set on the deployment, which I don't think we should do, especially since we run tests on the production deployments (we want users to be able to create accounts to test around with).

I'd also say it'd be alright that we register a bunch of customers in the CP as we already do this for orders right now. We could also pass in a custom provider/domain so we can clean them up eventually: `faker.person.email({ provider: 'mybigcommerce.com' })`

core/tests/test-data.ts
```diff
@@ -0,0 +1,13 @@
...
+  emailAddress: process.env.TEST_ACCOUNT_EMAIL || '',
+  password: process.env.TEST_ACCOUNT_PASSWORD || '',
+  streetAddress: faker.location.streetAddress(),
+  firstName: faker.person.firstName(),
+  lastName: faker.person.lastName(),
+  phoneNumber: faker.phone.number(),
+  state: faker.location.state(),
+  city: faker.location.city(),
+  zipCode: '76286',
+};
```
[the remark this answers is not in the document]
bc-azvierieva: I was trying to avoid code duplication by doing this, so we have everything in one place. But I agree that different tests have different needs in test data, so this way is not universal. I didn't want to use the way I've seen in other tests - to declare a variable for each input at the top of the file, having this code fragment duplicated in different files. 
Inline usage indeed seems to be the best option for now. I'll update code to use this approach. When we'll have more tests with various test data used, we could create a better solution for handling this. 
Thanks for your input! ☀️

core/instrumentation.ts
bc-azvierieva: Ah, good catch @chanceaclark  - so, this means Mock Service Worker is not actually suitable for us. 
Ok, I will close this PR and create a new one only with register customer test. 

> I'd also say it'd be alright that we register a bunch of customers in the CP as we already do this for orders right now. 

We can do that - looks like we don't have much of a choice at the moment. But before doing that, I think it would be nice to define some data clean up strategy - but it is a matter of a separate discussion, I believe.

PR #1017, opened 2024-06-20
Correct translation key for Warranty in Product Details

rlogank:
This pull request addresses an issue with the incorrect translation key used for the warranty section on product pages. The current code utilizes the key `Product.Details`, which results in displaying "Product.Details.warranty" on the product page instead of the correct translation.


chanceaclark (approved):
Appreciate it!

PR #1020, opened 2024-06-24
fix(chore): Fix flappy tests

avattipalli:
## What/Why?
Stablize test suite by fixing flappy tests.

## Testing
Locally and checks run against this PR.

### DIFF
--- a/core/tests/ui/desktop/e2e/account.spec.ts
+++ b/core/tests/ui/desktop/e2e/account.spec.ts
@@ -85,7 +85,7 @@ test('Add and remove new address', async ({ page }) => {
 
   await page.getByRole('textbox', { name: 'Suburb/City Required' }).fill(city);
   await page.getByRole('combobox', { name: 'Choose state or province' }).click();
-  await page.getByLabel(state).click();
+  await page.getByLabel(state, { exact: true }).click();
   await page.getByRole('textbox', { name: 'Zip/Postcode Required' }).fill(zipCode);
 
   await page.getByRole('button', { name: 'Add new address' }).click();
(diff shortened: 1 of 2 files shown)


core/tests/ui/desktop/e2e/account.spec.ts
```diff
@@ -85,7 +85,7 @@ test('Add and remove new address', async ({ page }) => {
 
   await page.getByRole('textbox', { name: 'Suburb/City Required' }).fill(city);
   await page.getByRole('combobox', { name: 'Choose state or province' }).click();
-  await page.getByLabel(state).click();
+  await page.getByLabel(state, { exact: true }).click();
```
avattipalli, line 88: Test fails to identify strict locator when Kansas is the selected state since`Kansas` and `Arkansas` match for this selector.

core/tests/ui/desktop/e2e/checkout.spec.ts
```diff
@@ -3,14 +3,19 @@ import { expect, Page, test } from '@playwright/test';
...
+const actionTimeout = 10000;
 
 async function enterShopperDetails(page: Page) {
-  await page.getByLabel('First Name').fill(testUser);
-  await page.getByLabel('Last Name').fill(faker.person.lastName());
-  await page.getByLabel('Company Name (Optional)').fill('BigCommerce');
-  await page.getByLabel('Phone Number (Optional)').fill(faker.phone.number());
-  await page.getByLabel('Address', { exact: true }).fill(faker.location.buildingNumber());
-  await page.getByLabel('City').fill('Natick');
+  await page.getByLabel('First Name').fill(testUser, { timeout: actionTimeout });
```
avattipalli, line 9: This test has been failing a few times unexpectedly due to either the targeted element is not filled with desired text or the focus moves too quick resulting in mistyped string or partial fill. The only alternative to `fill`, `type` has been deprecated so, I opted to use `timeout` on these specific actions. Also added, `retry` for this one test since it's determined to be flaky until we have a robust fix for this.

jorgemoya (approved):

PR #1031, opened 2024-06-25
feat(core): add dates field for account & addresses forms

bc-alexsaiannyi:
## What/Why?

This PR adds brings `DatesFormField` for Address & Account forms.

## Testing
Locally

Screenshot 2024-06-25 at 17 46 35

Screenshot 2024-06-25 at 17 48 01

### DIFF
--- a/core/components/ui/date-picker/date-picker.tsx
+++ b/core/components/ui/date-picker/date-picker.tsx
@@ -8,7 +8,7 @@ import { Calendar } from '../calendar';
 import { Input, InputIcon, InputProps } from '../input';
 import { Popover, PopoverContent, PopoverTrigger } from '../popover';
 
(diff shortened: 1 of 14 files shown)


core/components/ui/date-picker/date-picker.tsx
```diff
@@ -41,7 +41,7 @@ export const DatePicker = React.forwardRef<React.ElementRef<'div'>, DatePickerPr
           <PopoverTrigger asChild>
             <Input
               placeholder={placeholder}
-              readOnly={true}
+              readOnly={!required}
```
bc-alexsaiannyi, line 44: **Note**:  If the element is `read-only`, then the element's value can not be updated by the user, and does not participate in constraint validation. So I've made this change to validate against `missingValue` when the field is required.

yurytut1993 (approved):
👌

PR #1032, opened 2024-06-25
feat: prefetch product option on hover

deini:
## What/Why?
On product option hover -> prefetch PDP data. This change makes the UX better at the cost of a few more requests.

This change was only required on multiple choice options, are these are the only ones changing the url.

## Testing
Inspect network tab while hovering on product options.

### DIFF
--- a/core/components/product-form/fields/multiple-choice-field/index.tsx
+++ b/core/components/product-form/fields/multiple-choice-field/index.tsx
@@ -19,6 +19,12 @@ interface Props {
   option: FragmentOf<typeof MultipleChoiceFieldFragment>;
 }
 
+interface InteractionOptions {
+  optionId: number;
+  valueId: number;
+  prefetch?: boolean;
+}
+
 export const MultipleChoiceField = ({ option }: Props) => {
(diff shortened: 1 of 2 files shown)


core/components/product-form/fields/multiple-choice-field/index.tsx
```diff
@@ -27,12 +33,29 @@ export const MultipleChoiceField = ({ option }: Props) => {
...
 
     optionSearchParams.set(String(optionId), String(valueId));
 
-    router.replace(`${pathname}?${optionSearchParams.toString()}`, { scroll: false });
+    const newUrl = `${pathname}?${optionSearchParams.toString()}`;
+
+    if (prefetch) {
+      // PrefetchKind enum is not exported
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-expect-error
```
chanceaclark, line 46: 😠

chanceaclark (approved):

jorgemoya (approved):
🙇‍♂️

PR #1034, opened 2024-06-25
feat(core): add store selector page and link in footer

jorgemoya:
## What/Why?

- Add store selector page to change locales.
- Page can be accessed via footer link (temp design) or `/store-selector`

## Known issues
- Two latin americas

## Testing
Go to `/store-selector` and switch locales.

## Screenshot
Screenshot 2024-06-25 at 3 10 43 PM


core/app/[locale]/store-selector/_components/locale-link.tsx
```diff
@@ -0,0 +1,60 @@
...
+};
+
+const getCountryFlagEmoji = (countryCode: string) => {
+  const codePoints = countryCode
+    .toUpperCase()
+    .split('')
+    .map((char) => 127397 + char.charCodeAt(0));
+
+  return String.fromCodePoint(...codePoints);
+};
```
Comment (jorgemoya) [inline core/app/[locale]/store-selector/_components/locale-link.tsx:39]: I wonder if I should just add the emoji flags instead of using this. 🤔

core/components/footer/footer.tsx
```diff
@@ -69,10 +70,21 @@ export const Footer = ({ data }: Props) => {
...
+          {data.settings && <Copyright data={data.settings} />}
+        </div>
+      </FooterSection>
+
+      <FooterSection className="hidden justify-between gap-8 py-6 lg:flex">
         {data.settings && <Copyright data={data.settings} />}
+        <div className="flex gap-8">
+          <Locale />
+          <PaymentMethods />
+        </div>
```
jorgemoya, line 87: Design between mobile/tablet vs desktop differ enough to warrant show/hidding.

core/components/footer/locale.tsx
```diff
@@ -0,0 +1,14 @@
...
+
+import { Link } from '../link';
+
+export const Locale = async () => {
+  const locale = await getLocale();
+
+  return (
+    <Link className="flex gap-2" href="/store-selector">
+      <Globe /> {locale.toUpperCase()}
+    </Link>
```
jorgemoya, line 12: Will change.

core/app/[locale]/store-selector/_components/locale-link.tsx
Comment (deini) [reply core/app/[locale]/store-selector/_components/locale-link.tsx:43]: Do you mean adding them to the map? Yeah

core/components/footer/locale.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { locales } from '~/i18n';
+
+import { Link } from '../link';
+
+export const Locale = async () => {
+  const locale = await getLocale();
+
+  return (
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    locales.length > 1 && (
```
jorgemoya, line 13: Want to hide this link if there is only one locale, unfortunately had to prevent this rule from applying because when it is only one locale typescript doesn't see the need for this conditional.

core/app/[locale]/store-selector/page.tsx
```diff
@@ -0,0 +1,76 @@
...
+        </NavigationMenu>
+      </header>
+
+      <div className="flex flex-col gap-2 px-4 lg:container sm:px-10 lg:mx-auto lg:max-w-[1000px] lg:px-12">
+        <h1 className="text-3xl font-black lg:text-4xl">{t('heading')}</h1>
+
+        <div className="grid grid-cols-1 gap-6 py-6 md:grid-cols-3 md:gap-11 lg:grid-cols-4 lg:gap-8">
+          {locales.map((locale) => (
+            // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+            <LocaleLink key={locale} locale={locale} selected={selectedLocale === locale} />
```
Comment (jorgemoya) [inline core/app/[locale]/store-selector/page.tsx:64]: Similar issue here, unfortunately had to prevent this rule from applying because when it is only one locale typescript doesn't see the need for this conditional since selectedLocale can only match the current locale.

jorgemoya:
Found an issue with kv store when swapping locales, reverting to draft to fix this.

jorgemoya:
Ready for re-review @bigcommerce/team-catalyst

chanceaclark (approved):

PR #1055, opened 2024-07-02
feat(core): make fetch channel aware per locale

jorgemoya:
## What/Why?
Add channel <=> locale config example.
Make client.fetch channel aware by fetching channel id assigned to locale.
Fetch from default channel if no map is found.

## Testing
Locally, it requests per channel configured.

### DIFF
--- a/core/channels.config.ts
+++ b/core/channels.config.ts
@@ -1,0 +1,17 @@
+import { type LocaleType } from './i18n';
+
+export type RecordFromLocales = {
+  [K in LocaleType]: string;
+};
+
+// Set overrides per locale
+const localeToChannelsMappings: Partial<RecordFromLocales> = {
+  // es: '12345',
+};
+
+function getChannelIdFromLocale(locale?: string) {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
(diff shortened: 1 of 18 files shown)


core/channels.config.ts
```diff
@@ -0,0 +1,17 @@
...
+};
+
+// Set overrides per locale
+const localeToChannelsMappings: Partial<RecordFromLocales> = {
+  // es: '12345',
+};
+
+function getChannelIdFromLocale(locale?: string) {
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+  return localeToChannelsMappings[locale as LocaleType] ?? process.env.BIGCOMMERCE_CHANNEL_ID;
```
jorgemoya, line 14: Returning `process.env.BIGCOMMERCE_CHANNEL_ID` until we can figure out how to programatically detect if `getLocale` won't work in `getChannelId`.

core/client/index.ts
```diff
@@ -1,4 +1,8 @@
 import { createClient } from '@bigcommerce/catalyst-client';
+// import { headers } from 'next/headers';
```
deini, line 2: 👀

deini (approved):

PR #1056, opened 2024-07-03
fix(core): make selected account tab visible for user by default on mobile devices

bc-yevhenii-buliuk:
## What/Why?
This PR makes selected account tab visible for user by default on Mobile devices even after reloading the page.

## Testing
locally

**before:**

https://github.com/bigcommerce/catalyst/assets/82589781/dc82a192-ef8c-4131-ab55-9e60381590e5

**after:**

https://github.com/bigcommerce/catalyst/assets/82589781/25e83dce-6b51-43b5-b67f-691a10b68328

### DIFF
--- a/core/app/[locale]/(default)/account/[tab]/_components/account-status-provider.tsx
+++ b/core/app/[locale]/(default)/account/[tab]/_components/account-status-provider.tsx
@@ -1,10 +1,19 @@
 'use client';
 
-import { createContext, ReactNode, useContext, useEffect, useState } from 'react';
+import {
+  createContext,
+  ReactNode,
+  RefObject,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
 
 import { State as AccountState } from '../_actions/submit-customer-change-password-form';
 
 export const AccountStatusContext = createContext<{
+  activeTabRef: RefObject<HTMLAnchorElement>;
   accountState: AccountState;
   setAccountState: (state: AccountState | ((prevState: AccountState) => AccountState)) => void;
 } | null>(null);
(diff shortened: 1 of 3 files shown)


core/app/[locale]/(default)/account/[tab]/_components/account-status-provider.tsx
```diff
@@ -17,6 +26,7 @@ export const AccountStatusProvider = ({
   isPermanentBanner?: boolean;
 }) => {
   const [accountState, setAccountState] = useState<AccountState>({ status: 'idle', message: '' });
+  const elementRef = useRef<HTMLAnchorElement>(null);
```
Comment (bc-alexsaiannyi) [inline core/app/[locale]/(default)/account/[tab]/_components/account-status-provider.tsx:29]: Since we're using `AccountStatusContext` it probably makes more sense to specify which element we basically refer to. What do you think about using more specific name like `tabRef` or `activeTabRef`?

core/app/[locale]/(default)/account/[tab]/_components/account-tabs.tsx
```diff
@@ -1,20 +1,30 @@
...
+import { useAccountStatusContext } from './account-status-provider';
+
 interface Props extends PropsWithChildren {
   tabs: TabType[];
   activeTab?: TabType;
 }
 
 export const AccountTabs = ({ children, activeTab, tabs }: Props) => {
   const t = useTranslations('Account.Home');
+  const context = useAccountStatusContext();
```
Comment (bc-alexsaiannyi) [inline core/app/[locale]/(default)/account/[tab]/_components/account-tabs.tsx:20]: why would not use destructuring here? what is a benefit of  keeping an additional variable?

core/app/[locale]/(default)/account/[tab]/_components/account-status-provider.tsx
Comment (yurytut1993) [reply core/app/[locale]/(default)/account/[tab]/_components/account-status-provider.tsx:29]: + for `activeTabRef`

Comment (bc-yevhenii-buliuk) [reply core/app/[locale]/(default)/account/[tab]/_components/account-status-provider.tsx]: @bc-alexsaiannyi @yurytut1993 fixed!

bc-alexsaiannyi (approved):
looks 👍

yurytut1993 (approved):

chanceaclark (approved):

jorgemoya (approved):

PR #1074, opened 2024-07-08
feat(core): allow setting a channel id in kv

jorgemoya:
## What/Why?
With multi-channel work, we need to be able to get kv's for each specific channel. With this change we can set a channel id or use default as before.

## Testing
Locally each channel has individual kv pairs.

### DIFF
--- a/core/middlewares/with-routes.ts
+++ b/core/middlewares/with-routes.ts
@@ -71,6 +71,8 @@ let locale: string;
 const updateRouteCache = async (pathname: string, event: NextFetchEvent): Promise<RouteCache> => {
   const channelId = getChannelIdFromLocale(locale);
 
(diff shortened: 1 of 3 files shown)


core/middlewares/with-routes.ts
```diff
@@ -71,6 +71,8 @@ let locale: string;
 const updateRouteCache = async (pathname: string, event: NextFetchEvent): Promise<RouteCache> => {
   const channelId = getChannelIdFromLocale(locale);
 
+  kv.setChannelId(channelId);
```
deini, line 74: Would this be a problem if the code runs on the same isolate between requests for different channels? 🤔

jorgemoya:
Closing to try a different approach.

PR #1075, opened 2024-07-08
feat(core): append channel to kv key to make it channel aware

jorgemoya:
## What/Why?
We need for kv keys to be channel aware when using multiple locales.

## Testing
Locally each locale/channel has its own kv pairs.

### DIFF
--- a/core/lib/kv/keys.ts
+++ b/core/lib/kv/keys.ts
@@ -1,8 +1,10 @@
+const VERSION = 'v3';
+
 export const STORE_STATUS_KEY = 'storeStatus';
 
 export const kvKey = (key: string, channelId?: string) => {
   const namespace = process.env.KV_NAMESPACE ?? process.env.BIGCOMMERCE_STORE_HASH ?? 'store';
   const id = channelId ?? process.env.BIGCOMMERCE_CHANNEL_ID ?? '1';
 
-  return `${namespace}_${id}_v3_${key}`;
+  return `${namespace}_${id}_${VERSION}_${key}`;
 };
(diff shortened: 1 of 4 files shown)


core/lib/kv/keys.ts
```diff
@@ -1,2 +1,3 @@
 export const STORE_STATUS_KEY = 'storeStatus';
-export const routeCacheKvKey = (pathname: string) => `v3_${pathname}`;
+export const kvKey = (key: string, channelId?: string) =>
+  `${channelId ?? process.env.BIGCOMMERCE_CHANNEL_ID ?? '1'}_${key}`;
```
deini, line 3: Thoughts on moving the `KV` namespace logic to this fn? .

core/middlewares/with-routes.ts
```diff
@@ -76,7 +76,7 @@ const updateRouteCache = async (pathname: string, event: NextFetchEvent): Promis
     expiryTime: Date.now() + 1000 * 60 * 30, // 30 minutes
   };
 
-  event.waitUntil(kv.set(routeCacheKvKey(pathname), routeCache));
+  event.waitUntil(kv.set(kvKey(`v3_${pathname}`, channelId), routeCache));
```
deini, line 79: Thoughts on moving the key version `v3` to `kvKey()`?

jorgemoya: Do we need `v3`? 🤔

core/lib/kv/keys.ts
jorgemoya: Will do

core/middlewares/with-routes.ts
deini: It's an easy way for us to change all keys if needed. I think we should keep it.

deini (approved):

PR #1077, opened 2024-07-09
fix: use system environment variables for robots

chanceaclark:
## What/Why?
Uses the deployment URL for robots.txt.

## Testing
https://catalyst-latest-git-fix-robots-bigcommerce-platform.vercel.app/robots.txt

Screenshot 2024-07-09 at 17 09 52

### DIFF
--- a/core/app/robots.ts
+++ b/core/app/robots.ts
@@ -1,27 +1,30 @@
 import type { MetadataRoute } from 'next';
 
+function parseUrl(url?: string): URL {
+  let incomingUrl = '';
+  const defaultUrl = new URL('http://localhost:3000/');
+
+  if (url && !url.startsWith('http')) {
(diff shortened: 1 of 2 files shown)


core/app/robots.ts
```diff
@@ -1,27 +1,30 @@
...
-const robotsConfig: MetadataRoute.Robots = {
-  rules: [
-    {
-      userAgent: '*',
-      allow: ['/'],
-      disallow,
-    },
-  ],
-};
+export function parseUrl(url?: string): URL {
```
jorgemoya, line 3: Do we need to export this?

jorgemoya (approved):

core/app/robots.ts
chanceaclark: Good call!

deini (approved):

PR #1084, opened 2024-07-10
feat(functional): Add code coverage skeleton

avattipalli:
## What/Why?
Add the option to generate code coverage in dev environment.

## Testing
Tested locally.

Screenshot 2024-07-10 at 6 45 16 PM


core/tests/global-teardown.js
```diff
@@ -0,0 +1,60 @@
+import fs from 'fs';
+import { CDPClient } from 'monocart-coverage-reports';
+import { addCoverageReport } from 'monocart-reporter';
+import path from 'path';
+import { fileURLToPath } from 'url';
+
+const globalTeardown = async (config) => {
+  if (!process.env.CI) {
```
avattipalli, line 8: This is to prevent collecting coverage on CI as we the source file instrumentation is different on vercel builds which will be handled in a follow up PR.

core/next.config.js
```diff
@@ -22,6 +22,18 @@ const nextConfig = {
     ignoreDuringBuilds: !!process.env.CI,
     dirs: ['app', 'client', 'components', 'lib', 'middlewares'],
   },
+  webpack: (config) => {
```
avattipalli, line 25: This change is ineffective and won't impact anything unless we pass additional flags to update the devtool packaging during the build.

.github/workflows/playwright-visual-regression.yml
```diff
@@ -42,8 +42,7 @@ jobs:
           npx playwright test tests/visual-regression/components/ --project=tests-chromium
 
       - uses: actions/upload-artifact@v4
-        if: failure()
```
bc-azvierieva, line 45: does removing this mean we are now collecting test running reports not only upon test failure, but every test run?

avattipalli: I did this to test if reports were being generated and saved into artifacts since not many runs end up in failures. Good catch, I'll restate this.

jorgemoya (approved):

PR #1098, opened 2024-07-16
Move sitemap fetching into the client

bookernath:
## What/Why?
Moves sitemap index fetching into the client so it gets the benefits of everything the client offers (such as user agent), as well as making the client more useful within other systems who may also wish to use a sitemap.

I fixed some missing user agents on admin API requests while I was at it and also fixed hardcoded hostnames.

## Testing
Preview build + server logs

### DIFF
--- a/.changeset/afraid-coats-play.md
+++ b/.changeset/afraid-coats-play.md
@@ -1,0 +1,6 @@
+---
+"@bigcommerce/catalyst-client": minor
+"@bigcommerce/catalyst-core": patch
+---
+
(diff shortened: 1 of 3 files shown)


bookernath:
Storefront works ✅ 
Sitemap works https://catalyst-latest-kxcmrtm6w-bigcommerce-platform.vercel.app/sitemap.xml ✅ 
Shipping estimator on cart page works ✅

deini (approved):

PR #1133, opened 2024-07-22
fix: radio group error state in pdp

jorgemoya:
## What/Why?
Missing prop to show error state.

## Testing
Locally.


chanceaclark (approved):

jorgemoya:
Added variants back, with success and error.

PR #1136, opened 2024-07-22
refactor: counter

jorgemoya:
## What/Why?

- Remove cva

## Testing
Locally

### DIFF
--- a/core/components/ui/counter/counter.tsx
+++ b/core/components/ui/counter/counter.tsx
@@ -1,33 +1,23 @@
-import { cva } from 'class-variance-authority';
 import { ChevronDown, ChevronUp } from 'lucide-react';
-import { ComponentPropsWithRef, ElementRef, forwardRef, useRef, useState } from 'react';
+import {
+  ComponentPropsWithRef,
+  ElementRef,
+  forwardRef,
+  useImperativeHandle,
+  useRef,
+  useState,
+} from 'react';
 
 import { cn } from '~/lib/utils';
 
-const inputVariants = cva(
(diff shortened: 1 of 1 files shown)


core/components/ui/counter/counter.tsx
```diff
@@ -64,6 +54,10 @@ export const Counter = forwardRef<ElementRef<'div'>, CounterProps>(
   ) => {
     const [value, setValue] = useState<number | ''>(getDefaultValue(defaultValue, min, max));
     const inputRef = useRef<ElementRef<'input'>>(null);
+
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions, @typescript-eslint/non-nullable-type-assertion-style
+    useImperativeHandle(ref, () => inputRef.current as ElementRef<'input'>);
```
jorgemoya, line 59: Is there a better way to do this? I'm trying to use ref in sibling elements.

chanceaclark: Using this is fine tbh, but you can type it better like so:
```ts
type CounterRef = ElementRef<'input'> | null;

const inputRef = useRef<CounterRef>(null);

useImperativeHandle<CounterRef, CounterRef>(ref, () => inputRef.current);
```

chanceaclark (approved):

core/components/ui/counter/counter.tsx
jorgemoya: This seems to work perfectly, thanks @chanceaclark

chanceaclark (approved):

deini (approved):

PR #1166, opened 2024-07-28
SEO improvements

bookernath:
## What/Why?

- Take advantage of the new `seo` field on `site.settings` to set default metadata on the layout level. This metadata will be used, if specified in the channel settings, to set the page title, meta description, and meta keywords on any pages that do not specify more specific values.
- Normalize/standardize how we set up SEO metadata across page entity types for consistency
- Fix a bug where `generateMetadata` was not exported on our static pages for categories and brands

## Testing
Manually/locally tested different scenarios with/without store settings and page-local settings.

### DIFF
--- a/core/app/[locale]/layout.tsx
+++ b/core/app/[locale]/layout.tsx
@@ -26,6 +26,11 @@ const RootLayoutMetadataQuery = graphql(`
     site {
       settings {
         storeName
+        seo {
+          pageTitle
+          metaDescription
+          metaKeywords
+        }
       }
     }
   }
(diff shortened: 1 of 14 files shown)


core/app/[locale]/layout.tsx
```diff
@@ -37,13 +42,17 @@ export async function generateMetadata(): Promise<Metadata> {
...
-  const title = data.site.settings?.storeName ?? '';
+  const storeName = data.site.settings?.storeName ?? '';
+
+  const { pageTitle, metaDescription, metaKeywords } = data.site.settings?.seo || {};
 
   return {
     title: {
-      template: `${title} - %s`,
-      default: title,
+      template: `%s - ${storeName}`,
```
Comment (bookernath) [inline core/app/[locale]/layout.tsx:51]: intentional stylistic change here - page first, store name second (same thing we do on stencil)

chanceaclark (approved):

bookernath:
This should not be merged until the SEO node is rolled out to all stores.

PR #1167, opened 2024-07-29
feat(core): add orders all page for account

bc-alexsaiannyi:
## What/Why?
This PR brings all orders page for a customer.

## Testing
locally

https://github.com/user-attachments/assets/1fe75aca-7974-43b7-a9b7-5147f3f7ebbb

### DIFF
--- a/core/app/[locale]/(default)/account/[tab]/_components/product-snippet.tsx
+++ b/core/app/[locale]/(default)/account/[tab]/_components/product-snippet.tsx
@@ -1,0 +1,86 @@
+import { graphql, ResultOf } from '~/client/graphql';
+import { BcImage } from '~/components/bc-image';
+import { Pricing, PricingFragment } from '~/components/pricing';
+import { cn } from '~/lib/utils';
+import { Link } from '~/navigation';
+
+export const ProductCardFragment = graphql(
+  `
+    fragment ProductCardFragment on Product {
+      entityId
+      name
+      defaultImage {
+        altText
+        url: urlTemplate
+      }
+      path
+      brand {
+        name
+        path
+      }
+
+      ...PricingFragment
+    }
+  `,
+  [PricingFragment],
+);
+
+interface Props {
+  product: ResultOf<typeof ProductCardFragment>;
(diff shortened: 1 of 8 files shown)


core/app/[locale]/(default)/account/[tab]/_components/product-snippet.tsx
```diff
@@ -0,0 +1,149 @@
+import { useFormatter } from 'next-intl';
+
+import { graphql, ResultOf } from '~/client/graphql';
+import { BcImage } from '~/components/bc-image';
+import { PricingFragment } from '~/components/product-card';
+import { cn } from '~/lib/utils';
+import { Link } from '~/navigation';
+
+export const ProductCardFragment = graphql(
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/account/[tab]/_components/product-snippet.tsx:9]: Why not import `ProdutCardFragment` from `/components/product-card`? If it has less nodes, I would rename.

core/app/[locale]/(default)/account/[tab]/_components/product-snippet.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+interface Props {
+  product: ResultOf<typeof ProductCardFragment>;
+  imageSize?: 'tall' | 'wide' | 'square';
+  brandSize?: string;
+  productSize?: string;
+  imagePriority?: boolean;
+}
+
+export const ProductSnippet = ({
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/account/[tab]/_components/product-snippet.tsx:37]: Can you reuse `ProductCard` instead?

Comment (bc-alexsaiannyi) [reply core/app/[locale]/(default)/account/[tab]/_components/product-snippet.tsx:37]: well, it was my thought but we use kind of a simpler version for Orders pages. Besides that in Orders we have  a different design as well. We've discussed with Turner and Product title is smaller than on PLP since it is more for management things rather marketing as on PLP. I also try re-use ProductSnippet on Order Details Page which has extended version ( it has little bit different layout and product options). Taking this into account I've thought that creating a specific version of ProductCard for account pages makes more sense than to extend and add props for our ProductCard.

PR #1190, opened 2024-07-30
feat: remove title prop from Tabs

jorgemoya:
## What/Why?

- Remove title prop from Tabs.
- Remove use of Tabs in /account since it's not really a Tab component, but a set of links.
- Removed test since we don't have a current implementation of Tabs yet.

## Testing
Locally.


jorgemoya:
Made a change in how we set the Tab title from translations.

deini (approved):

PR #1194, opened 2024-07-31
Add basic support for Google Analytics via BODL

BC-krasnoshapka:
## What/Why?
Add basic support for Google Analytics via [Big Open Data Layer](https://developer.bigcommerce.com/docs/integrations/hosted-analytics). BODL and GA4 integration is encapsulated in `bodl` workspace package which hides current complexity and limitations that will be improved in future. Package can be extended with more events and integrations with other analytics providers later. Data transformation from Catalyst data models to BODL and firing events is done in client components, as only frontend events are supported by BODL for now.

Following events supported:
- View product category
- View product page
- Add product to cart
- View cart
- Remove product from cart

## For reviewers
Due to large amount of changes there are following recommendations:
- Start reviewing from `bodl` package. There is no need to check `google_analytics4.js` as entire file is taken from Stencil implementation.
- BODL initialization in Catalyst is done in `bodl.ts` 
- In the end check all events one by one, their code is usually close to a component that implements appropriate functionality.

## Testing
Localy. In order to test you need to specify `NEXT_PUBLIC_GOOGLE_ANALYTICS` environment variable which is essentially your GA4 ID.

https://github.com/user-attachments/assets/8a697a14-fbcf-4d9d-9b8a-5204dfb89805


packages/bodl/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@bigcommerce/bodl",
```
chanceaclark, line 2: I'm starting to think we should move this into `core/` for now as we'd have to publish this if we wanted to use it. I think we'd rather wait to publish this until we know it is 100% what we want to do. What do you think?

chanceaclark (commented):
Overall I am good with the code, however do you think we can split this up? One PR with the package code (or if we decide to move it in `core/` per my comment) and another with the implementation?

packages/bodl/package.json
BC-krasnoshapka: I see your point and it makes sense. Another alternative is to call package `catalyst-bodl` so it won't be so generic for now.

BC-krasnoshapka: @chanceaclark moved, please check

core/tsconfig.json
```diff
@@ -54,6 +54,7 @@
   ],
   "exclude": [
     "node_modules",
-    ".next"
+    ".next",
+    "**/google_analytics4.js"
```
chanceaclark, line 58: Do we need this if it's a `.js` file?

core/lib/bodl/index.ts
```diff
@@ -0,0 +1,11 @@
+import { Bodl } from './bodl';
+
+export const bodl = new Bodl({
+  // TODO: get channel id from locale in multi-lang setup
+  channel_id: Number(process.env.BIGCOMMERCE_CHANNEL_ID ?? 1),
+  // TODO: Replace with actual store configuration
+  ga4: {
+    gaId: process.env.NEXT_PUBLIC_GOOGLE_ANALYTICS ?? '',
```
chanceaclark, line 8: 🍹 Just to be a bit more descriptive.
```suggestion
    gaId: process.env.NEXT_PUBLIC_GOOGLE_ANALYTICS_ID ?? '',
```

core/lib/bodl/bodl.ts
```diff
@@ -0,0 +1,128 @@
...
+import {
+  bodl_v1_cart_product_added,
+  bodl_v1_cart_product_removed,
+  bodl_v1_cart_viewed,
+  bodl_v1_product_category_viewed,
+  bodl_v1_product_page_viewed,
+  BodlConfig,
+} from './types';
+
+const GA4_DEVELOPER_ID = 'dMjk3Nj';
```
chanceaclark, line 13: Can we just set this as a private property on the class?

chanceaclark (commented):
Small changes, otherwise lgtm

core/tsconfig.json
BC-krasnoshapka: Fixed

core/lib/bodl/index.ts
BC-krasnoshapka: Sure

chanceaclark (approved):
LGTM! Make sure to add all the environment variables to all the Vercel Projects that we want to start tracking!

PR #1202, opened 2024-08-01
chore: combine regression tests

chanceaclark:
## What/Why?
Combine the regression suites into one GitHub action. This should enable us to more easily figure out which deployments have failures in a combined view rather than having to search through multiple actions.

## Testing
See CI runs.


jorgemoya (approved):

avattipalli (approved):
:shipit:

PR #1209, opened 2024-08-02
Default webpack chunk limit of 5, env variable to customize

bookernath:
# What/Why?
Limit the number of chunks Webpack produces to a default of 5 to strike a balance between performance and total requests.

Can be customized via env variable.

We've noticed that the first limit you hit when hosting a Catalyst storefront on Vercel is Edge Requests, and the biggest contributing factor to that is the numerous chunked JS files. Putting a limit on this makes Catalyst less expensive to operate on Vercel and allows for more page views to fit within the limits of a Vercel plan before incurring overages.

# Testing
Preview deployment + lighthouse

## Before (26 chunks):
image
image
image

## After (8 chunks):
image
image
image

### DIFF
--- a/core/next.config.js
+++ b/core/next.config.js
@@ -18,7 +18,7 @@ const nextConfig = {
     ignoreDuringBuilds: !!process.env.CI,
     dirs: ['app', 'client', 'components', 'lib', 'middlewares'],
   },
-  webpack: (config) => {
+  webpack: (config, { isServer }) => {
     if (process.env.NODE_V8_COVERAGE) {
       Object.defineProperty(config, 'devtool', {
         get() {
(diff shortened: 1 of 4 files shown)


core/next.config.js
```diff
@@ -28,6 +28,17 @@ const nextConfig = {
       });
     }
 
+    // Limit the number of chunks to reduce CDN requests, which contribute to Edge Request costs
+    if (!isServer) {
+      config.plugins.push(
+        new (require('webpack').optimize.LimitChunkCountPlugin)({
+          maxChunks: process.env.WEBPACK_MAX_CHUNKS
+            ? parseInt(process.env.WEBPACK_MAX_CHUNKS, 5)
```
chanceaclark, line 36: It's pretty rare to see a base 5 number. Might want to make it base 10 🤔 
```suggestion
            ? parseInt(process.env.WEBPACK_MAX_CHUNKS, 10)
```

core/next.config.js
```diff
@@ -28,6 +28,17 @@ const nextConfig = {
       });
     }
 
+    // Limit the number of chunks to reduce CDN requests, which contribute to Edge Request costs
+    if (!isServer) {
+      config.plugins.push(
+        new (require('webpack').optimize.LimitChunkCountPlugin)({
+          maxChunks: process.env.WEBPACK_MAX_CHUNKS
```
chanceaclark, line 35: If this requires a new build, do we even need a environment variable for this? I'd be more inclined to just inline a value.

chanceaclark (approved):

deini (approved):

PR #1210, opened 2024-08-02
feat(functional): Send gh workflow results to slack

avattipalli:
## What/Why?
Send notifications to slack team channel with the necessary information to alert any job failures. When a job from the regression tests workflow fails, this step will send a message with failure notification and the pull request associated with the workflow which caused the failure.

For this to work, I've added a slack workflow which provides a webhook. It accepts the payload we send as part of this step on failure and dispatches the message to the configured slack channel.

## Testing
Tested by sending messages to channel.

Screenshot 2024-08-07 at 10 07 41 AM

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -151,7 +151,8 @@ jobs:
               "Job": "${{ github.job }}",
               "Status": "Fail",
               "Pull_Request": "https://github.com/bigcommerce/catalyst/pull/${{ steps.finder.outputs.pr }}",
-              "Commit_Message" : "${{ steps.finder.outputs.title }}"
+              "Commit_Message" : "${{ steps.finder.outputs.title }}",
+              "Job_Run": "https://github.com/bigcommerce/catalyst/actions/runs/${{ github.run_id }}"
             }
 
   visual-regression-tests:
@@ -207,5 +208,6 @@ jobs:
               "Job": "${{ github.job }}",
               "Status": "Fail",
               "Pull_Request": "https://github.com/bigcommerce/catalyst/pull/${{ steps.finder.outputs.pr }}",
-              "Commit_Message" : "${{ steps.finder.outputs.title }}"
+              "Commit_Message" : "${{ steps.finder.outputs.title }}",
+              "Job_Run": "https://github.com/bigcommerce/catalyst/actions/runs/${{ github.run_id }}"
             }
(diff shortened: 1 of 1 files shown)


.github/workflows/regression-tests.yml
```diff
@@ -136,6 +140,20 @@ jobs:
...
           retention-days: 30
 
+      - name: Send slack notification
+        uses: slackapi/slack-github-action@v1.26.0
+        if: failure()
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+        with:
+          payload: |
+            {
```
avattipalli, line 150: With the current setup we have for deployment, it is not possible to map the github user and bc user to tag them and so, we are not sending that information as part of the payload.

bc-azvierieva: Hey @avattipalli, it looks good! Is this possible to add the specific test run link in addition to the PR link? For example, https://github.com/bigcommerce/catalyst/actions/runs/10280244306

As if some check is failed on PR, we will go to this PR and look for this type of information to investigate the test run report, so would be really convenient to have this link directly in the Slack notification about test failure.

avattipalli: That's a good callout, let me try and add that.

.github/workflows/regression-tests.yml
```diff
@@ -136,6 +140,21 @@ jobs:
...
+              "Job": "${{ github.job }}",
+              "Status": "Fail",
```
chanceaclark, line 152: Let's change this to `Failed` as it's a past occurrence when we receive the notification. (Same below)

chanceaclark (commented):
If we are wanting to make these more reusable, we can create custom actions locally. See the next.js repo as an example: https://github.com/vercel/next.js/blob/e8dec1f3bcacd759c615fec0e07e8700f5aece36/.github/workflows/popular.yml#L17

^ No action needed

.github/workflows/regression-tests.yml
avattipalli: I changed this and probably set it back during rebase 😮‍💨

avattipalli:
> If we are wanting to make these more reusable, we can create custom actions locally. See the next.js repo as an example: https://github.com/vercel/next.js/blob/e8dec1f3bcacd759c615fec0e07e8700f5aece36/.github/workflows/popular.yml#L17
> 
> ^ No action needed

I'll read about it 👍

chanceaclark (approved):

PR #1211, opened 2024-08-02
feat(core): update pricing prop for ProductCard

jorgemoya:
## What/Why?
Instead of accepting a `ReactNode`, we pass in a custom object that has all pricing info and `ProductCard` is the one that has all styling logic.

## Testing
Locally it behaves the same as before.


chanceaclark (commented):
While you are updating this, would you mind fixing the issue where a logged in customer doesn't see their customer-specific pricing?

chanceaclark (approved):

PR #1233, opened 2024-08-06
test(core): add some stability to some account tests

chanceaclark:
## What/Why?
Attempt to add some stability to the account/auth pages to ensure the test runs completely.

## Testing
See CI.

### DIFF
--- a/core/app/[locale]/(default)/login/register-customer/_components/register-customer-form/_actions/register-customer.ts
+++ b/core/app/[locale]/(default)/login/register-customer/_components/register-customer-form/_actions/register-customer.ts
@@ -1,5 +1,7 @@
 'use server';
 
+import { BigCommerceAPIError } from '@bigcommerce/catalyst-client';
+
 import {
   Input,
   registerCustomer as registerCustomerClient,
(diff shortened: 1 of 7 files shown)


core/app/[locale]/(default)/login/register-customer/_components/register-customer-form/_actions/register-customer.ts
```diff
@@ -49,17 +51,34 @@ export const registerCustomer = async ({ formData, reCaptchaToken }: RegisterCus
...
+        status: 'error',
+        error: 'Looks like we are experience a server error, please try again in a few minutes.',
+      };
+    }
+
+    return {
+      status: 'error',
+      error: 'Something went wrong. Please try again later.',
+    };
+  }
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/login/register-customer/_components/register-customer-form/_actions/register-customer.ts:83]: This is supposed to help with things like rate limiting. I am not the best at writing, so please leave any feedback around what these error messages should be 🙏

core/app/[locale]/(default)/login/register-customer/_components/register-customer-form/index.tsx
```diff
@@ -188,17 +188,8 @@ export const RegisterCustomerForm = ({
...
-        status: 'success',
-        message: t('successMessage', {
-          firstName: submit.data?.firstName,
-          lastName: submit.data?.lastName,
-        }),
-      });
 
-      setTimeout(() => {
-        void login(formData);
-      }, 3000);
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/login/register-customer/_components/register-customer-form/index.tsx:201]: This timeout hurt our register customer test, so I opted to remove it for now. I think we should be showing this message on the account page instead, but will do so in another PR.

core/components/ui/dropdown/dropdown.tsx
```diff
@@ -37,17 +37,17 @@ const Dropdown = ({ align = 'center', className, items, sideOffset = 4, trigger
                 </Link>
               </DropdownMenu.Item>
             ) : (
-              <form action={item.action} key={item.name}>
-                <DropdownMenu.Item asChild>
+              <DropdownMenu.Item asChild key={item.name}>
+                <form action={item.action} className="hover:focus-visible:ring-0">
```
chanceaclark, line 41: This was causing the logout functionality to be flaky.

deini (approved):

PR #1236, opened 2024-08-07
Speed up builds by excluding node_modules from tailwind config

bookernath:
## What/Why?
https://github.com/vercel/next.js/issues/37825#issuecomment-1633364694

Our build time was spending over 60 seconds on `Creating an optimized production build...` so I tried removing the `node_modules` folder from the `content` in tailwind config, and I see much faster build times now.

## Testing
Preview build

Before:
image
image

After:
image

image

### DIFF
--- a/core/tailwind.config.js
+++ b/core/tailwind.config.js
@@ -1,6 +1,10 @@
 /** @type {import('tailwindcss').Config} */
 const config = {
-  content: ['./app/**/*.{ts,tsx}', './components/**/*.{ts,tsx}'],
+  content: [
+    './app/**/*.{ts,tsx}',
+    './components/**/*.{ts,tsx}',
+    '!./node_modules/**', // Exclude everything in node_modules to speed up builds
+  ],
   theme: {
     extend: {
       colors: {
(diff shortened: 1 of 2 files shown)


core/tailwind.config.js
```diff
@@ -1,6 +1,10 @@
 /** @type {import('tailwindcss').Config} */
 const config = {
-  content: ['./app/**/*.{ts,tsx}', './components/**/*.{ts,tsx}'],
+  content: [
+    './app/**/*.{ts,tsx}',
+    './components/**/*.{ts,tsx}',
+    '!./node_modules/**', // Exclude everything in node_modules to speed up builds
```
deini, line 6: Interesting, do you know why this happens? I don't understand why we have to exclude a path we aren't including. 🤔

deini (approved):

core/tailwind.config.js
bookernath: I wish I knew!

chanceaclark: Seems like it's expected: https://github.com/tailwindlabs/tailwindcss/issues/5014#issuecomment-1047038219

PR #1241, opened 2024-08-08
fix(functional): Add environment details to message

avattipalli:
## What/Why?
Add environment details to slack message to distinguish between the releases. Along with it, I updated the script to rely on `actions/github-script` to fetch the pull request information. 

## Testing
Screenshot 2024-08-09 at 11 01 39 AM

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -74,8 +74,17 @@ jobs:
             core.setOutput("comment", comment);
 
       - name: Find pull request
-        uses: jwalton/gh-find-current-pr@v1.3.3
-        id: finder
+        id: find_pull_request
+        uses: actions/github-script@v7
+        with:
+          script: |
+            return (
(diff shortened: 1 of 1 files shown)


chanceaclark (approved):

.github/workflows/regression-tests.yml
```diff
@@ -102,6 +111,9 @@ jobs:
     name: Playwright UI Tests
     timeout-minutes: 30
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
```
chanceaclark, line 116: Why do we need permissions for just this job and not the rest?

avattipalli: I read this was required to get it working but it worked without it 🤷‍♂️ . Let me remove it in the other jobs as well.

avattipalli: I removed it from everywhere.

chanceaclark (approved):

PR #1260, opened 2024-08-13
fix(chore): Fetch pull request details in script

avattipalli:
## What/Why?
Fall back to using github `action-script` to fetch the pull request details which was more reliable than what we currently have.

## Testing
Tested a few times as part of this change in the CI.

### DIFF
--- a/.github/workflows/regression-tests.yml
+++ b/.github/workflows/regression-tests.yml
@@ -21,9 +21,13 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
-      - name: Find pull request
-        uses: jwalton/gh-find-current-pr@v1.3.3
-        id: finder
+      - name: Pull Request Details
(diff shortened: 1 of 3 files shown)


.github/workflows/regression-tests.yml
```diff
@@ -21,9 +21,14 @@ jobs:
...
-      - name: Find pull request
-        uses: jwalton/gh-find-current-pr@v1.3.3
-        id: finder
+      - name: Pull Request Details
+        id: pr_details
+        uses: actions/github-script@v7
+        with:
+          retries: 3
+          script: |
+            const prData = require('./.github/workflows/prData.js');
```
chanceaclark, line 29: hyphen + expand the name
```suggestion
            const prData = require('./.github/workflows/pull-request-data.js');
```

chanceaclark (approved):

PR #1261, opened 2024-08-13
feat: remove the need of using the `BIGCOMMERCE_ACCESS_TOKEN`

chanceaclark:
## What/Why?

> [!TIP]
> Reviewing by commit will be helpful.

Removes the need of using the `BIGCOMMERCE_ACCESS_TOKEN` to fetch any data. If the environment variable is present, it'll fetch shipping zone from the management API, otherwise it'll regress back to fetching all countries and states.

### With an access token
Screenshot 2024-08-13 at 17 19 04

### Without an access token
Screenshot 2024-08-13 at 17 30 39


bookernath (approved):

core/components/ui/select/select.tsx
```diff
@@ -60,11 +60,11 @@ const Select = forwardRef<ElementRef<typeof SelectPrimitive.Trigger>, Props>(
             </SelectPrimitive.ScrollUpButton>
             <SelectPrimitive.Viewport className="h-[var(--radix-select-trigger-height)] w-full min-w-[var(--radix-select-trigger-width)]">
               {options.map((option) => {
-                const { value, label, ...optionProps } = option;
+                const { value, label, key, ...optionProps } = option;
 
                 return (
                   <SelectPrimitive.Item
-                    key={`${id}-${value}`}
+                    key={key ?? `${id}-${value}`}
```
jorgemoya, line 67: I guess we got to do this for component that iterate.

jorgemoya (approved):

core/components/ui/select/select.tsx
chanceaclark: Yeah unfortunately :(

jorgemoya (approved):

PR #1262, opened 2024-08-14
feat: use customer access tokens for customer logins

chanceaclark:
## What/Why?
Uses the new approach of using Customer Access Tokens (**CATs**)  when login in as a customer. This change utilized the Storefront API Token (**APIT**) instead of the Storefront Customer Impersonation Token (**CIT**) to authenticate requests as CATs will only be generated using a APIT.

> [!TIP]
> Reviewing by commit will be helpful.

## Testing
Every commit was individually tested so that Catalyst could keep working depending on the state. If need be, try to roll back to a previous commit if not working correctly.

## Releasing
1. Cut a release before merging
2. Merge this PR
3. Immediately cut another release


deini (approved):

jorgemoya (approved):

integrations/algolia/integration.patch
matthewvolk, line 1: Now that I'm thinking about it, this `patch` file is only compatible with the latest core tag. If we merge this and someone creates a new Catalyst storefront at the currently available core tag, this patch won't work for them since the new CAT changes are in `main` but not in the currently available core tag.

integrations/makeswift/integration.patch
matthewvolk, line 1: Same here

matthewvolk (changes requested):

matthewvolk (approved):

PR #1301, opened 2024-08-22
Fix: Decode Webpage Id to fix 404 on Webpages with nonsafe ids

onurstats:
## What/Why?

Pages with non url safe ids causes to 404 response.

Page Data and PageMetadata request sends the urlencoded id strings. this cause no response. I checked the  Contact page but it already been fix

Example id: `Tm9ybWFsUGFnZToxMA==` and graphgl has the variable `{id:"Tm9ybWFsUGFnZToxMA%3D%3D"}`

Screenshot 2024-08-22 at 17 24 41

Screenshot 2024-08-22 at 17 22 12

## Testing

Go to a page with nonurlsafe id

### DIFF
--- a/.changeset/few-panthers-judge.md
+++ b/.changeset/few-panthers-judge.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+fix: decode webpage id to fix 404 on some Webpages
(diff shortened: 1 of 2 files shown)


deini (approved):

jorgemoya (approved):

bookernath:
Thanks for your contribution!

PR #1304, opened 2024-08-23
Adjust WEBPACK_MAX_CHUNKS default and add more documentation

bookernath:
## What/Why?
Our reduction in webpack chunks _does_ negatively affect Lighthouse scores, so I'm increasing the default and adding documentation to inform users of why they might want to tweak it and how.

## Testing
Preview deployment

### DIFF
--- a/docs/environment-variables.md
+++ b/docs/environment-variables.md
@@ -1,4 +1,5 @@
 **Reference**
+
 # Environment variables
 
 The following reference lists the environment variables that a Catalyst storefront can use, along with a description of their data types and approximate values.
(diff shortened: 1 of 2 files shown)


docs/environment-variables.md
```diff
@@ -12,11 +13,11 @@ Catalyst storefronts require the following variables to function.
 
 ### BIGCOMMERCE_STORE_HASH
 
-| Attribute | Value |
-|:----------|:------|
-| Type | string |
-| Permanently required | true |
-| CLI-configurable | true |
+| Attribute            | Value  |
```
bookernath, line 16: Think this is prettier fixing things up for us, we should probably commit this

jorgemoya (approved):

PR #1314, opened 2024-08-28
Add Wishlist test and update wishlist naming

bc-azvierieva:
## What/Why?
- Create add/delete Wishlist test
- Fix Wishlist naming in some places, since we agreed with Turner that "wishlist" is the correct one over the "wish list"

## Testing
Screenshot 2024-08-27 at 11 23 01

### DIFF
--- a/core/messages/ja.json
+++ b/core/messages/ja.json
@@ -171,7 +171,7 @@
             "messages": "メッセージ",
             "addresses": "住所",
             "newAddress": "New address",
-            "wishlists": "Wish lists",
+            "wishlists": "Wishlists",
             "recentlyViewed": "Recently viewed",
             "settings": "アカウント設定",
             "changePassword": "パスワード変更",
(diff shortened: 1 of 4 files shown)


core/messages/ja.json
```diff
@@ -171,7 +171,7 @@
             "messages": "メッセージ",
             "addresses": "住所",
             "newAddress": "New address",
-            "wishlists": "Wish lists",
+            "wishlists": "Wishlists",
```
bc-azvierieva, line 174: Not sure if we should edit this translation file

yurytut1993 (approved):
👍

core/messages/ja.json
chanceaclark: According to our glossary, the correct version is `wishlist`: https://developer.bigcommerce.com/docs/start/design/ux-writing/glossary#x-y-z:~:text=word%2C%20no%20space.-,wishlist,-Do%20not%20use

chanceaclark (approved):

PR #1316, opened 2024-08-29
refactor(core): localization

jorgemoya:
Note: Review by commit if needed & remove white spaces.

## What/Why?

The PR addresses missing translations on some pages and aims to normalize how translations are handled throughout the project.

### Configuration Changes
- Updated `next-intl` version and adjusted the configuration accordingly.
- Renamed `navigation.ts` to `routing.ts` for consistency with documentation.
- Introduced the `defineRouting` function.
- Simplified the translation setup by using a single `NextIntlClientProvider` in RootLayout to pass all messages, making the use of `next-intl` hooks in client components easier.
- Standardized the use of translation hooks by:
  - Using `useTranslations` in both client and server components without async functions.
  - Using `await getTranslations` in server components with async functions.
  - Same case for `formatter`.
- Removed the need to pass the `locale` to these hooks (not needed).

### Translations
- Added translations for all strings, except those in `components/ui`.
- Shuffle translations to have pages as parent groups.
- Renamed some translations keys for clarity.
- Add `Components` root group for better organization.

### Docs
- Updated docs to match new use cases.

## Testing
Locally translations work.


jorgemoya:
@bc-andreadao For this PR I updated the docs to better showcase the simplified use of translations. Please let me know if you have any feedback. Thanks!

core/app/[locale]/not-found.tsx
```diff
@@ -66,30 +63,16 @@ export default async function NotFound() {
...
+          title={t('Carousel.featuredProducts')}
+        />
       </main>
 
       <Footer data={data.site} />
     </>
   );
 }
-
-export const runtime = 'edge';
```
Comment (chanceaclark) [inline core/app/[locale]/not-found.tsx:95]: Is there a reason we are removing the runtime from this page?

core/app/[locale]/store-selector/page.tsx
```diff
@@ -68,5 +68,3 @@ export default async function StoreSelector({ params: { locale: selectedLocale }
 export function generateStaticParams() {
   return locales.map((locale) => ({ locale }));
 }
-
-export const dynamic = 'force-static';
```
Comment (chanceaclark) [inline core/app/[locale]/store-selector/page.tsx:72]: Reason for this removal?

docs/localization.md
```diff
@@ -87,33 +85,40 @@ const locales = ['en', 'fr'] as const;
...
+```tsx
+'use client';
+
+import { useTranslations } from 'next-intl';
+// ...
 
-  const t = await getTranslations({ locale, namespace: 'Home' });
-  const messages = await getMessages({ locale });
+const Component = () => {
+  const t = await getTranslations('Cart');
```
chanceaclark, line 115:
```suggestion
  const t = await useTranslations('Cart');
```

chanceaclark:
ℹ️ Hide whitespace changes helps with review

core/app/[locale]/not-found.tsx
Comment (jorgemoya) [reply core/app/[locale]/not-found.tsx:95]: Mistake, thanks for catching it!

docs/localization.md
jorgemoya: 😆 I changed the import but not the translation.

chanceaclark (approved):

bc-andreadao (approved):

bc-andreadao:
Thanks @jorgemoya !

PR #1322, opened 2024-08-30
refactor(core): split auth pages to their own routes

jorgemoya:
## What/Why?
Split auth forms to four different pages:

- /login
- /register
- /reset
- /change-password

Additionally, moved shared form field components to `/components/form-fields/` and updated translations.

## Testing
Locally.


core/middlewares/with-auth.ts
```diff
@@ -1,7 +1,26 @@
...
+    // @ts-expect-error: The `auth` function doesn't have the correct type to support it as a MiddlewareFactory.
+    const authWithCallback = auth(async (req) => {
+      if (req.auth && AUTH_PATHS.includes(req.nextUrl.pathname)) {
+        return NextResponse.redirect(new URL('/account', req.nextUrl.origin));
+      }
+
+      // Continue the middleware chain
+      const response = await next(req, event);
+
+      return response;
```
deini, line 20: Since we are not doing anything with the response, could we `return next(req, event);` instead?

core/middlewares/with-auth.ts
```diff
@@ -1,7 +1,24 @@
...
+export const withAuth: MiddlewareFactory = (next) => {
+  return async (request, event) => {
+    // @ts-expect-error: The `auth` function doesn't have the correct type to support it as a MiddlewareFactory.
+    const authWithCallback = auth(async (req) => {
+      if (req.auth && AUTH_PATHS.includes(req.nextUrl.pathname)) {
+        return NextResponse.redirect(new URL('/account', req.nextUrl.origin));
+      }
+
+      // Continue the middleware chain
+      return await next(req, event);
```
deini, line 18: I believe we can remove the `async/await` here

deini (approved):

PR #1328, opened 2024-09-03
fix: remove pagination cursors from search params on sort change

willPrattUPL:
## What/Why?
removes pagination cursors from search params when changing sort option

old cursors were sticking around in the url after changing sort which messed up pagination.

## Testing

1. find a search with enough results for pagination
2. click through a few pages
3. change the sort option
4. it should reload on first page result of new sort selection & remove the cursors from params

------------------------------------

Extends from slack conversation in bigcommercedevs/#catalyst


chanceaclark (approved):
Thanks @willPrattUPL!

PR #1334, opened 2024-09-04
fix(core): resolve 500 server error on locale change

deini:
## What/Why?
With the recent changes to our middleware, there is an issue when switching locales form the UI. `intlResponse.redirected` is false when status is 307 🤔


chanceaclark (approved):

chanceaclark:
We should probably add some tests for locale switching soon

jorgemoya (approved):

PR #1338, opened 2024-09-05
fix(core): improve redirect behavior after change password on account page

bc-yevhenii-buliuk:
## What/Why?
This PR improves the redirect behavior after changing the password on the Account Page, namely adding a successful password change banner on Login Page.

## Testing
locally

**before**:

https://github.com/user-attachments/assets/f79e290f-0cde-4cff-ad32-318167ac1255

**after**:

https://github.com/user-attachments/assets/64ebda4a-4a96-4928-b6ec-cc0e25104d1c


core/app/[locale]/(default)/account/(tabs)/settings/_components/change-password-form.tsx
```diff
@@ -107,14 +108,20 @@ export const ChangePasswordForm = () => {
...
-        router.push('/login');
-      }, 2000);
+      void logout();
+
+      setAccountState({
+        status: 'success',
+        message: t('confirmChangePassword'),
+      });
+
+      router.push('/login');
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/account/(tabs)/settings/_components/change-password-form.tsx:122]: `logout` should redirect to login page, no?

yurytut1993 (approved):
👌

core/components/header/_actions/logout.ts
```diff
@@ -1,7 +1,11 @@
...
 
+import { redirect } from 'next/navigation';
+
 import { signOut } from '~/auth';
 
 export const logout = async () => {
-  await signOut();
+  await signOut({ redirect: false });
+
+  redirect('/login');
```
bc-alexsaiannyi, line 10: I think that `signOut` takes `redirectTo` as an option also. can we use it instead of `redirect` method?

core/app/[locale]/(default)/account/(tabs)/settings/_components/change-password-form.tsx
Comment (bc-yevhenii-buliuk) [reply core/app/[locale]/(default)/account/(tabs)/settings/_components/change-password-form.tsx]: I made some changes in the [logout](core/components/header/_actions/logout.ts) method, but now if the user decides to log out from any page, he will automatically be redirected to the Login page. Perhaps it is worth clarifying this behavior with Turner, because now when logging out, the user remains on the same page.

Otherwise, if we don't change the `logout` method, we can use `router.push('/login')` after successfully changing password.

core/components/header/_actions/logout.ts
bc-yevhenii-buliuk: I used this approach to be [consistent](https://github.com/bigcommerce/catalyst/blob/main/core/app/%5Blocale%5D/(default)/login/register-customer/_components/register-customer-form/_actions/login.ts#L17)  with Chance's approach to login

bc-alexsaiannyi (approved):

jorgemoya (approved):

bc-azvierieva (approved):
Nice job!

PR #1342, opened 2024-09-05
fix(core): update localeSwitcher to use link instead of form

jorgemoya:
## What/Why?
Use link instead of form. Simplifies logic and behaves better.

## Testing
Locally, it redirects to correct locale without use, even when always showing locale prefix.

### DIFF
--- a/core/components/ui/header/locale-switcher.tsx
+++ b/core/components/ui/header/locale-switcher.tsx
@@ -2,9 +2,10 @@
 
 import * as PopoverPrimitive from '@radix-ui/react-popover';
 import { useTranslations } from 'next-intl';
-import { FormEvent, useMemo, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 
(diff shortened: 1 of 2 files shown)


core/components/ui/header/locale-switcher.tsx
```diff
@@ -55,6 +53,18 @@ const LocaleSwitcher = ({ activeLocale, locales }: Props) => {
...
+    if (regionSelected && languageSelected) {
+      const nextLocale = locales.find(
+        (locale) => locale.language === languageSelected && locale.region === regionSelected,
+      );
+
+      if (nextLocale) {
+        setNewLocale(nextLocale.id);
+      }
+    }
+  }, [regionSelected, languageSelected, locales]);
```
deini, line 66: Can we compute `newLocale` instead of this effect? Or use a memo?

jorgemoya: Sure!

deini (approved):

PR #1359, opened 2024-09-11
refactor(core): functional test changes

chanceaclark:
## What/Why?
- Removes unused JS coverage collection as we are not even using it.
- Reorganizes/renames functional tests
- Reenable/update skipped tests
- Remove serial mode tests as it doesn't behave like I thought it would
- Add a fixture to create customer accounts, log them in and out, and delete them when unused.

## Testing
See regression test run


bc-azvierieva (approved):
That is fantastic improvement for account utils! 👏👏👏

jorgemoya (approved):

core/tests/fixtures/utils/account/index.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    this.autoCleanup = autoCleanup;
+    this.account = {
+      ...customer,
+      cleanup: this.cleanup.bind(this),
+      login: this.login.bind(this),
+      logout: this.logout.bind(this),
+    };
+
+    return this.account;
```
deini, line 31: 🍹 we can return `this` here and we can move the account info to the root maybe?

chanceaclark: sure 👍

deini (approved):

PR #1361, opened 2024-09-12
chore(core): enforce use of next-intl's navigation APIs

jorgemoya:
## What/Why?
Noticed we were not using next-intl's wrapper navigation APIs.

- Added eslint rule to enforce use of these APIs.
- Fixed issues.

## Testing
Locally, behaves the same.

### DIFF
--- a/core/.eslintrc.cjs
+++ b/core/.eslintrc.cjs
@@ -27,6 +27,21 @@ const config = {
             name: 'next/link',
             message: "Please import 'Link' from '~/components/Link' instead.",
           },
+          {
+            name: '~/i18n/routing',
+            importNames: ['Link'],
+            message: "Please import 'Link' from '~/components/Link' instead.",
+          },
+          {
+            name: 'next/router',
+            importNames: ['useRouter'],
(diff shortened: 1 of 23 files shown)


jorgemoya:
Investigating how to prevent `redirect` to append locale on default locale.

jorgemoya:
Opened an issue for this: https://github.com/amannn/next-intl/issues/1335.

Will need to expect locale prefix when using `redirect` until a solution is found.

core/.eslintrc.cjs
```diff
@@ -27,6 +27,21 @@ const config = {
...
           },
+          {
+            name: '~/i18n/routing',
+            importNames: ['Link'],
+            message: "Please import 'Link' from '~/components/Link' instead.",
+          },
+          {
+            name: 'next/router',
+            importNames: ['useRouter'],
+            message: 'Please import from `@/i18n/routing` instead.',
```
deini, line 38: Should these be `~/` instead of `@/`?

jorgemoya: Yes, will update.

deini (approved):

PR #1363, opened 2024-09-13
docs: add catalyst-client documentation

bc-andreadao:
## What/Why?

Document how to use the `client` for fetching data from BigCommerce Storefront API

Methods have their parameters, return values, etc. documented (e.g. how to use `client.fetch()`, etc)    

## Testing
N / A


docs/catalyst-client/overview.md
```diff
@@ -0,0 +1,28 @@
...
+```ts
+import { client } from '~/client';
+import { graphql } from '~/client/graphql'; // A utility function for handling GraphQL queries
+```
+
+## Methods
+
+- `client.fetch()`: allows you to interact with BigCommerce's GraphQL Storefront API.
+
+- `client.fetchShippingZones()`: retrieves shipping zones for the BigCommerce store. It sends an HTTP GET request to the BigCommerce Management API, but abstracts the complexity of making a direct API request to fetch shipping zones.
```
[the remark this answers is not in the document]
bc-andreadao: Ah thanks for the info. I'm inclined to have a separate `deprecated` section (rather than removing) so people using the old Catalyst repositories won't be confused. But I guess removing it might also be okay since Catalyst is new. @bookernath I wonder if you have a preferred direction?

docs/client.md
```diff
@@ -0,0 +1,156 @@
...
+
+- `fetch()`: allows you to interact with BigCommerce's [GraphQL Storefront API](https://developer.bigcommerce.com/docs/storefront/graphql). You can execute all queries and mutations available in [BigCommerce's Storefront graph](https://developer.bigcommerce.com/graphql-storefront/reference).
+
+- `fetchSitemapIndex(channelId?: string)`: fetches the sitemap index for the store, which provides a URL to the XML sitemap.
+
+- `getCanonicalUrl(channelId?: string)`: constructs and returns the canonical URL for the BigCommerce store. This URL is used for constructing various API endpoint URLs and is based on the store's hash and the provided or default channel ID.
+
+- `getGraphQLEndpoint(channelId?: string)`: returns the GraphQL endpoint URL for the store, incorporating the optional channel ID to specify which store channel's GraphQL API to access. This URL is used for sending GraphQL queries and mutations.
+
+- `requestLogger(document: string)`: provides a logging function to help debug and monitor performance. Logs the operation type, name, request duration, and query complexity.
```
chanceaclark, line 21: These are private methods and probably shouldn't be documented.

docs/client.md
```diff
@@ -0,0 +1,156 @@
...
+## Fetch
+
+The `fetch()` method lets you send GraphQL queries or mutations to the [GraphQL Storefront API](https://developer.bigcommerce.com/docs/storefront/graphql).
+
+### Parameters
+
+| Parameter name | Type | Required? | Description |
+| - | - | - | - |
+| `document` | object `DocumentDecoration<TResult, TVariables>` | Yes | The GraphQL query or mutation you want to execute. It must be in the form of a string or a GraphQL AST (Abstract Syntax Tree) that defines the query. <br /><br /> The `DocumentDecoration` interface supports types from `@graphql-typed-document-node/core` and `TypedQueryDocumentNode`. These ensure the types of variables and results match. The document could be a GraphQL query or mutation. |
+| `variables` | object `TVariables` | No | Variables to be passed to the GraphQL query or mutation. <br /><br />This is a generic type constrained to `Record<string, unknown>`, meaning it can be any object where the keys are strings and the values are of any type. This allows you to pass variables into the query/mutation dynamically. |
```
chanceaclark, line 38: So we have to type `TVariables` set to `Record<string, unknown>` for TypeScript to have a base type, but `TVariables` actually gets dynamically typed to whatever variables are defined within the `document` string/GraphQL AST. I don't think it's worth mentioning that `Record<string, unknown>` here as `variables` is always going to be `undefined | (some implicit dynamic type)`.

docs/client.md
```diff
@@ -0,0 +1,156 @@
...
+
+### Deprecated
+
+The following method is deprecated and requires the `BIGCOMMERCE_ACCESS_TOKEN` to be set in the `.env` file.
+
+- `fetchShippingZones()`: retrieves shipping zones for the BigCommerce store. It sends an HTTP GET request to the [BigCommerce Management API](https://developer.bigcommerce.com/docs/rest-management/shipping-v2/shipping-zones), but simplifies how you would make a direct request to fetch shipping zones.
+
+## Fetch
+
+The `fetch()` method lets you send GraphQL queries or mutations to the [GraphQL Storefront API](https://developer.bigcommerce.com/docs/storefront/graphql).
```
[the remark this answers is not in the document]
bc-andreadao: Just rephrased it to make it less repetitive. If I remove the sentence, it looks like the `## Fetch` section could be referring to `fetchSitemapIndex` or `fetchShippingZones()` (rather than just the `fetch()` method)

docs/client.md
bc-andreadao: Good point, just removed the 3 methods

docs/client.md
bc-andreadao: Just removed it 👍

bc-andreadao:
@chanceaclark Thanks, I incorporated your feedback, let me know if there's anything else

chanceaclark (approved):

PR #1365, opened 2024-09-13
fix: do not write access token to env file

matthewvolk:
## What/Why?
A BigCommerce access token is no longer required to run Catalyst. If a merchant wants to use an access token, they should manually create one following: https://www.catalyst.dev/docs/environment-variables#bigcommerce_access_token

## Testing
Built and ran `node packages/create-catalyst/bin/index.cjs` locally, checked `.env.local` in created project to confirm access token not present:

```shell
BIGCOMMERCE_STORE_HASH=storehash
BIGCOMMERCE_CHANNEL_ID=12345
BIGCOMMERCE_CUSTOMER_IMPERSONATION_TOKEN=eyJ...

AUTH_SECRET=e85b...
CLIENT_LOGGER=false
ENABLE_ADMIN_ROUTE=true
NEXT_PUBLIC_DEFAULT_REVALIDATE_TARGET=3600
```

### DIFF
--- a/.changeset/small-sheep-divide.md
+++ b/.changeset/small-sheep-divide.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/create-catalyst": minor
+---
+
+Changes the default behavior of the `create-catalyst` CLI such that it no longer writes the access token created by the OAuth device flow to the created project's `.env.local` file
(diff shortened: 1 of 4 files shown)


chanceaclark (approved):
Hmm... I thought I did this already, but seems not 😅 Thanks for doing this!

PR #1382, opened 2024-09-16
chore(core): add translations changeset workflow

jorgemoya:
## What/Why?
Auto generate patch changeset for translations PRs.

Screenshot 2024-09-16 at 3 21 52 PM

## Testing
Tested it locally in this PR, generates patch and commits as the bot. No longer shown since I rebased to previous commit.

### DIFF
--- a/.github/workflows/translations-changeset.yml
+++ b/.github/workflows/translations-changeset.yml
@@ -4,13 +4,12 @@ on:
   pull_request:
     types:
       - opened
-      - reopened
     branches:
       - main
 
 jobs:
   create-translations-patch:
-    if: github.actor == 'jorgemoya'
+    if: github.actor == 'bc-svc-local'
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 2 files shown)


deini (approved):

.github/workflows/translations-changeset.yml
```diff
@@ -0,0 +1,41 @@
...
+          \"@bigcommerce/catalyst-core\": patch
+          ---
+
+          Update translations." > .changeset/translations-patch.md
+
+      - name: Commit changeset
+        run: |
+          git config --global user.name 'bc-svc-local'
+          git config --global user.email 'bc-svc-local@users.noreply.github.com'
+          git add .changeset/translations-patch.md
```
chanceaclark, line 34: What if we don't release for a couple weeks and there are multiple translations PRs? Is there any way we can add a rng postfix to the file name so we don't run into any file conflicts?

jorgemoya: I can see the benefit now of having multiple changesets, in the change log when it links to the commit. I will look into it.

chanceaclark (approved):

PR #1386, opened 2024-09-18
refactor(core): homepage

jorgemoya:
## What/Why?

- Uses vibes components to build home page.
  - Subscribe is WIP and not hooked up.
  - Cart count is not working.
- Enabled two locales for testing purposes.
 
## Testing
Locally.


core/components/header/_actions/search.ts
```diff
@@ -0,0 +1,14 @@
...
+import { redirect } from '~/i18n/routing';
+
+// eslint-disable-next-line @typescript-eslint/require-await
+export const search = async (formData: FormData) => {
+  const searchTerm = z.string().parse(formData.get('searchTerm'));
+
+  if (searchTerm) {
+    redirect(`/search?term=${searchTerm}`);
+  }
+};
```
jorgemoya, line 14: Seems like server actions need to be async functions, but there's no async function inside this. Is this the best way to solve this?

deini: Do we need this to be a server action? Could it be a client side `router.push`?

jorgemoya: Pushed an update to just use the client.

deini (approved):

PR #1391, opened 2024-09-23
Get lossy image from API instead of setting param in code

bookernath:
## What/Why?
Simplify image loader implementation by using the lossy param from the API, instead of setting it in code.

## Testing
Preview deployment


chanceaclark (approved):

jorgemoya (approved):

deini (approved):
nice

PR #1400, opened 2024-09-25
refactor(core): cart

jorgemoya:
## What/Why?
- Use soul Cart component in Cart page
- Skip shipping consignment for now (and extra summary fields related to this, like `tax`, `grandTotal`, etc). Only show `subtotal` and rely on redirected checkout to finish checkout.
- Update soul Cart component to be functional and fit our needs. Not final version of component. Skeleton, etc, will change.

## Testing
Locally.


jorgemoya:
Converted to draft while I look into an issue with actions not triggering correctly.

chanceaclark (approved):

PR #1401, opened 2024-09-26
Add dynamic robots.txt from API

bookernath:
## What/Why?
Use new robots.txt node to dynamically populate robots.txt details from control panel, replacing previous hardcoded approach.

We still dynamically insert the sitemap URL based on the env.

## Testing
Preview deployment

image


core/app/robots.txt/route.ts
```diff
@@ -0,0 +1,61 @@
...
+ * https://nextjs.org/docs/app/api-reference/file-conventions/metadata/robots
+ *
+ */
+
+import { getChannelIdFromLocale } from '~/channels.config';
+import { client } from '~/client';
+import { graphql } from '~/client/graphql';
+import { defaultLocale } from '~/i18n/routing';
+
+const ROBOTS_TXT_QUERY = graphql(`
```
chanceaclark, line 18:
```suggestion
const RobotsTxtQuery = graphql(`
```

core/app/robots.txt/route.ts
chanceaclark, line 1: I'm curious if this file needs to be in `core/app/[locale]` so we can actually do a channel-specific query? I think the `next-intl` routing mechanism will help route the request here 🤔

core/app/robots.txt/route.ts
```diff
@@ -0,0 +1,61 @@
...
+  process.env.NEXTAUTH_URL || process.env.VERCEL_PROJECT_PRODUCTION_URL || '',
+);
+
+export const GET = async () => {
+  const { data } = await client.fetch({
+    document: ROBOTS_TXT_QUERY,
+    channelId: getChannelIdFromLocale(defaultLocale),
+  });
+
+  let robotsTxtText = data.site.settings?.robotsTxt || '';
```
chanceaclark, line 49: 🍹 Seems redundant to have `TxtText`
```suggestion
  let robotsTxt = data.site.settings?.robotsTxt || '';
```

core/app/robots.txt/route.ts
bookernath: robots.txt must live on the root (in the standard)... maybe I shouldn't even be trying to make the query channel specific?

chanceaclark: Well we do have channel-specific robot settings so I think we do need to support that. Normally, yeah `robots.txt` should live in the root, but I am curious if next-intl will handle the routing if it's nested under a locale, e.g. `/robots.txt` would route to `/en/robots.txt`? The other channels wouldn't necessarily "work" as no request would be sent to `/da/robots.txt` but it'll enable multi-domain -> single Catalyst instance in the future.

bookernath: Multi-domain is a great point.

bookernath: After poking at this, I'm going to recommend we push forward as-is and re-evaluate at the time we add multi-domain support; this ended up being more complex than I expected.

core/app/robots.txt/route.ts
```diff
@@ -0,0 +1,62 @@
...
+  const { data } = await client.fetch({
+    document: RobotsTxtQuery,
+    channelId: getChannelIdFromLocale(defaultLocale),
+    fetchOptions: { cache: 'no-store' }, // disable caching to get the latest robots.txt at build time
+  });
+
+  let robotsTxt = data.site.settings?.robotsTxt || '';
+
+  // add sitemap to robots.txt
+  robotsTxt += `\nSitemap: ${baseUrl.origin}/sitemap.xml\n`;
```
chanceaclark, line 53: 🍹 
```suggestion
  const robotsTxt = `${data.site.settings?.robotsTxt ?? ''}\nSitemap: ${baseUrl.origin}/sitemap.xml\n`;
```

chanceaclark (approved):

PR #1402, opened 2024-09-26
fix(core): fixing the problem with submitting the password change form

bc-yevhenii-buliuk:
## What/Why?
This PR fixes the problem of sending the password change form, namely, the form was not sent after the first click on the submit button due to the validation of the input fields.

## Testing
locally

**before:**

https://github.com/user-attachments/assets/116fbd62-b32b-4b33-a1c0-754c4ba2e6f6

**after:**

https://github.com/user-attachments/assets/983b9a26-1d9a-480a-8ee5-fefaedcf51b6

### DIFF
--- a/.changeset/swift-jeans-jam.md
+++ b/.changeset/swift-jeans-jam.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+fixing the problem with submitting the password change form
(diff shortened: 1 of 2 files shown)


jorgemoya (approved):

bc-alexsaiannyi (approved):
👌

chanceaclark (approved):

PR #1410, opened 2024-09-30
Namespace /reset route to /login/forgot-password

bookernath:
## What/Why?
Reduce the number of top-level routes by namespacing:
`/reset` --> `/login/forgot-password`

## Testing

Reset:

https://github.com/user-attachments/assets/965a742d-5e78-4369-afb0-d9d59f64e989


jorgemoya:
The more I think about this, I think this should be `forget-password` and not `reset-password`. Other than that it looks good to me, except we need to group translations within `Login` for the `en` dictionary.

chanceaclark (approved):

PR #1416, opened 2024-10-03
test(core): Add Orders API utils and UI tests

bc-azvierieva:
## What/Why?
- Add Orders API utils and fixture
- Add Order UI tests

## Additional changes
New access token is needed to make tests work in CI. List of scopes to be added to already selected:
- **Orders:modify**

## Testing
image


core/tests/ui/e2e/order.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+  await expect(page.getByRole('heading', { name: 'Orders' })).toBeVisible();
+  await expect(page.getByText('No orders yet')).toBeVisible();
+});
+
+test('Order details are visible on Orders page', async ({ page, account, product, order }) => {
+  const customer = await account.create();
+
+  await customer.login();
+
+  const orderProduct = await product.create();
```
chanceaclark, line 33: Have we tried to use existing products instead of creating them programmatically?

bc-azvierieva: I didn't try that, but will do so. Do you mean something like Get all products -> filter them by stock/availability/other -> randomly get suitable product for order ? As I don't think always taking the same existing product is a good idea.

What we might improve by this is decreasing API requests amount from create, assign to channel, delete (3) to get all, and filter them (1). And bigger test data diversity. Is this the improvement you meant?

chanceaclark: I was just thinking about keeping it simple as we already know all our product ids, e.g. 
```ts
const orderDetails = await order.create(customer.id, 111);
```

bc-azvierieva: If we are not planning to have store provisioning for every test run any time soon, then we can go with this approach. It is indeed simpler, but makes tests dependant on specific test data. But since plenty of UI tests already depend on specific test data, we can keep it consistent then 🙂

chanceaclark: Yeah, I see us just using catalyst-latest for functional test suite for now. We can always reevaluate in the future if we need something like this.

bc-azvierieva: updated to use existing product. should we keep create product functionality for future use or better to remove now and we will add it in future if needed?

bc-azvierieva: done ✅

core/tests/ui/e2e/order.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  await expect(page.getByRole('link', { name: `Order # ${orderDetails.id}` })).toBeVisible();
+  await expect(page.getByText('Order placed')).toBeVisible();
+  await expect(page.getByText(formattedDate)).toBeVisible();
+  await expect(page.getByText('Total')).toBeVisible();
+  await expect(page.getByText(formattedTotal)).toBeVisible();
+  await expect(page.getByText(orderDetails.status)).toBeVisible();
+  await expect(page.getByRole('link', { name: orderProduct.name })).toBeVisible();
+  await expect(page.getByText(formattedProductPrice)).toBeVisible();
+});
```
chanceaclark, line 52: Can we get rid of the `ProductFactory` and inline the product information? I see the test looking like this:
```ts
test('Order details are visible on Orders page', async ({ page, account, order }) => {
  const customer = await account.create();

  await customer.login();

  // Customer id last and make it optional – in case we want to test guest shopper orders?
  const orderDetails = await order.create(88, customer.id);

  await page.goto('/account/orders/');

  await expect(page.getByRole('link', { name: `Order # ${orderDetails.id}` })).toBeVisible();
  await expect(page.getByText('Order placed')).toBeVisible();
  await expect(page.getByText(orderDetails.date_created)).toBeVisible();
  await expect(page.getByText('Total')).toBeVisible();
  await expect(page.getByText(orderDetails.total_inc_tax)).toBeVisible();
  await expect(page.getByText(orderDetails.status)).toBeVisible();
  await expect(page.getByRole('link', { name: ''[Sample] Chemex Coffeemaker 3 Cup
 })).toBeVisible();
  await expect(page.getByText('$49.50')).toBeVisible();
});
```

Also, we haven't been testing internationalization yet, so I don't think we need to format the dates/currencies yet. I think we might want to hold off till we support Single-channel multi-locale in Catalyst for this.

bc-azvierieva: format functions needed to convert "Thu, 03 Oct 2024 09:41:52 +0000" to "October 3, 2024" and "59.4000" to "$59.40"

bc-azvierieva: Ok, I'll update it

bc-azvierieva: done ✅

chanceaclark: >format functions needed to convert "Thu, 03 Oct 2024 09:41:52 +0000" to "October 3, 2024" and "59.4000" to "$59.40"

Ahh okay! Makes sense!

chanceaclark (approved):

core/tests/fixtures/utils/order/create.ts
```diff
@@ -0,0 +1,91 @@
...
+  billing_address: Address,
+});
+
+export async function createOrder(productId: number, customerId?: number) {
+  const status_id = 1;
+  const firstName = faker.person.firstName();
+  const lastName = faker.person.lastName();
+  const street1 = faker.location.streetAddress();
+  const city = faker.location.city();
+  const state = faker.location.state();
```
avattipalli, line 35: 🍺 Just to avoid some unknow issues, there's probably way to make sure the state and city are within United States since you're defaulting the country to US.

core/tests/ui/e2e/order.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+  await expect(page.getByRole('heading', { name: 'Orders' })).toBeVisible();
+  await expect(page.getByText('No orders yet')).toBeVisible();
+});
+
+test('Order details are visible on Orders page', async ({ page, account, order }) => {
+  const customer = await account.create();
+
+  await customer.login();
+
+  const orderDetails = await order.create(88, customer.id);
```
avattipalli, line 34: 🍺 If we can create an order through the UI like the end user instead of a call, we should do that but since I don't know if that feature is fully functional this is fine for now. Otherwise, it feels not completely e2e testing to me.

avattipalli (approved):
I have a couple of comments that just are opinionated but not a blocker 👍

yurytut1993 (approved):

core/tests/fixtures/utils/order/create.ts
bc-azvierieva: Good catch! Updated to using

`import { faker } from '@faker-js/faker/locale/en_US';`

core/tests/ui/e2e/order.spec.ts
bc-azvierieva: I totally see your point. Since we already have a test, where user adds product to the cart, navigates to checkout and places order - I considered this path as covered. 
I was thinking about creating an order via UI here, but it seems to be too long user journey for repeating it in every order related test.

PR #1419, opened 2024-10-03
fix(core): update register customer test

chanceaclark:
## What/Why?
Fixes the register customer test to use the created account info.

## Testing
See CI.

### DIFF
--- a/core/tests/fixtures/utils/account/create.ts
+++ b/core/tests/fixtures/utils/account/create.ts
@@ -55,6 +55,7 @@ export async function createAccount() {
               postal_code,
             },
           ],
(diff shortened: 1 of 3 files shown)


core/tests/fixtures/utils/account/create.ts
```diff
@@ -55,6 +55,7 @@ export async function createAccount() {
               postal_code,
             },
           ],
+          origin_channel_id: process.env.BIGCOMMERCE_CHANNEL_ID,
```
chanceaclark, line 58: This is to ensure we create the account on the same channel we are making the register customer mutation to. In the future we need to be aware of multi-channel account creation issues.

jorgemoya (approved):

jorgemoya (approved):

PR #1431, opened 2024-10-09
refactor(cli): add isExecException to check if an error is an ExecException

matthewvolk:
Adds a typeguard to narrow Errors thrown by `execSync` to be of `ExecException` types.

## Testing
Locally


chanceaclark (approved):

packages/create-catalyst/src/utils/is-exec-exception.ts
chanceaclark, line 1: We have `zod` in this package, so we could enforce this type with that instead 🤔

PR #1432, opened 2024-10-09
refactor(cli): add hasGitHubSSH function to check if GitHub SSH authentication is successful

matthewvolk:
> [!IMPORTANT]
> This PR stacks on #1431 

Adds a function to check if a user has SSH authentication set up with GitHub

## Testing
Locally

### DIFF
--- a/packages/create-catalyst/src/utils/has-github-ssh.ts
+++ b/packages/create-catalyst/src/utils/has-github-ssh.ts
@@ -1,0 +1,30 @@
+import { execSync } from 'child_process';
+
+import { isExecException } from './is-exec-exception';
+
+export function hasGitHubSSH(): boolean {
+  try {
+    // Attempt to connect to GitHub via SSH and capture the output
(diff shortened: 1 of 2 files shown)


packages/create-catalyst/src/utils/has-github-ssh.ts
```diff
@@ -0,0 +1,32 @@
...
+      const stdout = error.stdout ? error.stdout.toString() : '';
+      const stderr = error.stderr ? error.stderr.toString() : '';
+      const combinedOutput = stdout + stderr;
+
+      // Check if the output indicates successful authentication
+      if (combinedOutput.includes('successfully authenticated')) {
+        return true;
+      } else if (combinedOutput.includes('Permission denied')) {
+        return false;
+      }
```
chanceaclark, line 27: I believe `else if` is not needed here as it'll break out of the try/catch and return `false` on line 30.

chanceaclark (approved):

PR #1434, opened 2024-10-09
refactor(cli): update required Node.js version

matthewvolk:
Previously, we allowed any Node.js version 20 or greater. We have been experiencing issues with people using less-stable odd-numbered Node.js versions (like `21`) or Node.js versions that are still new and not something we have fully tested with Catalyst (like `22`)

## Testing
Locally

### DIFF
--- a/.changeset/small-experts-divide.md
+++ b/.changeset/small-experts-divide.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/create-catalyst": minor
+---
+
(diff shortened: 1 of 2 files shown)


chanceaclark (approved):

bookernath:
We can refactor this later, but ideally it would handle a range and just adjust the message if the current range was singular.

bookernath (approved):

PR #1435, opened 2024-10-09
refactor(cli)!: ensure pnpm is installed on the machine running the CLI

matthewvolk:
> [!IMPORTANT]
> This PR stacks on #1433

Ensure `pnpm` is installed on the machine running the CLI

## Testing
Locally

PR #1436, opened 2024-10-09
refactor(cli)!: ensure git is installed on the machine running the CLI

matthewvolk:
> [!IMPORTANT]
> This PR stacks on #1435

Ensure `git` is installed on the machine running the CLI

## Testing
Locally

### DIFF
--- a/packages/create-catalyst/src/commands/create.ts
+++ b/packages/create-catalyst/src/commands/create.ts
@@ -67,6 +67,13 @@ export const create = new Command('create')
   // eslint-disable-next-line complexity
   .action(async (options) => {
     try {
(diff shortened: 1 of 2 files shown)


packages/create-catalyst/src/commands/create.ts
```diff
@@ -66,6 +66,13 @@ export const create = new Command('create')
   )
   // eslint-disable-next-line complexity
   .action(async (options) => {
+    try {
+      execSync('git --version', { stdio: 'ignore' });
```
chanceaclark, line 70: Can we just use `which git` instead?

chanceaclark (approved):

PR #1438, opened 2024-10-09
refactor(cli)!: make cloneCatalyst clone `bigcommerce/catalyst`

matthewvolk:
> [!IMPORTANT]
> This PR stacks on #1437

Refactor `cloneCatalyst` so that it simply clones the `bigcommerce/catalyst` repo, configures remotes, and checks out an optional ref

## Testing
Locally


bookernath:
Seems like we should accept an arg to specify another repo while we're in here... seems like a simple addition and this is the right time to make it

christensenep (approved):

bookernath (approved):

deini (approved):

PR #1445, opened 2024-10-10
feat: optimistic updates on add to cart

deini:
## What/Why?
`Add to cart` (in PDP, PLP, Compare) now supports some optimistic updates. We keep the loading spinner on the button but show the toast and update the badge right away.

If the request fails it will revert the changes and show the error toast.

## Testing

https://github.com/user-attachments/assets/8e8f494e-240c-44ff-b82b-e5a2b80136c0

### DIFF
--- a/core/app/[locale]/(default)/compare/_components/add-to-cart/index.tsx
+++ b/core/app/[locale]/(default)/compare/_components/add-to-cart/index.tsx
@@ -3,66 +3,72 @@
 import { FragmentOf } from 'gql.tada';
 import { AlertCircle, Check } from 'lucide-react';
 import { useTranslations } from 'next-intl';
-import { useFormStatus } from 'react-dom';
+import { useTransition } from 'react';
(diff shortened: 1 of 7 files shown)


core/app/[locale]/(default)/compare/_components/add-to-cart/index.tsx
```diff
@@ -3,66 +3,72 @@
...
+
+    const formData = new FormData(event.currentTarget);
+    const quantity = Number(formData.get('quantity'));
 
-        if (result.error) {
-          toast.error(t('error'), {
-            icon: <AlertCircle className="text-error-secondary" />,
-          });
+    // Optimistic update
+    cart.increment(quantity);
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/compare/_components/add-to-cart/index.tsx:29]: 🍹 Just thinking it might be nice to explicitly mention it's an optimistic update with the fn name.
```suggestion
    cart.optimisticlyIncrement(quantity);
```

chanceaclark (approved):

PR #1449, opened 2024-10-13
Add new env variable for Trusted Proxy

bookernath:
## What/Why?
If the store has Trusted Proxy enabled, the trusted proxy secret can be stored as an env variable in which case it will be sent as a header on all requests.

## Testing
Verified via logs

### DIFF
--- a/packages/client/src/client.ts
+++ b/packages/client/src/client.ts
@@ -33,6 +33,7 @@ class Client<FetcherRequestInit extends RequestInit = RequestInit> {
   private beforeRequest?: (
     fetchOptions?: FetcherRequestInit,
   ) => Partial<FetcherRequestInit> | undefined;
+  private trustedProxySecret = process.env.BIGCOMMERCE_TRUSTED_PROXY_SECRET;
 
   constructor(private config: Config<FetcherRequestInit>) {
     if (!config.channelId) {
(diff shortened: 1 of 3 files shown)


packages/client/src/client.ts
```diff
@@ -33,6 +33,7 @@ class Client<FetcherRequestInit extends RequestInit = RequestInit> {
   private beforeRequest?: (
     fetchOptions?: FetcherRequestInit,
   ) => Partial<FetcherRequestInit> | undefined;
+  private trustedProxySecret?: string;
```
chanceaclark, line 36: I would move the setting of this outside the constructor. The type should resolve as `string | undefined` here, but just double check.

```suggestion
  private trustedProxySecret = process.env.BIGCOMMERCE_TRUSTED_PROXY_SECRET;
```

packages/client/src/client.ts
```diff
@@ -45,6 +46,7 @@ class Client<FetcherRequestInit extends RequestInit = RequestInit> {
       ? config.getChannelId
       : (defaultChannelId) => defaultChannelId;
     this.beforeRequest = config.beforeRequest;
+    this.trustedProxySecret = process.env.BIGCOMMERCE_TRUSTED_PROXY_SECRET;
```
chanceaclark, line 49: We don't need this now

chanceaclark (approved):

jorgemoya (approved):

PR #1451, opened 2024-10-13
Set explicit cache policy on favicon fetch at build time

bookernath:
## What/Why?
Explicitly no-store the favicon fetch at build time; ensures we get a fresh favicon every build.

## Testing
N/A


chanceaclark (commented):
Can you provide how you tested this? IIRC the cache should be fresh on each build 🤔

PR #1461, opened 2024-10-14
feat(cli): add ability to pass arbitrary GitHub repository to clone from

matthewvolk:
## What/Why?
This PR adds the ability to pass in an arbitrary repository in the form of `org-or-user/repository-name` that the `cloneCatalyst` function will use to determine which GitHub repository to clone locally. 

This PR also refactors the types for the arguments passed to the `cloneCatalyst` function; we want Typescript to complain if the arguments are not passed at all, but also want the ability to pass `undefined` as values for those arguments.

## Testing
Tested + verified locally

### DIFF
--- a/packages/create-catalyst/src/commands/create.ts
+++ b/packages/create-catalyst/src/commands/create.ts
@@ -23,7 +23,8 @@ export const create = new Command('create')
   .option('--access-token <token>', 'BigCommerce access token')
   .option('--channel-id <id>', 'BigCommerce channel ID')
   .option('--customer-impersonation-token <token>', 'BigCommerce customer impersonation token')
-  .option('--gh-ref <ref>', 'Clone a specific ref from the bigcommerce/catalyst repository')
+  .option('--gh-ref <ref>', 'Clone a specific ref from the source repository')
+  .option('--repository <repository>', 'GitHub repository to clone from', 'bigcommerce/catalyst')
   .addOption(
     new Option('--bigcommerce-hostname <hostname>', 'BigCommerce hostname')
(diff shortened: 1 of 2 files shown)


packages/create-catalyst/src/commands/create.ts
```diff
@@ -24,6 +24,7 @@ export const create = new Command('create')
   .option('--channel-id <id>', 'BigCommerce channel ID')
   .option('--customer-impersonation-token <token>', 'BigCommerce customer impersonation token')
   .option('--gh-ref <ref>', 'Clone a specific ref from the bigcommerce/catalyst repository')
```
matthewvolk, line 26: Should I adjust this description since now any repository can be cloned?

chanceaclark (approved):

packages/create-catalyst/src/utils/clone-catalyst.ts
```diff
@@ -7,12 +7,14 @@ export const cloneCatalyst = ({
   projectName,
   projectDir,
   ghRef,
+  repository,
 }: {
   projectName: string;
   projectDir: string;
-  ghRef?: string;
+  ghRef: string | undefined;
+  repository: string | undefined;
```
matthewvolk, line 15: Not entirely confident that this is best practice, I just thought it would be more helpful to explicitly require all params be passed to the function, even if the values for some params is `undefined`.

packages/create-catalyst/src/commands/create.ts
chanceaclark: Yeah, I'd adjust it.

PR #1464, opened 2024-10-15
Pass through BC channel ID as env var in basic workflow

jmahony:
## What/Why?
Currently, generate.cjs always defaults to channel ID 1. In our case, the Catalyst channel isn't 1 which caused the basic workflow to fail. This allows the channel to be overridden via a GH action secret.

## Testing
N/A


chanceaclark (approved):
Great catch! Thank you!

jmahony:
Thank you for the quick merge!

PR #1474, opened 2024-10-17
Wishlist Sheet on Product Card

yurytut1993:
## What/Why?
This pr adds Wishlist Sheet on Product Card
Also, as Wishlist Sheet now used not only in Account scope, component was replaced

https://github.com/user-attachments/assets/2adf9c52-9878-4fed-a050-a14174759a1e

## Testing
Locally


core/app/[locale]/(default)/(faceted)/category/[slug]/page-data.ts
```diff
@@ -6,6 +7,8 @@ import { graphql, VariablesOf } from '~/client/graphql';
 import { revalidate } from '~/client/revalidate-target';
 import { BreadcrumbsFragment } from '~/components/breadcrumbs/fragment';
 
+import { WishlistSheetFragment } from '../../../../../../components/wishlist-sheet/fragment';
```
Comment (bc-alexsaiannyi) [inline core/app/[locale]/(default)/(faceted)/category/[slug]/page-data.ts:10]: If it's possible maybe it would look prettier to use:

```
import { WishlistSheetFragment } from '~/components/wishlist-sheet/fragment';
```

that looks more consistent with line above 💅

core/components/product-card/index.tsx
```diff
@@ -48,6 +53,16 @@ export const ProductCard = ({
       price={price}
       showCompare={showCompare}
       subtitle={brand?.name}
+      wishlistSheet={
+        showWishlistSheet &&
+        wishlistsList && (
```
bc-alexsaiannyi, line 58: do we need to check `wishlistsList.length > 0` here ?

core/app/[locale]/(default)/(faceted)/category/[slug]/page-data.ts
Comment (yurytut1993) [reply core/app/[locale]/(default)/(faceted)/category/[slug]/page-data.ts]: for sure, I should rely less on import autocomplete 😁

core/components/product-card/index.tsx
yurytut1993: yes, because be we still need to be sure that wishlistsList was fetched and provided from upper scope

yurytut1993: no, because we always will have favourites wishlist, so we only need to be sure that `wishlistsList` is not `undefined`

core/app/[locale]/(default)/product/[slug]/page-data.ts
```diff
@@ -8,7 +8,7 @@ import { graphql, VariablesOf } from '~/client/graphql';
 import { revalidate } from '~/client/revalidate-target';
 import { BreadcrumbsFragment } from '~/components/breadcrumbs/fragment';
 
-import { WishlistSheetFragment } from '../../account/(tabs)/wishlists/_components/wishlist-sheet/fragment';
+import { WishlistSheetFragment } from '../../../../../components/wishlist-sheet/fragment';
```
Comment (bc-alexsaiannyi) [inline core/app/[locale]/(default)/product/[slug]/page-data.ts:11]: dont forget to rewrite here as well)

bc-alexsaiannyi (approved):
lgtm

PR #1475, opened 2024-10-17
Add distinct CLI UA

bookernath:
## What/Why?
Add CLI UA to make it easier to detect CLI API calls and troubleshoot issues

## Testing
Observed UA in logs


deini (approved):

packages/create-catalyst/src/utils/user-agent.ts
```diff
@@ -0,0 +1,64 @@
...
+  return 'Unknown OS';
+};
+
+const getEnv = () => {
+  if (isDevelopment) return 'Development';
+  if (isTest) return 'Test';
+  if (isProduction) return 'Production';
+
+  return 'Unknown Env';
+};
```
matthewvolk, line 32: When would the CLI be running in Test/Production? Or are we just adding them because we can? (with `std-env`)

bookernath: CI/CD

PR #1478, opened 2024-10-17
Update CLI scopes

bookernath:
## What/Why?
Update scopes requested by CLI
- Remove cart (no longer needed)
- Add both storefront API token scopes to ease transition to customer access tokens

## Testing
Tested locally


chanceaclark (approved):
I'm generally fine with the adding of the `store_storefront_api` but it's not really needed tbh... I am adding it in #1262 and I "ease" people into it per each commit on that PR

PR #1484, opened 2024-10-18
feat(client): handle GQL errors

jorgemoya:
> [!CAUTION]  
> Do not merge

## What/Why?
GQL requests that respond as `200` but have an `errors` field will now be properly handled by the client and throw a proper `BigCommerceGQLError` response with the message reason from the API. This will provide a more detailed description of why the GQL request errored out.

API errors will still be handled and attribute the errored status as the message with this change as `BigCommerceAPIError`.

## Testing
Tested this and now our requests will actually `throw` now if they error from the GQL request and show a valid reason.


chanceaclark:
Closing for now, @jorgemoya we can always come back to this one.

PR #1486, opened 2024-10-21
(REFERENCE) Add gift certificates page with purchase, check, and redeem tabs

becomevocal:
## What/Why?

Adds a gift certificate page with purchase, check, and redeem tab similar to what Cornerstone has.

Uses our REST API for checking the certificate balance, which requires an additional `GIFT_CERTIFICATE_V3_API_TOKEN` env var. Suggest creating a read-only token if you go this route. Adding a gift certificate to a cart uses GraphQL.

The route is at `/gift-certificates`.

### Here's a demo:

https://github.com/user-attachments/assets/c7458958-e89e-4b31-8b29-d06c623676e7

Under the hood it supports localization using our existing standards within Catalyst.

## Backstory

I've been looking for something tangible to test v0.dev with and this fit the bill! It ended up doing a pretty good job out of the box, with my first prompt that asked for a gift certificate balance checker using the BigCommerce API.

### Initial v0 prompt and result
Screenshot 2024-10-21 at 2 22 17 AM

Ended up having to switch out the components since it defaults to shadcn components, fix the Gift Certificate API request that was using an invalid route, and manually polish areas to follow our standards. Like using our GraphQL client.

Overall it's a good pairing experience, especially when thinking through app router / RSCs.

## Testing

🎥 See video ^

### DIFF
--- a/core/app/[locale]/(default)/cart/_components/cart-item.tsx
+++ b/core/app/[locale]/(default)/cart/_components/cart-item.tsx
@@ -4,7 +4,7 @@ import { FragmentOf, graphql } from '~/client/graphql';
 import { BcImage } from '~/components/bc-image';
 
 import { ItemQuantity } from './item-quantity';
-import { RemoveItem } from './remove-item';
+import { RemoveItem, RemoveGiftCertificate } from './remove-item';
 
 const PhysicalItemFragment = graphql(`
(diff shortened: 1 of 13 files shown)


chanceaclark:
We merged in Gift Certificates and are going to release it via Catalyst 1.3

PR #1502, opened 2024-10-22
chore(functional): Add tests for checkout and cart

avattipalli:
## What/Why?
Adds the following tests and assertions

- Verify cart is session-synched. This will be skipped until we have this feature ready.
- Return from checkout completion to verify logged in customer remains logged in.

## Testing
Verified locally and is also tested as part of the CI.


avattipalli:
Waiting for further improvements before expanding tests for these usecases. Closing this PR for now.

PR #1508, opened 2024-10-23
feat(core): Next 15 upgrade

deini:
## What/Why?
Upgrade to Next 15, started with codemods and did many manual changes.

## Testing
CI 👀


core/app/[locale]/(default)/product/[slug]/static/page.tsx
```diff
@@ -56,5 +57,42 @@ export async function generateStaticParams() {
   });
 }
 
+interface Props {
+  params: Promise<{ slug: string; locale: LocaleType }>;
+}
+
+export async function generateMetadata(props: Props): Promise<Metadata> {
```
Comment (deini) [inline core/app/[locale]/(default)/product/[slug]/static/page.tsx:64]: This one took a while to figure out, turns our our `generateMetadata` from the non-static page which we used to re-export is using `searchParams` which marks the page as dynamic even if we use `force-static`.

Hard to track as this doesn't fail on dev, only in production builds.

Had to duplicate this fn and remove the usage of searchParams, since it's a static page we never hit it if we have searchParams set anyway.

I would like to know how this used to work in Next 14 and if there are plans to add `searchParams` to the list of `force-static` and return empty on static builds.

packages/eslint-config-catalyst/react.js
```diff
@@ -6,6 +6,7 @@ const config = {
   rules: {
     'react/react-in-jsx-scope': 'off',
     'react/prefer-read-only-props': 'off',
+    'react/destructuring-assignment': 'off',
```
deini, line 9: With the new changes to Next 15 you need to await things like `searchParams` and `params`. Destructuring props makes this a bit annoying in some scenarios.

core/lib/bodl/index.ts
```diff
@@ -9,6 +9,8 @@ const bodl = new Bodl({
   },
 });
 
-bodl.initialize();
+if (typeof window !== 'undefined') {
+  bodl.initialize();
+}
```
chanceaclark, line 14: 💯

chanceaclark (approved):

core/app/[locale]/(default)/cart/_components/item-quantity/update-item-quantity.ts
```diff
@@ -36,10 +36,15 @@ export async function updateItemQuantity({
   variantEntityId,
   selectedOptions,
 }: UpdateProductQuantityParams) {
+  console.log('updateItemQuantity:start');
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/cart/_components/item-quantity/update-item-quantity.ts:39]: Remove logs?

PR #1511, opened 2024-10-24
fix(core): update styles for reset password validation

bc-alexsaiannyi:
## What/Why?
The PR improves styles for reset password page like:

- fix invalid email validation
- update colour & size for validation errors
- fix ReCaptcha colour

## Testing
locally

https://github.com/user-attachments/assets/e82af807-ba97-429a-8b9b-02f9035dc16e

change_password_before

change_password_after


yurytut1993 (approved):
👌

PR #1517, opened 2024-10-28
test(core): Add wishlist related tests

bc-azvierieva:
## What/Why?
- moved add/delete wishlist test from account.spec.ts to separate file wishlist.spec.ts
- test with adding product to wishlist
- test with creating new wishlist from drawer
- test to make sure guest user cannot use wishlist
- test to verify Favorites wishlist is default and cannot be deleted

## Testing
See regression test run in CI (there is separate ticket to fix button test in visual regression)

### DIFF
--- a/core/tests/ui/e2e/wishlist.spec.ts
+++ b/core/tests/ui/e2e/wishlist.spec.ts
@@ -9,6 +9,22 @@ test('Guest user is redirected to login upon adding product to wishlist', async 
   await expect(page).toHaveURL(/login/);
 });
 
+test('Favorites wishlist present by default and cannot be deleted', async ({ page, account }) => {
+  const customer = await account.create();
+
+  await customer.login();
+
+  await page.goto('/account/wishlists/');
+
+  await expect(page.getByText('Favorites')).toBeVisible();
+  await expect(page.getByRole('button', { name: 'Delete' })).toBeHidden();
+
+  await page.getByRole('link', { name: 'Favorites' }).click();
+  await page.getByRole('heading', { name: 'Favorites' }).waitFor();
+
+  // Need to check there is no Wishlist Actions to edit/delete wishlist
+});
+
 test('Add product to Favorites wishlist from PDP', async ({ page, account }) => {
   const customer = await account.create();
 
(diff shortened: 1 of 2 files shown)


yurytut1993 (approved):

core/tests/ui/e2e/wishlist.spec.ts
```diff
@@ -0,0 +1,87 @@
+import { expect, test } from '~/tests/fixtures';
+
+test('Guest user is redirected to login upon adding product to wishlist', async ({ page }) => {
```
avattipalli, line 3:
```suggestion
test('Guest user is required to register to create a wishlist', async ({ page }) => {
```

core/tests/ui/e2e/wishlist.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+  const customer = await account.create();
+
+  await customer.login();
+
+  await page.goto('/account/wishlists/');
+
+  await expect(page.getByText('Favorites')).toBeVisible();
+  await expect(page.getByRole('button', { name: 'Delete' })).toBeHidden();
+
+  await page.getByRole('link', { name: 'Favorites' }).click();
```
avattipalli, line 22: L21-22 feels like it's incomplete. I would rather remove that for now and just leave a comment to suggest(like you added below) that we need to add missing part of the test.

avattipalli (changes requested):
🍹 I would also suggest we cleanup the accounts at the end of the test wherever we are creating them.

bc-azvierieva:
> 🍹 I would also suggest we cleanup the accounts at the end of the test wherever we are creating them.

@avattipalli this is already done automatically for each test as a part of account fixture, see core/tests/fixtures/index.ts

`await accountFactory.cleanup();`

core/tests/ui/e2e/wishlist.spec.ts
bc-azvierieva: Makes sense, done!

core/tests/ui/e2e/wishlist.spec.ts
bc-azvierieva: Makes sense, done!

avattipalli (approved):

PR #1518, opened 2024-10-28
feat(cli): add ability to pass arbitrary env vars to write in .env.local

matthewvolk:
## What/Why?
Adds the ability to pass in arbitrary environment variables to the CLI

## Testing
Running the following command:
```
pnpm create @bigcommerce/catalyst@latest --env HI=BYE --env BYE=HI
```

Results in the following `.env.local` in the project created by the `create` command:
```
# .env.local

# ...
# Other environment variables
# ...

HI=BYE
BYE=HI
```


bookernath (approved):

chanceaclark:
Could we also allow a comma separated list? e.g. `--env FOO=BAR,BAZ=BUZZ`?

PR #1532, opened 2024-11-03
chore: APPSEC-705 update SECURITY.md

bc-donfran:
## What/Why?
Update SECURITY.md

## Testing
NA


chanceaclark (approved):
Thank you! The regression test failing is nothing to worry about here as we are going through a major version update of Next.js.

PR #1561, opened 2024-11-07
Wire up product description, custom fields, warranty

bookernath:
## What/Why?
Move away from hardcoded accordion to using the API data; insert HTML where relevant.

Next step - embedded component.

## Testing
image


migueloller (approved):
LGTM!

PR #1564, opened 2024-11-07
Wire up plain text description and custom fields on compare

bookernath:
## What/Why?
Use plaintextdescription and custom fields on compare page.

Need to look at styling next; this should be table-like.

## Testing
Reference:

https://cornerstone-light-demo.mybigcommerce.com/compare/111/110/109

image


migueloller (approved):
LGTM

PR #1572, opened 2024-11-07
Chore: Update FeateredImage CTA href

jamesqquick:
## What/Why?
The cta button on the featured image of the homepage was linking back to the home page. Now, it links to `/shop-all`. 

## Testing
Go to home page, scroll down to the featured image and click the `Shop Now` button.

CleanShot 2024-11-07 at 13 26 17@2x

### DIFF
--- a/core/app/[locale]/(default)/page.tsx
+++ b/core/app/[locale]/(default)/page.tsx
@@ -68,7 +68,7 @@ export default async function Home({ params: { locale } }: Props) {
       <FeaturedProductsCarousel products={newestProducts} title={t('Carousel.title')} />
 
       <FeaturedImage
-        cta={{ href: '/products', label: t('FeaturedImage.cta') }}
+        cta={{ href: '/shop-all', label: t('FeaturedImage.cta') }}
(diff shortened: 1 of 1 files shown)


core/app/[locale]/(default)/page.tsx
```diff
@@ -68,7 +68,7 @@ export default async function Home({ params: { locale } }: Props) {
       <FeaturedProductsCarousel products={newestProducts} title={t('Carousel.title')} />
 
       <FeaturedImage
-        cta={{ href: '/#', label: t('FeaturedImage.cta') }}
+        cta={{ href: '/shop-all', label: t('FeaturedImage.cta') }}
```
Comment (migueloller) [inline core/app/[locale]/(default)/page.tsx:71]: @jorgemoya, can you confirm if this is the way we want to add links? Do they need to be localized using `next-intl`?

Comment (jamesqquick) [reply core/app/[locale]/(default)/page.tsx:71]: I did a quick scan. I searched through `href` references and I didn't see anything existing that included a localized URL as far as I could tell.

migueloller (approved):

PR #1584, opened 2024-11-08
fix: add z-index to select to prevent it from being hidden

jorgemoya:
## What/Why?
Sets a z-index to prevent select from being hidden by other elements.

Before:

https://github.com/user-attachments/assets/b1db962f-b8b9-455c-9123-3bec2100839c

After:

https://github.com/user-attachments/assets/ca1177e3-4eec-4b43-8c16-acbab613d1b3

## Testing
Locally.


migueloller (commented):
We generally try to above z-index or only use them in stacking contexts where the z-indexes are in the same file. What particular component was this one fighting with?

jorgemoya:
> We generally try to above z-index or only use them in stacking contexts where the z-indexes are in the same file. What particular component was this one fighting with?

Generally yes I agree, I think Select is one of two cases were we need it to overlap whatever is being rendered behind. In this case, it clashes with the Navigation component.

migueloller (approved):
I'm good w/ this for now. Let's find a better approach later (if possible).

migueloller (approved):

PR #1587, opened 2024-11-08
Hide some Makeswift builtins

migueloller:
## What/Why?
We don't want the following builtins to appear in the Makeswift builder:
- Carousel
- Countdown
- Form
- Navigation

Also, we want the Button component to appear in the Makeswift toolbar instead of the menu

### Before

CleanShot 2024-11-08 at 15 17 59

### After

https://github.com/user-attachments/assets/1f3124d5-a213-4d6b-9e36-a3028b85f610


apledger (commented):
LGTM!

bookernath (approved):

bookernath (approved):

PR #1603, opened 2024-11-10
feat: ugprade to React 19 RC

migueloller:
Note, this commit also regenerated the pnpm-lock.yaml file to force @types/react and @types/react-dom to be overridden with types-react@rc and types-react-dom@rc, respectively.

## What/Why?
We're upgrading React to 19 RC so that we can use `useActionState` instead of `useFormState`. We need this because `useFormState` does not have an `isPending` value, which we rely on extensively.

We also had to upgrade to Next.js 15 to use React 19 RC.

### Before

https://github.com/user-attachments/assets/9caa2945-cbe2-45bd-8d60-8c53f38970a3

### Ater

https://github.com/user-attachments/assets/0b52292f-d0f1-4942-8924-3b22c3cde8da


migueloller:
Unfortunately this is causing a 500 in the PDP. I'm going to have to pivot in approach and attempt to build the `useActionState` interface using `useFormState` together with `useTransition`.

PR #1610, opened 2024-11-11
fix: show correct image on variants

jorgemoya:
## What/Why?
+ Append default image if found to Gallery, show result first.
+ Use first image from image in description.

https://github.com/user-attachments/assets/883b6435-69e6-408e-86cf-174c4d1ebb44

## Testing
Locally.


bookernath:
🍹 consider appending at the end, not the beginning

bookernath (approved):

PR #1614, opened 2024-11-11
feat: visually editable site theme

agurtovoy:
https://github.com/user-attachments/assets/5438eed0-9bf2-4ca5-898e-3c1fa38f93d6


core/app/[locale]/layout.tsx
```diff
@@ -113,20 +115,21 @@ export default function RootLayout({ children, params: { locale } }: Props) {
...
     >
-      <head>
-        <DraftModeScript appOrigin={process.env.MAKESWIFT_APP_ORIGIN} />
-      </head>
-      <body className="flex h-screen min-w-[375px] flex-col font-body">
-        <Notifications />
-        <NextIntlClientProvider locale={locale} messages={messages}>
-          <NuqsAdapter>
-            <MakeswiftProvider previewMode={draftMode().isEnabled}>
+      <MakeswiftProvider previewMode={draftMode().isEnabled}>
```
Comment (agurtovoy) [inline core/app/[locale]/layout.tsx:118]: Moved up `<MakeswiftProvider>` so that it wraps the whole document, including `<head>` where `<CssTheme>` is injecting the styles.

core/app/globals.css
```diff
@@ -3,19 +3,6 @@
...
-  --foreground: 0 0% 7%;
-  --success: 116 78% 65%;
-  --error: 0 100% 60%;
-  --warning: 40 100% 60%;
-  --info: 220 70% 45%;
-  --contrast-100: 0 0% 93%;
-  --contrast-200: 0 0% 82%;
-  --contrast-300: 0 0% 70%;
-  --contrast-400: 0 0% 54%;
-  --contrast-500: 0 0% 34%;
```
agurtovoy, line 18: Moved out of CSS to a Typescript file so we can reuse these values in theme component registration to provide props with default values.

fikrikarim (approved):

fikrikarim (approved):

core/app/globals.css
apledger: I'm wondering if there's a better way than deleting things for when we need to separate Makeswift related integration code from main. Ideally we don't have to delete anything from the base repo, just add Makeswift related code. Maybe that's wishful thinking.

apledger (approved):

core/app/globals.css
migueloller: I think this is a good goal and we can accomplish it by overriding these values with Makeswift CSS variables further down. That being said, sometimes conflicts are good as we would want to know when this list changes to that we can do the equivalent change on `integrations/makeswift`.

PR #1619, opened 2024-11-11
feat: make product detail visually editable

fikrikarim:
https://github.com/user-attachments/assets/2330174c-58eb-47c6-af98-d87ea25f5aa9

Notes:
- I haven't removed the description from the product description accordion, since I'm not sure what to show on the left side of the product description by default.
- I set the characterLimit of the plainTextDescription to 1200. This is adjustable.

### DIFF
--- a/core/makeswift/components/product-detail/client.tsx
+++ b/core/makeswift/components/product-detail/client.tsx
@@ -1,0 +1,84 @@
+'use client';
+
+import React, {
+  type ComponentPropsWithoutRef,
+  createContext,
+  forwardRef,
+  type PropsWithChildren,
+  type ReactNode,
+  type Ref,
+  useContext,
+  useMemo,
+} from 'react';
+
+import { ProductDetail } from '@/vibes/soul/sections/product-detail';
+
+type VibesProductDetailProps = ComponentPropsWithoutRef<typeof ProductDetail>;
+
+type Props = VibesProductDetailProps & {
+  product: VibesProductDetailProps['product'] & {
+    description: string;
+    plainTextDescription: string;
+  };
+};
+
+const PropsContext = createContext<Props | null>(null);
+
+export const PropsContextProvider = ({ value, children }: PropsWithChildren<{ value: Props }>) => (
+  <PropsContext.Provider value={value}>{children}</PropsContext.Provider>
+);
+
+export const DescriptionType = {
+  PlainText: 'PlainText',
+  RichText: 'RichText',
+  Custom: 'Custom',
+} as const;
+
+type DescriptionType = (typeof DescriptionType)[keyof typeof DescriptionType];
+
+export const MakeswiftProductDetail = forwardRef(
+  (
+    {
+      descriptionType,
+      slot,
+    }: {
+      descriptionType: DescriptionType;
+      slot: ReactNode;
+    },
+    ref: Ref<HTMLDivElement>,
+  ) => {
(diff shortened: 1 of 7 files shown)


bookernath (dismissed):

core/makeswift/components/product-detail/client.tsx
```diff
@@ -0,0 +1,84 @@
...
+          );
+        case DescriptionType.Custom:
+          return slot;
+      }
+    }, [descriptionType, context?.product, slot]);
+
+    if (context == null) {
+      console.error('No context provided for MakeswiftProductDetail');
+      return <p ref={ref}>There was an error rendering the product detail.</p>;
+    }
```
agurtovoy, line 70: I'd remove this check, personally. You are already handling the null context/product gracefully, and even those we are registering these components as private, I've been structuring them in a way where you can use them as regular editable client components should decide to make them public.

fikrikarim: I'd love to as well. I need to do this because I didn't want to provide the default props for the `useContext`, since we have to define this:
```
type Props<F extends Field> = {
  breadcrumbs?: Breadcrumb[];
  product: ProductDetailProduct;
  action: ProductDetailFormAction<F>;
  fields: F[];
};
```

Or do you think it's still better to setup an initial data there? @agurtovoy

core/makeswift/components/product-detail/register.tsx
```diff
@@ -0,0 +1,25 @@
...
+runtime.registerComponent(MakeswiftProductDetail, {
+  type: COMPONENT_TYPE,
+  label: 'MakeswiftProductDetail (private)',
+  hidden: true,
+  props: {
+    descriptionType: Select({
+      label: 'Description',
+      options: [
+        { label: 'From BC (plain text)', value: DescriptionType.PlainText },
+        { label: 'From BC (rich text)', value: DescriptionType.RichText },
```
agurtovoy, line 18: I know this is what @apledger suggested in the thread, but "From BC..." feels off to me. May be "From the catalog ..."?

core/vibes/soul/sections/product-detail/index.tsx
```diff
@@ -55,7 +57,7 @@ export function ProductDetail<F extends Field>({ product, action, fields, breadc
             </div>
 
             {product.description != null && product.description !== '' && (
-              <p className="mb-6 text-contrast-500">{product.description}</p>
+              <div className="mb-6 text-contrast-500">{product.description}</div>
```
agurtovoy, line 60: Need a corresponding PR for this in Vibes.

agurtovoy (dismissed):
Left some minor comments, thank you for taking this! 🙏

core/makeswift/components/product-detail/client.tsx
agurtovoy: Hmm, I see your dilemma now. I actually can't think of a better way to handle this, I guess we just need to be content that some embedded components can never become freestanding.

fikrikarim: Sounds good.

core/makeswift/components/product-detail/register.tsx
agurtovoy: Changed to "Catalog (plain text)" etc.

agurtovoy (approved):

PR #1623, opened 2024-11-11
chore(core): update to next@15

chanceaclark:
## What/Why?
Updates next to v15.

## Testing
Manual + Regression Suite


bookernath (approved):
💚

PR #1639, opened 2024-11-13
feat(core): add customer account orders functionality

bc-alexsaiannyi:
## What/Why?
This PR combines and brings Order functionality to Customer's Account.

## Testing
locally

https://github.com/user-attachments/assets/2f0bb8a1-3394-403d-8cad-b5868a5a3a4a


bookernath:
@bc-alexsaiannyi I think we can squash this into one commit with a single changeset

core/app/[locale]/(default)/account/(tabs)/orders/page-data.ts
```diff
@@ -0,0 +1,369 @@
...
+      return {
+        lineItems: extendedLineItems,
+        shipments: removeEdgesAndNodes(shipments),
+        ...otherItems,
+      };
+    }),
+  );
+
+  return shippingConsignments;
+};
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/account/(tabs)/orders/page-data.ts:248]: 🍹 This page is going to be really slow with a large amount of `consignments` & `lineItems` due to the `O(n^2)` amount of promises happening here. We probably need to refactor this to be more efficient. Maybe we can use some suspense boundaries instead to fetch while the page is rendered.

core/app/[locale]/(default)/account/(tabs)/orders/page-data.ts
```diff
@@ -0,0 +1,369 @@
...
+      paymentInfo: {
+        billingAddress: order.billingAddress,
+        // TODO: add payments data
+      },
+      consignments: {
+        shipping,
+      },
+    };
+
+    return data;
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/account/(tabs)/orders/page-data.ts:365]: 🍹 We should probably do all the remapping in application code rather than at the boundary.

chanceaclark (approved):
I think there's a bit of refactoring needed here to optimize for performance. I left 🍹's where I think that is the case. I'm not blocking this PR from going through though, but we should probably tackle these optimizations during release freeze.

core/app/[locale]/(default)/account/(tabs)/orders/_components/orders-content.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+import { Pagination } from '~/components/ui/pagination';
+
+import { TabHeading } from '../../_components/tab-heading';
+import { getCustomerOrders, getOrderDetails } from '../page-data';
+
+import { OrderDetails } from './order-details';
+import { OrdersList } from './orders-list';
+
+type CustomerOrders = NonNullable<Awaited<ReturnType<typeof getCustomerOrders>>>;
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/account/(tabs)/orders/_components/orders-content.tsx:12]: Can you use `ExistingResultType` util?

core/app/[locale]/(default)/account/(tabs)/orders/page.tsx
```diff
@@ -0,0 +1,32 @@
+import { notFound } from 'next/navigation';
+
+import { OrdersContent } from './_components/orders-content';
+import { getCustomerOrders } from './page-data';
+
+interface Props {
+  searchParams: Promise<{
+    [key: string]: string | string[] | undefined;
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/account/(tabs)/orders/page.tsx:8]: Is this required? 

Edit: Ohh, it's because we get the order id. I think we should use slugs and individual pages per order, not search params.

So to view an individual order it should be `/account/orders/[id]`

core/app/[locale]/(default)/account/(tabs)/orders/page.tsx
```diff
@@ -0,0 +1,32 @@
...
+    before?: string;
+    after?: string;
+    order?: string;
+  }>;
+}
+
+export default async function OrdersPage({ searchParams }: Props) {
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/account/(tabs)/orders/page.tsx:15]:
```suggestion
export default async function Orders({ searchParams }: Props) {
```

jorgemoya (dismissed):
For order details, we should rely on slugs to fetch details. Lets create a page for `/account/orders/[id]` to display order content, instead of using search params to know what to render.

We also should rely more on Suspense boundaries and loading states to make sure this page is speedy.

core/app/[locale]/(default)/account/(tabs)/orders/page-data.ts
Comment (bc-alexsaiannyi) [reply core/app/[locale]/(default)/account/(tabs)/orders/page-data.ts]: already fixed

bc-alexsaiannyi:
Hey @jorgemoya, I think that mentioned concerns have been addressed in separate PRs that are already merged in this branch. Can you please look at it once more if we're good to go🙏

chanceaclark (approved):

PR #1640, opened 2024-11-14
refactor(core): add filters to orders query

bc-alexsaiannyi:
## What/Why?
This PR extends `CustomerOrders` query with ability to filter out results.

## Testing
locally

**Filtering by Status**:
Screenshot 2024-11-14 at 13 58 11

Screenshot 2024-11-14 at 13 58 22

### DIFF
--- a/core/app/[locale]/(default)/account/(tabs)/orders/page-data.ts
+++ b/core/app/[locale]/(default)/account/(tabs)/orders/page-data.ts
@@ -31,9 +31,15 @@ const OrderShipmentFragment = graphql(`
 
 const CustomerAllOrders = graphql(
   `
-    query CustomerAllOrders($after: String, $before: String, $first: Int, $last: Int) {
+    query CustomerAllOrders(
+      $after: String
+      $before: String
+      $first: Int
+      $last: Int
+      $filters: OrdersFiltersInput
+    ) {
       customer {
-        orders(after: $after, before: $before, first: $first, last: $last) {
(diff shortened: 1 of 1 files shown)


core/app/[locale]/(default)/account/(tabs)/orders/page-data.ts
```diff
@@ -241,20 +247,37 @@ const addProductAttributesToShippingConsignments = async (consignments: Shipping
...
 
 export const getCustomerOrders = cache(
-  async ({ before = '', after = '', limit = 2 }: CustomerOrdersArgs) => {
+  async ({
+    before = '',
+    after = '',
+    filterByStatus,
+    filterByDateRange,
+    limit = 2,
+  }: CustomerOrdersArgs) => {
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/account/(tabs)/orders/page-data.ts:269]: 🍹 We might be able to get away with using the variables as the arguments. No action, but just a thought.
```suggestion
  }: VariablesOf<typeof CustomerAllOrders>) => {
```

chanceaclark (approved):

core/app/[locale]/(default)/account/(tabs)/orders/page-data.ts
Comment (bc-alexsaiannyi) [reply core/app/[locale]/(default)/account/(tabs)/orders/page-data.ts:269]: Yeah, I agree. But for now I'd prefer to leave it and update later since we use quite similar number of arguments across all functions where we have pagination data. It might be worth to go and replace all of them at once🤔

PR #1646, opened 2024-11-15
fix: product-carousel and product-list now sort properly, included append checkbox control

drewpledger:
`products-list.makeswift.tsx`

https://github.com/user-attachments/assets/c7921d2b-f532-4c11-8bfc-cf8aea8c7dac

`products-carousel.makeswift.tsx`

https://github.com/user-attachments/assets/a2dd2fb1-ba6c-4903-afdd-947ee90c7694

### DIFF
--- a/core/makeswift/components/products-carousel/products-carousel.makeswift.tsx
+++ b/core/makeswift/components/products-carousel/products-carousel.makeswift.tsx
@@ -2,6 +2,7 @@
 
 import { removeEdgesAndNodes } from '@bigcommerce/catalyst-client';
 import {
+  Checkbox,
   Combobox,
   Image,
   Link,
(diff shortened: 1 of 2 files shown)


core/makeswift/components/products-carousel/products-carousel.makeswift.tsx
```diff
@@ -116,7 +141,7 @@ runtime.registerComponent(
         }
 
         return {
-          id: title ?? '',
+          id: index.toString(),
```
migueloller, line 144: We should use a stable identifier here instead of the index

migueloller (changes requested):

core/makeswift/components/products-carousel/products-carousel.makeswift.tsx
drewpledger: Should we be forcing users to enter an ID or something? Alan was saying using index is the most intuitive way outside of entering one in manually

migueloller: Don't we have the ID of the product?

PR #1647, opened 2024-11-16
refactor(core): update submit create account errors message

bc-alexsaiannyi:
## What/Why?
This PR adds helper for submit error messages for `create-account` form with aim to extend it further on other account related forms. As a result we have a list with separated submit errors.  

## Testing
locally

https://github.com/user-attachments/assets/0d2c9d06-e8bb-43d9-b481-aa77f913db00


core/app/[locale]/(default)/(auth)/register/_actions/register-customer.ts
```diff
@@ -84,7 +86,7 @@ export const registerCustomer = async ({ formData, reCaptchaToken }: RegisterCus
 
     return {
       status: 'error',
-      error: result.errors.map((error) => error.message).join('\n'),
+      error: createErrorsList(result.errors),
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/(auth)/register/_actions/register-customer.ts:89]: Why don't we just change `error` to be `errors` and update the UI to render a list `ul` instead. It'll be better for accessibility instead of using a unicode character to fake it.

Comment (bc-alexsaiannyi) [reply core/app/[locale]/(default)/(auth)/register/_actions/register-customer.ts]: Redone. Please check and if it looks appropriate I will extend to other account related forms

chanceaclark (approved):
Thanks for updating it!

PR #1651, opened 2024-11-18
fix(core): refresh the entire list of addresses after deleting an address

bc-yevhenii-buliuk:
…

## What/Why?
This PR fixes the display of the existing address list after removing all visible addresses on the page.
Also this PR updates the entire list of addresses after deleting an address.

## Testing
locally

**before:**

https://github.com/user-attachments/assets/bdc81450-2b76-47a5-9ea9-ff202c2d4e43

**after:**

https://github.com/user-attachments/assets/dd9f46b3-d90d-4954-a8aa-fd3010d48ebb

### DIFF
--- a/core/app/[locale]/(default)/account/(tabs)/addresses/_actions/delete-address.ts
+++ b/core/app/[locale]/(default)/account/(tabs)/addresses/_actions/delete-address.ts
@@ -8,13 +8,6 @@ import { client } from '~/client';
 import { graphql } from '~/client/graphql';
 
 import { State } from '../../settings/change-password/_actions/change-password';
-import { Addresses } from '../_components/address-book';
-import { SearchParams } from '../page';
-import { getCustomerAddresses } from '../page-data';
-
-interface DeleteAddressResponse extends State {
-  addresses?: Addresses;
-}
 
 const DeleteCustomerAddressMutation = graphql(`
   mutation DeleteCustomerAddressMutation(
@@ -39,13 +32,9 @@ const DeleteCustomerAddressMutation = graphql(`
   }
 `);
 
-export const deleteAddress = async (
-  addressId: number,
-  searchParams?: SearchParams,
-): Promise<DeleteAddressResponse> => {
+export const deleteAddress = async (addressId: number): Promise<State> => {
   const t = await getTranslations('Account.Addresses.Delete');
   const customerAccessToken = await getSessionCustomerAccessToken();
-  const { before, after } = searchParams || {};
 
   try {
     const response = await client.fetch({
@@ -64,16 +53,7 @@ export const deleteAddress = async (
     revalidatePath('/account/addresses', 'page');
 
     if (result.errors.length === 0) {
-      const updatedData = await getCustomerAddresses({
-        ...(after && { after }),
-        ...(before && { before }),
-      });
-
-      return {
-        status: 'success',
-        message: t('success'),
-        addresses: updatedData?.addresses || [],
-      };
+      return { status: 'success', message: t('success') };
     }
 
     return {
--- a/core/app/[locale]/(default)/account/(tabs)/addresses/_actions/delete-address.ts
+++ b/core/app/[locale]/(default)/account/(tabs)/addresses/_actions/delete-address.ts
@@ -8,6 +8,13 @@ import { client } from '~/client';
 import { graphql } from '~/client/graphql';
 
 import { State } from '../../settings/change-password/_actions/change-password';
+import { Addresses } from '../_components/address-book';
+import { SearchParams } from '../page';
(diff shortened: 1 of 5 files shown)


core/app/[locale]/(default)/account/(tabs)/addresses/page.tsx
```diff
@@ -24,13 +26,17 @@ export async function generateMetadata() {
   };
 }
 
+export enum AddressPageLimit {
+  Default = 3,
```
Comment (bc-alexsaiannyi) [inline core/app/[locale]/(default)/account/(tabs)/addresses/page.tsx:30]: Just curious if you discuss the qty with Turner or Eric? I believe previously we have increased it to 10🤔

core/app/[locale]/(default)/account/(tabs)/addresses/page.tsx
```diff
@@ -24,13 +26,17 @@ export async function generateMetadata() {
...
+}
+
 export default async function Addresses({ searchParams }: Props) {
   const { before, after } = await searchParams;
 
   const data = await getCustomerAddresses({
     ...(after && { after }),
     ...(before && { before }),
-    limit: 10,
+    limit: AddressPageLimit.Default,
```
Comment (bc-alexsaiannyi) [inline core/app/[locale]/(default)/account/(tabs)/addresses/page.tsx:39]: Also we have a default value for `limit` argument in `getCustomerAddresses` so maybe we can little bit simplify here and instead of `enum` use its default value, wdyt?💅

core/app/[locale]/(default)/account/(tabs)/addresses/page.tsx
```diff
@@ -43,7 +49,13 @@ export default async function Addresses({ searchParams }: Props) {
   return (
     <>
       <TabHeading heading="addresses" />
-      <AddressBook addressesCount={addressesCount} customerAddresses={addresses} key={endCursor}>
+      <AddressBook 
+        addressesCount={addressesCount}
+        customerAddresses={addresses}
+        key={endCursor}
+        pageInfo={pageInfo}
```
Comment (bc-alexsaiannyi) [inline core/app/[locale]/(default)/account/(tabs)/addresses/page.tsx:56]: Small suggestion here - we use key for [resetting](https://react.dev/reference/react/useState#resetting-state-with-a-key) state. So if we do component's refactoring then we can go further and optimize its props in the next way.

on deleting all addresses on a Page we check if there a Previous page and if so we move there:
```
if (addressBook.length === 0 && hasPreviousPage) {
      router.push(`/account/addresses/?before=${startCursor}`);
    }
```
So basically we just need `hasPreviousPage` & `startCursor` from `PageInfo` data. Based on this we can add them to `AddressBook` Props directly and remove `key` and `PageInfo` instead. Maybe I missed smth, wdyt?

core/app/[locale]/(default)/account/(tabs)/addresses/page.tsx
Comment (bc-yevhenii-buliuk) [reply core/app/[locale]/(default)/account/(tabs)/addresses/page.tsx]: You are right, I changed it to 3 during work and local testing. I changed it back to 10, thanks for noticing.

core/app/[locale]/(default)/account/(tabs)/addresses/page.tsx
Comment (bc-yevhenii-buliuk) [reply core/app/[locale]/(default)/account/(tabs)/addresses/page.tsx]: Makes sense. Done!

core/app/[locale]/(default)/account/(tabs)/addresses/page.tsx
Comment (bc-yevhenii-buliuk) [reply core/app/[locale]/(default)/account/(tabs)/addresses/page.tsx]: I optimized the props passing, but removing the `key` breaks the pagination, so I left it to reset the component state when switching between pages. Thanks for the suggestion!

bc-alexsaiannyi (approved):

core/app/[locale]/(default)/account/(tabs)/addresses/_actions/delete-address.ts
```diff
@@ -53,7 +64,16 @@ export const deleteAddress = async (addressId: number): Promise<State> => {
...
+      const updatedData = await getCustomerAddresses({
+        ...(after && { after }),
+        ...(before && { before }),
+      });
+
+      return {
+        status: 'success',
+        message: t('success'),
+        addresses: updatedData?.addresses || [],
+      };
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/account/(tabs)/addresses/_actions/delete-address.ts:76]: Why don't we just use Next.js [revalidation](https://nextjs.org/docs/app/api-reference/functions/revalidateTag) + reload the page?

deini (approved):

core/app/[locale]/(default)/account/(tabs)/addresses/_actions/delete-address.ts
Comment (bc-yevhenii-buliuk) [reply core/app/[locale]/(default)/account/(tabs)/addresses/_actions/delete-address.ts]: @chanceaclark thanks for the suggestion, I simplified the logic and left `revalidatePath` as it is done in adding and editing address. Could you double check the implementation?

chanceaclark (approved):

PR #1655, opened 2024-11-18
chore(core): update tailwind config and layout to use Soul config

jorgemoya:
## What/Why?

- Add tailwind setup per VIBES/Soul docs
- Use electric theme
- Update layout to use new fonts and remove container classes
- Remove styles from Link component
- Fix eslint conflicts from new tailwind config

## Testing
Locally, pages use new tailwind config.

### DIFF
--- a/core/app/[locale]/layout.tsx
+++ b/core/app/[locale]/layout.tsx
@@ -1,7 +1,7 @@
 import { Analytics } from '@vercel/analytics/react';
 import { SpeedInsights } from '@vercel/speed-insights/next';
 import type { Metadata } from 'next';
-import { Inter } from 'next/font/google';
+import { DM_Serif_Text, Inter } from 'next/font/google';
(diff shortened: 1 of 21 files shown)


chanceaclark (approved):

core/app/[locale]/layout.tsx
```diff
@@ -16,9 +16,16 @@ import { Notifications } from '../notifications';
...
+  variable: '--font-family-body',
+});
+
+const dm_serif_text = DM_Serif_Text({
   display: 'swap',
-  variable: '--font-inter',
+  subsets: ['latin'],
+  weight: '400',
+  variable: '--font-family-heading',
 });
```
Comment (migueloller) [inline core/app/[locale]/layout.tsx:29]: Looks like we didn't bring in `--font-family-mono`

Comment (jorgemoya) [reply core/app/[locale]/layout.tsx:29]: I pulled this in from the latest soul docs. Will add, but we need to fix in the docs.

PR #1659, opened 2024-11-19
Option to provide non translated strings in a fallback locale

thebigrick:
## What/Why?
If a new string is added, the translation must be available in any language. If not, the dotted path notation will be displayed in its place. I introduced the capability of using a fallback language for the missing strings.

## Testing
- Add a new string to the `messages/en.json` file
- Reference the string from any component
- Switch to a locale different from `en`
- The string should appear in English
- Add the string to the correct locale
- The string should appear in the correct locale

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -30,3 +30,8 @@ TURBO_REMOTE_CACHE_SIGNATURE_KEY=
 # https://nextjs.org/docs/app/building-your-application/caching#data-cache
 # This sets a sensible revalidation target for cached requests
 DEFAULT_REVALIDATE_TARGET=3600
+
+# Default locale fallback to use when a string is not found in the current locale.
+# This should be a valid locale code, such as `en` or `it`.
+# If not set, the default locale is `en`.
+LOCALE_FALLBACK=
--- a/core/i18n/request.ts
+++ b/core/i18n/request.ts
@@ -1,7 +1,9 @@
 import { notFound } from 'next/navigation';
 import { getRequestConfig } from 'next-intl/server';
 
-import { locales } from './routing';
+import { fallbackLocale, locales } from './routing';
+
+import deepmerge from 'deepmerge';
 
 export default getRequestConfig(async ({ requestLocale }) => {
   const locale = await requestLocale;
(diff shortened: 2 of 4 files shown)


.env.example
```diff
@@ -30,3 +30,8 @@ TURBO_REMOTE_CACHE_SIGNATURE_KEY=
 # https://nextjs.org/docs/app/building-your-application/caching#data-cache
 # This sets a sensible revalidation target for cached requests
 DEFAULT_REVALIDATE_TARGET=3600
+
+# Default locale fallback to use when a string is not found in the current locale.
+# This should be a valid locale code, such as `en` or `it`.
+# If not set, the default locale is `en`.
+LOCALE_FALLBACK=
```
deini, line 37: Can we use `defaultLocale` instead of this env var?

core/i18n/request.ts
```diff
@@ -11,7 +13,12 @@ export default getRequestConfig(async ({ requestLocale }) => {
...
   return {
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-member-access
-    messages: (await import(`../messages/${locale}.json`)).default,
+    locale,
+    messages: deepmerge(
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument,@typescript-eslint/no-unsafe-member-access
+      (await import(`../messages/${fallbackLocale}.json`)).default,
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-argument,@typescript-eslint/no-unsafe-member-access
+      (await import(`../messages/${locale}.json`)).default,
+    ),
```
deini, line 22: Can we skip this if `locale === fallbackLocale`?

thebigrick: Yep, good point :) ... I was so focused on having a fallback different from local that I missed the most obvious scenario ;). I will fix it right now.

.env.example
chanceaclark: We run into the same issue here – we can't guarantee the translations are going to be present for the fallback locale. And looking at the logic of this code, we are essentially just providing two "defaults": `defaultLocale` and `fallbackLocale`. We should probably just have one, which should just be `defaultLocale`.

thebigrick: I see your point. My concern is about the following scenario I have been facing, and I would like your point of view on that:

I was working on some missing translations in the search bar, so I prepared a contribution to add them, providing both the implementation of `useTranslation` and the entries in the `en.json`. It is a common practice for every contributor to provide at least the English version.

The translation team will take time to translate my new entries into every language, but the contribution fixed one issue and could be available. Now imagine I need this fix on a store with a `defaultLocale` pointed to `it`. It will display the string as a dotted path notation — much worse than having it in English.

I 100% agree with you that we cannot ensure the fallback language is complete, so I set it to `en` as the default. Because, in most cases, it is full.

I also agree that having two variables could be confusing. One option is not to provide an environment variable and force the fallback to English.

jorgemoya: Couldn't we just hardcore `en`? Considering Catalyst is built in `en` principally, this would make it work as expected most of the time. It will be up to the users of Catalyst to change this fallback (since they will be maintaining their locale strings anyway, or rebasing).

thebigrick: I agree with @jorgemoya. The best option is removing it and hardcoding `en` as a fallback. Having the env variable is just confusing.

thebigrick: I modified the pull request accordingly.

deini:
Looks good, can we add a changeset please?

chanceaclark (approved):

PR #1670, opened 2024-11-21
Remove public submodule

jmahony:
## What/Why?
The submodule doesn't seem to be used and it's causing my build in Netlify to fail with the error: 
```
6:48:19 AM: Failed during stage 'preparing repo': Error checking out submodules: fatal: No url found for submodule path 'public' in .gitmodules
: exit status 128: fatal: No url found for submodule path 'public' in .gitmodules
: exit status 128
6:48:19 AM: Error checking out submodules: fatal: No url found for submodule path 'public' in .gitmodules
: exit status 128
6:48:19 AM: Failing build: Failed to prepare repo
```


chanceaclark (approved):
Thanks for removing this! Definitely glanced over this.

PR #1682, opened 2024-11-22
feat(core): add subscribe form to Homepage

jorgemoya:
## What
+ Add VIBES components
+ Install conform library
+ Use component
+ Update dictionary
+ Action is not included because this feature is not yet available.

Screenshot 2024-11-22 at 4 34 16 PM

## Testing
Locally it renders.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -180,9 +180,6 @@ importers:
       sharp:
         specifier: ^0.33.5
         version: 0.33.5
-      sonner:
-        specifier: ^1.7.0
-        version: 1.7.1(react-dom@19.0.0-rc-5c56b873-20241107(react@19.0.0-rc-5c56b873-20241107))(react@19.0.0-rc-5c56b873-20241107)
       tailwind-merge:
         specifier: ^2.5.5
         version: 2.5.5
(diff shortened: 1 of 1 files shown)


core/app/[locale]/(default)/page.tsx
```diff
@@ -89,6 +90,14 @@ export default async function Home({ params }: Props) {
         products={bestSellingProducts}
         title={t('BestSellingProducts.title')}
       />
+
+      <Subscribe
+        // @ts-expect-error - We don't have an action ready for this component. TODO: add action
+        action={null}
```
Comment (jamesqquick) [inline core/app/[locale]/(default)/page.tsx:96]: Should this action be defined before we merge this?

core/app/[locale]/(default)/page.tsx
```diff
@@ -89,6 +90,14 @@ export default async function Home({ params }: Props) {
...
+        description="Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor."
```
Comment (jamesqquick) [inline core/app/[locale]/(default)/page.tsx:97]: What did we decide on lorem ipsum text?

Comment (jorgemoya) [reply core/app/[locale]/(default)/page.tsx:97]: Good callout, will change this!

core/app/[locale]/(default)/page.tsx
Comment (migueloller) [reply core/app/[locale]/(default)/page.tsx]: Instead of passing `null` with a `@ts-expect-error` and the very least we should implement a stub for the action that shows an alert or logs something to the console server-side, no?

migueloller (changes requested):
The code looks good to me but the interaction feels incomplete. I feel that even if we don't have the functionality to add the email to a newsletter just yet, the UI should show some success message after submission, even if the message is "We still don't have a newsletter, that feature is coming soon!". Once we have the feature it will just be a matter of updating the implementation of the server action.

Also, I feel like we should test and make sure that the errors states are handled properly. Perhaps it's time to implement alerts? We could make it so if the email has `+error` (e.g., `miguel+error@makeswift.com`) it errors so we can test it?

jorgemoya:
Valid points @migueloller I think we definitely need to implement alerts and error validation.

migueloller (approved):

jorgemoya:
Bumped the components to include latest changes and added a dummy server action for now.

core/components/subscribe/_actions/subscribe.ts
```diff
@@ -0,0 +1,25 @@
...
+export const subscribe = async (
+  _lastResult: { lastResult: SubmissionResult | null },
+  formData: FormData,
+) => {
+  const t = await getTranslations('Components.Subscribe');
+
+  const submission = parseWithZod(formData, { schema });
+
+  if (submission.status !== 'success') {
+    return { lastResult: submission.reply() };
```
migueloller, line 18: Should we reply with a form error here? This will show, for example, if the user provides an invalid email address

migueloller (approved):

core/components/subscribe/_actions/subscribe.ts
jorgemoya: This is what I don't get, it already shows this error by just replying.

migueloller: Ah yes you're correct. We only need to pass a form error if we handle an error with try/catch and we want to send an error message that wasn't from an error as a result of parsing. So there's no need to do so in this case. My bad

PR #1687, opened 2024-11-25
fix: slide not in view when element inside is focused

hunterbecton:
## What/Why?

- Fixes issue where the slide was not visible when the button was focused
- Replaces `text-pretty` with `text-balance` for text-wrapping

## Testing

https://github.com/user-attachments/assets/f4b91351-0f8c-428a-a6fa-ffcae2a096ba


migueloller (approved):
This looks great! Can we get a companion PR for the same change in https://github.com/makeswift/vibes?

chanceaclark (approved):

PR #1690, opened 2024-11-28
Missing translation strings

thebigrick:
## What/Why?
Some of the `aria-label` attributes and a few strings were hardcoded.
I added new entries to the `en.json` file and applied the translations in the TSX files.

## Testing
Strings should yet be there. And should be translatable.


bookernath (approved):

bookernath:
Thanks @thebigrick!

thebigrick:
I just fixed a typo in the `en.json` file causing a build failure

core/app/[locale]/not-found.tsx
```diff
@@ -32,6 +32,7 @@ const NotFoundQuery = graphql(
 
 export default async function NotFound() {
   const t = await getTranslations('NotFound');
+  const ct = await getTranslations('Components.Header.MiniCart');
```
Comment (chanceaclark) [inline core/app/[locale]/not-found.tsx:35]: Would love to know if there are performance implications doing this double `await` for translations. 

Also, I think our pattern for things like this is just throw another translation string in the `NotFound` namespace instead of creating a new translation getter.

core/components/header/cart-icon.tsx
```diff
@@ -40,12 +41,12 @@ export const CartIcon = ({ count: serverCount }: CartIconProps) => {
   }, [serverCount, locale, setCount]);
 
   if (!count) {
-    return <ShoppingCart aria-label="cart" />;
+    return <ShoppingCart aria-label={t('cart')} />;
```
chanceaclark, line 44: I know you were just updating the current strings, but going off the rules of aria we should only use `aria-` roles if it's absolutely necessary. We should be using the `<title>` element for SVGs instead:
```jsx
<ShoppingCart>
  <title>{t('cart')}</title>
</ShoppingCart>
```

chanceaclark (commented):
Thanks for updating these! I know you already merged, but left some comments if you want to go back and clean some of them up.

PR #1696, opened 2024-12-02
Add Makeswift integration: catch all route

fikrikarim:
## Note
- Only review the last commit. The other commits are from `soul/main`

## What/Why?
- [x] Basic editing working
- [x] No `force-dynamic` needed
- [x] Revalidation working
- [ ]  ~404 path -> add and publish a Makeswift page at that path -> render the page at that path~ This is a bug from Next.js https://github.com/vercel/next.js/issues/73973

## Testing
https://github.com/user-attachments/assets/4ab1f163-828d-4437-821f-1cc8d56ebe7a


core/middlewares/with-makeswift.ts
```diff
@@ -0,0 +1,39 @@
+import { NextRequest } from 'next/server';
+import { parse as parseSetCookie } from 'set-cookie-parser';
+
+import { MiddlewareFactory } from './compose-middlewares';
+
+export const withMakeswift: MiddlewareFactory = (middleware) => {
+  return async (request, event) => {
+    const apiKey = request.nextUrl.searchParams.get('x-makeswift-draft-mode');
```
migueloller, line 8:
```suggestion
    const apiKey = request.nextUrl.searchParams.get('x-makeswift-draft-mode') ?? request.headers.get('x-makeswift-draft-mode');
```

This change is necessary for a "proper" proxy implementation. @agurtovoy, I wonder if this has anything to do with server actions misbehaving in Catalyst

core/middlewares/with-makeswift.ts
migueloller, line 1: I wonder if we should export this middleware from `@makeswift/runtime` at some point in the future so that we can maintain it properly. Nothing to do for this PR but I wanted to share that.

core/app/[locale]/(default)/[...rest]/page.tsx
```diff
@@ -1,5 +1,42 @@
...
+    siteVersion: await getSiteVersion(),
+    locale: locale === defaultLocale ? undefined : locale,
+  });
+
+  if (snapshot == null) return notFound();
+
+  return <MakeswiftPage snapshot={snapshot} />;
 }
+
+export const runtime = 'nodejs';
```
Comment (migueloller) [inline core/app/[locale]/(default)/[...rest]/page.tsx:42]: This runtime entry should be unnecessary now that we've removed all edge runtime occurrences

migueloller (approved):
LGTM 👍🏻

core/app/[locale]/(default)/[...rest]/page.tsx
Comment (fikrikarim) [reply core/app/[locale]/(default)/[...rest]/page.tsx:42]: Will update this on the next PR

jorgemoya (approved):

PR #1698, opened 2024-12-02
feat: add footer with streamable sitemap links

hunterbecton:
## What/Why?
- Adds Soul footer to Catalyst
- Includes skeleton for streaming sitemap link sections

## Testing

https://github.com/user-attachments/assets/7eb9b41a-ee87-444b-8542-bc5c665fb4ac


core/components/footer/footer.tsx
```diff
@@ -1,39 +1,56 @@
...
+const logo = 'SOUL';
+
+const paymentIconsArray: React.ReactNode[] = [
+  <Visa key="Visa" />,
+  <Amex key="Amex" />,
+  <Mastercard key="Mastercard" />,
+  <Paypal key="Paypal" />,
+  <GooglePay key="GooglePay" />,
+  <ApplePay key="ApplePay" />,
+  <Bitcoin key="Bitcoin" />,
```
migueloller, line 33: These icon seem to be different from the original ones. In particular, we've removed the Amazon and added the Bitcoin and Google Pay icons. Are these payment methods actually supported during checkout? We probably want to keep what was already there before in `main`.

Or @bookernath, maybe you can speak to what we should and shouldn't have here.

core/components/footer/footer.tsx
```diff
@@ -1,39 +1,56 @@
...
+  },
+  {
+    href: '#',
+    icon: <Instagram />,
+  },
+  {
+    href: '#',
+    icon: <Youtube />,
+  },
+];
```
migueloller, line 53: These seem to be changing from what we had before. These don't seem as critical as the payment icons but still wanted to flag it

core/components/footer/footer.tsx
```diff
@@ -70,29 +87,12 @@ export const Footer = async () => {
...
+      copyright={copyright}
+      logo={logo}
+      paymentIcons={paymentIconsArray}
       sections={sections}
-      socialMediaLinks={data.settings?.socialMediaLinks
-        .filter((socialMediaLink) => Boolean(socialIcons[socialMediaLink.name]))
-        .map((socialMediaLink) => ({
-          href: socialMediaLink.url,
-          icon: socialIcons[socialMediaLink.name]?.icon,
-        }))}
```
migueloller, line 95: Looks like the API has data on the social media links that should be available but we're completely ignoring that. We should preserve that functionality

core/components/footer/footer.tsx
```diff
@@ -70,29 +87,12 @@ export const Footer = async () => {
   ];
 
   return (
-    <ComponentsFooter
-      contactInformation={data.settings?.contact ?? undefined}
-      copyright={
-        data.settings
-          ? `© ${new Date().getFullYear()} ${data.settings.storeName} – Powered by BigCommerce`
-          : undefined
-      }
```
migueloller, line 79: We seem to be hardcoding "SOUL" here instead of using the store name. We should preserve the functionality that was there in `main` as much as possible while making it fit with the VIBE

core/components/footer/footer.tsx
```diff
@@ -70,29 +87,12 @@ export const Footer = async () => {
   ];
 
   return (
-    <ComponentsFooter
-      contactInformation={data.settings?.contact ?? undefined}
```
migueloller, line 74: We don't seem to be passing any contact information anymore. This feels like a regression

core/components/footer/footer.tsx
```diff
@@ -70,29 +87,12 @@ export const Footer = async () => {
...
-      copyright={
-        data.settings
-          ? `© ${new Date().getFullYear()} ${data.settings.storeName} – Powered by BigCommerce`
-          : undefined
-      }
-      logo={data.settings ? logoTransformer(data.settings) : undefined}
```
migueloller, line 80: We seem to be hardcoding the logo instead of using the store's logo

migueloller (changes requested):
@hunterbecton, I think it's very important that as we bring in the Soul VIBE we preserve all the functionality as far as data and BigCommerce API integration as we can. We want 100% feature parity with `main` but we seem to be deleting existing functionality that should stay there

jorgemoya (changes requested):
I think we should be streaming in all the values that are being fetched, and remove the wrapping Suspense for the Catalyst Footer component. Similar that what I'm trying to implement in #1702

core/components/footer/footer.tsx
```diff
@@ -25,27 +18,71 @@ import { MastercardIcon } from './payment-icons/mastercard';
...
+  <VisaIcon key="visa" />,
+];
+
 const socialIcons: Record<string, { icon: JSX.Element }> = {
-  Facebook: { icon: <SiFacebook title="Facebook" /> },
-  Twitter: { icon: <SiX title="Twitter" /> },
-  X: { icon: <SiX title="X" /> },
-  Pinterest: { icon: <SiPinterest title="Pinterest" /> },
-  Instagram: { icon: <SiInstagram title="Instagram" /> },
-  LinkedIn: { icon: <SiLinkedin title="LinkedIn" /> },
```
migueloller, line 34: Seems like we're still missing LinkedIn and Pinterest

core/components/footer/payment-icons/amazon.tsx
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 
 export const AmazonIcon: React.FC<React.ComponentPropsWithoutRef<'svg'>> = ({ ...props }) => (
   <svg
+    aria-hidden="true"
```
migueloller, line 5: Wouldn't it be better to use an `aria-label` instead of `aria-hidden`?

core/vibes/soul/sections/footer/index.tsx
migueloller, line 1: I'm assuming this file was copy/pasted from VIBES without modification?

migueloller (commented):
Looking good! I just think we're missing a couple of social icons

core/components/footer/payment-icons/amazon.tsx
hunterbecton: I intended to remove this from the accessibility API because it's not part of interactive elements. However, I shouldn't be hardcoding aria-hidden into the SVG after looking at this again. It would make more sense to pass this as a prop because there might be a case where this icon should not be hidden.

The icons do need a title, description, and other accessibility attributes. I already have an issue for that, so I think it would make more sense to do that in a separate PR.

core/vibes/soul/sections/footer/index.tsx
hunterbecton: The only thing I had to change was the Next.js Link component because Catalyst forces you to use its Link component.

core/components/footer/payment-icons/amazon.tsx
chanceaclark: In the realm of accessibility, it's better to use the `<title>` element inside an `<svg>` element rather than `aria-label`. We could provide a prop, `title` and set `aria-hidden` to `true` if `title` is not provided. We do this in BigDesign: https://github.com/bigcommerce/big-design/blob/main/packages/big-design-icons/src/components/AddIcon.tsx#L10-L11

core/vibes/soul/sections/footer/index.tsx
```diff
@@ -0,0 +1,372 @@
+import { clsx } from 'clsx';
+import Image from 'next/image';
```
jorgemoya, line 2:
```suggestion
import { BcImage as Image } from '~/components/bc-image';
```

jorgemoya: We need to use BcImage here for the logo image.

core/vibes/soul/sections/footer/index.tsx
migueloller: Cool 👍🏻

core/components/footer/payment-icons/amazon.tsx
migueloller: TIL. Thanks @chanceaclark!

migueloller (approved):
This LGTM 👍🏻 

I know @jorgemoya asked us to add `BcImage` but I think we need to do a pass on all components anyway. I think we need an ESLint rule as we have for `Link` that forces us to _always_ use `BcImage` and we should update `BcImage` so that it can use the default loader if `src` doesn't include the BigCommerce CDN origin. cc: @chanceaclark

core/components/footer/footer.tsx
```diff
@@ -25,6 +25,15 @@ import { MastercardIcon } from './payment-icons/mastercard';
 import { PayPalIcon } from './payment-icons/paypal';
 import { VisaIcon } from './payment-icons/visa';
 
+const paymentIcons: React.ReactNode[] = [
```
jorgemoya, line 28:
```suggestion
const paymentIcons = [
```

I don't think you need to type this, should be inferred.

core/vibes/soul/sections/footer/index.tsx
jorgemoya: We will leave as is since we will change this after #1710 changes gets added to the `soul/main` branch.

jorgemoya (approved):

migueloller (approved):
Did you intend to squash all commits into 1?

hunterbecton:
> Did you intend to squash all commits into 1?

@migueloller Yeah, I thought we were supposed to squash commits to keep our history less cluttered with commits from minor changes.

migueloller:
The changes look good. Just gotta fix the lint errors

PR #1702, opened 2024-12-03
feat: add header

jorgemoya:
## What/Why?
+ Use `Header` section component from VIBES.
+ Install react-headroom
+ Pass in streamable props to component.

## Testing
Locally.

Screenshot 2024-12-03 at 3 33 29 PM

### DIFF
--- a/core/vibes/soul/primitives/banner/index.tsx
+++ b/core/vibes/soul/primitives/banner/index.tsx
@@ -1,0 +1,69 @@
+'use client';
+
+import { clsx } from 'clsx';
+import { X } from 'lucide-react';
+import { ForwardedRef, forwardRef, ReactNode, useCallback, useEffect, useState } from 'react';
+
+export const Banner = forwardRef(
+  (
+    {
+      children,
+      hideDismiss = false,
+      className,
+      onDismiss,
+    }: {
+      children: ReactNode;
+      hideDismiss?: boolean;
+      className?: string;
+      onDismiss?(): void;
+    },
+    ref: ForwardedRef<HTMLDivElement>,
+  ) => {
+    const [banner, setBanner] = useState({ dismissed: false, initialized: false });
+
+    useEffect(() => {
+      const hidden = localStorage.getItem('hidden-banner') === 'true';
+
+      setBanner({ dismissed: hidden, initialized: true });
+    }, []);
+
+    const hideBanner = useCallback(() => {
+      setBanner((prev) => ({ ...prev, dismissed: true }));
+      localStorage.setItem('hidden-banner', 'true');
+      onDismiss?.();
+    }, [onDismiss]);
+
+    if (!banner.initialized) return null;
+
+    return (
+      <div
+        className={clsx(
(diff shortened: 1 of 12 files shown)


core/vibes/soul/primitives/banner/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+    useEffect(() => {
+      const hidden = localStorage.getItem('hidden-banner') === 'true';
+
+      setBanner({ dismissed: hidden, initialized: true });
+    }, []);
+
+    const hideBanner = useCallback(() => {
+      setBanner((prev) => ({ ...prev, dismissed: true }));
+      localStorage.setItem('hidden-banner', 'true');
```
chanceaclark, line 32: 🍹 Should these banners have a unique id? That way if a merchants starts a new promotion the banner will show up again?

core/vibes/soul/sections/header-section/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+  ({ navigation, banner }, ref) => {
+    const [bannerElement, setBannerElement] = useState<HTMLElement | null>(null);
+    const [bannerHeight, setBannerHeight] = useState(0);
+    const [isFloating, setIsFloating] = useState(false);
+
+    useEffect(() => {
+      if (bannerElement) {
+        setBannerHeight(bannerElement.getBoundingClientRect().height);
+      }
+    }, [bannerElement]);
```
chanceaclark, line 24: Can we compute the banner height instead of using state here?

core/vibes/soul/primitives/banner/index.tsx
migueloller: Yeah, they should. And the ID should be a prop so that if using Makeswift, for example, the ID can be changed to have the promotion show up again. I think we probably need a PR in VIBES that fixes this

core/vibes/soul/primitives/banner/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+      hideDismiss?: boolean;
+      className?: string;
+      onDismiss?(): void;
+    },
+    ref: ForwardedRef<HTMLDivElement>,
+  ) => {
+    const [banner, setBanner] = useState({ dismissed: false, initialized: false });
+
+    useEffect(() => {
+      const hidden = localStorage.getItem('hidden-banner') === 'true';
```
migueloller, line 25: I think right now there's a pretty bad layout shift because we're using local storage for banners. I think we need to move to using cookies for this. We can either remove banner support for this PR and make a separate issue for it to track progress on that or we could update VIBES to fix the banner before we merge this.

migueloller (commented):
@jorgemoya, this is look good 👍🏻 

I think there might be some changes we want to do in VIBES, though, before we ship this. Mainly for the banner feature.

Also, have you been having to change much code when copy/pasting VIBES to Catalyst? Are there any other changes I need to make to reduce changes that happen when copy/pasting VIBES components?

jorgemoya:
Added id changes from VIBES.

chanceaclark (approved):

PR #1705, opened 2024-12-04
test(core): add order details test

bc-alexsaiannyi:
## What/Why?
This Pr adds test for Order page to validate that  necessary sections are presented on the page.

## Testing
locally

Screenshot 2024-12-04 at 14 30 25
Screenshot 2024-12-04 at 14 29 33

### DIFF
--- a/core/tests/fixtures/utils/order/index.ts
+++ b/core/tests/fixtures/utils/order/index.ts
@@ -10,8 +10,8 @@ export class OrderFactory {
 
   constructor(private readonly page: Page) {}
 
-  async create(customerId: number, productId: number) {
-    const orderData = await createOrder(customerId, productId);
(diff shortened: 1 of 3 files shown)


core/tests/fixtures/utils/order/index.ts
```diff
@@ -11,7 +11,7 @@ export class OrderFactory {
   constructor(private readonly page: Page) {}
 
   async create(customerId: number, productId: number) {
-    const orderData = await createOrder(customerId, productId);
+    const orderData = await createOrder(productId, customerId);
```
chanceaclark, line 14: Can we align the method params with this change?

bc-alexsaiannyi: thanks, done

chanceaclark (approved):

PR #1706, opened 2024-12-04
fix: don't rewrite Next.js server actions' requests

agurtovoy:
## What/Why?

Behind the scenes, Next.js server actions are [translated into `POST` requests](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations#behavior) that get handled by the framework. `withRoutes` middleware was intercepting and rewriting these requests as page renders.

### DIFF
--- a/core/middlewares/with-routes.ts
+++ b/core/middlewares/with-routes.ts
@@ -252,8 +252,58 @@ const getRouteInfo = async (request: NextRequest, event: NextFetchEvent) => {
   }
 };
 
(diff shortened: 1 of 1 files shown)


bookernath:
@agurtovoy nice catch. should this PR perhaps be to `main`? I think this is relevant to all users of Catalyst.

agurtovoy:
@bookernath Good point, closing this one in favor of https://github.com/bigcommerce/catalyst/pull/1707

PR #1708, opened 2024-12-05
check for variant stock levels before disabling 'add to cart' button

EoinConway:
## What/Why?

This PR fixes an issue where the `add to cart` button only checks parent stock inventory. If the product has variants that are in stock, the button will incorrectly be disabled and show the `outOfStock` button text.

## Testing

Tested on a product with variant stock levels. The `add to cart` button isn't disabled and user can add to cart as expected. Also tested on product with two variants- one with stock, and one out of stock. User can add the stocked variant to cart, but still won't be able to add unstocked variant, as expected.


jorgemoya (commented):
Thanks for looking into this! This looks great, could we just add a changeset for this fix? Or I can do it, let me know.

PR #1710, opened 2024-12-05
feat(core): default to local image component

chanceaclark:
## What/Why?
I attempted to try and use the `images.loaderFile` from `next.config.ts` but there are some internals that Next.js does with the build to replace all references to the default loader import. This causes infinite loops trying to import the defined `loaderFile`.

The PR makes the default image component `CatalystImage` which uses the default loader as a fallback. `BcImage` was renamed to `CatalystImage`.

## Testing
Screenshot 2024-12-05 at 18 55 42

### DIFF
--- a/core/.eslintrc.cjs
+++ b/core/.eslintrc.cjs
@@ -28,6 +28,12 @@ const config = {
             message: "Please import 'Link' from '~/components/Link' instead.",
           },
           {
+            name: 'next/image',
+            importNames: ['default'],
+            message:
+              "Please import 'BcImage' from '~/components/bc-image' instead. This component handles CDN and static image optimization.",
+          },
+          {
             name: '~/i18n/routing',
             importNames: ['Link'],
             message: "Please import 'Link' from '~/components/Link' instead.",
--- a/core/.eslintrc.cjs
+++ b/core/.eslintrc.cjs
@@ -31,7 +31,7 @@ const config = {
             name: 'next/image',
             importNames: ['default'],
             message:
-              "Please import 'BcImage' from '~/components/bc-image' instead. This component handles CDN and static image optimization.",
+              "Please import 'Image' from '~/components/image' instead. This component handles CDN and static image optimization.",
           },
           {
             name: '~/i18n/routing',
--- a/core/components/bc-image/index.tsx
+++ b/core/components/bc-image/index.tsx
@@ -1,34 +1,26 @@
 'use client';
 
-import Image from 'next/image';
-import { ComponentPropsWithRef } from 'react';
+// eslint-disable-next-line @typescript-eslint/no-restricted-imports
+import Image, { ImageProps } from 'next/image';
 
 import bcCdnImageLoader from '~/lib/cdn-image-loader';
 
-type NextImageProps = Omit<ComponentPropsWithRef<typeof Image>, 'quality'>;
+const cdnHostname = process.env.BIGCOMMERCE_CDN_HOSTNAME ?? 'cdn11.bigcommerce.com';
 
-interface BcImageOptions {
-  lossy?: boolean;
+function shouldUseLoaderProp(props: ImageProps): boolean {
+  return typeof props.src === 'string' && props.src.startsWith(`https://${cdnHostname}`);
 }
 
-type Props = NextImageProps & BcImageOptions;
-
 /**
- * This `<BcImage>` component is a wrapper for Next's `<Image>` component, designed to
- * specifically handle images that are served from the BigCommerce CDN. It makes the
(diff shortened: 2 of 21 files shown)


core/lib/cdn-image-loader.ts
```diff
@@ -1,21 +1,9 @@
...
-  src: string;
-  width: number;
-  height?: number;
-}): string {
-  let url;
+import { ImageLoaderProps } from 'next/image';
 
-  if (height) {
-    url = src.replace('{:size}', `${width}x${height}`);
-  }
```
chanceaclark, line 16: Height isn't actually supported by the image loader, so opted to remove.

core/components/bc-image/index.tsx
```diff
@@ -1,34 +0,0 @@
...
-import Image from 'next/image';
-import { ComponentPropsWithRef } from 'react';
-
-import bcCdnImageLoader from '~/lib/cdn-image-loader';
-
-type NextImageProps = Omit<ComponentPropsWithRef<typeof Image>, 'quality'>;
-
-interface BcImageOptions {
-  lossy?: boolean;
-}
```
chanceaclark, line 12: Lossy is done through the API now, so we can omit this prop.

chanceaclark (commented):
I'll need to throw a PR up for https://www.catalyst.dev/docs/cdn-and-images.

core/components/catalyst-image/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-restricted-imports
+import Image, { ImageProps } from 'next/image';
+
+import bcCdnImageLoader from '~/lib/cdn-image-loader';
+
+const cdnHostname = process.env.BIGCOMMERCE_CDN_HOSTNAME ?? 'cdn11.bigcommerce.com';
+
+function shouldUseLoaderProp(props: ImageProps): boolean {
+  return typeof props.src === 'string' && props.src.includes(cdnHostname);
```
bookernath, line 11: 🍹 would a startsWith check be slightly safer?

core/components/catalyst-image/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+}
+
+/**
+ * This `<CatalystImage>` component is a wrapper for Next's `<Image>` component, designed to
+ * specifically handle images that are served from the BigCommerce CDN. It makes the
+ * assumption that it has been supplied with a `urlTemplate` field from GraphQL, which
+ *  contains a `{:size}` placeholder that will be replaced with the appropriate width
+ * and height values. It also adds a `compression` query parameter based on the `lossy`
```
bookernath, line 19: out of date comment

core/.eslintrc.cjs
```diff
@@ -27,6 +27,12 @@ const config = {
             name: 'next/link',
             message: "Please import 'Link' from '~/components/Link' instead.",
           },
+          {
+            name: 'next/image',
+            importNames: ['default'],
+            message:
+              "Please import 'CatalystImage' from '~/components/catalyst-image' instead. This components handles CDN and static image optimization.",
```
bookernath, line 34:
```suggestion
              "Please import 'CatalystImage' from '~/components/catalyst-image' instead. This component handles CDN and static image optimization.",
```

core/components/catalyst-image/index.tsx
migueloller: If we want to go super safe `new URL(props.src, 'http://n').hostname === cdnHostname` could do the trick. We'd also probably want to try/catch `new URL` in case an invalid URL is passed in.

Doing `startsWith(`https://${cdnHostname}`) is probable the sane thing to do

migueloller (approved):
If we have an ESLint rule to enforce the proper import I wonder if we should just call the component `Image` instead of `CatalystImage` since we call the link component `Link`, not `CatalystLink` 🤔 

Otherwise, LGTM 👍🏻

core/components/catalyst-image/index.tsx
chanceaclark: I wanted to avoid `startsWith` but I guess we should. Went with that instead of `new URL`.

apledger:
Do we want to merge this into `soul/main` first and then update all the VIBES components?

chanceaclark:
>Do we want to merge this into soul/main first and then update all the VIBES components?

Can we just do a rebase onto `soul/main`?

deini (approved):

jorgemoya (approved):
Lets merge this and we can rebase for other branches. 👍

migueloller (approved):

PR #1716, opened 2024-12-09
FEAT: added reset branch to ref command

jamesqquick:
## What/Why?
When moving from OCC to development, users should have the ability to check out a ref that is then set to the `main` branch so they can continue smoothly in their development journey. This will put them in a state where they have the exact code they originally deployed in OCC but will be set to the `main` branch.

- create `reset-main` boolean option for CLI
- create new resetBranchToRef to be called if `ghRef` and `resetMain` are present
- update checkoutRef to return a boolean status to determine whether or not it was run successfully

## Testing
- test existing OCC generated command in isolation - should work as expected
- test existing OCC generated command  with `--reset-main` flag with the end result of being in the `main` branch with code matching the `gh-ref` provided

### DIFF
--- a/packages/create-catalyst/src/utils/clone-catalyst.ts
+++ b/packages/create-catalyst/src/utils/clone-catalyst.ts
@@ -33,9 +33,10 @@ export const cloneCatalyst = ({
 
   if (ghRef) {
     if (resetMain) {
-      return resetBranchToRef(projectDir, 'main', ghRef);
+      resetBranchToRef(projectDir, 'main', ghRef);
+      return;
     }
-    
+
     checkoutRef(projectDir, ghRef);
 
     console.log();
--- a/packages/create-catalyst/src/utils/clone-catalyst.ts
+++ b/packages/create-catalyst/src/utils/clone-catalyst.ts
@@ -33,10 +33,10 @@ export const cloneCatalyst = ({
 
   if (ghRef) {
     if (resetMain) {
-      resetBranchToRef(projectDir, 'main', ghRef);
-    } else {
-      checkoutRef(projectDir, ghRef);
+      return resetBranchToRef(projectDir, 'main', ghRef);
     }
+    
+    checkoutRef(projectDir, ghRef);
 
     console.log();
   }
--- a/packages/create-catalyst/src/utils/clone-catalyst.ts
+++ b/packages/create-catalyst/src/utils/clone-catalyst.ts
@@ -37,6 +37,7 @@ export const cloneCatalyst = ({
     } else {
       checkoutRef(projectDir, ghRef);
     }
+
     console.log();
   }
 };
--- a/packages/create-catalyst/src/utils/clone-catalyst.ts
+++ b/packages/create-catalyst/src/utils/clone-catalyst.ts
@@ -33,27 +33,17 @@ export const cloneCatalyst = ({
 
   if (ghRef) {
     if (resetMain) {
-      try {
-        resetBranchToRef(projectDir, 'main', ghRef);
-      } catch (error: unknown) {
-        if (error instanceof Error) {
-          console.error(error.message);
-        } else {
-          console.error('Unknown error occurred while resetting branch to ref');
-        }
-      }
+      checkoutRef(projectDir, 'main');
+
+      resetBranchToRef(projectDir, ghRef);
+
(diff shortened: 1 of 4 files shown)


packages/create-catalyst/src/utils/reset-branch-to-ref.ts
```diff
@@ -0,0 +1,36 @@
...
+
+    // Reset current branch to the specified ref
+    if (checkoutSuccess) {
+      execSync(`git reset --hard ${ref}`, {
+        cwd: repoDir,
+        stdio: 'inherit',
+        encoding: 'utf8',
+      });
+      console.log(`Reset ${targetBranch} to ${ref} successfully.`);
+    }
```
chanceaclark, line 19: Can we rely on errors being thrown instead of boolean conditions? If `checkoutSuccess` is `false`, we essentially do "nothing".

packages/create-catalyst/src/utils/clone-catalyst.ts
```diff
@@ -29,7 +32,12 @@ export const cloneCatalyst = ({
   console.log();
 
   if (ghRef) {
-    checkoutRef(projectDir, ghRef);
+    if (resetMain) {
+      resetBranchToRef(projectDir, 'main', ghRef);
+    } else {
+      checkoutRef(projectDir, ghRef);
+    }
```
chanceaclark, line 39:
```suggestion
    if (resetMain) {
      return resetBranchToRef(projectDir, 'main', ghRef);
    }
    
    checkoutRef(projectDir, ghRef);
```

packages/create-catalyst/src/utils/reset-branch-to-ref.ts
jamesqquick: Do you mean checking the boolean condition of `checkoutSuccess`? Right now, `checkoutRef` function handles errors gracefully, meaning it doesn't surface it's errors to a caller. Because of that, I don't know how I would know if it ran successfully or not?

Right now, if `checkoutSuccess` is `false` the error handling and logging has already been handled in `checkoutRef`.

Does that address your question?

chanceaclark: Ummm... so to me CLI application logic flow should be handled by errors being thrown instead of conditional statements like you have currently. It allows us to surface errors and handle them through the catch statement, rather than `if x, do y, else noop`.

I think it should be something like this:
```ts
try {
  checkoutRef(...);

  execSync(...);
} catch (error) {
  if (error instanceof CheckoutRefError) {
    // handle fallback logic
  }

  // ...
}
```

jamesqquick: Yeah I understand that. The way things are written already, this would mean reworking `checkoutRef` to throw the errors that it's already handling? Then, we have to handle that error every time it's called, which is only once for now and potentially replicating logging for it?

matthewvolk (approved):
Looks good, I met with James and functionally this PR does what we want; there are improvements we can make to automatically set the upstream of local main to pull from the branch deployed by OCC. But to get this across the finish line, I'm going to approve this one.

packages/create-catalyst/src/utils/reset-branch-to-ref.ts
```diff
@@ -0,0 +1,9 @@
+import { execSync } from 'node:child_process';
+
+export function resetBranchToRef(projectDir: string, ghRef: string): void {
+  execSync(`git reset --hard ${ghRef}`, {
+    cwd: projectDir,
+    stdio: 'inherit',
+    encoding: 'utf8',
+  });
+}
```
matthewvolk, line 9: As discussed internally, in the unlikely event that a bad `ghRef` is passed, this will throw and kill the CLI process. This will leave the user running the CLI in a state where the repository was cloned successfully, but dependencies wouldn't be installed nor would the environment file be written (see: https://github.com/bigcommerce/catalyst/blob/ad8c86d574474eb5ed18d99265fe4001d267fb5f/packages/create-catalyst/src/commands/create.ts#L257-L266) 

For now, we'll accept this since it's unlikely the `ghRef` provided by OCC will be rejected. That said, we do need to handle errors in a follow up PR.

matthewvolk (approved):
Looks good, but let's make note of the comment below 👍

PR #1726, opened 2024-12-10
feat: stream in products for carousel/featured in Home

jorgemoya:
## What/Why?
Builds off #1725 

Passes in a promise of products to these components to allow for streaming.

## Testing
Locally components are streamed in showing a skeleton while info loads.


core/components/featured-products-carousel/index.tsx
migueloller, line 1: Having this additional component is unnecessary IMO and we can do the mapping of the products in the page's `getNewestProducts` and `getFeaturedProducts` functions

migueloller (approved):

core/components/featured-products-carousel/index.tsx
jorgemoya: Valid observation! Why I thinks this is still useful is because this component is shared and if we move the transform logic we'd be repeating this transform every time we want to use it. What do you think?

migueloller: That fair. I was thinking we could share the transformation logic, but I'm fine sharing the component. We can always come back to this later and re-asses 👍🏻

jorgemoya: Another benefit is that we define the fragment in this folder as well. But yeah I'm open to reassess as we develop more patterns.

jorgemoya: I have modified the code to follow your suggestions, after consideration I agree!

PR #1728, opened 2024-12-10
feat(core): use toasts for address deletions

chanceaclark:
## What/Why?
Converts feedback message when deleting an address to use toasts instead of the page messages.

## Testing

https://github.com/user-attachments/assets/1633607f-dca3-4f39-9418-017a846131fb


jorgemoya (approved):
👌

PR #1743, opened 2024-12-11
feat(core): redirect account overview page to orders

chanceaclark:
## What/Why?
Redirects the customer to the accounts orders page instead of having an overview page.

I tried a few different approaches here @apledger:
- Went with a `page.tsx` redirect which caused the page to initially render, then redirect, which isn't an ideal experience.
- Tried `next.config.ts` redirect and occasionally ran into some weird issue in conjunction with next-intl.

The best and most simple one was just having the redirect after `signIn` was called.

## Testing

https://github.com/user-attachments/assets/97f67723-1631-4a06-b579-b64a9bce940b


core/middlewares/with-auth.ts
```diff
@@ -1,24 +1,7 @@
...
-      }
-
-      // Continue the middleware chain
-      return next(req, event);
-    });
-
-    // @ts-expect-error: The `auth` function doesn't have the correct type to support it as a MiddlewareFactory.
-    return authWithCallback(request, event);
-  };
-};
```
chanceaclark, line 24: We were doing the redirect here, but I think it's probably better to handle in the `layout.tsx` file that we did in #1741

deini (approved):

PR #1748, opened 2024-12-12
chore: cleanup header components

jorgemoya:
## What/Why?
Remove unused components from Header.

## Testing
Eslint no issues.


core/components/header/_actions/get-search-results.ts
```diff
@@ -1,67 +0,0 @@
-'use server';
```
jorgemoya, line 1: Now that I think about, we might want to keep this one since it'll be used for search. 🤔

chanceaclark (approved):

jorgemoya:
I will cleanup once we have all the functionality of Header working.

PR #1749, opened 2024-12-12
feat(core): remove account status provider with toast messages

chanceaclark:
## What/Why?
Remove the rest of the account status provider usages and move over to using toasts.

I also had to adjust a few things for all the pages to work correctly:
- The `login` route was statically generated, which prevented an auth check on the page. Needed to remove that static generation.

## Testing
Updated and ran the test suite to verify all functionality worked the same.


core/tests/ui/e2e/register.spec.ts
```diff
@@ -35,5 +35,5 @@ test('Account is not created if email is already in use', async ({ page, account
 
   await page.getByRole('button', { name: 'Create account' }).click();
 
-  await expect(page.getByText('The email address is already in use.')).toBeVisible();
```
chanceaclark, line 38: I just had to fix an issue in the wordpress issue, but this is a security concern allowing bad actors find out if email addresses are valid. Changed it over to a more generic error.

jorgemoya (approved):

PR #1751, opened 2024-12-13
feat: add ProductDetails to PDP

jorgemoya:
## What/Why?

- Import or update VIBES components.
- Adds `<ProductDetail />` to PDP.
- Removes no longer used components.

https://github.com/user-attachments/assets/71bfea73-3eab-4aff-965c-3385ae1462ba

\* Ignore toaster issue it has been fixed *


core/app/[locale]/layout.tsx
```diff
@@ -117,6 +118,7 @@ export default async function RootLayout({ params, children }: Props) {
         <Notifications />
         <NextIntlClientProvider locale={locale} messages={messages}>
           <NuqsAdapter>
+            <Toaster position="top-right" />
```
Comment (chanceaclark) [inline core/app/[locale]/layout.tsx:121]: This should be provided in https://github.com/bigcommerce/catalyst/blob/soul/main/core/app/providers.tsx#L12 Do we need to add this?

core/app/[locale]/(default)/product/[slug]/page-data.ts
```diff
@@ -53,18 +269,18 @@ const ProductPageQuery = graphql(
...
-    DetailsFragment,
-    ProductItemFragment,
-    DescriptionFragment,
-    WarrantyFragment,
+    ProductDetailsFragment,
+    ProductViewedFragment,
+    ProductSchemaFragment,
+    // DetailsFragment,
+    // DescriptionFragment,
+    // WarrantyFragment,
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/product/[slug]/page-data.ts:277]: Are we going to be revisiting these later?

core/app/[locale]/(default)/product/[slug]/_actions/add-to-cart.tsx
```diff
@@ -61,13 +83,15 @@ export async function handleAddToCart(
...
+          optionEntityId: Number(field.name),
           optionValueEntityId:
-            Number(optionValueEntityId) !== 0
-              ? option.checkedOptionValueEntityId
-              : option.uncheckedOptionValueEntityId,
+            optionValueEntityId === 'true'
+              ? // @ts-expect-error Need to fix types
+                Number(field.checkedValue)
+              : // @ts-expect-error Need to fix types
+                Number(field.uncheckedValue),
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/product/[slug]/_actions/add-to-cart.tsx:94]: Do we need to fix these right now?

core/app/[locale]/(default)/product/[slug]/page-data.ts
Comment (jorgemoya) [reply core/app/[locale]/(default)/product/[slug]/page-data.ts:277]: Yes, didn't want to delete because I might need it for the next PRs. Good call tho, I should've left a comment.

core/app/[locale]/layout.tsx
Comment (jorgemoya) [reply core/app/[locale]/layout.tsx:121]: Ahh, probably what is causing my toaster issue. Good catch, I missed this in the rebase.

core/app/[locale]/(default)/product/[slug]/_actions/add-to-cart.tsx
Comment (jorgemoya) [reply core/app/[locale]/(default)/product/[slug]/_actions/add-to-cart.tsx:94]: I forgot about these, let me take a look 👀

Comment (jorgemoya) [reply core/app/[locale]/(default)/product/[slug]/_actions/add-to-cart.tsx:94]: Other than updating the schema which I think @migueloller was hesitant to include because this is a very BC pattern:

```ts
type CheckboxField = {
  type: 'checkbox';
  defaultValue?: string;
  checkedValue?: string;
  uncheckedValue?: string
} & FormField;
```

I can solve by passing the "unparsed" product options to my action:

```ts
// Action I pass to ProductDetails
const action = async (prevState: State, payload: FormData) => {
  'use server';

  const product = await productPromise;

  return await addToCart(prevState, payload, product.productOptions);
};
  
// In action function
case 'checkbox':
// eslint-disable-next-line no-case-declarations <-- still need this or I can move outside of the switch I guess.
const checkbox = checkboxOptions.find((option) => option.entityId === Number(field.name));

if (!checkbox) {
  return { ...accum };
}

checkboxOptionInput = {
  optionEntityId: Number(field.name),
  optionValueEntityId:
    optionValueEntityId === 'true'
      ? checkbox.checkedOptionValueEntityId
      : checkbox.uncheckedOptionValueEntityId,
};
....
```

What do you think?

Comment (migueloller) [reply core/app/[locale]/(default)/product/[slug]/_actions/add-to-cart.tsx:94]: The approach we had decided to take on this was to allow fields to be generic so you could pass additional fields that aren't necessarily used by VIBES and those would be passed back to the server action. Perhaps we also need to introduce the ability to pass the Zod schema for the form or we can simply make the Zod schema in VIBES use `passthrough` and in the server action, the VIBES schema can be extended to validate those fields are being added.

```ts
async function action() {
  const submission = parseWithZod(formData, { schema: schema.extend({ /* extend schema here */}) })
}
```

core/components/header/index.tsx
migueloller, line 1: I see you sneaked in some cart functionality 😏

core/components/header/index.tsx
```diff
@@ -52,6 +67,32 @@ const getLogo = async () => {
...
+    document: GetCartCountQuery,
+    variables: { cartId },
+    customerAccessToken,
+    fetchOptions: {
+      cache: 'no-store',
+      next: {
+        tags: [TAGS.cart],
+      },
+    },
+    // channelId,
```
migueloller, line 86: Do we need to address this later or can we delete?

core/data-transformers/product-options-transformer.ts
```diff
@@ -0,0 +1,185 @@
...
+        return {
+          persist: option.isVariantOption,
+          type: 'number',
+          label: option.displayName,
+          required: option.isRequired,
+          name: option.entityId.toString(),
+          defaultValue: option.defaultNumber?.toString(),
+          min: option.lowest ?? undefined,
+          max: option.highest ?? undefined,
+          // TODO: other props?
```
migueloller, line 146: 🍹 Can we add a bit more detail here? Not sure what "other props?" means. Is it about extending the VIBES field to support Catalyst-only props using generics?

migueloller (approved):
It's happening! 👐🏻

core/components/header/index.tsx
jorgemoya: You know what, I actually forgot 😆

core/components/header/index.tsx
jorgemoya: Addressing this.

jorgemoya: Actually, I don't think this is needed.

PR #1758, opened 2024-12-15
feat: add props and css variables to `Slideshow`

apledger:
## What/Why?
Added props to `Slideshow`:
- `playOnInit`
- `showDescription`
- `showCta`
- `variant`, `size`, and `shape` to `cta`

Added css variables to `Slideshow` for future theming.

## Testing
Verified on homepage.

CleanShot 2024-12-15 at 18 22 24@2x

### DIFF
--- a/core/app/[locale]/layout.tsx
+++ b/core/app/[locale]/layout.tsx
@@ -1,5 +1,6 @@
 import { Analytics } from '@vercel/analytics/react';
 import { SpeedInsights } from '@vercel/speed-insights/next';
+import { clsx } from 'clsx';
 import type { Metadata } from 'next';
 import { DM_Serif_Text, Inter, Roboto_Mono } from 'next/font/google';
 import { NextIntlClientProvider } from 'next-intl';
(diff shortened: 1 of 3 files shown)


core/app/[locale]/layout.tsx
```diff
@@ -107,8 +108,11 @@ export default async function RootLayout({ params, children }: Props) {
   const messages = await getMessages();
 
   return (
-    <html lang={locale}>
-      <body className={`${inter.variable} ${dm_serif_text.variable} ${roboto_mono.variable}`}>
+    <html
+      className={clsx(inter.variable, dm_serif_text.variable, roboto_mono.variable)}
```
Comment (migueloller) [inline core/app/[locale]/layout.tsx:112]: nit: do we really need `clsx` here? 🙄

migueloller (approved):

core/app/[locale]/layout.tsx
Comment (apledger) [reply core/app/[locale]/layout.tsx:113]: idk we just never string interpolate classes like that anywhere so I changed it to be consistent

PR #1768, opened 2024-12-17
fix: aria-label and placeholder in Select

hunterbecton:
## What/Why
- Adds `aria-label` into Radix `SelectPrimitive.Trigger`
- Moves `placeholder` to `SelectPrimitive.Value`
- Add `placeholder` to prop to parent component

Reference: [Radix Select Docs](https://www.radix-ui.com/primitives/docs/components/select#with-a-placeholder)

## Testing
Locally

### DIFF
--- a/core/vibes/soul/sections/products-list-section/sorting.tsx
+++ b/core/vibes/soul/sections/products-list-section/sorting.tsx
@@ -64,6 +64,7 @@ function SortingInner({
 
   return (
     <Select
+      hideLabel
       label={label}
       name={paramName}
       onValueChange={(value) => {
--- a/core/vibes/soul/sections/products-list-section/sorting.tsx
+++ b/core/vibes/soul/sections/products-list-section/sorting.tsx
@@ -18,11 +18,13 @@ export function Sorting({
   options,
   paramName = 'sort',
   defaultValue = '',
+  placeholder,
 }: {
   label?: Streamable<string | null>;
   options: Streamable<Option[]>;
   paramName?: string;
   defaultValue?: string;
+  placeholder?: Streamable<string | null>;
 }) {
   return (
     <Suspense fallback={<SortingSkeleton />}>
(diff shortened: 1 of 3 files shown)


core/vibes/soul/sections/products-list-section/sorting.tsx
```diff
@@ -66,7 +67,6 @@ function SortingInner({
         });
       }}
       options={options}
-      placeholder={label}
```
apledger, line 69: Are you sure we can just get rid of the `placeholder` prop here? I'm passing it through to `<SelectPrimitive.Value>` so it can be rendered in the button.

hunterbecton: The `Select` component still accepts a `placeholder` prop, but I removed it because I assumed it would just use the default value for the `placeholder`, which is `Select an item`

jorgemoya: Can you include an image of how that select looks after these changes? If we are now using `label` to set the `aria-label`, we now need a placeholder prop we can pass from the parent section component no?

migueloller: Yes, exactly. We need to pass in `placeholder` from the parent

migueloller (commented):
We're changing the behavior of the component, can we make sure we're passing the `placeholder` prop where the product list section component is being used?

core/vibes/soul/sections/products-list-section/sorting.tsx
```diff
@@ -41,11 +44,13 @@ function SortingInner({
   defaultValue,
   options: streamableOptions,
   label: streamableLabel,
+  placeholder: streamablePlaceholder,
```
migueloller, line 47: Looks like we're still not passing this value from the places that reference this component

migueloller (approved):

PR #1772, opened 2024-12-17
Inline default CSS variables, add props to Carousels

apledger:
## What/Why?
With the new component based CSS variables, the number of variables is exploding. This is great for theming control, but the DX of installing VIBES would require developers to copy paste a set of variables in addition to the component code to make sure things rendered properly. 

This PR moves default theme values into inline tailwind styles so developers can skip this step as long as they've followed the top level VIBES installation, ie setup their main design tokens (`--primary`, `--accent`, `--contrast-100` etc). I've done this for these components:
- `Button`
- `ButtonLink`
- `Accordions`
- `Carousel`
- `Slideshow`

In addition I've added some requested props and style updates to the `ProductsCarousel`:
- Added props for showing / hiding the carousel arrows and/or scrollbar.
- Added prop for product card aspect ratio.
- Updated CSS of product cards in the carousel so that offscreen cards "peek" into view.

### PLP
image

### Home
image


core/vibes/soul/primitives/products-carousel/index.tsx
```diff
@@ -58,17 +66,32 @@ export function ProductsCarousel({
...
+                    product={product}
+                  />
                 </CarouselItem>
               ))}
             </CarouselContent>
-            <div className="flex w-full items-center justify-between">
-              <CarouselScrollbar label={scrollbarLabel} />
-              <CarouselButtons nextLabel={nextLabel} previousLabel={previousLabel} />
-            </div>
+            {(showButtons || showScrollbar) && (
```
migueloller, line 80: If we're combining the values here why not just have 1 prop? Shouldn't we be using this separately for the scrollbar and the buttons?

apledger: It's because I want to hide the entire row (that contains the margin) if both controls are hidden. Andrew wants to be able to hide them independently though.

migueloller: I missed we were using these in the classes below as well 👍🏻

migueloller (approved):

PR #1775, opened 2024-12-18
Add /login/token to support customer login API

bookernath:
## What/Why?
Support [Customer Login API](https://developer.bigcommerce.com/docs/start/authentication/customer-login) on the same path Stencil uses (`/login/token`) for backwards compatibility

Also adds a util method for generating tokens which is useful as a stub for plugging in 3rd party auth providers and exchanging Customer Login API tokens for Customer Access Tokens.

## Testing

- [x] Valid tokens work
- [x] Non-JWT tokens error out and take you back to login page
- [x] Invalid or expired well-formed JWTs fail and take you back to login page
- [x] Arbitrary `redirect_to` parameters work
- [x] If `redirect_to` is absent we get taken to `/account/orders`
- [x] `channel_id` passed in the token is respected
- [x] Localization redirect behavior is acceptable

https://github.com/user-attachments/assets/f42523f8-ec3d-4cb1-8778-fa191d9d5e70


core/app/login/token/[token]/route.ts
```diff
@@ -0,0 +1,47 @@
...
+    const claims = decodeJwt(token);
+    const redirectTo = claims.redirect_to || '/account/orders';
+
+    // sign in with token which will check validity against BigCommerce API
+    // and redirect to redirectTo
+    await signIn('credentials', {
+      type: 'jwt',
+      jwt: token,
+        redirectTo
+    });
```
Comment (chanceaclark) [inline core/app/login/token/[token]/route.ts:34]: We've had a few cases where the `redirectTo` doesn't work correctly. Can you should a screen recording of this working correctly?

core/app/login/token/[token]/route.ts
```diff
@@ -0,0 +1,47 @@
...
+  } catch (error) {
+    // https://github.com/nextauthjs/next-auth/discussions/9389
+    if (isRedirectError(error)) {
+      throw error;
+    }
```
Comment (chanceaclark) [inline core/app/login/token/[token]/route.ts:39]: `import { unstable_rethrow as rethrow } from 'next`;

```suggestion
    rethrow(error)
```

core/auth.ts
```diff
@@ -85,37 +118,73 @@ const config = {
...
+            variables: { email, password, cartEntityId },
+            fetchOptions: {
+              cache: 'no-store',
+            },
+          });
 
-        if (response.errors && response.errors.length > 0) {
-          return null;
+          if (response.errors?.length > 0) {
+            console.error('Login error:', response.errors);
```
chanceaclark, line 142:
```suggestion
```

core/auth.ts
```diff
@@ -85,37 +118,73 @@ const config = {
...
+          const response = await client.fetch({
+            document: LoginWithTokenMutation,
+            variables: { jwt, cartEntityId },
+            fetchOptions: {
+              cache: 'no-store',
+            },
+          });
+
+          if (response.errors?.length > 0) {
+            console.error('Login error:', response.errors);
```
chanceaclark, line 170:
```suggestion
```

core/app/login/token/[token]/route.ts
Comment (chanceaclark) [inline core/app/login/token/[token]/route.ts:34]: Maybe we can avoid the `jose` library and handle the redirect in `auth.ts` 🤔

Comment (bookernath) [reply core/app/login/token/[token]/route.ts:34]: redirect needs to be passed into `signIn` I think to do things the next-auth way... and I have future plans for `jose`, I might add to this PR :)

core/lib/customer-login-api.ts
```diff
@@ -0,0 +1,57 @@
...
+  const clientSecret = process.env.BIGCOMMERCE_CLIENT_SECRET;
+  const storeHash = process.env.BIGCOMMERCE_STORE_HASH;
+
+  if (!clientSecret) {
+    throw new Error('BIGCOMMERCE_CLIENT_SECRET is not set in environment variables');
+  }
+
+  if (!clientId) {
+    throw new Error('BIGCOMMERCE_CLIENT_ID is not set in environment variables');
+  }
```
chanceaclark, line 35: Might want to validate `storeHash` here too.

core/lib/customer-login-api.ts
```diff
@@ -0,0 +1,57 @@
...
+    ...(redirectTo && { redirect_to: redirectTo }),
+  };
+
+  // Convert client secret to Uint8Array for jose library
+  const secretKey = new TextEncoder().encode(clientSecret);
+
+  // Create and sign the JWT
+  const token = await new SignJWT(payload).setProtectedHeader({ alg: 'HS256' }).sign(secretKey);
+
+  return token;
```
chanceaclark, line 54:
```suggestion
  return await new SignJWT(payload).setProtectedHeader({ alg: 'HS256' }).sign(secretKey);
```

core/lib/customer-login-api.ts
```diff
@@ -0,0 +1,57 @@
...
+ * from a client that has the Customer Login scope enabled
+ *
+ * @param {number} customerId - The BigCommerce customer ID to generate the login token for
+ * @param {number} [channelId] - Optional channel ID that the customer will be logged into
+ * @param {string} [redirectTo] - Optional relative URL to redirect to after successful login
+ * @returns {Promise<string>} A JWT token that can be used to authenticate the customer
+ * @throws {Error} If BIGCOMMERCE_CLIENT_SECRET is not set in environment variables
+ * @throws {Error} If BIGCOMMERCE_CLIENT_ID is not set in environment variables
+ */
+const generateCustomerLoginApiJwt = async (
```
chanceaclark, line 20: Can we remove the default export in favor of the named export?
```suggestion
export const generateCustomerLoginApiJwt = async (
```

core/auth.ts
```diff
@@ -85,37 +119,72 @@ const config = {
...
+            },
+          });
 
-        if (response.errors && response.errors.length > 0) {
-          return null;
+          if (response.errors && response.errors.length > 0) {
+            return null;
+          }
+
+          const result = response.data.login;
```
chanceaclark, line 146: I think now is the time to extract a lot of this into its own function:
```ts
const login = async (credentials: Partial<Record<"email" | "password", unknown>>) => {
  // ...
}

// ...
async authorize(credentials) {
  const result = await login(credentials);

  if (!result.customer || !result.customerAccessToken) {
    return null;
  }

  return {
    name: `${result.customer.firstName} ${result.customer.lastName}`,
    email: result.customer.email,
    customerAccessToken: result.customerAccessToken.value,
  };
}
```

🍹  I personally would love the `parsed.type` check to be a `switch` statement here.

core/auth.ts
```diff
@@ -31,10 +47,28 @@ const LogoutMutation = graphql(`
...
+    type: z.literal('password'),
+    email: z.string().email(),
+    password: z.string().min(1),
+  }),
+  z.object({
+    type: z.literal('jwt'),
+    jwt: z.string(),
+    channelId: z.string().optional(),
+  }),
+]);
```
chanceaclark, line 61:
```suggestion
const PasswordCredentials = z.object({
  type: z.literal('password'),
  email: z.string().email(),
  password: z.string().min(1),
});

const JwtCredentials = z.object({
  type: z.literal('jwt'),
  jwt: z.string(),
  channelId: z.string().optional(),
})
 
export const Credentials = z.discriminatedUnion('type', [PasswordCredentials, JwtCredentials]);
```

core/app/login/token/[token]/route.ts
```diff
@@ -0,0 +1,46 @@
...
+    await signIn('credentials', {
+      type: 'jwt',
+      jwt: token,
+      redirectTo,
+      channelId,
+    });
+  } catch (error) {
+    if (isRedirectError(error)) {
+      rethrow(error);
+    }
```
Comment (chanceaclark) [inline core/app/login/token/[token]/route.ts:39]: Rethrow should handle the redirect error.

```suggestion
    rethrow(error);
```

core/app/login/token/[token]/route.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+import { decodeJwt } from 'jose';
+import { isRedirectError } from 'next/dist/client/components/redirect-error';
+// eslint-disable-next-line @typescript-eslint/no-restricted-imports
+import { redirect, unstable_rethrow as rethrow } from 'next/navigation';
+
+import { signIn } from '~/auth';
+
+export async function GET(request: Request, { params }: { params: Promise<{ token: string }> }) {
```
Comment (chanceaclark) [inline core/app/login/token/[token]/route.ts:14]: 🍹 Extract the type into an interface

core/lib/customer-login-api.ts
chanceaclark, line 1: 🍹 If we have a few utilities for auth related functions, I am tempted to move any auth related things into an `auth/` folder:
```bash
auth/
  index.ts # rename of auth.ts
  customer-login-api.ts
```

Curious what you and others think about that?

chanceaclark:
I pushed some changes, so need a review from other @bigcommerce/team-catalyst members 🙏

core/auth/customer-login-api.ts
```diff
@@ -0,0 +1,60 @@
...
+    throw new Error('BIGCOMMERCE_CLIENT_ID is not set in environment variables');
+  }
+
+  if (!storeHash) {
+    throw new Error('BIGCOMMERCE_STORE_HASH is not set in environment variables');
+  }
+
+  const payload = {
+    iss: clientId,
+    iat: Math.floor(Date.now() / 1000),
```
jordanarldt, line 45: Wondering if `Date.now()` will suffice here. Usually the most reliable way to get `iat` is the `v2/time` endpoint since it will always return the correct time synced with BC servers. 🤔

bookernath: Race conditions are possible in case of clock skew, yes. I can always just subtract 1 😛

bookernath: I'm voting "do nothing" for now, we can iterate on it if users have issues.

jordanarldt (approved):

chanceaclark (approved):

PR #1779, opened 2024-12-18
Ap/footer variables

apledger:
## What/Why?
Added CSS variables to `Footer` and these props:
- `logoWidth`
- `contactTitle`

The "Contact Us" title was previously hardcoded so now that is being pulled from the localization service. In addition I've added a new primitive called `Logo` that is shared between the `Navigation` and `Footer`. It encapsulates the fallback streaming logic and the idea that a logo can be either text or an image.

### Footer
CleanShot 2024-12-18 at 17 54 32@2x

### Navigation
CleanShot 2024-12-18 at 17 54 51@2x


core/messages/de.json
migueloller, line 1: This file shouldn't have changed?

migueloller (approved):

PR #1781, opened 2024-12-19
fix: fix image priority on PLP to improve lighthouse

bookernath:
## What/Why?
Improve lighthouse finding around above-the-fold image being loaded lazily; prioritizing just the first image is enough to eliminate the finding and maps to the mobile experience where you can only see one image. On desktop this might ideally be more like the first 3-4 images, but the finding on desktop goes away also so I'm fine to start with 1 image.

## Testing
Before:

image

After (finding gone):

image


bookernath:
I'll defer this to @migueloller when he's looking at image perf this week

PR #1782, opened 2024-12-19
refactor(makeswift): userland RSC `Page` component etc.

agurtovoy:
## What/Why?

Implements "userland" RSC `Page` component and related API abstractions as discussed [here](https://makeswifthq.slack.com/archives/C014SCS4G75/p1734553802046819?thread_ts=1734538843.665429&cid=C014SCS4G75) and [here](https://makeswifthq.slack.com/archives/C080BNSNEG7/p1734401070722499?thread_ts=1734384162.728269&cid=C080BNSNEG7).

## Testing

https://github.com/user-attachments/assets/fcb2c2a1-ccee-46aa-95ce-31baeec92343

### DIFF
--- a/core/app/[locale]/(default)/[...rest]/page.tsx
+++ b/core/app/[locale]/(default)/[...rest]/page.tsx
@@ -1,42 +1,27 @@
-import { Page as MakeswiftPage } from '@makeswift/runtime/next';
-import { getSiteVersion } from '@makeswift/runtime/next/server';
-import { notFound } from 'next/navigation';
+import { locales } from '~/i18n/routing';
+import { client, Page } from '~/lib/makeswift';
 
-import { defaultLocale, locales } from '~/i18n/routing';
-import { client } from '~/lib/makeswift/client';
+interface PageParams {
+  locale: string;
+  rest: string[];
+}
 
-export async function generateStaticParams() {
+export async function generateStaticParams(): Promise<PageParams[]> {
   const pages = await client.getPages().toArray();
 
-  return pages
-    .filter((page) => page.path !== '/')
-    .flatMap((page) =>
-      locales.map((locale) => ({
-        rest: page.path.split('/').filter((segment) => segment !== ''),
-        locale: locale === defaultLocale ? undefined : locale,
-      })),
-    );
+  return pages.filter((page) => page.path !== '/').flatMap((page) => localesFanOut(page.path));
 }
 
-interface Props {
-  params: Promise<{
-    locale: string;
-    rest: string[];
-  }>;
-}
-
-export default async function CatchAllPage({ params }: Props) {
+export default async function CatchAllPage({ params }: { params: Promise<PageParams> }) {
   const { rest, locale } = await params;
   const path = `/${rest.join('/')}`;
 
-  const snapshot = await client.getPageSnapshot(path, {
-    siteVersion: await getSiteVersion(),
-    locale: locale === defaultLocale ? undefined : locale,
-  });
-
-  if (snapshot == null) return notFound();
-
-  return <MakeswiftPage snapshot={snapshot} />;
+  return <Page locale={locale} path={path} />;
 }
 
-export const runtime = 'nodejs';
+function localesFanOut(path: string): PageParams[] {
+  return locales.map((locale) => ({
+    rest: path.split('/').filter((segment) => segment !== ''),
+    locale,
+  }));
+}
--- a/core/lib/makeswift/client.ts
+++ b/core/lib/makeswift/client.ts
@@ -1,7 +1,11 @@
 import { Makeswift } from '@makeswift/runtime/next';
+import { getSiteVersion } from '@makeswift/runtime/next/server';
 import { strict } from 'assert';
+import { getLocale } from 'next-intl/server';
 
-import { runtime } from '~/lib/makeswift/runtime';
+import { defaultLocale } from '~/i18n/routing';
+
+import { runtime } from './runtime';
 
 strict(process.env.MAKESWIFT_SITE_API_KEY, 'MAKESWIFT_SITE_API_KEY is required');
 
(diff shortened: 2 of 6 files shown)


core/lib/makeswift/page.tsx
```diff
@@ -0,0 +1,35 @@
...
+    rest: path.split('/').filter((segment) => segment !== ''),
+    locale,
+  }));
+}
+
+export async function generateStaticParams(): Promise<PageParams[]> {
+  const pages = await client.getPages().toArray();
+
+  return pages.filter((page) => page.path !== '/').flatMap((page) => localesFanOut(page.path));
+}
```
migueloller, line 24: This feels like it should be kept in the page file since it's coupled to the route and where the page file is in the filesystem.

migueloller (changes requested):
I think we might be abstracting too much IMO. I think the page component should take `path` as a parameter and abstract getting the snapshot but not make assumptions about the route. The way it is now it won't be usable in the home page static route

core/lib/makeswift/page.tsx
agurtovoy: > This feels like it should be kept in the page file since it's coupled to the route

Yeah, I debated it, but I think the implementation is going to be the same sans the page prefix (which I forgot to pass here).

> where the page file is in the filesystem

It's re-exported from the page file.

agurtovoy: On second thought, I think I see what you mean: specifically the `rest` param is page-location specific.

agurtovoy:
Good catch, I've seen the `rest` pattern so many times that in my brain it became the default Next behavior detached from the actual page route, haha. Reworked.

fikrikarim (approved):
LGTM

fikrikarim:
I'm merging this since Aleksey addressed Miguel's requested changes.

core/app/[locale]/(default)/[...rest]/page.tsx
```diff
@@ -1,42 +1,27 @@
...
+  return <Page locale={locale} path={path} />;
 }
 
-export const runtime = 'nodejs';
+function localesFanOut(path: string): PageParams[] {
+  return locales.map((locale) => ({
+    rest: path.split('/').filter((segment) => segment !== ''),
+    locale,
+  }));
+}
```
Comment (migueloller) [inline core/app/[locale]/(default)/[...rest]/page.tsx:27]: nit: I think extracting out this function might be a bit harder for users of Catalyst to follow than leaving it inline in `generateStaticParams`. The reason is because the return type is coupled to the route params and so the function isn't reusable in other contexts. Also, I think it's helpful for `generateStaticParams` to be fully contained to remove the need to jump around to figure out what's being statically generated

core/lib/makeswift/client.ts
```diff
@@ -1,11 +1,34 @@
...
 export const client = new Makeswift(process.env.MAKESWIFT_SITE_API_KEY, {
   runtime,
   apiOrigin: process.env.MAKESWIFT_API_ORIGIN,
 });
+
+export const getPageSnapshot = async ({ path, locale }: { path: string; locale: string }) =>
+  await client.getPageSnapshot(path, {
+    siteVersion: await getSiteVersion(),
+    locale: normalizeLocale(locale),
+  });
```
migueloller, line 21: This function can only be called on a React Server Component or Route Handler since it uses `getSiteVersion`. We might want to inline this in the `<Page>` RSC so users don't mistakenly try to use this elsewhere. What do you think?

fikrikarim:
Sorry, I merged this early. I thought that was your only comment @migueloller. 

Do you want to handle the changes @agurtovoy?

core/lib/makeswift/client.ts
agurtovoy: @migueloller Are you saying that they might see `getPageSnapshot` and try to call it instead `client.getPageSnapshot` in contexts outside of rendering? I can theoretically see that happening, but I don't think it's an issue worth sacrificing the function for. It mirrors `getComponentSnapshot` and I like that all of our uses of `getSiteVersion` and `normalizeLocale` are self-contained in the single file. If somewhere were to copy & paste the `lib/maskeswift` dir in a different project, potentially w/ a different router, this would be the only file they need to change.

migueloller: I think that all uses should be hidden behind a `<Page>` or `<MakeswiftComponent>` RSC anyway. There should be no need to call `getPageSnapshot` or `getComponentSnapshot` outside those two RSCs

migueloller: Also, if this were to be copy/pasted to a Pages Router app, it would be incompatible. If it's App Router, then my previous comment stands: the RSCs can be the API, hiding `getComponentSnapshot` and `getPageSnapshot` completely

core/app/[locale]/(default)/[...rest]/page.tsx
Comment (migueloller) [reply core/app/[locale]/(default)/[...rest]/page.tsx:27]: If the function was incomprehensible because of the chained `map` and `flatMap` perhaps separating it out into different variables might help with readability. I just think that there's a cost to having to go find the function being called to understand how the path is being mapped to the route params (in particular because it doesn't take `locales` as a param and reads it from module context)

core/lib/makeswift/client.ts
agurtovoy: > Also, if this were to be copy/pasted to a Pages Router app, it would be incompatible

Yes, but the incompatibility/point of variability would be limited to a single file rather than spread through the dir.

> I think that all uses should be hidden behind a <Page> or <MakeswiftComponent> RSC anyway. There should be no need to call `getPageSnapshot` or `getComponentSnapshot` outside those two RSCs

That's true. We don't have a userland `<MakeswiftComponent>` right now; I'll make one/refactor the `getComponentSnapshot` usage once https://github.com/bigcommerce/catalyst/pull/1783 is merged.

core/app/[locale]/(default)/[...rest]/page.tsx
Comment (agurtovoy) [reply core/app/[locale]/(default)/[...rest]/page.tsx:27]: > I just think that there's a cost to having to go find the function being called to understand how the path is being mapped

Sure, there is a cost to everything; inlining the logic or putting it into a variable makes you pay the cost of parsing and understanding code with multiple variables coming from different scopes upfront. Giving a piece of logic a mnemonic name and tacking it nearby gives you a chance to understand the bigger picture at glance and make a decision whether you want to explore the details deeper now or later (or never). Breaking down the problem into smaller semantic steps with lesser parts is worth a lot more to me than the cost of clicking on a function to jump to its definition when I'm ready to see the details.

> in particular because it doesn't take locales as a param and reads it from module context

The global `locales` being buried inside of an arrow function inside of `flatMap` is worse, isn't it?

PR #1785, opened 2024-12-19
feat: add pagination and breadcrumbs to /blog and /blog/tag/[tagId]

jordanarldt:
## What/Why?
- Added breadcrumbs to `/blog` and `/blog/tag/[tagId]` routes - when testing yesterday I realized that there was not an easy way to navigate after filtering for a blog tag
- Added `CursorPagination` to `FeaturedBlogPostList`
- Refactored `CursorPagination` component and its uses, as I noticed we were using `suspense`, even though the component already uses suspense internally. 
- Created utility function to transform GraphQL pagination info into the `CursorPaginationInfo` type

## Testing

https://github.com/user-attachments/assets/2224741c-2d9e-4d0a-8771-ed82bcab5694

### DIFF
--- a/core/app/[locale]/(default)/blog/tag/[tagId]/page.tsx
+++ b/core/app/[locale]/(default)/blog/tag/[tagId]/page.tsx
@@ -2,6 +2,7 @@ import type { Metadata } from 'next';
 import { notFound } from 'next/navigation';
 
 import { FeaturedBlogPostList } from '@/vibes/soul/sections/featured-blog-post-list';
+import { pageInfoTransformer } from '~/data-transformers/page-info-transformer';
 
 import { getBlog, getBlogMetaData, getBlogPosts } from '../../page-data';
 
@@ -16,19 +17,25 @@ export async function generateMetadata(): Promise<Metadata> {
   return await getBlogMetaData();
 }
 
-async function listBlogPosts(
-  paramsPromise: Promise<{ tagId: string }>,
-  searchParamsPromise: Promise<SearchParams>,
-) {
-  const { tagId } = await paramsPromise;
-  const searchParams = await searchParamsPromise;
+async function listBlogPosts(props: Props) {
+  const { tagId } = await props.params;
+  const searchParams = await props.searchParams;
   const blogPosts = await getBlogPosts({ tagId, ...searchParams });
   const posts = blogPosts?.posts ?? [];
 
   return posts;
 }
 
+async function getPaginationInfo(props: Props) {
+  const { tagId } = await props.params;
+  const searchParams = await props.searchParams;
+  const blogPosts = await getBlogPosts({ tagId, ...searchParams });
+
+  return pageInfoTransformer(blogPosts?.pageInfo);
+}
+
 export default async function Tag(props: Props) {
+  const { tagId } = await props.params;
   const blog = await getBlog();
 
   if (!blog) {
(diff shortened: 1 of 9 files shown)


core/data-transformers/page-info-transformer.ts
```diff
@@ -0,0 +1,23 @@
...
+  endCursor: string | null;
+}
+
+/**
+ * Utility function to convert GraphQL PageInfo object into CursorPaginationInfo
+ *
+ * @param {PageInfo | undefined} pageInfo - The PageInfo object to transform
+ * @returns {CursorPaginationInfo}
+ */
+export function pageInfoTransformer(pageInfo: PageInfo | undefined): CursorPaginationInfo {
```
migueloller, line 16: 🍹 Is there a situation where we'd pass `undefined` here? I would've thought it would come as `null` from GraphQL and never undefined

core/app/[locale]/(default)/blog/tag/[tagId]/page.tsx
```diff
@@ -16,19 +17,25 @@ export async function generateMetadata(): Promise<Metadata> {
...
 
   return posts;
 }
 
+async function getPaginationInfo(props: Props) {
+  const { tagId } = await props.params;
+  const searchParams = await props.searchParams;
+  const blogPosts = await getBlogPosts({ tagId, ...searchParams });
+
+  return pageInfoTransformer(blogPosts?.pageInfo);
```
Comment (migueloller) [inline core/app/[locale]/(default)/blog/tag/[tagId]/page.tsx:34]: 🍹 If we do change the `| undefined` in `pageInfoTransformer` we could change this as follows
```suggestion
  return pageInfoTransformer(blogPosts?.pageInfo ?? null);
```

migueloller (changes requested):
Looks good! I just think we need to address the issues on the VIBES side https://github.com/makeswift/vibes/pull/430 and bring those back here

core/data-transformers/page-info-transformer.ts
jordanarldt: @migueloller In the scenario where `blog` node comes back `null`, this would mean that `blog.posts.pageInfo` would be `undefined`.

Perhaps just for ease-of-use, we could do `PageInfo | null | undefined`?

core/app/[locale]/(default)/blog/tag/[tagId]/page.tsx
Comment (jordanarldt) [reply core/app/[locale]/(default)/blog/tag/[tagId]/page.tsx:34]: @migueloller I think it makes sense to go with `PageInfo | undefined | null`, as it will cover most (if not all) possible cases for the different GraphQL contracts/responses, and prevent the need to use nullish coalescing operators (or other operators) to convert between `undefined` or `null` before calling the function.

Comment (chanceaclark) [reply core/app/[locale]/(default)/blog/tag/[tagId]/page.tsx:34]: I think we should have the argument type be `ResultOf<typeof PaginationFragment>` and enforce the data exists before being passed into the function.

migueloller (approved):
Looking great!

core/data-transformers/page-info-transformer.ts
```diff
@@ -0,0 +1,22 @@
...
+};
+
+export function pageInfoTransformer(
+  pageInfo: ResultOf<typeof PaginationFragment>,
+): CursorPaginationInfo {
+  return {
+    startCursorParamName: 'before',
+    startCursor: pageInfo.startCursor,
+    endCursorParamName: 'after',
+    endCursor: pageInfo.endCursor,
```
jorgemoya, line 20: Could there be a situation where we have two paginations in the same page, and since we can't modify the param names using this transformer, we'd be screwed 🤔 .

I guess for now lets not think about this because it is not the case.

jorgemoya (approved):

PR #1786, opened 2024-12-19
feat: add reviews to PDP

jorgemoya:
## What/Why?
+ Add reviews section to PDP
+ Include pagination
+ Issue with layout overflow in Review section component is unrelated and being worked on a different PR

Screenshot 2024-12-19 at 12 39 01 PM

Screenshot 2024-12-19 at 12 39 13 PM


core/app/[locale]/(default)/product/[slug]/page.tsx
```diff
@@ -164,6 +165,10 @@ export default async function Product(props: Props) {
         quantityLabel={t('ProductDetails.quantity')}
       />
 
+      <Stream fallback={null} value={productPromise}>
+        {(product) => <Reviews productId={product.entityId} />}
+      </Stream>
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/product/[slug]/page.tsx:170]: I wonder if I should use the `ReviewSkeleton` here, or if we don't really care since this should live below the fold. 🤔

Comment (migueloller) [reply core/app/[locale]/(default)/product/[slug]/page.tsx]: Yeah, should definitely use a skeleton. Feel free to decouple from this PR if you want though

Comment (migueloller) [reply core/app/[locale]/(default)/product/[slug]/page.tsx]: Wait, aren't we already handling streaming inside the `Reviews` component? Why do we need to use `Stream` here?

migueloller (approved):
LGTM, only have 1 comment about using `Stream` without it being needed since the `Reviews` component already handles it

core/app/[locale]/(default)/product/[slug]/page.tsx
Comment (jorgemoya) [reply core/app/[locale]/(default)/product/[slug]/page.tsx]: 🤦 You're right! Updated!

PR #1787, opened 2024-12-19
feat: Soul ProductsListSection for search results

migueloller:
## What/Why?

Uses the Soul product list section component for the `/search` page.

## Testing

https://github.com/user-attachments/assets/cf36d5df-7128-41be-a49b-ac7013669852


jorgemoya (approved):
LGTM!

jorgemoya:
I'm guessing we're okay with removing the search form inside this page? cc @bookernath 

I believe having the one on the header is enough, but just want to make sure we're in sync regarding the flow of search.

migueloller:
> I'm guessing we're okay with removing the search form inside this page? cc @bookernath
> 
> I believe having the one on the header is enough, but just want to make sure we're in sync regarding the flow of search.

I created an issue to track all this stuff so we can come back later and make the search experience more bespoke than generic PLP

PR #1797, opened 2024-12-20
feat: add support for custom blog url routing, update blog tag filtering, delete /blog/tag/[tagId] route

jordanarldt:
## What/Why?
- Add support for custom blog URL routing - we were generating links to the blog based on the BC URL for the blog index, but had no routing for it.
- Now that custom blog URL routing will work, we needed to refactor blog tag filtering to be a query parameter
- Delete `/blog/tag/[tagId]` route (replaced with `/blog?tag=[tag]` query param

## Testing

Blog posts, blog index, tag filtering, and blog pagination all work as expected:

https://github.com/user-attachments/assets/1d89f201-0781-4727-a686-ac5170d25d7f


core/app/[locale]/(default)/blog/page.tsx
```diff
@@ -4,9 +4,9 @@ import { notFound } from 'next/navigation';
 import { FeaturedBlogPostList } from '@/vibes/soul/sections/featured-blog-post-list';
 import { defaultPageInfo, pageInfoTransformer } from '~/data-transformers/page-info-transformer';
 
-import { getBlog, getBlogMetaData, getBlogPosts } from './page-data';
+import { BlogPostsFiltersInput, getBlog, getBlogMetaData, getBlogPosts } from './page-data';
 
-type SearchParams = Record<string, string | string[] | undefined>;
+type SearchParams = Record<string, string | string[] | undefined> & BlogPostsFiltersInput;
```
Comment (migueloller) [inline core/app/[locale]/(default)/blog/page.tsx:9]: I don't think it's safe to add types to `SearchParams` like this since a user could type anything. Instead, I think we should use `nuqs` to properly parse search params to get type safety and parsing of external inputs

migueloller (commented):
Looks good but I think we should use `nuqs` for search param parsing and serialization to avoid things blowing up when encountering an unexpected value for the `tag` search param (e.g., passing multiple tags or passing an empty value, etc.)

migueloller (approved):

PR #1806, opened 2024-12-23
feat: update /blog/[blogId] route to use skeleton and streamable props

jordanarldt:
## What/Why?
- Update blog post route `/blog/[blogId]` to use the streamable props
- This will display the blog post content skeleton during load

### Mobile

image

image

### Desktop

image

image


migueloller (approved):
💯

PR #1807, opened 2024-12-24
refactor: move app fonts from root layout to their own file

agurtovoy:
## What/Why?

We want to be able to import the fonts elsewhere, plus this arrangement matches Next.js examples as well: https://nextjs.org/docs/pages/building-your-application/optimizing/fonts#using-multiple-fonts

There are no semantic changes in this PR.

## Testing

Screenshot 2024-12-23 at 8 10 46 PM

https://github.com/user-attachments/assets/bf2831ec-cb79-48c3-be69-ed52ac5bba80


migueloller (approved):
LGTM 👍🏻

I hadn't seen this pattern before. Can you share some of the examples you found that use it (other than docs)?

agurtovoy:
@migueloller https://github.com/search?q=%22app%2Ffonts%22+next&type=code

PR #1808, opened 2024-12-25
feat: visually editable PDP

agurtovoy:
## What/Why?

Make product detail page visually editable.

## Testing

https://github.com/user-attachments/assets/c210f664-8145-403c-b792-e104d2e10612

### DIFF
--- a/core/app/[locale]/(default)/product/[slug]/page.tsx
+++ b/core/app/[locale]/(default)/product/[slug]/page.tsx
@@ -4,10 +4,10 @@ import { getFormatter, getTranslations, setRequestLocale } from 'next-intl/serve
 
 import { Stream } from '@/vibes/soul/lib/streamable';
 import { FeaturedProductsCarousel } from '@/vibes/soul/sections/featured-products-carousel';
-import { ProductDetail } from '@/vibes/soul/sections/product-detail';
 import { pricesTransformer } from '~/data-transformers/prices-transformer';
 import { productCardTransformer } from '~/data-transformers/product-card-transformer';
 import { productOptionsTransformer } from '~/data-transformers/product-options-transformer';
+import { ProductDetail } from '~/lib/makeswift/components/product-detail';
 
 import { addToCart } from './_actions/add-to-cart';
 import { ProductSchema } from './_components/product-schema';
@@ -100,6 +100,7 @@ const getProduct = async (productPromise: ReturnType<typeof getProductData>) => 
     description: (
       <div className="prose" dangerouslySetInnerHTML={{ __html: product.description }} />
     ),
+    plainTextDescription: product.plainTextDescription,
     href: product.path,
     images: product.defaultImage
       ? [{ src: product.defaultImage.url, alt: product.defaultImage.altText }, ...images]
@@ -232,6 +233,7 @@ export default async function Product(props: Props) {
         incrementLabel={t('ProductDetails.increaseQuantity')}
         prefetch={true}
         product={getProduct(productPromise)}
+        productId={productId}
         quantityLabel={t('ProductDetails.quantity')}
         thumbnailLabel={t('ProductDetails.thumbnail')}
       />
--- a/core/lib/makeswift/component.tsx
+++ b/core/lib/makeswift/component.tsx
@@ -4,13 +4,14 @@ import { getComponentSnapshot } from '~/lib/makeswift/client';
 
 export const Component = async ({
   snapshotId,
+  label,
   ...props
 }: {
   type: string;
-  label: string;
+  label: string | Promise<string>;
   snapshotId: string;
 }) => {
   const snapshot = await getComponentSnapshot(snapshotId);
 
-  return <MakeswiftComponent snapshot={snapshot} {...props} />;
+  return <MakeswiftComponent label={await label} snapshot={snapshot} {...props} />;
 };
--- a/core/lib/makeswift/components/product-detail/client.tsx
+++ b/core/lib/makeswift/components/product-detail/client.tsx
@@ -1,0 +1,125 @@
+'use client';
+
+import React, {
+  type ComponentPropsWithoutRef,
+  createContext,
+  forwardRef,
+  type PropsWithChildren,
+  type ReactNode,
+  type Ref,
+  useCallback,
+  useContext,
+} from 'react';
+
+import { Stream, type Streamable } from '@/vibes/soul/lib/streamable';
+import { ProductDetail, ProductDetailSkeleton } from '@/vibes/soul/sections/product-detail';
+import { mergeSections } from '~/lib/makeswift/utils/merge-sections';
+
+type VibesProductDetailProps = ComponentPropsWithoutRef<typeof ProductDetail>;
+type VibesProductDetail = Exclude<Awaited<VibesProductDetailProps['product']>, null>;
+
+export type ProductDetail = VibesProductDetail & {
+  plainTextDescription?: string;
+};
+
+export type Props = Omit<VibesProductDetailProps, 'product'> & {
+  product: Streamable<ProductDetail>;
+};
+
+const PropsContext = createContext<Props | null>(null);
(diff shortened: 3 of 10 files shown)


core/lib/makeswift/component.tsx
```diff
@@ -4,13 +4,14 @@ import { getComponentSnapshot } from '~/lib/makeswift/client';
 
 export const Component = async ({
   snapshotId,
+  label,
   ...props
 }: {
   type: string;
-  label: string;
+  label: string | Promise<string>;
```
agurtovoy, line 11: Allow label to be a promise so we can provide a descriptive label [here](https://github.com/bigcommerce/catalyst/pull/1808/files#diff-a4b550ad0e2dccf6642504f8ce1f2bff09d72d78cf355d59127fabbf1609aeb4R11).

Note that waiting for the label to resolve below feels a bit off as strictly speaking label is not required for rendering the snapshot element tree and could theoretically be resolved in parallel. Might be something we want to pursue later.

core/lib/makeswift/components/product-detail/client.tsx
```diff
@@ -0,0 +1,123 @@
...
+          }))
+        : undefined,
+    [accordions],
+  );
+
+  return (
+    <Stream value={streamableProduct}>
+      {(product) => (
+        <Stream value={product.accordions}>
+          {(productAccordions) => (
```
agurtovoy, line 87: @migueloller Nested streamables still cause a bit of complexity here.

core/vibes/soul/lib/streamable.tsx
```diff
@@ -1,7 +1,84 @@
 import { Suspense, use } from 'react';
+import { v4 as uuid } from 'uuid';
```
agurtovoy, line 2: Here and below, pulled https://github.com/makeswift/vibes/pull/455 changes from VIBES.

core/lib/makeswift/component.tsx
migueloller: Should this be `Streamable<string>`?

core/lib/makeswift/components/product-detail/client.tsx
migueloller: All good. Don't we need to provide appropriate fallbacks here so that we don't see FOUC?

core/vibes/soul/lib/streamable.tsx
migueloller: Did Catalyst already have the `uuid` dependency?

core/app/[locale]/(default)/product/[slug]/page.tsx
```diff
@@ -232,6 +233,8 @@ export default async function Product(props: Props) {
         incrementLabel={t('ProductDetails.increaseQuantity')}
         prefetch={true}
         product={getProduct(productPromise)}
+        productId={productId}
+        productTitle={productPromise.then((product) => product.name)}
```
Comment (migueloller) [inline core/app/[locale]/(default)/product/[slug]/page.tsx:237]: How come we have `productId` and `productTitle` props? Shouldn't the `product` prop have `id` and `name`/`title` that we can use?

migueloller (approved):

core/lib/makeswift/component.tsx
agurtovoy: I didn't want to make this dependent on Vibes.

core/vibes/soul/lib/streamable.tsx
agurtovoy: > Did Catalyst already have the uuid dependency?

core/app/[locale]/(default)/product/[slug]/page.tsx
Comment (agurtovoy) [reply core/app/[locale]/(default)/product/[slug]/page.tsx:237]: `product` is a promise here. Awaiting on `product` to get `productId` is conceptually backwards, IMHO, regardless of whether we have or don't have to await on it to get other props. `productTitle` resolution can be moved inside of `<ProductDetail/>`.

Comment (agurtovoy) [reply core/app/[locale]/(default)/product/[slug]/page.tsx]: Done.

core/lib/makeswift/components/product-detail/client.tsx
agurtovoy: It didn't seem to make a difference, but I've added them anyway.

core/vibes/soul/sections/product-detail/index.tsx
```diff
@@ -284,7 +283,7 @@ function ProductAccordionsSkeleton() {
   );
 }
 
-function ProductDetailSkeleton() {
+export function ProductDetailSkeleton() {
```
agurtovoy, line 286: TODO: put up a PR for VIBES that exposes this.

core/app/[locale]/(default)/product/[slug]/page.tsx
Comment (migueloller) [reply core/app/[locale]/(default)/product/[slug]/page.tsx]: In practice, we want to stream in the whole product, though. We don't have to have to fetch the entire product just to pass the ID as that defeats the purpose of streamability, no?

It's not like the component fetches the product based on the `productId` prop

migueloller (approved):

core/lib/makeswift/component.tsx
agurtovoy: > Catalyst itself depends on VIBES so this shouldn't be an issue.

I know it does, but I view the collection of "userland" Makeswift boilerplate in this dir as a prototype for our updated integration guidelines, and I don't think they should depend on VIBES.

migueloller: Would we ever want to make `label` a promise in that userland implementation? If we did, wouldn't that effectively be introducing the streamable pattern, at which point we'd want to introduce the concept and the library, etc.?

agurtovoy: > Would we ever want to make label a promise in that userland implementation?

I think we should; it's likely that giving a "dynamic" component a meaningful label would require a fetch.

> If we did, wouldn't that effectively be introducing the streamable pattern

I personally don't think the label should be "streamed" here (awaited while delaying rendering/rendering a placeholder). I think it should be resolved in parallel and assigned to the snapshot's element tree asynchronously, independently of what happens in React.

core/vibes/soul/sections/product-detail/index.tsx
agurtovoy: Done: https://github.com/makeswift/vibes/pull/458

core/lib/makeswift/component.tsx
migueloller: > I think we should; it's likely that giving a "dynamic" component a meaningful label would require a fetch.

Given that the snapshot is not a promise, though, and that we `await` label without a suspense boundary, there seems to be no benefit to taking `label` as a promise. I think that taking in promises like this (i.e., the streamable pattern) is only relevant when the component uses `Suspense` and handles the fallback appropriately. Simply making the prop a promise won't make it stream

> I personally don't think the label should be "streamed" here (awaited while delaying rendering/rendering a placeholder). I think it should be resolved in parallel and assigned to the snapshot's element tree asynchronously, independently of what happens in React.

See my point above, accepting a promise here is useless since we `await` it immediately and don't use `Suspense`. The user can `await` it themselves before passing it to us.

agurtovoy: @migueloller It's useless now, what I was proposing in my [original comment](https://github.com/bigcommerce/catalyst/pull/1808#discussion_r1907690554) is that, because label is not required for rendering the snapshot element tree and could theoretically be resolved in parallel, we make label an optional promise in the runtime and do resolve it in parallel.

migueloller: When you say "in parallel", do you mean "lazily" so that we do it at a later time and not in the path of rendering the component? For example, perhaps we resolve the promise right before we send the label via `postMessage` to the builder?

agurtovoy: Yes, exactly.

PR #1820, opened 2024-12-30
Improve CLI user experience & create continuity between control panel and CLI

bookernath:
## What/Why?
- Update CLI to use OCC APIs
- Restructuring & simplifying code
- Use new OCC client ID
- Add sample data prompt
- Create `.catalyst` file with store hash/access token after login action to be used for other CLI commands without re-prompting for login
- Ensure CLI-created storefronts have the same end state in the control panel as storefronts created in the control panel
- Various UX improvements like automatically opening the device auth URL

## Testing
Create wizard (the extraneous log statement is not there in this actual PR)

https://github.com/user-attachments/assets/4b173c35-000f-4724-8c86-92f4f685b091

Starting from app (I invalidated these credentials before posting this)

https://github.com/user-attachments/assets/368e8a4d-e578-4f25-b92d-2a6df7c38a63

Init command

https://github.com/user-attachments/assets/b1f67f0c-a0dd-42a7-92c2-e5f772d811b7

### DIFF
--- a/packages/create-catalyst/src/commands/create.ts
+++ b/packages/create-catalyst/src/commands/create.ts
@@ -151,6 +151,8 @@ export const create = new Command('create')
     if (!projectName) throw new Error('Something went wrong, projectName is not defined');
     if (!projectDir) throw new Error('Something went wrong, projectDir is not defined');
 
+    let envVars: Record<string, string> = {};
+
     if (!options.storeHash || !options.accessToken) {
       const credentials = await login(`https://login.${options.bigcommerceHostname}`);
 
@@ -240,23 +242,7 @@ export const create = new Command('create')
         const channelData = (await response.json()) as CreateChannelResponse;
         channelId = channelData.data.id;
         storefrontToken = channelData.data.storefront_api_token;
-
-        const envVars = { ...channelData.data.envVars };
-
-        // Add any CLI-provided env vars as overrides
-        if (options.env) {
-          const cliEnvVars = options.env.reduce((acc, env) => {
-            const [key, value] = env.split('=');
-            if (key && value) {
-              acc[key] = value;
-            }
-            return acc;
-          }, {} as Record<string, string>);
-
-          Object.assign(envVars, cliEnvVars);
-        }
-
-        writeEnv(projectDir, envVars);
+        envVars = { ...channelData.data.envVars };
       }
 
       if (!shouldCreateChannel) {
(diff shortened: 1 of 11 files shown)


packages/create-catalyst/src/utils/config.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  return true;
+}
+
+export class Config {
+  private configPath: string;
+  private config: CatalystConfig;
+
+  constructor(projectDir: string) {
+    this.configPath = join(projectDir, '.catalyst');
```
becomevocal, line 46: This is being used to handle .env vars for the CLI, right? Why wouldn't we use a clearer naming scheme for the file that makes it clear it's an additional .env file. To prevent potential user error of committing it in the future.

Note: `.env-catalyst-cli.local` should be caught automatically by the main [.gitignore](https://github.com/bigcommerce/catalyst/blob/main/.gitignore) (could add another catch to make it a cleaner filename)

I am pulling from my recent experience with Sentry, where their CLI automatically creates a `.env.sentry-build-plugin`:
```
# DO NOT commit this file to your repository!
# The SENTRY_AUTH_TOKEN variable is picked up by the Sentry Build Plugin.
# It's used for authentication when uploading source maps.
# You can also set this env variable in your own `.env` files and remove this file.
SENTRY_AUTH_TOKEN=***********
```

packages/create-catalyst/src/utils/config.ts
```diff
@@ -0,0 +1,100 @@
...
+    this.config.auth = { storeHash, accessToken };
+    this.save();
+  }
+
+  save(): void {
+    const configObj: TomlRecord = {};
+
+    if (this.config.auth?.storeHash && this.config.auth.accessToken) {
+      configObj.auth = {
+        storeHash: this.config.auth.storeHash,
```
becomevocal, line 64: Will the same store hash be in the main .env file and this config? If it is, do we need the store hash in both?

becomevocal (changes requested):
Looks good. I just want to make sure we are thinking about the .env file makeup.

packages/create-catalyst/src/utils/config.ts
bookernath: This filename was chosen a while ago and is guaranteed to be in `.gitignore` for all previous users. I think we'd probably name it slightly differently today (this was inspired by `.stencil`) but I think it's better to stick with this previous decision.

https://github.com/bigcommerce/catalyst/pull/575

Agree we should add some preamble to the file though, good idea

packages/create-catalyst/src/utils/config.ts
bookernath: The utility of the `.catalyst` file is you can regenerate the .env.local, so it makes sense for it to have no dependencies on .env.local

packages/create-catalyst/src/commands/create.ts
```diff
@@ -23,6 +55,175 @@ function getPlatformCheckCommand(command: string): string {
...
+
+  return {
+    channelId: channelData.data.id,
+    storefrontToken: channelData.data.storefront_api_token,
+    envVars: channelData.data.envVars,
+  };
+}
+
+async function handleChannelSelection(bc: Https) {
+  const channelSortOrder = ['catalyst', 'next', 'bigcommerce'];
```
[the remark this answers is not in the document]
bookernath: Legacy code, but I updated it so other storefront types will appear at the end.

packages/create-catalyst/src/utils/config.ts
bookernath: Preamble added.

becomevocal (approved):

becomevocal:
Appreciate the changes post review. Flow looks 🧈

PR #1822, opened 2024-12-31
Improve security of github action

bookernath:
## What/Why?
https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections

## Testing
N/A


.github/workflows/translations-changeset.yml
```diff
@@ -11,13 +11,27 @@ jobs:
...
+        id: validate
+        run: |
+          # Validate ref name against allowed pattern (alphanumeric, dash, underscore, and forward slash only)
+          if ! [[ "${{ github.event.pull_request.head.ref }}" =~ ^[a-zA-Z0-9/_-]+$ ]]; then
+            echo "Error: Invalid branch name format"
+            exit 1
+          fi
+          echo "ref=${{ github.event.pull_request.head.ref }}" >> $GITHUB_OUTPUT
+
       - name: Use commit SHA for filename
```
becomevocal, line 35: Was this removed intentionally?

.github/workflows/translations-changeset.yml
```diff
@@ -11,13 +11,27 @@ jobs:
...
       - name: Checkout repo
         uses: actions/checkout@v4
         with:
           fetch-depth: 2
 
+      - name: Validate inputs
+        id: validate
+        run: |
+          # Validate ref name against allowed pattern (alphanumeric, dash, underscore, and forward slash only)
+          if ! [[ "${{ github.event.pull_request.head.ref }}" =~ ^[a-zA-Z0-9/_-]+$ ]]; then
```
bc-donfran, line 29: With this line i think we should still env the `${{ github.event.pull_request.head.ref }} `as if you enter something like `"]]asdf"` it can escape the shell condition which can lead to RCE.

Using intermediate env vars like here: https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-an-intermediate-environment-variable should treat any special characters as escaped which we can safely use later

chanceaclark: I think we could use the default environment variable `GITHUB_HEAD_REF` instead? Should probably have the same escaped value 🤔 

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables

bookernath:
Will reopen against `canary`

PR #1827, opened 2024-12-31
feat: add Error page and ErrorBoundary

jordanarldt:
## What/Why?
- Use the `Error` component from VIBES
- Modify existing `error` page to use this component
- Add an ErrorBoundary in `(default)` route to maintain layout structure when error occurs

## Testing

image


core/app/[locale]/(default)/error.tsx
```diff
@@ -0,0 +1,5 @@
+'use client';
+
+import { lazy } from 'react';
+
+export default lazy(() => import('../_components/error'));
```
Comment (migueloller) [inline core/app/[locale]/(default)/error.tsx:5]: How come we're doing this? Next.js files are already lazily loaded so this shouldn't be necessary

Comment (migueloller) [reply core/app/[locale]/(default)/error.tsx:5]: We should define a new component here and render the error component so it's clear we're passing `retry` as a function and not a server action

migueloller (changes requested):
Looks good. Do we have a way to test this?

jordanarldt:
> Looks good. Do we have a way to test this?

@migueloller I tested on my sandbox store, since my category pages are still returning 500 errors due to that weird `nuqs` error when I have no product filters 👍  That's how I got the screenshots.

If we were to make functional tests for this behavior though, not entirely sure how we'd go about testing - we'd have to force an internal error.

migueloller:
> > Looks good. Do we have a way to test this?
> 
> @migueloller I tested on my sandbox store, since my category pages are still returning 500 errors due to that weird `nuqs` error when I have no product filters 👍 That's how I got the screenshots.
> 
> If we were to make functional tests for this behavior though, not entirely sure how we'd go about testing - we'd have to force an internal error.

We should be able to do this by passing a streamable prop that throws whenever we implement the test

jordanarldt:
> We should be able to do this by passing a streamable prop that throws whenever we implement the test

@migueloller Yeah, that's what I had in mind also. For a functional / E2E test though, not sure how we'd force the error there 💭

core/app/[locale]/error.tsx
```diff
@@ -1,5 +1,16 @@
 'use client';
 
-import { lazy } from 'react';
+import { useTranslations } from 'next-intl';
 
-export default lazy(() => import('./_components/error'));
```
Comment (jorgemoya) [inline core/app/[locale]/error.tsx:5]: We were doing this to prevent `next-intl` from loading as part of the initial bundle, but at this point I don't know if this is happening anyway since we use the context. More info [here](https://next-intl.dev/docs/environments/error-files).

jorgemoya (approved):
LGTM. Having two error boundaries makes sense, even though we duplicate code, which seems off.

migueloller (approved):

PR #1839, opened 2025-01-03
fix: revert expireTag to revalidateTag

jorgemoya:
## What/Why?
Next has tagged `expireTag/expireTag` as `unstable_`. This PR reverts to use `revalidateTag/revalidatePath`.

## Testing
Locally no issues.


jorgemoya:
Closing since we we'll tackle this when we upgrade Next.js.

PR #1844, opened 2025-01-03
feat(orders): update the orders list to use vibes section

chanceaclark:
## What/Why?
Updates the orders list page to use Vibes. Also, had to do some cleanup on some translation strings around the order details views.

## Testing
Screenshot 2025-01-03 at 16 47 34


core/data-transformers/orders-transformer.ts
```diff
@@ -0,0 +1,96 @@
...
+  // all the orders on the page and fetch them all at once.
+  const { data } = await client.fetch({
+    document: ProductAttributes,
+    // We only display the first 5 orders on the page and we truncate the amount of
+    // products are shown per order. The max products shown per order is 8 based on
+    // the largest screen size, so we can safely fetch 50 products and feel comfortable.
+    // 50 is the max we can fetch at once.
+    variables: { entityIds: productIds, first: 50 },
+    fetchOptions: { next: { revalidate } },
+  });
```
chanceaclark, line 56: This is unfortunate, but the only way I can see us circumventing this is adding a node on our GQL API. Pretty sure the team wanted to avoid adding it though as it may causes an N+1, but I'd need to reach out to the team to confirm that.

migueloller (approved):
Nice! 🚀

PR #1846, opened 2025-01-03
Set default ghref for CLI to @bigcommerce/catalyst-core@latest

bookernath:
## What/Why?
Set default ghref to use the latest tag, so that CLI users will not be exposed to canary code by default.

## Testing
Ran `create` with no args:
image

### DIFF
--- a/packages/create-catalyst/src/commands/create.ts
+++ b/packages/create-catalyst/src/commands/create.ts
@@ -244,7 +244,11 @@ export const create = new Command('create')
   .option('--access-token <token>', 'BigCommerce access token')
   .option('--channel-id <id>', 'BigCommerce channel ID')
   .option('--storefront-token <token>', 'BigCommerce storefront token')
-  .option('--gh-ref <ref>', 'Clone a specific ref from the source repository')
+  .option(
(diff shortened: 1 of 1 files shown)


migueloller (approved):

packages/create-catalyst/src/commands/create.ts
```diff
@@ -244,7 +244,11 @@ export const create = new Command('create')
   .option('--access-token <token>', 'BigCommerce access token')
   .option('--channel-id <id>', 'BigCommerce channel ID')
   .option('--storefront-token <token>', 'BigCommerce storefront token')
-  .option('--gh-ref <ref>', 'Clone a specific ref from the source repository')
+  .option(
+    '--gh-ref <ref>',
+    'Clone a specific ref from the source repository',
+    '@bigcommerce/catalyst-core@latest',
```
migueloller, line 250: I would've thought this was set to `main` before?

Is there a place where `main` is hardcoded that we need to change? Or was it just using the GitHub default branch and thus now it would be `canary` so by always passing a value for `--gh-ref`, even if the default, then the default GitHub branch won't be used implicitly anymore?

PR #1849, opened 2025-01-06
Makeswift bug template

jamesqquick:
This adds a bug report template that is more specific to Makeswift. Right now, it is pretty similar to the existing bug report, but this gives us the ability to evolve this one separately as needed and also track issues with Makeswift more easily.


bookernath:
@jamesqquick please re-create this PR into `canary` as the default branch has changed.

PR #1854, opened 2025-01-06
initial Klevu integration

jerrypena1:
## What/Why?
This is the initial integration of Klevu webcomponents into Catalys

### DIFF
--- a/core/components/ui/klevu/init.jsx
+++ b/core/components/ui/klevu/init.jsx
@@ -1,0 +1,33 @@
+'use client'
+
+import { KlevuInit } from '@klevu/ui-react';
+import React from 'react';
+
+export function KlevuInitWrapper({ children }) {
+
+  return (
+    <KlevuInit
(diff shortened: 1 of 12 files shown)


core/components/ui/klevu/init.jsx
matthewvolk, line 1: My only review comment is to do with this file - and I think you probably already saw it coming from what you mentioned in your README 🙂 
```
This file contains the KlevuInit component which needs two attributes specific to your Klevu account `apiKey` & `url`. 
```

Is there any chance we can replace the hard coded `apiKey` and `url` with a `NEXT_PUBLIC_` environment variable? This will help with developers wanting to swap out keys, but will also help with the OCC deployment - we can programmatically add these variables potentially during deployment. 

This is my only change requested, this PR looks good for approval after this!

matthewvolk (changes requested):
Just one! Thanks Juan 😄

PR #1860, opened 2025-01-07
refactor(product detail): no nested `Streamable` props

agurtovoy:
## What/Why?

Rework Soul's `<ProductDetail>` section to use a flat set of `Streamable` props. See [this thread](https://makeswifthq.slack.com/archives/C07DAL1D8JK/p1735272158243889?thread_ts=1735186593.042939&cid=C07DAL1D8JK) for more context.

### Before

https://github.com/user-attachments/assets/1b8e756d-a758-4b5f-bba1-91e58f03288d

### After

https://github.com/user-attachments/assets/4567c36b-00c1-42dc-933e-cd328aca9864


core/app/[locale]/(default)/product/[slug]/page.tsx
```diff
@@ -28,16 +28,10 @@ const getOptionValueIds = ({ searchParams }: { searchParams: Awaited<Props['sear
     );
 };
 
-const getProduct = async (productPromise: ReturnType<typeof getProductData>) => {
```
Comment (agurtovoy) [inline core/app/[locale]/(default)/product/[slug]/page.tsx:31]: Split up monolith `getProduct` into individual `getAccordions`, `getTitle`, `getSubtitle`, `getRating`, `getPrice`, `getDescription`, and `getImages` streamable getters.

core/vibes/soul/lib/streamable.tsx
```diff
@@ -2,6 +2,15 @@ import { Suspense, use } from 'react';
 
 export type Streamable<T> = T | Promise<T>;
 
+export function all<T extends readonly unknown[] | []>(
+  streamables: T,
+): Streamable<{ -readonly [P in keyof T]: Awaited<T[P]> }> {
+  return streamables.some((streamable) => streamable instanceof Promise)
+    ? Promise.all(streamables)
```
agurtovoy, line 9: Ideally this would be `Streamable.all`, but we can't namespace it like that in TS. Returns `Promise.all(streamables)` if at least one of the `streamables` is a promise, otherwise simply returns the `streamables` array.

This avoids a flash of loading indicator(s) when a component with multiple resolved streamable props that are combined using `Promise.all` is re-rendered on the client.

core/vibes/soul/sections/product-detail/index.tsx
```diff
@@ -8,154 +8,147 @@ import { ProductGallery } from '@/vibes/soul/sections/product-detail/product-gal
 import { ProductDetailForm, ProductDetailFormAction } from './product-detail-form';
 import { Field } from './schema';
 
-interface ProductDetailProduct {
```
agurtovoy, line 11: Flattened into `Props`. Will carry this over to the VIBES repo once this is approved.

core/vibes/soul/sections/product-detail/index.tsx
```diff
@@ -8,154 +8,147 @@ import { ProductGallery } from '@/vibes/soul/sections/product-detail/product-gal
...
   ctaDisabled: streamableCtaDisabled,
+  action,
   prefetch,
   thumbnailLabel,
 }: Props<F>) {
   return (
     <section className="@container">
       <div className="mx-auto w-full max-w-screen-2xl px-4 py-10 @xl:px-6 @xl:py-14 @4xl:px-8 @4xl:py-20">
         {breadcrumbs && <Breadcrumbs breadcrumbs={breadcrumbs} className="mb-6" />}
-
```
agurtovoy, line 61: Best reviewed with the "Hide whitespace" toggle on:

hide whitespace

core/vibes/soul/lib/streamable.tsx
migueloller: This is something I was waiting to introduce. The reason is that we can't use `Promise.all` in client components because we can't create new promises in client components and call `use` as React will blow up. I would like to implement a `Streamable.all` utility that works in the client and the server, but to do that we need to be smart about using `Promise.all` and `use` depending on the context. Also, perhaps the answer isn't `Streamable.all` but a `<Stream>` variation that can take in many streamables (or a overload on the existing `<Stream>`)

migueloller: > This avoids a flash of loading indicator(s) when a component with multiple resolved streamable props that are combined using Promise.all is re-rendered on the client.

This shouldn't be happening 🤔

core/vibes/soul/sections/product-detail/index.tsx
```diff
@@ -8,154 +8,147 @@ import { ProductGallery } from '@/vibes/soul/sections/product-detail/product-gal
...
   breadcrumbs,
+  title: streamableTitle,
+  subtitle: streamableSubtitle,
+  rating: streamableRating,
+  price: streamablePrice,
+  images: streamableImages,
+  fields: streamableFields,
+  summary: streamableSummary,
+  description: streamableDescription,
+  accordions: streamableAccordions,
```
migueloller, line 49: I think we should still take a single `product` prop that is streamable. I think the problem with the current implementation is we have too many fallbacks and skeletons which was implemented that way because of the nested streamables. Instead, we should have a single `product` prop and a single `<Streamable>`

migueloller (changes requested):
These changes got rid of the nested streamables but went in what I think is the opposite direction where we should go, which is very granular skeletons. I think that naturally there will be a single request for a product and we just need to make the product prop streamable

agurtovoy:
Closing, we decided to support nested `Streamable` props as a valid pattern after all, see https://github.com/makeswift/vibes/pull/455 and https://github.com/bigcommerce/catalyst/pull/1808

PR #1868, opened 2025-01-07
chore: remove store selector page that is no longer being used

jorgemoya:
## What/Why?
This page was used in the previous footer link that is no longer being used with Soul.


chanceaclark (approved):
Are we adding this back later?

jorgemoya:
> Are we adding this back later?

I don't think we are afaik.

PR #1870, opened 2025-01-07
feat(core): update address list to use VIBES

chanceaclark:
## What/Why?
Migrates the account addresses pages over to using the new address component from VIBES.

### Update Address:

https://github.com/user-attachments/assets/9505f8e9-2628-4ff5-ba83-4bcfbc4b290e

### Add Address:

https://github.com/user-attachments/assets/2747625c-abc7-40e1-ae5d-dd294e2f4577

### Delete Address:

https://github.com/user-attachments/assets/91657e3e-e768-4fd6-be75-74e6e2affead


core/app/[locale]/(default)/account/addresses/_actions/create-address.ts
```diff
@@ -0,0 +1,263 @@
...
+          String(FieldNameToFieldId.lastName),
+          String(FieldNameToFieldId.address1),
+          String(FieldNameToFieldId.address2),
+          String(FieldNameToFieldId.city),
+          String(FieldNameToFieldId.company),
+          String(FieldNameToFieldId.countryCode),
+          String(FieldNameToFieldId.stateOrProvince),
+          String(FieldNameToFieldId.phone),
+          String(FieldNameToFieldId.postalCode),
+        ].includes(field.id),
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/account/addresses/_actions/create-address.ts:120]: These functions/schemas are essentially the same as https://github.com/bigcommerce/catalyst/blob/soul/main/core/app/%5Blocale%5D/(default)/(auth)/register/_actions/register-customer.ts, but in this file I key off the field `id` rather than the `name` as the address form has a specific schema it adheres to.

core/app/[locale]/(default)/account/addresses/_actions/update-address.ts
```diff
@@ -0,0 +1,272 @@
...
+      }
+    }
+  }
+`);
+
+const stringToNumber = z.coerce.string().pipe(z.coerce.number());
+
+const inputSchema = z.object({
+  addressEntityId: stringToNumber,
+  data: z.object({
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/account/addresses/_actions/update-address.ts:49]: The only big difference between this schema and the add address schema is that the fields are optional and are nested in this `data` key.

core/app/[locale]/(default)/account/addresses/_actions/update-address.ts
```diff
@@ -0,0 +1,272 @@
...
+  }),
+});
+
+function parseUpdateAddressInput(
+  value: Record<string, unknown>,
+  fields: Array<Field | FieldGroup<Field>>,
+): VariablesOf<typeof UpdateCustomerAddressMutation>['input'] {
+  const customFields = fields
+    .flatMap((f) => (Array.isArray(f) ? f : [f]))
+    .filter((field) => Boolean(value[field.name]))
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/account/addresses/_actions/update-address.ts:115]: Difference between the add address function is that we filter out empty values here.

core/data-transformers/form-field-transformer/index.ts
```diff
@@ -91,3 +111,14 @@ export const formFieldTransformer = (
       return null;
   }
 };
+
+// TODO: See if we can merge this is with the above function
+// Will require testing and refactoring of register customer functionality
```
chanceaclark, line 116: Not going to do this in this PR but something we should look at.

core/vibes/soul/form/checkbox/index.tsx
chanceaclark, line 1: Already merged into VIBES

core/vibes/soul/primitives/dynamic-form/index.tsx
chanceaclark, line 1: The changes in the following files has a PR in VIBES already: https://github.com/makeswift/vibes/pull/434

core/data-transformers/form-field-transformer/index.ts
```diff
@@ -56,7 +62,18 @@ export const formFieldTransformer = (
...
     case 'PicklistFormField':
+      if (field.entityId === FieldNameToFieldId.countryCode) {
+        return {
+          id: String(field.entityId),
+          type: 'text',
+          name: String(field.entityId),
+          label: field.label,
+          required: field.isRequired,
+        };
+      }
```
chanceaclark, line 73: In a follow up PR, I think we need to use our `Geography` node to populate a dropdown here instead of a text field. The API requires a country code (`US`), but autofill will copy over the entire string (`United States`).

migueloller (approved):
Looks good!

Something I noticed from the testing videos is that we have no pending indicator while the addresses are being changed. This is more of a VIBES concern, but I think that we should have a pending indicator even if we have optimistic updates. Something we had discussed doing before was having a spinner next to the addresses title while there was pending state.

core/data-transformers/form-field-transformer/index.ts
chanceaclark: Added in https://github.com/bigcommerce/catalyst/pull/1878

PR #1880, opened 2025-01-09
fix: use lazy promise on getCartCount to defer cookies() execution

fikrikarim:
## What/Why?
This fixes 500 error:
```
needs to bail out of prerendering at this point because it used cookies()
```

Without the lazy promise, the `getCartCount()` is immediately executed, and calls `cookies()` on the Header.

With lazy promise, the `getCartCount()` execution is deferred, until `await`/`then` is used.

## Testing
This bug was only reproducible if we reverted the [temporary fix commit](https://github.com/bigcommerce/catalyst/pull/1856) on the `soul/makeswift/integration` branch. See this [deployment](https://vercel.com/bigcommerce-platform/catalyst-soul/BmZk2nfaemBus1EQzeJP5JrCGcH2).

Makeswift pages are PPRed correctly instead of being server-rendered
```
@bigcommerce/catalyst-core:build: ├ ◐ /[locale]                                  1.8 kB          441 kB
@bigcommerce/catalyst-core:build: ├   ├ /[locale]
@bigcommerce/catalyst-core:build: ├   └ /en
@bigcommerce/catalyst-core:build: ├ ◐ /[locale]/[...rest]                        1.8 kB          441 kB
@bigcommerce/catalyst-core:build: ├   ├ /[locale]/[...rest]
@bigcommerce/catalyst-core:build: ├   ├ /en/[...rest]
@bigcommerce/catalyst-core:build: ├   ├ /en/blog
@bigcommerce/catalyst-core:build: ├   └ [+5 more paths]
```

No 500 error [logs](https://vercel.com/bigcommerce-platform/catalyst-soul/BmZk2nfaemBus1EQzeJP5JrCGcH2/logs?page=2&levels=):
image

### DIFF
--- a/core/components/header/index.tsx
+++ b/core/components/header/index.tsx
@@ -1,5 +1,6 @@
 import { cookies } from 'next/headers';
 import { getTranslations } from 'next-intl/server';
+import PLazy from 'p-lazy';
 import { cache } from 'react';
 
 import { HeaderSection } from '@/vibes/soul/sections/header-section';
@@ -111,7 +112,7 @@ export const Header = async () => {
         mobileMenuTriggerLabel: t('toggleNavigation'),
         openSearchPopupLabel: t('Search.openSearchPopup'),
         logoLabel: t('home'),
-        cartCount: getCartCount(),
+        cartCount: PLazy.from(getCartCount),
       }}
     />
   );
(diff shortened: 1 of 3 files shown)


migueloller (approved):
Thanks!

Looking at this makes me wonder if we should handle this in VIBES at some point, where we take the approach of accepting functions and we turn them into lazy promises 🤔

PR #1884, opened 2025-01-09
chore: use setRequestLocale only where needed

jorgemoya:
## What/Why?
While working on adding locale dropdown, I noticed we are inconsistent with the use of `setRequestLocale`.

- No needed in metadata.
- Removed from pages that are no longer static (we need to fix this in different PR)
- Added to pages that are static or PPR.

### DIFF
--- a/.changeset/popular-ducks-bake.md
+++ b/.changeset/popular-ducks-bake.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Use `setRequestLocale` only where needed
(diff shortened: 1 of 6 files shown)


core/app/[locale]/not-found.tsx
```diff
@@ -41,8 +42,8 @@ async function getFeaturedProducts(): Promise<CarouselProduct[]> {
   return productCardTransformer(featuredProducts, format);
 }
 
-export default async function NotFound() {
-  const t = await getTranslations('NotFound');
+export default function NotFound() {
+  const t = useTranslations('NotFound');
```
Comment (chanceaclark) [inline core/app/[locale]/not-found.tsx:46]: Doesn't this still have to be async as this page is not a client component?

Comment (jorgemoya) [reply core/app/[locale]/not-found.tsx:46]: Yeah, I think you're right.

Comment (jorgemoya) [reply core/app/[locale]/not-found.tsx:46]: I mean, it works fine this way, `useTranslations` can be used in shared components, but I will revert.

chanceaclark (approved):

migueloller (approved):

PR #1885, opened 2025-01-09
feat: lazyload slideshow images

joshuawootonn:
## What/Why?
Embla carousels in the soul vibe aren't lazy loading images. This PR updates them to lazy load images other than the primary one. 

> [!NOTE]  
> I don't lazy load the `carousel.tsx` component which is used in the product carousel because it renders the images to the edge of the screen. The built in lazy loading of images by Next.js takes care of making sure these don't load until you rotate the carousel and they are in view. 

## Testing
Testing this was difficult because the Vercel toolbar slows down the preview deployment load time. For my testing I ran against a local production build. 

### Commit 1 diff
CleanShot 2025-01-09 at 15 36 46@2x

### Commit 2 diff
CleanShot 2025-01-10 at 09 45 47@2x
CleanShot 2025-01-10 at 09 47 21@2x

### DIFF
--- a/core/vibes/soul/sections/product-detail/product-gallery.tsx
+++ b/core/vibes/soul/sections/product-detail/product-gallery.tsx
@@ -1,8 +1,9 @@
 'use client';
 
 import { clsx } from 'clsx';
+import { EmblaCarouselType } from 'embla-carousel';
 import useEmblaCarousel from 'embla-carousel-react';
-import { useEffect, useState } from 'react';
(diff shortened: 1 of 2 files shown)


migueloller (approved):
This looks good to me! 👍🏻 

I find it a bit weird that we have to do this given all the effort `next/image` goes through to lazily load images, though 😞 

I feel like Embla Carousel should work better with the browser's built-in `loading` attribute, which is what `next/image` uses.

Don't forget about the companion PR in VIBES!

PR #1886, opened 2025-01-10
feat: use generateStaticParams to pregenerate some routes

jorgemoya:


core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx
```diff
@@ -249,3 +251,14 @@ export default async function Brand(props: Props) {
     />
   );
 }
+
+export async function generateStaticParams() {
+  const brands = await getBrands();
+
+  return routing.locales.map((locale) => {
+    return removeEdgesAndNodes(brands).map((brand) => ({
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx:259]: This is something I have been thinking about recently when migrating the account pages. I think we should probably have the pattern of removing the edges and nodes closer to the boundary, in this case inside `getBrands`. This way consumers don't need to worry about it and can focus on just massaging a normalized payload into what is needed for the page.

jorgemoya:
Closing PR because it is not providing value as expected.

PR #1887, opened 2025-01-10
Add Web Pages from vibes

jordanarldt:
## What/Why?
Add Web Page component from vibes. This is only the first iteration.

### Next steps:
- Add support for nested web pages with children, and a way to navigate between them. This will bring full feature parity with BigCommerce.

## Testing

image

### DIFF
--- a/core/app/[locale]/(default)/webpages/normal/[id]/page.tsx
+++ b/core/app/[locale]/(default)/webpages/normal/[id]/page.tsx
@@ -1,26 +1,53 @@
 import type { Metadata } from 'next';
 import { notFound } from 'next/navigation';
 
+import { Breadcrumb } from '@/vibes/soul/primitives/breadcrumbs';
+
+import { WebPageContent, WebPage as WebPageData } from '../../_components/web-page';
+
 import { getWebpageData } from './page-data';
 
 interface Props {
   params: Promise<{ id: string }>;
 }
 
-export async function generateMetadata({ params }: Props): Promise<Metadata> {
-  const { id } = await params;
-
+async function getWebPage(id: string): Promise<WebPageData> {
   const data = await getWebpageData({ id: decodeURIComponent(id) });
   const webpage = data.node?.__typename === 'NormalPage' ? data.node : null;
 
   if (!webpage) {
-    return {};
+    return notFound();
   }
 
+  return {
+    title: webpage.name,
+    content: webpage.htmlBody,
+    seo: webpage.seo,
+  };
+}
+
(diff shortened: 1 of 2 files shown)


core/app/[locale]/(default)/webpages/normal/[id]/page.tsx
```diff
@@ -1,26 +1,52 @@
...
 import { getWebpageData } from './page-data';
 
 interface Props {
   params: Promise<{ id: string }>;
 }
 
-export async function generateMetadata({ params }: Props): Promise<Metadata> {
-  const { id } = await params;
-
+async function getWebPage(id: string): Promise<WebPageData> {
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/webpages/normal/[id]/page.tsx:13]: We need to wrap this in `cache` to memoize the response.

Comment (jordanarldt) [reply core/app/[locale]/(default)/webpages/normal/[id]/page.tsx:13]: @chanceaclark The `getWebpageData({ id: decodeURIComponent(id) })` function is wrapped in `cache` in the `page-data.ts` file 👍

Comment (chanceaclark) [reply core/app/[locale]/(default)/webpages/normal/[id]/page.tsx:13]: Why do we have two fns to fetch the data then. Can we make it just one function to fetch the data? I'm not sure we should be calling `notFound()` within `generateMetadata`.

Comment (jordanarldt) [reply core/app/[locale]/(default)/webpages/normal/[id]/page.tsx:13]: @chanceaclark I noticed this is a similar pattern we follow in most other pages/components, the `getWebPage` function is responsible for making the GQL call and then converting it into an object ready to be passed into the vibes component as the promise. 

I can update it to not return `notFound` though 👍  My line of thinking there was that if the webpage isn't found, then returning `notFound` should trump everything else. If the page isn't found, why would we want to do anything with the metadata?

Comment (jorgemoya) [reply core/app/[locale]/(default)/webpages/normal/[id]/page.tsx:13]: I believe this is the pattern. @chanceaclark

core/vibes/soul/sections/web-page-content/index.tsx
```diff
@@ -0,0 +1,84 @@
+import { Stream, Streamable } from '@/vibes/soul/lib/streamable';
```
jorgemoya, line 1: I don't think this should be a vibes component or section, since webpages are only pertinent to Catalyst, but maybe someone disagrees!

jordanarldt: @jorgemoya I agree, I closed the vibes PR 👍

jorgemoya (approved):

PR #1899, opened 2025-01-11
Add Contact Us page, refactoring

jordanarldt:
## What/Why?
- Add **Contact Us** page parity to Catalyst.
- Refactor `SignUpForm` to `DynamicFormSection`
- Fix a layout issue on pages

## Testing

https://github.com/user-attachments/assets/a4088897-30e9-40b3-a075-395da54becd2


core/app/[locale]/(default)/webpages/_components/web-page.tsx
```diff
@@ -16,35 +16,40 @@ interface Props {
...
-          }}
-        </Stream>
-      </div>
+              <div
+                className="prose mx-auto w-full max-w-4xl space-y-4 [&_h2]:font-heading [&_h2]:text-3xl [&_h2]:font-normal [&_h2]:leading-none [&_h2]:@xl:text-4xl [&_img]:mx-auto [&_img]:max-h-[600px] [&_img]:w-fit [&_img]:rounded-2xl [&_img]:object-cover"
+                dangerouslySetInnerHTML={{ __html: content }}
+              />
+              {children}
+            </>
+          );
```
Comment (migueloller) [inline core/app/[locale]/(default)/webpages/_components/web-page.tsx:50]: All these styles should be handled in VIBES so that we can swap out different VIBES and things won't look visually off.

core/app/[locale]/(default)/webpages/contact/[id]/page.tsx
```diff
@@ -1,54 +1,189 @@
...
+    return (
+      <WebPageContent
+        breadcrumbs={getWebPageBreadcrumbs(id)}
+        webPage={getWebPageWithSuccessContent(id, t('success'))}
+      >
+        <div className="mx-auto w-full max-w-4xl">
+          <ButtonLink className="mt-8" href="/" size="large" type="submit" variant="primary">
+            {t('successCta')}
+          </ButtonLink>
+        </div>
```
Comment (migueloller) [inline core/app/[locale]/(default)/webpages/contact/[id]/page.tsx:175]: We should avoid defining CSS like this in Catalyst, otherwise we won't we able to swap out VIBES while keeping things visually consistent

core/app/[locale]/(default)/webpages/_components/web-page.tsx
Comment (jordanarldt) [reply core/app/[locale]/(default)/webpages/_components/web-page.tsx]: @migueloller Can you clarify this a bit? We decided that the base web page component shouldn't be in VIBES, so what's the best way to ensure the styles in Catalyst are sync'd with the vibes styles?

core/vibes/soul/sections/dynamic-form-section/index.tsx
```diff
@@ -0,0 +1,39 @@
+import { clsx } from 'clsx';
```
jorgemoya, line 1: Will this be added to VIBES?

core/vibes/soul/sections/sign-up-section/index.tsx
```diff
@@ -1,27 +0,0 @@
-import { DynamicForm, DynamicFormAction } from '@/vibes/soul/primitives/dynamic-form';
```
jorgemoya, line 1: I'm not sure if we should remove this yet.

core/app/[locale]/(default)/webpages/_components/web-page.tsx
Comment (jordanarldt) [reply core/app/[locale]/(default)/webpages/_components/web-page.tsx]: @migueloller For some context, I grabbed that styling directly from the blog post content VIBES component here: https://github.com/makeswift/vibes/blob/85e9a51900472a0b5796731b9437781ef264b0b4/apps/web/vibes/soul/sections/blog-post-content/index.tsx#L92

Comment (jorgemoya) [reply core/app/[locale]/(default)/webpages/_components/web-page.tsx]: I was under the impression that webpage is not a section that will live in VIBES, so I asked Jordan to just style this page as close to the vibe as possible (maybe using `SectionLayout`).

However, I like the concept Jordan introduced of `DynamicFormSection` that has a title and subtitle and the form. This is something that should live in VIBES I think, since there are many use cases were we just need a title and form. 

What do you think @migueloller ?

Comment (jordanarldt) [reply core/app/[locale]/(default)/webpages/_components/web-page.tsx]: @jorgemoya I did actually add `DynamicFormSection` into VIBES: https://github.com/makeswift/vibes/pull/468

However, I ended up not using that section for Contact Page specifically, because of styling issues. For the contact page, we don't need the whole section, just the form, so it made more sense to just go with `DynamicForm` directly.

core/vibes/soul/sections/dynamic-form-section/index.tsx
jordanarldt: Yeah, it's already in vibes:
https://github.com/makeswift/vibes/pull/468/files#diff-03ae676bb50aba31a007bb4d8af6379c8e033b92b67d2957d72ed0b9b3aead83R1

core/vibes/soul/sections/sign-up-section/index.tsx
jordanarldt: @jorgemoya Since the sign up page is so simple, we're able to replace the page to use the DynamicFormSection, so we don't need these imports.

jorgemoya: Gotcha. Sorry was out of the loop!

core/app/[locale]/(default)/webpages/contact/[id]/_actions/submit-contact-form.ts
```diff
@@ -0,0 +1,119 @@
+'use server';
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/webpages/contact/[id]/_actions/submit-contact-form.ts:1]: 🍹 Any reason we moved this action outside of `/contact-us`?

core/app/[locale]/(default)/webpages/contact/[id]/page.tsx
```diff
@@ -1,54 +1,195 @@
...
+        <DynamicForm
+          action={submitContactForm}
+          fields={await getContactFields(id)}
+          submitLabel={t('submitFormText')}
+        />
       </div>
-
-      <ContactUs node={webpage} reCaptchaSettings={recaptchaSettings} />
-    </>
+    </WebPageContent>
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/webpages/contact/[id]/page.tsx:193]: Why not use `DynamicFormSection` here directly?

core/app/[locale]/(default)/webpages/contact/[id]/_actions/submit-contact-form.ts
Comment (jordanarldt) [reply core/app/[locale]/(default)/webpages/contact/[id]/_actions/submit-contact-form.ts:1]: @jorgemoya Just following the same pattern we use for the `register` route. It felt a bit redundant / unnecessary to have a `/contact-us` directory when we're already in the `/contact/[id]` directory. So after cleaning up the old implementation, this would have been the only file in there (`/contact-us/_actions/submit-contact-form.ts`, so I figured it made more sense to just move `_actions` under the `/contact/[id]` route 👍

core/app/[locale]/(default)/webpages/contact/[id]/page.tsx
Comment (jordanarldt) [reply core/app/[locale]/(default)/webpages/contact/[id]/page.tsx:193]: @jorgemoya because the `<SectionLayout>` of the dynamic form section messes with the padding. Since this is a child of the `WebPageContent` component, that would mean we have a section within a section, and by default, `SectionLayout` adds additional x-padding, so when using that it messes up the layout. Also, in this particular case, we don't need the whole section - just the form.

Here's an example of how it looks when we do use the `DynamicFormSection`:
image

As you can see, the `SectionLayout` within a `SectionLayout` messes it up, and the `SectionLayout` component offers no way to override the inner `div` that adds this padding.

Comment (jordanarldt) [reply core/app/[locale]/(default)/webpages/contact/[id]/page.tsx:193]: @jorgemoya Using the `DynamicForm` by itself allows us to piggy-back off the web page section layout:
image

Comment (jorgemoya) [reply core/app/[locale]/(default)/webpages/contact/[id]/page.tsx:193]: Ah, this is because the `We're happy to answer...` is the html body from the webpage response?

Comment (jordanarldt) [reply core/app/[locale]/(default)/webpages/contact/[id]/page.tsx:193]: > Ah, this is because the `We're happy to answer...` is the html body from the webpage response?

@jorgemoya Yes, exactly

jorgemoya (approved):
This looks good to me but if @migueloller feels this `WebPageContent` component needs to be in VIBES, lets work to add it as well.

core/app/[locale]/(default)/webpages/_components/web-page.tsx
Comment (migueloller) [reply core/app/[locale]/(default)/webpages/_components/web-page.tsx]: Sorry for chiming in late here. Basically, the idea is that we want to be able to "hot-swap" VIBES when installing Catalyst in the future. 

For this to work, we'll need to make sure Catalyst is using a curated interface that all VIBES will follow. 

If we write custom CSS, while that might look good for the Soul VIBE it might not for another VIBE. 

So, we'd need to identify what is that "shared" aspect that makes sense to move to VIBES. Perhaps it's a container/layout component, or something else. But by having it be part of the standard interface it would make it so VIBES as "hot-swappable". 

Does that make sense?

Comment (jorgemoya) [reply core/app/[locale]/(default)/webpages/_components/web-page.tsx]: In this case, which is contrary to what I suggested Jordan so it's my bad, we need to build a VIBES component that has a `title` and `subtitle` and renders `children`.

PR #1900, opened 2025-01-13
fix: element tree introspection on Makeswift API routes

agurtovoy:
## What/Why?

Makeswift API route needs a components import, otherwise it won't know about any of the custom components and won't be able to introspect them/replace the resources. See [this thread](https://makeswifthq.slack.com/archives/C014SCS4G75/p1736555455966869) for more context.

### Before

The typography resource inside of a `section-layout` component is not replaced:
before

### After

The typography resource is correctly replaced:
after


fikrikarim (approved):
Nice

PR #1901, opened 2025-01-13
fix: use lazy promise for searchParams to defers execution

fikrikarim:
## What/Why?
Use lazy promise on `searchParams` to defers execution.

This is needed to fix the 500 error:

```
Route /[locale]/search needs to bail out of prerendering at this point because it used `await searchParams`, `searchParams.then`, or similar.
```

## Testing
We had the same problem with `cookies()` on the `Header`: https://github.com/bigcommerce/catalyst/pull/1880

### DIFF
--- a/core/app/[locale]/(default)/(faceted)/search/page.tsx
+++ b/core/app/[locale]/(default)/(faceted)/search/page.tsx
@@ -1,5 +1,6 @@
 import { getFormatter, getTranslations } from 'next-intl/server';
 import { createSearchParamsCache } from 'nuqs/server';
+import PLazy from 'p-lazy';
 import { cache } from 'react';
 
 import { Breadcrumb } from '@/vibes/soul/primitives/breadcrumbs';
@@ -244,7 +245,10 @@ export async function generateMetadata() {
   };
 }
 
-export default async function Search(props: Props) {
+export default async function Search(pageProps: Props) {
+  const searchParams = PLazy.from(() => pageProps.searchParams);
+  const props = { searchParams };
+
   return (
     <ProductsListSection
       breadcrumbs={getBreadcrumbs()}


bookernath (approved):

core/app/[locale]/(default)/(faceted)/search/page.tsx
```diff
@@ -244,7 +245,10 @@ export async function generateMetadata() {
   };
 }
 
-export default async function Search(props: Props) {
+export default async function Search(pageProps: Props) {
+  const searchParams = PLazy.from(() => pageProps.searchParams);
+  const props = { searchParams };
```
Comment (migueloller) [inline core/app/[locale]/(default)/(faceted)/search/page.tsx:250]: If we're seeing problems with `searchParams` similar to `cookies` I think that we might need a different approach here...

Having to use a lazy promise here feels wrong to me. In particular, `Search` should be dynamic always.

Comment (jorgemoya) [reply core/app/[locale]/(default)/(faceted)/search/page.tsx:250]: I agree with @migueloller on this. Search is being rendered dynamically though.

jorgemoya:
Is this only happening when running the makeswift integration?

migueloller:
> Is this only happening when running the makeswift integration?

Yeah. It's a race condition that happens and it's really a problem with the streamable pattern and the way React implements "postpone". 

We might need to do this lazy evaluation but that might need to happen at the VIBES API layer so Catalyst isn't doing this. Or we might need to figure out a different approach. 

I should probably set up a repro in the suspense-patterns repo so we can explore ideas

core/app/[locale]/(default)/(faceted)/search/page.tsx
Comment (fikrikarim) [reply core/app/[locale]/(default)/(faceted)/search/page.tsx:250]: >Search should be dynamic always.
Wait, with this change, search is still dynamic. This change just defers the `searchParams` execution to component that await the function.

But I agree with the general sentiment though. This is a landmine. I don't want developers to think about this quirk while developing Catalyst. I'll close

fikrikarim:
> Is this only happening when running the makeswift integration?

No. This happens on the canary branch too. 

image (21)

But this happens more often on the integrations branch.

PR #1907, opened 2025-01-14
feat(core): add locale switcher to Header

jorgemoya:
## What/Why?
Add locale switcher to the Header, using server action and `redirect`. In this use case we need to revalidate the `/` to make sure the cookies get updated with the new locale. We also had to workaround an issue by redirecting to the `/en` when the default locale is used.

## Testing
Locale switcher works as expected.

https://github.com/user-attachments/assets/b78fcca2-1fc5-4608-b2e9-517797fcafab

### DIFF
--- a/core/components/header/_actions/switch-locale.ts
+++ b/core/components/header/_actions/switch-locale.ts
@@ -1,0 +1,34 @@
+'use server';
+
+import { SubmissionResult } from '@conform-to/react';
+import { parseWithZod } from '@conform-to/zod';
+import { revalidatePath } from 'next/cache';
+
+import { localeSchema } from '@/vibes/soul/primitives/navigation/schema';
+import { defaultLocale, redirect } from '~/i18n/routing';
+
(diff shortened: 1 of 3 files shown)


bookernath (approved):
Tested this, LGTM ✅

core/components/header/_actions/switch-locale.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  if (submission.status !== 'success') {
+    return submission.reply({ formErrors: [t('invalidLocale')] });
+  }
+
+  await Promise.resolve();
+
+  revalidatePath('/');
+
+  if (submission.value.id === defaultLocale) {
```
chanceaclark, line 24: I would just add a comment why we need to do this.
```suggestion

  // Since `redirect` doesn't prepend the local to the redirect url
  // when navigating the a default locale link, we need to prepend
  // it ourselves to ensure the redirect happens.
  if (submission.value.id === defaultLocale) {
```

chanceaclark (approved):

PR #1912, opened 2025-01-16
feat(core): Add currency selector to header

bookernath:
## What/Why?

- Adds currency selector following the same pattern used for locale selector
- Establishes new `currencyCode` httpOnly cookie
- New lib functions to customize/extend currency selection
- Moved cart to a lib as well while I was at it
- Decorates all queries that touch pricing to include currency code
  - ... except Cart and Orders, as currency lives on those models directly

Only transactional currencies are supported for now, because we don't have a way to render prices for the cart for a display currency from the API. Theoretically we could use the exchange rate we have access to in the API to calculate the prices within next.js, but my preference would be for all monetary values, even estimates, to come from the API. So for now, I have filtered the list of currencies to only the transactional ones.

## Todo

- [x] Handle default currency properly once exposed in GQL API
- [x] Run the mutation to change the currency of an active cart if the currency is changed (handle display vs transactional currencies properly)
- [x] Update VIBES https://github.com/makeswift/vibes/pull/472
- [x] ~Consider adding a middleware to set this cookie with a query param such as ?setCurrency=USD for the benefit of deep linking and crawlers~
   - Can be a future PR
- [x] Add translations

## Testing
Preview deployment:

https://github.com/user-attachments/assets/5bb057dd-f2d2-4303-a18e-a4539aa551fa

### DIFF
--- a/core/components/header/_actions/switch-currency.ts
+++ b/core/components/header/_actions/switch-currency.ts
@@ -1,18 +1,52 @@
 'use server';
 
+import { BigCommerceGQLError } from '@bigcommerce/catalyst-client';
 import { SubmissionResult } from '@conform-to/react';
 import { parseWithZod } from '@conform-to/zod';
-import { revalidatePath } from 'next/cache';
+import { revalidatePath, revalidateTag } from 'next/cache';
 import { getTranslations } from 'next-intl/server';
 import { z } from 'zod';
 
+import { client } from '~/client';
+import { graphql } from '~/client/graphql';
+import { TAGS } from '~/client/tags';
+import { getCartId, setCartId } from '~/lib/cart';
 import { setPreferredCurrencyCode } from '~/lib/currency';
 
+import { CurrencyCode } from '../fragment';
 import { CurrencyCodeSchema } from '../schema';
 
 const currencySwitchSchema = z.object({
   id: CurrencyCodeSchema,
 });
+
+// Note: this results in a new cart being created in the new currency, so the cart ID will change
+const UpdateCartCurrencyMutation = graphql(`
+  mutation UpdateCartCurrency($input: UpdateCartCurrencyInput!) {
+    cart {
+      updateCartCurrency(input: $input) {
+        cart {
+          currencyCode
+          entityId
+        }
+      }
+    }
+  }
+`);
+
+export const updateCartCurrency = async (cartId: string, currencyCode: CurrencyCode) => {
+  const result = await client.fetch({
+    document: UpdateCartCurrencyMutation,
+    variables: { input: { data: { currencyCode }, cartEntityId: cartId } },
+  });
+  const newCartId = result.data.cart.updateCartCurrency?.cart?.entityId;
+
+  if (newCartId) {
+    await setCartId(newCartId);
+  } else {
+    throw new Error('Failed to update cart currency', { cause: result });
+  }
+};
 
 export const switchCurrency = async (_prevState: SubmissionResult | null, payload: FormData) => {
   const t = await getTranslations('Components.Header.Currency');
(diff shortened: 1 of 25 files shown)


core/components/header/index.tsx
```diff
@@ -80,12 +83,7 @@ const getCartCount = async () => {
     document: GetCartCountQuery,
     variables: { cartId },
     customerAccessToken,
-    fetchOptions: {
-      cache: 'no-store',
-      next: {
-        tags: [TAGS.cart],
```
bookernath, line 86: No point tagging something that's `no-store`, right?

core/components/header/schema.ts
```diff
@@ -0,0 +1,12 @@
+import { z } from 'zod';
+
+import type { CurrencyCode } from './fragment';
+
+// eslint-disable-next-line @typescript-eslint/consistent-type-assertions
```
bookernath, line 5: Feedback welcome, I tried many approaches to avoid a type assertion but I think it's actually safer than alternatives

core/components/header/schema.ts
```diff
@@ -0,0 +1,12 @@
...
+export const currencyCodeSchema = z
+  .string()
+  .length(3)
+  .toUpperCase()
+  .refine((val) => /^[A-Z]{3}$/.test(val), {
```
chanceaclark, line 10: You can use a type guard instead of casting:
```suggestion
  .refine((val): val is CurrencyCode => /^[A-Z]{3}$/.test(val), {
```

core/components/header/schema.ts
```diff
@@ -0,0 +1,12 @@
+import { z } from 'zod';
+
+import type { CurrencyCode } from './fragment';
+
+// eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+export const currencyCodeSchema = z
```
chanceaclark, line 6: :nit: CapitalCase for zod schemas
```suggestion
export const CurrencyCodeSchema = z
```

core/vibes/soul/primitives/navigation/index.tsx
```diff
@@ -92,6 +96,9 @@ interface Props<S extends SearchResult> {
   locales?: Locale[];
   activeLocaleId?: string;
   localeAction?: LocaleAction;
+  currencies?: Currency[];
+  activeCurrencyCode?: string;
```
chanceaclark, line 100: I think we should follow the same pattern as the locale above and use the nomenclature of `activeCurrencyId`. The interface for `Currency` would be this instead:
```ts
interface Currency {
  id: string; // The currency code
  label: string // Separates the display value from the identifier. For now, could be the same value as `id`.
}
```

core/vibes/soul/primitives/navigation/index.tsx
chanceaclark, line 1: 🍹 Just make sure we have a corresponding PR in VIBES

core/components/header/index.tsx
```diff
@@ -95,15 +92,34 @@ const getCartCount = async () => {
...
   const locale = await getLocale();
+  const currencyCode = await getPreferredCurrencyCode();
 
   const locales = routing.locales.map((enabledLocales) => ({
     id: enabledLocales,
     label: enabledLocales.toLocaleUpperCase(),
   }));
 
+  const currencies = await getCurrencies();
+  // todo handle default currency properly once added to API
```
chanceaclark, line 120: 🍹 Easier for searchability and works ootb with the `TODO Highlight` plugin
```suggestion
  // TODO: handle default currency properly once added to API
```
Screenshot 2025-01-16 at 10 38 19

chanceaclark:
When it's ready to merge, don't forget a changelog 🙏

core/components/header/index.tsx
chanceaclark: It's still valid as `revalidateTag` will refetch that data without reloading the page.

core/vibes/soul/primitives/navigation/index.tsx
jorgemoya: Agreed!

core/components/header/index.tsx
bookernath: thanks for explaining, I'll revert this change

core/components/header/schema.ts
chanceaclark: You should be able to remove this since you are doing `val is CurrencyCode` below.

chanceaclark (approved):
LGTM pending the default currency stuff

jorgemoya (approved):

core/lib/cart.ts
chanceaclark, line 1: 🍹 Nothing actionable here, just noting that maybe we should also move `auth` into `/lib` too 🤔

core/components/header/_actions/switch-currency.ts
```diff
@@ -0,0 +1,79 @@
...
+  const newCartId = result.data.cart.updateCartCurrency?.cart?.entityId;
+
+  if (newCartId) {
+    await setCartId(newCartId);
+  } else {
+    throw new Error('Failed to update cart currency', { cause: result });
+  }
+};
+
+export const switchCurrency = async (_prevState: SubmissionResult | null, payload: FormData) => {
```
jorgemoya, line 50: This action needs some error handling like we now do in https://github.com/bigcommerce/catalyst/blob/canary/core/app/%5Blocale%5D/(default)/product/%5Bslug%5D/_actions/add-to-cart.tsx#L243. LMK if you have any questions.

PR #1914, opened 2025-01-16
feat(client): handle GQL errors

jorgemoya:
## What/Why?
GQL requests that respond as `200` but have an `errors` field will now be properly handled by the client and throw a proper `BigCommerceGQLError` response with the message reason from the API. This will provide a more detailed description of why the GQL request errored out.

API errors will still be handled and attribute the errored status as the message with this change as `BigCommerceAPIError`.

## Testing
Tested this and now our requests will actually `throw` now if they error from the GQL request and show a valid reason.

Screenshot 2025-01-16 at 3 35 03 PM

Screenshot 2025-01-16 at 3 34 59 PM


jorgemoya:
Testing changes for now, will convert to PR when ready.

chanceaclark (approved):

PR #1923, opened 2025-01-21
Add support for nested web pages

jordanarldt:
## What/Why?
- Add support for nested web page children
- Refactor `AccountLayout` and instead use `StickySidebarLayout` with `SidebarMenu` component
- Refactor web page routing structure to `/webpages/{pageId}/normal` or `/webpages/{pageId}/contact`. This allows us to have a web page layout file that can fetch the child links.
- Update footer to only add links for top-level web pages
- Add breadcrumb truncation capability on deeply nested pages

## Testing

https://github.com/user-attachments/assets/bfa07ea3-c2b6-46f0-9c40-1cf5aa6e9dbe


core/app/[locale]/(default)/webpages/[id]/contact/_actions/submit-contact-form.ts
```diff
@@ -96,18 +96,9 @@ export async function submitContactForm<F extends Field>(
     // eslint-disable-next-line no-console
     console.error(error);
 
-    if (error instanceof BigCommerceGQLError) {
+    if (error instanceof BigCommerceAPIError) {
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/webpages/[id]/contact/_actions/submit-contact-form.ts:99]: Is there no way to get a GQL error from this action?

core/components/footer/fragment.ts
```diff
@@ -26,7 +26,7 @@ export const FooterFragment = graphql(`
       }
     }
     content {
-      pages(filters: { isVisibleInNavigation: true }) {
+      pages(filters: { parentEntityIds: [0] }) {
```
jorgemoya, line 29: What does this change? Only fetch the root pages?

jordanarldt: @jorgemoya Yes, exactly. For the footer, we only want the top-level pages, and also we want to include web pages that don't have `isVisibleInNavigation: true` - this value is only pertinent to what is displayed in the nav bar, not the footer.

core/app/[locale]/(default)/webpages/[id]/contact/_actions/submit-contact-form.ts
Comment (jordanarldt) [reply core/app/[locale]/(default)/webpages/[id]/contact/_actions/submit-contact-form.ts:99]: @jorgemoya Ah, looks like this portion might have been a rebasing / conflict error on my part after your PR to update the GQL errors. I'll fix this

jorgemoya (approved):

PR #1927, opened 2025-01-23
fix(core): dedupe requests in various pages

jorgemoya:
## What/Why?

- Dedupe requests in various pages by properly caching/memoizing the function per page render.
- Fix an issue with quick search result.

## Testing
Tested and see less function calls that call `fetch`.

PR #1931, opened 2025-01-24
fix(core): dedupe requests in webpages

jorgemoya:
## What/Why?
Same as #1927, but for webpages pages.

## Testing
Requests is being properly memoized.


bookernath (approved):
LGTM, but I do have the question of whether memoization needs to include channel ID also or if that is already handled well upstream of the fetch

PR #1934, opened 2025-01-26
fix: use x-makeswift-draft-mode header for proxy

migueloller:
## What/Why?

The URL search param is only set on the very first request. For the proxy to work as intended, the `x-makeswift-draft-mode` header must be used to determine whether to proxy Draft Mode or not.

## Testing

https://github.com/user-attachments/assets/475ac733-4d7c-47d9-8e84-6db712d1ed58


bookernath (approved):
LGTM

PR #1944, opened 2025-01-30
fix(core): handle invalid credentials error message in Login

jorgemoya:
## What/Why?
Properly look for the type of error we get from next-auth and show the correct message when credentials are invalid.

### Before

Screenshot 2025-01-30 at 2 37 19 PM

### After
Screenshot 2025-01-30 at 3 12 15 PM

### DIFF
--- a/core/app/[locale]/(default)/(auth)/login/_actions/login.ts
+++ b/core/app/[locale]/(default)/(auth)/login/_actions/login.ts
@@ -3,6 +3,7 @@
 import { BigCommerceGQLError } from '@bigcommerce/catalyst-client';
 import { SubmissionResult } from '@conform-to/react';
 import { parseWithZod } from '@conform-to/zod';
+import { AuthError } from 'next-auth';
 import { getLocale, getTranslations } from 'next-intl/server';
 
 import { schema } from '@/vibes/soul/sections/sign-in-section/schema';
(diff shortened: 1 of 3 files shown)


core/app/[locale]/(default)/(auth)/login/_actions/login.ts
```diff
@@ -42,11 +43,17 @@ export const login = async (_lastResult: SubmissionResult | null, formData: Form
...
-    if (error instanceof Error) {
-      return submission.reply({ formErrors: [error.message] });
+    if (error instanceof AuthError) {
+      if (error.name === 'CallbackRouteError') {
+        if (error.cause && error.cause.err?.message.includes('Invalid credentials')) {
+          return submission.reply({ formErrors: [t('Form.invalidCredentials')] });
+        }
+      }
+
+      return submission.reply({ formErrors: [error.name] });
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/(auth)/login/_actions/login.ts:53]: Can we remove the nested conditionals?
```suggestion
    if (
      error instanceof AuthError &&
      error.name === 'CallbackRouteError' &&
      error.cause &&
      error.cause.err?.message.includes('Invalid credentials')
      ) {
          return submission.reply({ formErrors: [t('Form.invalidCredentials')] });
      }

      return submission.reply({ formErrors: [error.name] });
```

chanceaclark (approved):

core/app/[locale]/(default)/(auth)/login/_actions/login.ts
Comment (jorgemoya) [reply core/app/[locale]/(default)/(auth)/login/_actions/login.ts]: Fixed!

chanceaclark (approved):

PR #1945, opened 2025-01-30
feat(core): only make links and cart count streamed in in header

jorgemoya:
## What/Why?
Splits the navigation links and footer sections from the Layout request, to be able to fetch dynamically via PPR.

https://github.com/user-attachments/assets/9959eafd-20d1-4af5-8c6b-6d0d21654d97

## Testing
Shows skeletons on render, the rest of the header is a static component.


jorgemoya:
Closing for other solutions.

PR #1946, opened 2025-01-31
fix: disable ppr in auth pages to make them SSG

jorgemoya:
## What/Why?
Disables ppr by `export const experimental_ppr = false;` since there are no dynamic elements in these pages.

## Before

Screenshot 2025-02-04 at 2 56 33 PM

## After

Screenshot 2025-02-04 at 2 57 40 PM

## Testing
Locally, behavior works the same, but during build time the page is SSG.


jorgemoya:
Closing for other solutions.

PR #1958, opened 2025-02-05
fix: rename price label's prop to match its CSS variables' prefix

agurtovoy:
## What/Why?

Price label color changes had no effect because of the CSS variables mismatch.

### Before

before, shop all

before, PDP

### After

after

after

### DIFF
--- a/core/lib/makeswift/components/site-theme/components/index.ts
+++ b/core/lib/makeswift/components/site-theme/components/index.ts
@@ -5,7 +5,7 @@ import { carousel } from './carousel';
 import { footer } from './footer';
 import { header } from './header';
 import { logo } from './logo';
-import { priceLabel } from './price-label';
+import { price } from './price-label';
 import { productCard } from './product-card';
 import { section } from './section';
 import { slideshow } from './slide-show';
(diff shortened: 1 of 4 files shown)


core/lib/makeswift/components/site-theme/components/index.ts
```diff
@@ -18,7 +18,7 @@ export default {
   footer,
   header,
   logo,
-  priceLabel,
```
agurtovoy, line 21: The `<SiteTheme>` component auto-generates CSS variable names based on the corresponding prop names. The [CSS variable prefix for the `<PriceLabel>` component is `price`](https://github.com/bigcommerce/catalyst/blob/canary/core/vibes/soul/primitives/price-label/index.tsx#L30), not `price-label`:

CSS vars

migueloller (approved):

core/lib/makeswift/components/site-theme/to-css.ts
```diff
@@ -1,6 +1,5 @@
 /* eslint-disable @typescript-eslint/consistent-type-definitions */
 
-// eslint-disable-next-line @typescript-eslint/consistent-indexed-object-style
```
agurtovoy, line 3: Remove obsolete ESLint error suppression.

PR #1959, opened 2025-02-06
fix: configure `NEXT_LOCALE` cookie to work in the builder

agurtovoy:
## What/Why?

Override `next-intl`'s default cookie options to ensure that the `NEXT_LOCALE` cookie is set when a Catalyst site is loaded in the Makeswift Builder. See [this document](https://www.notion.so/makeswift/iframe-cookies-1886f737ce30809595a8dd7bffbf79a6) for additional context.

### Before

before, network

before, cookies

### After

after, network

after, cookies

### DIFF
--- a/.changeset/spotty-seahorses-type.md
+++ b/.changeset/spotty-seahorses-type.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+fix: configure `NEXT_LOCALE` cookie to work inside of the Makeswift Builder's canvas
--- a/core/i18n/routing.ts
+++ b/core/i18n/routing.ts
@@ -56,6 +56,12 @@ export const routing = defineRouting({
   locales,
   defaultLocale,
   localePrefix,
+  // configure `NEXT_LOCALE` cookie to work inside of the Makeswift Builder's canvas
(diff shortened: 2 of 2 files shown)


fikrikarim (approved):
Nice. I didn't know they provide a config for the locale cookie.

This is not related to this PR, but for Pages Router, to be able to edit localized homepage, we needed to disable the [Automatic Locale Detection](https://nextjs.org/docs/pages/building-your-application/routing/internationalization#automatic-locale-detection), because otherwise you'll be redirected to a single locale, even though you're trying to edit the other locales.

Catalyst also had this issue the last time I tried it. Again, not related for this PR, but something to keep in mind for the full localization feature.

migueloller (approved):
Thanks @agurtovoy 🙏🏻 

Something I noticed from your screenshots is that there seems to be cookies being set by [Auth.js](https://authjs.dev/) that are also not sticking. Perhaps we can fix those as well?

agurtovoy:
> Something I noticed from your screenshots is that there seems to be cookies being set by [Auth.js](https://authjs.dev/) that are also not sticking. Perhaps we can fix those as well?

@migueloller yes, a separate PR for those is coming.

PR #1966, opened 2025-02-07
fix: configure NextAuth cookies to work inside the builder

agurtovoy:
## What/Why?

Configure NextAuth cookies to ensure they are set when a Catalyst site is loaded in the Makeswift Builder. See [this document](https://www.notion.so/makeswift/iframe-cookies-1886f737ce30809595a8dd7bffbf79a6) for additional context.

Related: https://github.com/bigcommerce/catalyst/pull/1959

### Before

before, network

before, orders

### After

after

### DIFF
--- a/core/auth/index.ts
+++ b/core/auth/index.ts
@@ -144,6 +144,16 @@ async function authorize(credentials: unknown): Promise<User | null> {
   }
 }
 
+const partitionedCookie = (name?: string) =>
+  ({
+    ...(name !== undefined ? { name } : {}),
+    options: {
+      partitioned: true,
+      secure: true,
+      sameSite: 'none',
+    },
+  }) as const;
+
 const config = {
   session: {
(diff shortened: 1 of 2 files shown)


core/auth/index.ts
```diff
@@ -144,6 +144,16 @@ async function authorize(credentials: unknown): Promise<User | null> {
   }
 }
 
+const partitionedCookie = (name?: string) =>
+  ({
+    ...(name !== undefined ? { name } : {}),
```
fikrikarim, line 149: I assume this will be used on other PR?

fikrikarim (approved):

core/auth/index.ts
agurtovoy: @fikrikarim Great question, no, this is just to make life easier for whomever inherits this code from us.

migueloller (approved):

PR #1970, opened 2025-02-10
feat: customer-group-slots component

drewpledger:
This component can map groups to slots in Makeswift

https://github.com/user-attachments/assets/e32ab87b-7939-4f55-a5c6-4e4d652ccd81

### DIFF
--- a/core/app/api/customer/group/route.ts
+++ b/core/app/api/customer/group/route.ts
@@ -1,0 +1,9 @@
+import { NextResponse } from 'next/server';
+
+import { getCustomerGroupId } from '~/client/queries/get-customer';
+
+export async function GET(): Promise<NextResponse> {
+  const { data } = await getCustomerGroupId();
+
+  return NextResponse.json(data);
+}
--- a/core/app/api/customer/groups/route.ts
+++ b/core/app/api/customer/groups/route.ts
@@ -1,0 +1,27 @@
+import { NextResponse } from 'next/server';
+
+export async function GET(): Promise<NextResponse> {
+  if (!process.env.BIGCOMMERCE_ACCESS_TOKEN) {
+    throw new Error('BIGCOMMERCE_ACCESS_TOKEN is not set');
+  }
+
+  const response = await fetch(
+    `https://api.bigcommerce.com/stores/${process.env.BIGCOMMERCE_STORE_HASH}/v2/customer_groups`,
+    {
+      method: 'GET',
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json',
+        'X-Auth-Token': process.env.BIGCOMMERCE_ACCESS_TOKEN,
+      },
+    },
+  );
+
+  if (!response.ok) {
+    throw new Error(`Failed to fetch customer groups: ${response.statusText}`);
+  }
+
+  const jsonResponse: unknown = await response.json();
+
+  return NextResponse.json(jsonResponse);
+}
--- a/core/client/queries/get-customer.ts
+++ b/core/client/queries/get-customer.ts
@@ -1,0 +1,28 @@
+import { getSessionCustomerAccessToken } from '~/auth';
+import { client } from '~/client';
+import { graphql, ResultOf } from '~/client/graphql';
+
+export const GetCustomerGroup = graphql(`
+  query CustomerGroup {
+    customer {
+      customerGroupId
+    }
+  }
+`);
+
+export type GetCustomerGroupResponse = ResultOf<typeof GetCustomerGroup>;
+
+export async function getCustomerGroupId() {
+  const customerAccessToken = await getSessionCustomerAccessToken();
+
+  const { data, errors } = await client.fetch({
+    document: GetCustomerGroup,
+    customerAccessToken,
+  });
+
+  if (errors) {
+    throw new Error(errors.map((error) => error.message).join(', '));
+  }
+
+  return { data };
+}
--- a/core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
+++ b/core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
@@ -1,0 +1,189 @@
+'use client';
+
+import { Combobox, Group, List, Slot, Style } from '@makeswift/runtime/controls';
+import { useIsInBuilder } from '@makeswift/runtime/react';
+import { clsx } from 'clsx';
+import { ReactNode } from 'react';
+import useSWR from 'swr';
+
+import { GetCustomerGroupResponse } from '~/client/queries/get-customer';
+import { runtime } from '~/lib/makeswift/runtime';
+
+import { CustomerGroupSchema, CustomerGroupsSchema, CustomerGroupsType } from './schema';
+
+const NO_GROUP_ID = 'no-group';
+
+async function getAllCustomerGroups(): Promise<CustomerGroupsType | null> {
+  const response = await fetch('/api/customer/groups');
+
+  if (!response.ok) {
+    throw new Error(`HTTP error! status: ${response.status}`);
+  }
+
+  const data: unknown = await response.json();
+  const groups = CustomerGroupsSchema.parse(data);
+
+  return groups;
+}
+
+async function fetchCustomerGroupData(): Promise<GetCustomerGroupResponse | undefined> {
+  const response = await fetch('/api/customer/group');
+
+  if (!response.ok) {
+    throw new Error(`HTTP error! status: ${response.status}`);
+  }
+
+  const data: unknown = await response.json();
+  const group = CustomerGroupSchema.parse(data);
+
+  return group;
+}
+
+function UntargetedGroup() {
+  return (
+    <div className="p-4 text-center text-lg text-gray-400">
+      This group needs to be added to "Targeted customer groups".
+    </div>
+  );
+}
+
+function getGroupSlot(
+  allSlots: Array<{ group?: string; slot: ReactNode }> | undefined,
+  simulateGroup: boolean,
+  simulatedGroup: string,
+  customerGroupId: number | undefined,
+  noGroupSlot: ReactNode,
+): ReactNode {
+  const simulatedSlot = allSlots?.find((s) => s.group === simulatedGroup)?.slot ?? (
+    <UntargetedGroup />
+  );
+  const actualSlot = allSlots?.find((s) => s.group === `${customerGroupId}`)?.slot ?? (
+    <UntargetedGroup />
+  );
+
+  if (!customerGroupId && !simulateGroup) {
+    return noGroupSlot;
+  }
+
+  return simulateGroup ? simulatedSlot : actualSlot;
+}
+
+export function SlotSkeleton({ className }: { className?: string }) {
+  return (
+    <div className={clsx(className, 'relative w-full animate-pulse p-2')}>
+      <div className="line-clamp-1 h-20 w-full rounded-lg bg-contrast-100" />
+    </div>
+  );
+}
+
+interface Props {
+  className: string;
+  slots?: Array<{ group?: string; slot: ReactNode }>;
+  simulatedGroup?: string;
+  noGroupSlot: ReactNode;
+}
+
+function CustomerGroupSlot({ className, slots, simulatedGroup = NO_GROUP_ID, noGroupSlot }: Props) {
+  const allSlots = slots?.concat({ group: NO_GROUP_ID, slot: noGroupSlot });
+  const isInBuilder = useIsInBuilder();
+
+  const { data, isLoading, error } = useSWR<GetCustomerGroupResponse | undefined, Error>(
+    '/api/customer/group',
+    fetchCustomerGroupData,
+  );
+
+  if (isLoading) return <SlotSkeleton className={className} />;
+
+  if (error) {
+    return (
+      <p className={clsx(className, 'p-4 text-center text-gray-500')}>
+        An error occurred trying to fetch the customer's group.
+      </p>
+    );
+  }
+
+  const customerGroupId = data?.customer?.customerGroupId;
+  const groupSlot = getGroupSlot(
+    allSlots,
+    isInBuilder,
+    simulatedGroup,
+    customerGroupId,
+    noGroupSlot,
+  );
+
+  return <div className={className}>{groupSlot}</div>;
+}
+
+runtime.registerComponent(CustomerGroupSlot, {
+  type: 'catalyst-customer-group-slot',
+  label: 'Catalyst / Customer Group Slot',
+  props: {
+    className: Style(),
+    slots: List({
+      label: 'Targeted customer groups',
+      type: Group({
+        label: 'Group',
+        props: {
+          group: Combobox({
+            label: 'Name',
+            getOptions: async (query) => {
+              try {
+                const data = await getAllCustomerGroups();
+
+                if (!data) return [];
+
+                return data
+                  .map((d) => ({
+                    id: `${d.id}`,
+                    label: d.name,
+                    value: `${d.id}`,
+                  }))
+                  .filter((option) => option.label.toLowerCase().includes(query.toLowerCase()));
+              } catch (error) {
+                // eslint-disable-next-line no-console
+                console.error('Error fetching customer group options:', error);
+
+                return [];
+              }
+            },
+          }),
+          slot: Slot(),
+        },
+      }),
+      getItemLabel(item) {
+        return item?.group.label ?? 'Unselected group';
+      },
+    }),
+    simulatedGroup: Combobox({
+      label: 'Simulated group',
+      getOptions: async (query) => {
+        try {
+          const data = await getAllCustomerGroups();
+
+          if (!data) return [];
+
+          return [
+            {
+              id: NO_GROUP_ID,
+              label: 'No group',
+              value: NO_GROUP_ID,
+            },
+            ...data
+              .map((d) => ({
+                id: `${d.id}`,
+                label: d.name,
+                value: `${d.id}`,
+              }))
+              .filter((option) => option.label.toLowerCase().includes(query.toLowerCase())),
+          ];
+        } catch (error) {
+          // eslint-disable-next-line no-console
+          console.error('Error fetching customer group options:', error);
+
+          return [];
+        }
+      },
+    }),
+    noGroupSlot: Slot(),
+  },
+});
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
+⚠️ **IMPORTANT:** As of January 6, 2025, the `main` branch is frozen. The default branch is now `canary`. Please make sure to work with and submit PRs to the `canary` branch.
+
+
 <a href="https://catalyst.dev" target="_blank" rel="noopener norerrer">
   <img src="https://storage.googleapis.com/bigcommerce-developers/images/catalyst_readme_banner.png" alt="Catalyst for Composable Commerce Image Banner" title="Catalyst">
 </a>
--- a/core/lib/makeswift/components.ts
+++ b/core/lib/makeswift/components.ts
@@ -3,6 +3,7 @@ import './components/button-link/button-link.makeswift';
 import './components/card-carousel/card-carousel.makeswift';
 import './components/card/card.makeswift';
 import './components/carousel/carousel.makeswift';
+import './components/customer-group-slot/customer-group-slot.makeswift';
 import './components/product-card/product-card.makeswift';
 import './components/products-carousel/products-carousel.makeswift';
 import './components/products-list/products-list.makeswift';
--- a/core/lib/makeswift/components/customer-group-slot/schema.ts
+++ b/core/lib/makeswift/components/customer-group-slot/schema.ts
@@ -1,0 +1,20 @@
+import { z } from 'zod';
+
+export const CustomerGroupsSchema = z.optional(
+  z
+    .object({
+      id: z.number(),
+      name: z.string(),
+    })
+    .array(),
+);
+
+export type CustomerGroupsType = z.infer<typeof CustomerGroupsSchema>;
+
+export const CustomerGroupSchema = z.object({
+  customer: z
+    .object({
+      customerGroupId: z.number(),
+    })
+    .nullable(),
+});


core/app/api/customer/groups/route.ts
```diff
@@ -0,0 +1,32 @@
+import { strict } from 'assert';
+import { NextResponse } from 'next/server';
+import { z } from 'zod';
+
+export const Customers = z
```
jamesqquick, line 5: Is this exported because it's used somewhere else?

core/app/api/customer/groups/route.ts
```diff
@@ -0,0 +1,32 @@
...
+      method: 'GET',
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json',
+        'X-Auth-Token': process.env.BIGCOMMERCE_ACCESS_TOKEN,
+      },
+    },
+  );
+  const jsonResponse: unknown = await response.json();
+  const groups = Customers.parse(jsonResponse);
```
jamesqquick, line 29: How is error handling happening for this?

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { ReactNode } from 'react';
+import useSWR from 'swr';
+
+import { Customers, CustomersType } from '~/app/api/customer/groups/route';
+import { GetCustomerGroupResponse } from '~/client/queries/get-customer';
+import { runtime } from '~/lib/makeswift/runtime';
+
+async function fetchData(): Promise<CustomersType> {
+  const response = await fetch('/api/customer/groups');
+  const data: unknown = await response.json();
```
jamesqquick, line 12: How do we handle errors here?

core/client/queries/get-customer.ts
```diff
@@ -0,0 +1,23 @@
...
+export type GetCustomerGroupResponse = ResultOf<typeof GetCustomerGroup>;
+
+export async function getCustomer() {
+  const customerAccessToken = await getSessionCustomerAccessToken();
+  const response = await client.fetch({
+    document: GetCustomerGroup,
+    customerAccessToken,
+  });
+
+  return { data: response.data };
```
jamesqquick, line 22: Not required, but you could destructure `{ data }` from the response directly then use shorthand object notation. Don't know if we have opinions on doing this.

```
const { data } = await client.fetch({
    document: GetCustomerGroup,
    customerAccessToken,
  });

  return { data };
```

core/app/api/customer/group/route.ts
```diff
@@ -0,0 +1,9 @@
+import { NextResponse } from 'next/server';
+
+import { getCustomer } from '~/client/queries/get-customer';
+
+export async function GET() {
+  const { data } = await getCustomer();
```
jamesqquick, line 6: How do we handle errors?

core/app/api/customer/group/route.ts
```diff
@@ -0,0 +1,9 @@
...
+
+  return NextResponse.json(data);
```
migueloller, line 8: Is it okay to expose the customer like this in an unauthenticated endpoint? We should probably expose only what's needed to implement the customer group functionality to avoid leaking sensitive user info

core/app/api/customer/groups/route.ts
```diff
@@ -0,0 +1,32 @@
...
+  .object({
+    id: z.number(),
+    name: z.string(),
+  })
+  .array();
+
+export type CustomersType = z.infer<typeof Customers>;
+
+export async function GET() {
+  strict(process.env.BIGCOMMERCE_ACCESS_TOKEN, 'BIGCOMMERCE_ACCESS_TOKEN is required');
```
migueloller, line 15: `strict` isn't used in this codebase, let's change this to be consistent with environment variable use elsewhere

core/app/api/customer/groups/route.ts
```diff
@@ -0,0 +1,32 @@
...
+    `https://api.bigcommerce.com/stores/${process.env.BIGCOMMERCE_STORE_HASH}/v2/customer_groups`,
+    {
+      method: 'GET',
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json',
+        'X-Auth-Token': process.env.BIGCOMMERCE_ACCESS_TOKEN,
+      },
+    },
+  );
```
migueloller, line 27: Should we be using the GraphQL API here? In particular, the Catalyst GraphQL client?

core/client/queries/get-customer.ts
```diff
@@ -0,0 +1,23 @@
...
+  query CustomerGroup {
+    customer {
+      customerGroupId
+    }
+  }
+`);
+
+export type GetCustomerGroupResponse = ResultOf<typeof GetCustomerGroup>;
+
+export async function getCustomer() {
```
migueloller, line 15: This name is too generic for what it's doing. IMO we should name this `getCustomerGroupId` and return the customer group ID only

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
```diff
@@ -0,0 +1,99 @@
...
+interface Props {
+  className: string;
+  slots?: Array<{ group?: string; slot: ReactNode }>;
+  selectedGroup?: string;
+  noGroupSlot: ReactNode;
+}
+
+function CustomerGroupSlot({ className, slots, selectedGroup = 'no-group', noGroupSlot }: Props) {
+  const allSlots = slots?.concat({ group: 'no-group', slot: noGroupSlot });
+  const isInBuilder = useIsInBuilder();
```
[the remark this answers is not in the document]
migueloller: Yeah, this is something that I wouldn't want to promote usage of. It does make the UX of the controls better but I would like to think of perhaps a different API that isn't as problematic as this one

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
```diff
@@ -0,0 +1,99 @@
+import { Combobox, Group, List, Slot, Style } from '@makeswift/runtime/controls';
+import { useIsInBuilder } from '@makeswift/runtime/react';
+import { ReactNode } from 'react';
+import useSWR from 'swr';
+
+import { Customers, CustomersType } from '~/app/api/customer/groups/route';
+import { GetCustomerGroupResponse } from '~/client/queries/get-customer';
+import { runtime } from '~/lib/makeswift/runtime';
+
+async function fetchData(): Promise<CustomersType> {
```
migueloller, line 10: Can we give this a more descriptive name?

migueloller (changes requested):

core/app/api/customer/groups/route.ts
drewpledger: I couldn't find anything in the schema, and after digging around I found [this](https://developer.bigcommerce.com/docs/rest-management/customers-v2/customer-groups#get-all-customer-groups), and confirmed it with Nathan that this is the way.

core/app/api/customer/groups/route.ts
drewpledger: Yes

core/app/api/customer/groups/route.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  if (!response.ok) {
+    throw new Error(`HTTP error! status: ${response.status}`);
+  }
+
+  const jsonResponse: unknown = await response.json();
+  const groups = CustomersGroup.parse(jsonResponse);
```
jamesqquick, line 35: This can still throw an error. Alternatively, you can use `safeParse()` which will return an error but not throw one. I would wrap this whole function  in try/catch. Then, for the error, you can do checking on the type of the error to return specific messages. In this case, you can check for instance of `ZodError`.

You can see an example in the Zod docs here -  https://zod.dev/ERROR_HANDLING?id=a-demonstrative-example

core/app/api/customer/group/route.ts
```diff
@@ -0,0 +1,13 @@
...
+
+import { getCustomerGroupId } from '~/client/queries/get-customer';
+
+export async function GET() {
+  try {
+    const { data } = await getCustomerGroupId();
+
+    return NextResponse.json(data);
+  } catch (error) {
+    return NextResponse.json({ error }, { status: 500 });
```
jamesqquick, line 11: I don't know if we want to return the full error here. What details we return depends on the intended use case of the caller. Who/what would be calling this endpoint?

core/app/api/customer/groups/route.ts
```diff
@@ -0,0 +1,38 @@
...
+      method: 'GET',
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json',
+        'X-Auth-Token': process.env.BIGCOMMERCE_ACCESS_TOKEN,
+      },
+    },
+  );
+
+  if (!response.ok) {
```
jamesqquick, line 30: `fetch` can still throw an error which isn't covered here. `response.ok` only checks for a status between 200-300. I would wrap this whole function in a try/catch. Then we can check for different instances of errors and return appropriate data. Looking at `core/app/api/products/group/[group]/route.ts`, here's the general error handling there.

```
  if (error instanceof Error) {
      return { status: 'error', error: error.message };
    }

    return { status: 'error', error: 'Something went wrong. Please try again.' };
```

core/app/api/customer/groups/route.ts
jamesqquick: So you would check in your top-level try/catch instanced of a zod specific error as well as a generic error.

core/app/api/customer/groups/route.ts
```diff
@@ -0,0 +1,38 @@
...
+    throw new Error(`HTTP error! status: ${response.status}`);
```
jamesqquick, line 31: I don't think we want to throw an error here. Assuming this endpoint is being called from some frontend, we need return appropriate data which would include a message and http status code. If you wrap this whole function in try/catch and looking instance of generic `Error` as mentioned above this would work out.

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
```diff
@@ -0,0 +1,123 @@
...
+  simulatedGroup = 'no-group',
+  noGroupSlot,
+}: Props) {
+  const allSlots = slots?.concat({ group: 'no-group', slot: noGroupSlot });
+
+  const { data, isLoading } = useSWR<GetCustomerGroupResponse>('/api/customer/group', async () => {
+    const response = await fetch('/api/customer/group');
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+    return response.json();
```
jamesqquick, line 43: I think this response needs to be validated before parsing it as json. You can start with checking `status.ok`. I think you'll still need to surround with a try/catch as well in case the network request itself fails. Any idea what should happen if that network request fails?

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+import { CustomerGroupType, CustomersGroup } from '~/app/api/customer/groups/route';
+import { GetCustomerGroupResponse } from '~/client/queries/get-customer';
+import { runtime } from '~/lib/makeswift/runtime';
+
+async function fetchData(): Promise<CustomerGroupType> {
+  const response = await fetch('/api/customer/groups');
+
+  if (!response.ok) {
+    throw new Error(`HTTP error! status: ${response.status}`);
```
jamesqquick, line 13: I would only throw an error here if we expect each caller to handle the actual error handling. Is that the intent? The one caller I see below uses try/catch. As long as the expectation is that the caller should handle it, this is ok.

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
```diff
@@ -0,0 +1,123 @@
+import { Checkbox, Combobox, Group, List, Slot, Style } from '@makeswift/runtime/controls';
+import { ReactNode } from 'react';
+import useSWR from 'swr';
+
+import { CustomerGroupType, CustomersGroup } from '~/app/api/customer/groups/route';
+import { GetCustomerGroupResponse } from '~/client/queries/get-customer';
+import { runtime } from '~/lib/makeswift/runtime';
+
+async function fetchData(): Promise<CustomerGroupType> {
```
jamesqquick, line 9: I think there was a previous comment about making this function name more explicit and agree. `getCustomerGroupType()`?

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
```diff
@@ -0,0 +1,123 @@
...
+      }),
+      getItemLabel(item) {
+        return item?.group.label ?? 'Unselected group';
+      },
+    }),
+    simulateGroup: Checkbox({ label: 'Simulate group', defaultValue: false }),
+    simulatedGroup: Combobox({
+      label: 'Simulated group',
+      getOptions: async () => {
+        const data = await fetchData();
```
jamesqquick, line 106: Need to handle errors and validate data before mapping

jamesqquick (changes requested):

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
drewpledger: I talked to miguel about this, we aren't going to try to error handle here. Console logging the error is enough

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
```diff
@@ -0,0 +1,141 @@
...
+  simulateGroup,
+  simulatedGroup = 'no-group',
+  noGroupSlot,
+}: Props) {
+  const allSlots = slots?.concat({ group: 'no-group', slot: noGroupSlot });
+
+  const { data, isLoading } = useSWR<GetCustomerGroupResponse>(
+    '/api/customer/group',
+    fetchCustomerGroupData,
+  );
```
migueloller, line 62: We need to handle errors here

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  if (!response.ok) {
+    throw new Error(`HTTP error! status: ${response.status}`);
+  }
+
+  const data: unknown = await response.json();
+  const group = CustomerGroupSchema.parse(data);
+
+  return group;
+}
```
migueloller, line 48: What's the intended way to handle errors when calling this function? There's a lot of stuff that can go wrong here and the one error we're throwing is a generic error that doesn't leave much room for control flow

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
```diff
@@ -0,0 +1,141 @@
...
+        // If the simulatedGroup does not exist in the list of slots
+        <div className="p-4 text-center text-lg text-gray-400">
+          This group needs to be added to "Targeted customer groups".
+        </div>
+      ))
+    : (allSlots?.find((s) => s.group === `${data?.customer?.customerGroupId}`)?.slot ?? (
+        <div className="p-4 text-center text-lg text-gray-400">
+          This group needs to be added to "Targeted customer groups".
+        </div>
+      ));
```
migueloller, line 76: Can we have some intermediate values to make this a bit easier to reason about? In particular, can we extract the slots we're finding?

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  const slot = !data?.customer?.customerGroupId && !simulateGroup ? noGroupSlot : groupSlot;
```
migueloller, line 78: This ternary is very difficult to reason about, can we rewrite this?

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+  if (isLoading) return 'Loading...'; // Replace me with an appropriate loading state
```
migueloller, line 80: Looks like there's a todo here to replace with a loading state?

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
```diff
@@ -0,0 +1,141 @@
...
+  const group = CustomerGroupSchema.parse(data);
+
+  return group;
+}
+
+function CustomerGroupSlot({
+  className,
+  slots,
+  simulateGroup,
+  simulatedGroup = 'no-group',
```
migueloller, line 54: Can we extract this into a constant instead of using a mage string here and [below](https://github.com/bigcommerce/catalyst/pull/1970/files#diff-a30ece14842040e4b499de1748068dbdf68baab6e40c80cee449e227cfc7e27aR133)?

migueloller (commented):
I agree with a lot of @jamesqquick's comments here. We need to do a lot better when it comes to error handling

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
drewpledger: SWR is returning errors from the request. It's destructured and used to render an error state for the component ```if (error)
    return (
      <p```

core/app/api/customer/groups/route.ts
```diff
@@ -0,0 +1,41 @@
+import { NextResponse } from 'next/server';
+
+import { CustomerGroupsSchema } from '~/lib/makeswift/components/customer-group-slot/schema';
+
+export async function GET() {
+  if (!process.env.BIGCOMMERCE_ACCESS_TOKEN) {
+    // eslint-disable-next-line no-console
+    console.error('BIGCOMMERCE_ACCESS_TOKEN is not set');
+
+    return NextResponse.json({ status: 'error', error: 'Internal server error' }, { status: 500 });
```
migueloller, line 10: This can be replaced with a thrown error. It will be logged to the console and remove the need to disable the ESLint rule and manually craft an error response

core/app/api/customer/groups/route.ts
```diff
@@ -0,0 +1,41 @@
...
+      throw result.error;
+    }
+
+    return NextResponse.json(result.data);
+  } catch (error) {
+    if (error instanceof Error) {
+      return { status: 'error', error: error.message };
+    }
+
+    return { status: 'error', error: 'Something went wrong. Please try again.' };
```
migueloller, line 39: This function should return a response. Add types to avoid this. A test would also catch this.

core/app/api/customer/groups/route.ts
```diff
@@ -0,0 +1,41 @@
...
+        },
+      },
+    );
+
+    const jsonResponse: unknown = await response.json();
+    const result = CustomerGroupsSchema.safeParse(jsonResponse);
+
+    if (!result.success) {
+      throw result.error;
+    }
```
migueloller, line 31: If we're just throwing the error like this, then call `parse` instead of `safeParse`

core/app/api/customer/group/route.ts
```diff
@@ -0,0 +1,17 @@
+import { NextResponse } from 'next/server';
+
+import { getCustomerGroupId } from '~/client/queries/get-customer';
+
+export async function GET() {
```
migueloller, line 5: Add types to this function. We're returning an object, which is incorrect.

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
```diff
@@ -0,0 +1,161 @@
...
+import useSWR from 'swr';
+
+import { GetCustomerGroupResponse } from '~/client/queries/get-customer';
+import { runtime } from '~/lib/makeswift/runtime';
+
+import { CustomerGroupSchema, CustomerGroupsSchema, CustomerGroupsType } from './schema';
+
+const NO_GROUP_ID = 'no-group';
+
+async function getAllCustomerGroups(): Promise<CustomerGroupsType | undefined> {
```
migueloller, line 15: We should use `| null` instead of `| undefined`

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
```diff
@@ -0,0 +1,161 @@
...
+    '/api/customer/group',
+    fetchCustomerGroupData,
+  );
+
+  const customerGroupId = data?.customer?.customerGroupId;
+
+  const simulatedSlot = allSlots?.find((s) => s.group === simulatedGroup)?.slot ?? (
+    <UntargetedGroup />
+  );
+  const actualSlot = allSlots?.find((s) => s.group === `${customerGroupId}`)?.slot ?? (
```
migueloller, line 68: Why are we using string interpolation here?

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
```diff
@@ -0,0 +1,161 @@
...
+    <UntargetedGroup />
+  );
+
+  // If we're simulating the group, render the selected group slot, otherwise use the group from the API. Coalesce to noGroupSlot if no group is found
+  const groupSlot = simulateGroup ? simulatedSlot : actualSlot;
+  const slot = !customerGroupId && !simulateGroup ? noGroupSlot : groupSlot;
```
migueloller, line 74: This logic is still very hard to follow. The fact we have to write a comment means we probably want to rework it. We should express the logic so that we don't need the comment IMO

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
```diff
@@ -0,0 +1,161 @@
...
+
+  if (error)
+    return (
+      <p className={clsx(className, 'p-4 text-center text-gray-500')}>
+        An error occurred trying to fetch the customer's group.
+      </p>
+    );
```
migueloller, line 81: Use braces for a multi-line `if` block

core/app/api/customer/groups/route.ts
drewpledger: @migueloller are you saying to write a test for this as well?

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
drewpledger: @migueloller the `s.group` is a string from the combobox, and `customerGroupId` is a number from graphql. Is there a better approach for this?

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
drewpledger: I turned this into a function, I hope it clarifies the logic

core/app/api/customer/groups/route.ts
migueloller: Types should be enough for now. I was just making a comment that if we had tests we would've caught the bug

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
migueloller: Ah, understood. `customerGroupId.toString()` makes the intent more explicit IMO

core/app/api/customer/group/route.ts
```diff
@@ -0,0 +1,16 @@
...
+  try {
+    const { data } = await getCustomerGroupId();
+
+    return NextResponse.json(data);
+  } catch (error) {
+    const errorMessage =
+      error instanceof Error ? error.message : 'Something went wrong fetching the customer group.';
+
+    return NextResponse.json({ customer: null, error: errorMessage }, { status: 500 });
+  }
```
migueloller, line 15: The try/catch seems unnecessary here. The message, "Something went wrong fetching the customer group", is somewhat obvious if we get a 500: something went wrong sending a GET to get customer groups. This function can probably just be this:

```ts
export async function GET() {
  const { data } = await getCustomerGroupId();

  return NextResponse.json(data);
}
```

What, other than `data`, does the function return?

migueloller (approved):
Looking great!

PR #1971, opened 2025-02-11
fix: pagination for PDP reviews

migueloller:
## What/Why?

Handles pagination parameters in PDP `Reviews` component.

## Testing

https://github.com/user-attachments/assets/10ab8d36-1cd0-478d-80b7-4e496aed7796

### DIFF
--- a/core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx
+++ b/core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx
@@ -8,6 +8,7 @@ import { client } from '~/client';
 import { PaginationFragment } from '~/client/fragments/pagination';
 import { graphql } from '~/client/graphql';
 import { revalidate } from '~/client/revalidate-target';
+import { defaultPageInfo, pageInfoTransformer } from '~/data-transformers/page-info-transformer';
 
 import { ProductReviewSchemaFragment } from './product-review-schema/fragment';
 import { ProductReviewSchema } from './product-review-schema/product-review-schema';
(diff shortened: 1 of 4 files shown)


core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx
```diff
@@ -107,8 +132,8 @@ const getPaginationInfo = async (productId: number) => {
 
   return hasNextPage || hasPreviousPage
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx:133]: Oversight on my part, but we should be using `pageInfoTransformer` for this, like the other components.

core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx
```diff
@@ -48,18 +57,34 @@ const ReviewsQuery = graphql(
...
+
+  if (after != null) {
+    variables.first = pageSize;
+    variables.after = after;
+  }
+
+  if (after == null && before != null) {
+    variables.last = pageSize;
+    variables.before = before;
+  }
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx:75]: Why not just use this `const paginationArgs = before ? { last: limit, before } : { first: limit, after };` like in other queries?

jorgemoya (commented):
Thanks for looking into this! just a couple of comments.

core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx
Comment (migueloller) [reply core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx:75]: I didn't look at other queries 😅 

I'll make this change

core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx
Comment (migueloller) [reply core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx]: Done. This also uncovered a bug where we weren't properly passing pagination button i18n labels

jorgemoya (approved):

PR #1972, opened 2025-02-11
feat: improve styling of html product descriptions

becomevocal:
## What/Why?

Updates the product description styles to space out elements and support the same header styling as the web page route.

This helps the content blend into the aesthetic better. The content is smashed together now, which is strange to have as a default!

Note: I could totally see this being inside `vibe` components themselves. If that makes more sense, I'll close this PR and make the changes there.

## Testing

Before:
Screenshot 2025-02-11 at 11 03 04 AM

After:
Screenshot 2025-02-11 at 11 03 34 AM

### DIFF
--- a/core/app/[locale]/(default)/product/[slug]/page.tsx
+++ b/core/app/[locale]/(default)/product/[slug]/page.tsx
@@ -113,7 +113,10 @@ const getProduct = async (props: Props) => {
     id: product.entityId.toString(),
     title: product.name,
     description: (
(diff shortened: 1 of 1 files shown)


migueloller (commented):
@becomevocal, this looks a lot better! I do think this should be a change that's applied in VIBES so that the `prose` class applies these styles consistently

bookernath:
Recommend we close this in favor of https://github.com/bigcommerce/catalyst/pull/2007

PR #1975, opened 2025-02-11
fix: localized home page routes are rewritten to the "catch all" page

agurtovoy:
## What/Why?

Fix `withRoutes` middleware to properly rewrite localized home page routes (`/fr`, `/es-MX`, etc.)

### Before

https://github.com/user-attachments/assets/e52ab236-5ff1-4ce4-add2-42bc8bc926b0

### After

https://github.com/user-attachments/assets/88b973e1-ec24-41bb-bb19-2e3c6e169936

### DIFF
--- a/core/middlewares/with-routes.ts
+++ b/core/middlewares/with-routes.ts
@@ -205,6 +205,10 @@ const updateStatusCache = async (
 };
 
 const clearLocaleFromPath = (path: string, locale: string) => {
+  if (path === `/${locale}` || path === `/${locale}/`) {
+    return '/';
+  }
+
   if (path.startsWith(`/${locale}/`)) {
     return path.replace(`/${locale}`, '');
   }
--- a/.changeset/new-books-drop.md
+++ b/.changeset/new-books-drop.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+fix: localized home page routes are rewritten to the "catch all" page


core/middlewares/with-routes.ts
```diff
@@ -205,6 +205,10 @@ const updateStatusCache = async (
 };
 
 const clearLocaleFromPath = (path: string, locale: string) => {
+  if (path === `/${locale}`) {
+    return '/';
+  }
+
   if (path.startsWith(`/${locale}/`)) {
```
agurtovoy, line 212: The existing condition wasn't catching localized home page routes that didn't end with a trailing slash, e.g. `/fr`.

core/middlewares/with-routes.ts
```diff
@@ -205,6 +205,10 @@ const updateStatusCache = async (
 };
 
 const clearLocaleFromPath = (path: string, locale: string) => {
+  if (path === `/${locale}`) {
+    return '/';
+  }
```
agurtovoy, line 210: Added the case for `/${locale}` as a separate if to return `/` instead of '' here. Rewrites work correctly for both returns, but the function's return value is also used to get a non-localized pathname for the cache look up [in line 220 below](https://github.com/bigcommerce/catalyst/pull/1975/files#diff-fb9cfefddb5920bd983d193be75fdf2cf75455e3e42f28cffc500c6f72997a61L220), and I *think* the expectation is that it would return `/` for all home pages.

bookernath: Needs to handle trailing slash flag: https://github.com/bigcommerce/catalyst/blob/canary/core/next.config.ts#L40

migueloller: What do you guys think if we get a bit more structured here and parse the path into segments and then look at the first segment and compare it with the locale identifier?

bookernath (approved):

core/middlewares/with-routes.ts
agurtovoy: > What do you guys think if we get a bit more structured here and parse the path into segments and then look at the first segment and compare it with the locale identifier?

So we can do this:

```typescript
const clearLocaleFromPath = (path: string, locale: string) => {
  const segments = path.split('/').filter(Boolean);

  if (segments.length === 0 || segments[0] !== locale) {
    return path;
  }

  return `/${segments.slice(1).join('/')}`;
};
```

It's nice and simple, but, with the exception of the home page, it always removes the trailing slash. To fix that, we can do something along the line of

```typescript
const trailingSlash = process.env.TRAILING_SLASH !== 'false'

const clearLocaleFromPath = (path: string, locale: string) => {
  const segments = path.split('/').filter(Boolean);

  if (segments.length === 0 || segments[0] !== locale) {
    return path;
  }

  if (segments.length === 1) {
    return '/';
  }

  const suffix = trailingSlash ? '/' : '';
  return `/${segments.slice(1).join('/')}${suffix}`;
};
```

Or we can keep the current search & replace approach:

```typescript
const clearLocaleFromPath = (path: string, locale: string) => {
  if (path === `/${locale}` || path === `/${locale}/`) {
    return '/';
  }

  if (path.startsWith(`/${locale}/`)) {
    return path.replace(`/${locale}`, '');
  }

  return path;
};
```

Preferences?

migueloller: I think that given this is an encapsulated piece of functionality, clearing the locale from the path, I'd opt for the latter since it's easier to follow. Although the former could be made easier to follow if some of the logic were extracted into utilities. I vote for the latter.

That being said, I have a question. Does `path.replace` only replace the first occurrence? If it replaces all occurrences then we might want to take a different approach either with a regex or some other method that replaces only the first occurrence.

agurtovoy: > Does `path.replace` only replace the first occurrence?

I had the same q, because we're passing a string, it replaces [the first occurrence only](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace).

PR #1977, opened 2025-02-12
Add persistent cart support

jordanarldt:
## What/Why?
- Add support for Persistent Carts (restoring or merging carts during login)
- Ensure `/logout` link refreshes the layout with a force-refresh cookie component

## Testing

https://github.com/user-attachments/assets/fc85a346-0d8e-431f-8dad-61186d9daf8d

### DIFF
--- a/core/app/notifications.tsx
+++ b/core/app/notifications.tsx
@@ -1,6 +1,13 @@
+'use client';
+
+import { useEffect, useRef } from 'react';
 import { Toaster } from 'react-hot-toast';
 
+import { toast } from '@/vibes/soul/primitives/toaster';
+import { type ServerToastData } from '~/lib/server-toast';
+
 export const Notifications = () => {
+  // TODO: Remove this component after product comparisons are migrated to VIBES
   return (
     <Toaster
       containerClassName="px-4 2xl:container sm:px-10 lg:px-12 2xl:mx-auto 2xl:px-0"
@@ -11,4 +18,19 @@ export const Notifications = () => {
       }}
     />
   );
+};
+
+export const CookieNotifications = (notification: ServerToastData) => {
+  const lastRendered = useRef<ServerToastData>(null);
+
+  useEffect(() => {
+    const { message, variant, position, description } = notification;
+
+    if (notification.id !== lastRendered.current?.id) {
+      toast[variant](message, { position, description });
+      lastRendered.current = notification;
+    }
+  }, [notification]);
+
+  return null;
 };
--- a/core/lib/client-cookies.ts
+++ b/core/lib/client-cookies.ts
@@ -1,0 +1,57 @@
+export const FORCE_REFRESH_COOKIE = 'force-refresh';
+
+interface ClientCookieOptions {
+  expires: Date;
+  path: string;
+  domain: string;
+  secure: boolean;
+  sameSite: 'Strict' | 'Lax' | 'None';
+  maxAge: number;
+}
+
+const cookiePropertyMap: Record<string, string> = {
+  expires: 'Expires',
+  path: 'Path',
+  domain: 'Domain',
+  secure: 'Secure',
+  sameSite: 'SameSite',
+  maxAge: 'Max-Age',
+};
+
+export function getCookieValue(name: string): string | null {
+  const cookies = document.cookie.split('; ');
+  const cookie = cookies.find((c) => c.startsWith(`${name}=`));
+
+  if (!cookie) {
+    return null;
+  }
+
+  const cookieValue = cookie.split('=')[1];
+
+  return cookieValue ?? null;
+}
+
+export function setCookie(
+  name: string,
+  value: string,
+  options?: Partial<ClientCookieOptions>,
+): void {
+  const opts = options ?? {};
+  const cookieOptions = Object.entries(opts)
+    .reduce<string[]>((acc, [key, optValue]) => {
+      const propKey = cookiePropertyMap[key];
+
+      if (!propKey) {
+        return acc;
+      }
+
+      if (propKey === 'Secure' && optValue === true) {
+        return [...acc, propKey];
+      }
+
+      return [...acc, `${propKey}=${optValue.toString()}`];
+    }, [])
+    .join('; ');
+
+  document.cookie = `${name}=${value}; ${cookieOptions};`;
+}
--- a/core/lib/server-toast.ts
+++ b/core/lib/server-toast.ts
@@ -1,0 +1,78 @@
+import { cookies } from 'next/headers';
+import { z } from 'zod';
+import 'server-only';
+
+const serverToastSchema = z.object({
+  id: z.number(),
+  message: z.string(),
+  description: z.string().optional(),
+  variant: z.enum(['success', 'error', 'warning', 'info']),
+  position: z
+    .enum(['top-left', 'top-right', 'top-center', 'bottom-left', 'bottom-right', 'bottom-center'])
+    .optional(),
+});
+
+export type ServerToastData = z.infer<typeof serverToastSchema>;
+
+type ServerToastOptions = Pick<ServerToastData, 'position' | 'description'>;
+
+const TOAST_COOKIE = 'toast-notification';
+
+function encode(data: ServerToastData): string {
+  return btoa(JSON.stringify(data));
+}
+
+function decode(data: string): ServerToastData {
+  return serverToastSchema.parse(JSON.parse(atob(data)));
+}
+
+async function setToastCookie(data: ServerToastData) {
+  const cookieStore = await cookies();
+
+  cookieStore.set(TOAST_COOKIE, encode(data), {
+    httpOnly: true,
+    secure: true,
+    sameSite: 'strict',
+    path: '/',
+    partitioned: true,
+    maxAge: 0,
+  });
+}
+
+/**
+ * Server-side toast message propagator.
+ * Allows queuing a toast message from a server action which will be displayed when the next route renders.
+ * Only allows data serializable as JSON.
+ */
+export const serverToast = {
+  success: async (message: string, options?: ServerToastOptions) => {
+    await setToastCookie({ id: Date.now(), message, variant: 'success', ...options });
+  },
+  error: async (message: string, options?: ServerToastOptions) => {
+    await setToastCookie({ id: Date.now(), message, variant: 'error', ...options });
+  },
+  warning: async (message: string, options?: ServerToastOptions) => {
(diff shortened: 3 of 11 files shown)


bookernath (commented):
Nice job going above and beyond with the toasts 👏

core/messages/en.json
```diff
@@ -298,6 +298,8 @@
     "increment": "Increase quantity",
     "decrement": "Decrease quantity",
     "removeItem": "Remove item",
+    "cartCombined": "You left some items from a previous cart. We have added those items to your current cart.",
+    "cartRestored": "You had started a cart on another device. We have restored that cart for you.",
```
bookernath, line 302:
```suggestion
    "cartRestored": "You started a cart on another device, and we've restored it here so you can pick up where you left off.",
```

core/vibes/soul/form/select/index.tsx
```diff
@@ -16,7 +16,9 @@ type Props = {
   label?: string;
   hideLabel?: boolean;
   variant?: 'round' | 'rectangle';
-  options: Array<{ label: string; value: string }>;
+  options: Array<
+    { label: string; value: string } | { label: string; action: () => void | Promise<void> }
+  >;
```
migueloller, line 21: I don't think it's appropriate for the `Select` component to take a server action. Instead, a `<form>` _outside_ `Select` should be used to handle the action. If we use a `<form>` inside `Select` then it will result in nested forms in other places where we use `Select` within a form

core/lib/server-toast.ts
```diff
@@ -0,0 +1,84 @@
...
+  message: z.string(),
+  description: z.string().optional(),
+  variant: z.enum(['success', 'error', 'warning', 'info']),
+  position: z
+    .enum(['top-left', 'top-right', 'top-center', 'bottom-left', 'bottom-right', 'bottom-center'])
+    .optional(),
+});
+
+const TOAST_COOKIE = 'toast-notification';
+const getExpiry = () => new Date(Date.now() + 5000);
```
migueloller, line 26: Expiration needs to be instant (i.e., `maxAge: 0`) to guarantee the cookie is set only for the request right after the notification is set to avoid duplication of messages

core/lib/server-toast.ts
```diff
@@ -0,0 +1,84 @@
...
+
+interface ServerToastOptions {
+  description?: string;
+  position?: ToasterProps['position'];
+}
+
+export interface ServerToastData extends ServerToastOptions {
+  message: string;
+  variant: 'success' | 'error' | 'warning' | 'info';
+}
```
migueloller, line 14: We should derive these types from the Zod schema using `z.infer`

core/lib/server-toast.ts
```diff
@@ -0,0 +1,84 @@
...
+async function setToastCookie(data: ServerToastData) {
+  const cookieStore = await cookies();
+
+  cookieStore.set(TOAST_COOKIE, encode(data), {
+    httpOnly: true,
+    secure: process.env.NODE_ENV === 'production',
+    sameSite: 'strict',
+    path: '/',
+    expires: getExpiry(),
+  });
```
migueloller, line 45: Can we please set `partitioned: true` so this behaves well in iframes (e.g., Makeswift builder)? We could also do it in the `integrations/makeswift` branch only once this is merged

core/lib/server-toast.ts
```diff
@@ -0,0 +1,84 @@
...
+function decode(data: string): ServerToastData {
+  return serverToastSchema.parse(JSON.parse(atob(data)));
+}
+
+async function setToastCookie(data: ServerToastData) {
+  const cookieStore = await cookies();
+
+  cookieStore.set(TOAST_COOKIE, encode(data), {
+    httpOnly: true,
+    secure: process.env.NODE_ENV === 'production',
```
migueloller, line 41: I think that we want this to always be `true`. Most browsers treat `localhost` as secure even if not HTTPS. Furthermore, `NODE_ENV` being `production` doesn't signify this is running in HTTPS as one could run `next start` locally and it `NODE_ENV` would be `production` anyway.

core/app/notifications.tsx
```diff
@@ -12,3 +19,18 @@ export const Notifications = () => {
...
   );
 };
+
+export const CookieNotifications = (notification: ServerToastData) => {
+  const hasRendered = useRef(false);
+
+  useEffect(() => {
+    const { message, variant, position, description } = notification;
+
+    if (!hasRendered.current) {
```
migueloller, line 29: Won't this cause this component to never show a message unless the component remounts? If there is a re-render as a result from a cookie being set that includes a message, wouldn't that mean the message isn't shown? For example, if a new `notification` is passed in, even though the `useEffect` would run again, a message wouldn't be shown.

I wonder if we want to set this cookie not to be `httpOnly` so that we can "consume" the cookie on render and show it only once

migueloller (changes requested):
I don't think the changes to VIBES here are following patterns that we want to have. In particular, taking in `action` props on low-level primitives and rendering `<form>`s within them to handle them is a no-go IMO. Instead, we should be using forms together `useActionState` in sections that can use the primitives as form input fields that determine the form action payloads

core/lib/server-toast.ts
jordanarldt: @migueloller Ah nice, I wasn't sure if setting that to 0 would work. I imagine the cookie would have expired before the redirect from the server action occurred, but it seems to work!

core/lib/server-toast.ts
jordanarldt: @migueloller Good call, although I'm unsure how this will affect scenarios where we run Catalyst against a CDVM store.

jordanarldt: I did notice that we are using this same pattern for the cart ID cookie though.

core/app/notifications.tsx
jordanarldt: @migueloller That's the idea. If we don't have this, the component will display any cookie toast message two times. Cookies can only be set in a server action or a route handler, so if there _is_ a server-side cookie toast message to display, the component will have re-rendered anyways in order to show it.

If we need to display a toast message outside of that boundary, we would just use the client-side implementation like we do when we add a product to the cart.

jordanarldt: Furthermore, this `serverToast` is intended specifically for server actions, which will always make sure that the `CookieNotifications` component is re-rendered in the layout.

You can set cookies in an API route handler as well, but there would still be no need because you can just spawn a toast message based on the result of the API call in the client-code.

migueloller: > If we don't have this, the component will display any cookie toast message two times.

I think that this might be a result of `<ScricttMode>` which causes effects to run twice in development mode. I think the way to address this would be to consume the cookie when showing the message so that even if the effect runs twice, it only shows it once.

> Furthermore, this `serverToast` is intended specifically for server actions, which will always make sure that the `CookieNotifications` component is re-rendered in the layout.

Re-rendered, yes, but not remounted. The current implementation seems to only ever show one message and no more unless the components remounts. This can probably be tested easily by creating a button that calls a server action that sets the notification. I would expect this button to only work the first time it's clicked.

jordanarldt: @migueloller Ah yeah, that is true. StrictMode is the reason it fires off twice - I forget about that sometimes.

> Re-rendered, yes, but not remounted. The current implementation seems to only ever show one message and no more unless the components remounts. This can probably be tested easily by creating a button that calls a server action that sets the notification. I would expect this button to only work the first time it's clicked.

Also a good point! This would be the case. My implementation was intending for it to be fired in cases of a server-action that ends with a redirect, but it would be better to make it work outside of that scenario as well.

Consuming the cookie after the notification is fired would be ideal 🤔 the roadblock I encountered when trying to implement that was the fact that you can't modify cookies in a component, it has to be a server action or API route. Since we're firing the toast from a client component, we'd probably need to make an API endpoint specifically for clearing the notification cookie.

migueloller: Since this will always be consumed in a client component (the toast API is a client API) then we would be doing this via `document.cookie`, not `cookies()`. That's perfectly fine to do in a component in an event callback or in a `useEffect`.

core/lib/server-toast.ts
migueloller: Yeah, the `Set-Cookie` header will be in the redirect response and so the browser will set it for that response but it will immediately expire. This guarantees that it's seen only in one response's document

core/lib/client-cookies.ts
```diff
@@ -0,0 +1,57 @@
...
+    .reduce<string[]>((acc, [key, optValue]) => {
+      const propKey = cookiePropertyMap[key];
+
+      if (!propKey) {
+        return acc;
+      }
+
+      if (propKey === 'Secure' && optValue === true) {
+        return [...acc, propKey];
+      }
```
jorgemoya, line 50: Can you explain why we're doing this? 🙏

jorgemoya (approved):
LGTM!

core/lib/client-cookies.ts
jordanarldt: @jorgemoya It's just a simple client-side cookie function, when setting a cookie on the client, the `Secure` flag has no value so `{ secure: true }` becomes `my_cookie=value; Secure` and `{ secure: false }` would be `my_cookie=value;`

jordanarldt: A better example:
```
setCookie('my_cookie', 'test_value', { secure: true, maxAge: 1000 })
```

would become:
```
my_cookie=test_value; Secure; Max-Age=1000;`
```

core/lib/client-cookies.ts
migueloller, line 1: IMO we should use some OSS library to handle this for us so we don't have to maintain this type of "commodity" code. Not something I'd want to block this PR for, though.

core/lib/refresh.ts
migueloller, line 1: IMO it's a bit confusing we've for `lib/refresh.ts` and also `components/refresh/index.tsx`. What if we had a single `lib/refresh.tsx` file that had all related utilities in one place (i.e., the hook and the component). We might need multiple modules for server/client but it would still be nice to have it all in a single place instead of spread out

Also, what if we renamed this something a bit more explicit like "force-refresh"?

migueloller (approved):
I'd rather we consume the cookie client-side for the server-side toasts but I think this is ready to go! 👍🏻

chanceaclark (approved):
🍹 Would be nice if we could shift this functionality onto middleware, but I totally understand the performance impacts there.

PR #1979, opened 2025-02-13
feat(core): add coupon code to Cart

jorgemoya:
## What/Why?
Adds coupon code form to Cart.

Pending approval of https://github.com/makeswift/vibes/pull/505

https://github.com/user-attachments/assets/5667127c-9dd0-4b79-b6f5-7355d1c8d88f

## Testing
Locally.

### DIFF
--- a/core/app/[locale]/(default)/cart/_actions/apply-coupon-code.ts
+++ b/core/app/[locale]/(default)/cart/_actions/apply-coupon-code.ts
@@ -1,0 +1,63 @@
+'use server';
+
+import { revalidateTag } from 'next/cache';
+import { z } from 'zod';
+
+import { getSessionCustomerAccessToken } from '~/auth';
+import { client } from '~/client';
+import { graphql, VariablesOf } from '~/client/graphql';
+import { TAGS } from '~/client/tags';
+
+const ApplyCouponCodeSchema = z.object({
+  checkoutEntityId: z.string(),
+  couponCode: z.string(),
+});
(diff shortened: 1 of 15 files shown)


jorgemoya:
I wonder if I should change the error message we get from the BE to the generic "Please enter a valid coupon code". 🤔

core/app/[locale]/(default)/cart/_actions/apply-coupon-code.ts
```diff
@@ -0,0 +1,51 @@
...
+    variables: {
+      applyCheckoutCouponInput: {
+        checkoutEntityId,
+        data: {
+          couponCode,
+        },
+      },
+    },
+    customerAccessToken,
+    fetchOptions: { cache: 'no-store' },
```
Comment (migueloller) [inline core/app/[locale]/(default)/cart/_actions/apply-coupon-code.ts:43]: Not something to address on this PR, but perhaps we should look at `document`'s operation and if it's a mutation we automatically set `fetchOptions.cache` to `no-store`. We never want to cache mutations.

migueloller (approved):
The code looks very clean! Nice work @jorgemoya 

I skipped reviewing code in the `vibes/soul` directory in favor of looking at the VIBES PR.

core/app/[locale]/(default)/cart/_actions/apply-coupon-code.ts
Comment (jorgemoya) [reply core/app/[locale]/(default)/cart/_actions/apply-coupon-code.ts:43]: Good idea!

PR #1986, opened 2025-02-18
feat: manage locales when creating storefronts from the CLI

RomanKrasinskyi:
## What/Why?
Manage locales via CLI when creating a channel.

To rich a goal I created multi-select solution to select additional locales.

### About the Multi-Select Solution.
To enable the selection of multiple locales, a custom multi-select component is required with the following features:

- **Multiple Selection**: Users must be able to select more than one locale.
- **Search Functionality**: Due to the extensive list of available locales, a search feature is essential for efficient navigation and selection.
- **Selection Limit**: A maximum of five locales per channel is permitted. The component should enforce this limit and provide clear feedback to the user when they attempt to exceed it.
- **Informative Tooltips**: Helpful tooltips should be displayed to guide users and provide additional context about each locale.

No one of existing libraries fully met all requirements. While the `@inquirer/prompts` library offers similar `checkbox` functionality, it lacks the necessary search capability. Therefore, I developed a custom solution, combining the `select` and` checkbox` behaviors of `@inquirer/prompts` to meet these specific needs.

## Testing
Run channel creating

### DIFF
--- a/packages/create-catalyst/src/commands/create.ts
+++ b/packages/create-catalyst/src/commands/create.ts
@@ -6,10 +6,12 @@ import { pathExistsSync } from 'fs-extra/esm';
 import kebabCase from 'lodash.kebabcase';
 import { join } from 'path';
 
+import { multiSelect } from '../prompts/multi-select';
 import { CliApi } from '../utils/cli-api';
 import { cloneCatalyst } from '../utils/clone-catalyst';
 import { Https } from '../utils/https';
 import { installDependencies } from '../utils/install-dependencies';
+import { getAvailableLocales } from '../utils/localization';
 import { login, storeCredentials } from '../utils/login';
 import { Telemetry } from '../utils/telemetry/telemetry';
 import { writeEnv } from '../utils/write-env';
@@ -55,10 +57,62 @@ function getPlatformCheckCommand(command: string): string {
 
 const telemetry = new Telemetry();
 
-async function handleChannelCreation(cliApi: CliApi) {
+async function handleChannelCreation(bc: Https, cliApi: CliApi) {
   const newChannelName = await input({
     message: 'What would you like to name your new channel?',
   });
+
+  let availableLocales = [];
+
+  try {
+    availableLocales = await getAvailableLocales(bc);
+  } catch (error) {
+    if (error instanceof Error) {
+      console.error(chalk.red(error.message));
+    }
+
+    process.exit(1);
+  }
+
+  const storefrontLocale = await select({
+    message: 'Which default language would you like to set for your channel?',
+    default: 'en',
+    choices: availableLocales,
+    theme: {
+      style: {
+        help: () => chalk.dim('(Select locale from the list or start typing the name)'),
+      },
+    },
+  });
+
+  const shouldAddAdditionalLocales = await select({
+    message: 'Would you like to add additional languages?',
+    choices: [
+      { name: 'Yes', value: true },
+      { name: 'No', value: false },
+    ],
+  });
+
+  let additionalLocales: string[] = [];
+
+  if (shouldAddAdditionalLocales) {
+    additionalLocales = await multiSelect({
+      choices: availableLocales.filter(({ value }) => value !== storefrontLocale),
+      message: 'Which additional languages would you like to add to your channel?',
+      theme: {
+        style: {
+          help: () => chalk.dim('(Select locale from the list or start typing the name)'),
+        },
+      },
+      validate: (selections) => {
+        if (selections.length > 4) {
+          return 'You can only select up to 4 additional languages';
+        }
+
+        return true;
+      },
+    });
+  }
 
   const shouldInstallSampleData = await select({
     message: 'Would you like to install sample data?',
@@ -68,7 +122,12 @@ async function handleChannelCreation(cliApi: CliApi) {
     ],
   });
 
-  const response = await cliApi.createChannel(newChannelName, shouldInstallSampleData);
+  const response = await cliApi.createChannel(
+    newChannelName,
+    storefrontLocale,
+    additionalLocales,
+    shouldInstallSampleData,
+  );
 
   if (!response.ok) {
     console.error(
@@ -395,7 +454,7 @@ export const create = new Command('create')
           }
 
           if (shouldCreateChannel) {
-            const channelData = await handleChannelCreation(cliApi);
+            const channelData = await handleChannelCreation(bc, cliApi);
 
             channelId = channelData.channelId;
             storefrontToken = channelData.storefrontToken;
--- a/packages/create-catalyst/src/utils/cli-api.ts
+++ b/packages/create-catalyst/src/utils/cli-api.ts
@@ -28,7 +28,12 @@ export class CliApi {
     });
   }
 
-  async createChannel(name: string, installSampleData = false) {
+  async createChannel(
+    name: string,
+    storefrontLocale: string,
+    additionalLocales: string[],
+    installSampleData = false,
+  ) {
     return this.client.fetch('/channels/catalyst', {
       method: 'POST',
       headers: { 'Content-Type': 'application/json' },
@@ -39,6 +44,8 @@ export class CliApi {
         },
         deployStorefront: true,
         devOrigin: 'http://localhost:3000',
+        storefrontLanguage: storefrontLocale,
+        additionalLocales,
       }),
     });
   }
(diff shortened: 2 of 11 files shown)


packages/create-catalyst/src/utils/localization.ts
```diff
@@ -0,0 +1,51 @@
...
+  Array.isArray(response.data) &&
+  response.data.every(
+    (item) =>
+      typeof item === 'object' &&
+      item !== null &&
+      'id' in item &&
+      'name' in item &&
+      'fallback' in item &&
+      'is_supported' in item,
+  );
```
chanceaclark, line 29: Can we use `zod` to validate/define these types?

packages/create-catalyst/src/utils/localization.ts
```diff
@@ -0,0 +1,51 @@
...
+      ),
+    );
+    process.exit(1);
+  }
+
+  const availableLocales: unknown = await availableLocalesResponse.json();
+
+  if (!isAvailableLocalesResponse(availableLocales)) {
+    console.error(chalk.red('\nUnexpected response format from available-locales endpoint\n'));
+    process.exit(1);
```
chanceaclark, line 47: Can we have these utils throw an error rather than exiting the process. `process.exit` should be apart of the application flow, rather than "utilities".

packages/create-catalyst/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "display": "Node",
   "compilerOptions": {
     "strict": true,
-    "target": "es6",
+    "target": "es2023",
```
chanceaclark, line 6: Why was this needed?

packages/create-catalyst/src/utils/cli-api.ts
```diff
@@ -28,7 +28,12 @@ export class CliApi {
     });
   }
 
-  async createChannel(name: string, installSampleData = false) {
+  async createChannel(
+    name: string,
+    storefrontLanguage: string,
+    additionalLocales: string[],
```
chanceaclark, line 34: Can we align on either "language" or "locale" here? e.g. could it be `storefrontLocale` instead?

packages/create-catalyst/src/commands/create.ts
```diff
@@ -55,11 +58,47 @@ function getPlatformCheckCommand(command: string): string {
...
+      { name: 'No', value: false },
+    ],
+  });
+
+  let additionalLocales: string[] = [];
+
+  if (shouldAddAdditionalLocales) {
+    additionalLocales = await multiSelect({
+      choices: availableLocales.filter(({ value }) => value !== storefrontLanguage),
+      maxSelections: 4,
```
chanceaclark, line 92: Could we follow a similar API to some of the other `@inquirer` prompts? They usually have a `validate` function which might be more extendable if we need to use this prompt in the future.

packages/create-catalyst/src/promts/multi-select/helpers.ts
chanceaclark, line 1: I think "promts" is misspelled in the path. Should be `prompts`.

chanceaclark (commented):
Hey @RomanKrasinskyi would you mind providing some more information in the PR description on your reasoning behind building your own prompt helper vs. finding a package? I already see why you did, but want to provide some context for other reviewers or our future selves as to why we decided to go down this path.

🍹 Something to think about... If we think the multi-select prompt is useful, we should think about creating a OSS repo for it as I think other engineers outside of our org would appreciate it.

packages/create-catalyst/src/utils/localization.ts
RomanKrasinskyi: Yeah, I will update it, to use `zod`

packages/create-catalyst/src/utils/localization.ts
RomanKrasinskyi: Yeah, it make sense for me 👍 
Should I catch it outside and existing the process?

packages/create-catalyst/tsconfig.json
RomanKrasinskyi: `const last = items.findLastIndex(isSelectable);`

findLastIndex() was added since 2022 year and works in all latest browsers and NodeJS 18+ versions
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findLastIndex

packages/create-catalyst/src/utils/cli-api.ts
RomanKrasinskyi: I will rename `storefrontLanguage` to `storefrontLocale`

packages/create-catalyst/src/commands/create.ts
RomanKrasinskyi: Make sense for me. I will check it

packages/create-catalyst/src/promts/multi-select/helpers.ts
RomanKrasinskyi: Good catch 👍

packages/create-catalyst/src/commands/create.ts
```diff
@@ -5,11 +5,14 @@ import { execSync } from 'child_process';
 import { pathExistsSync } from 'fs-extra/esm';
 import kebabCase from 'lodash.kebabcase';
 import { join } from 'path';
+import colors from 'yoctocolors-cjs';
```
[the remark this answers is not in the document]
RomanKrasinskyi: I'll check if it has a similar color, because the main thing is that the color is the same as the library uses

RomanKrasinskyi:
@chanceaclark I believe that multi-select should be in separate repository, because it can be useful in some cases. So I'm going to move it

packages/create-catalyst/src/commands/create.ts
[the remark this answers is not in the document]
chanceaclark: I would say get it close to the color you are using - it doesn't need to be exact.

packages/create-catalyst/src/utils/localization.ts
chanceaclark: Just throw an error, the whole process should stop as I don't think we are properly wrapping the code in `try/catch`.

packages/create-catalyst/src/commands/create.ts
RomanKrasinskyi: The validation triggers once the enter is pressed. I guess it's too late and better to check it once space is pressed. 
But If I change when validation is triggered, it will have the different behaviour in compare with other `@inquirer` prompts.

chanceaclark: I think that logic is what the other `@inquirer/prompts` do: https://github.com/SBoudrias/Inquirer.js/blob/main/packages/checkbox/src/index.ts#L96C3-L98

RomanKrasinskyi: and it calls here
https://github.com/SBoudrias/Inquirer.js/blob/main/packages/checkbox/src/index.ts#L189

but I would like to check amount of selected choices when user press space button:
```
if (isSpaceKey(key)) {
  ...
  if (maxSelections && items.filter(isChecked).length >= maxSelections) {
    setError(`You can only select up to ${maxSelections} choices`);

    return;
  }

  setItems(items.map((choice, i) => (i === active ? toggle(choice) : choice)));
}
```

chanceaclark: If we always show help text on the max selections + validating it on enter, I think it's _good enough_. If it's actually crucial to do validation on select too, then we can throw another `await validate()` under the space key as well.

RomanKrasinskyi:
@chanceaclark I updated PR. Please take a look.
Due the technical limitation, I filter locales and show only from the allowed list

packages/create-catalyst/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "display": "Node",
   "compilerOptions": {
     "strict": true,
-    "target": "es6",
+    "target": "esnext",
```
chanceaclark, line 6:
```suggestion
    "lib": ["esnext"],
    "target": "es6",
```

chanceaclark (commented):
Looks good to me, just update the tsconfig with my suggestion.

packages/create-catalyst/tsconfig.json
RomanKrasinskyi: @chanceaclark Thanks, updated.

chanceaclark (approved):

bookernath:
@RomanKrasinskyi I added a changelog for you - this is good to go from my perspective

PR #1990, opened 2025-02-20
fix: switching locales redirects user to the home page

agurtovoy:
## What/Why?

When switching to a different locale, the user should remain on their current page.

### Before

https://github.com/user-attachments/assets/1f7fc23e-5ec1-4dd9-af86-4aca89d42bbe

### After

https://github.com/user-attachments/assets/575b827f-e3c3-486f-8008-35d8317b8c48


core/components/header/header-section.tsx
```diff
@@ -0,0 +1,34 @@
...
+type Props = ComponentProps<typeof VibesHeaderSection>;
+
+const useSwitchLocale = () => {
+  const pathname = usePathname();
+  const router = useRouter();
+  const params = useParams();
+
+  return useCallback(
+    (locale: string) =>
+      router.push(
```
agurtovoy, line 18: Using `next-intl`'s [recommended way](https://next-intl.dev/docs/routing/navigation#userouter) to switch between locales, with the exception of pushing the URL change to history (`router.push` vs `router.replace`).

core/vibes/soul/primitives/navigation/index.tsx
```diff
@@ -824,26 +822,26 @@ function SearchResults({
...
+  onChange: (localeId: string) => void;
   locales: [Locale, ...Locale[]];
 }) {
-  const [lastResult, formAction] = useActionState(action, null);
   const activeLocale = locales.find((locale) => locale.id === activeLocaleId);
-
-  useEffect(() => {
-    if (lastResult?.error) console.log(lastResult.error);
-  }, [lastResult?.error]);
+  const [isPending, startTransition] = useTransition();
```
agurtovoy, line 835: `useTransition` allows us to show an affordance while the transition is in progress.

core/vibes/soul/primitives/navigation/index.tsx
```diff
@@ -824,26 +822,26 @@ function SearchResults({
...
 
   return (
     <DropdownMenu.Root>
       <DropdownMenu.Trigger
-        className={clsx('flex items-center gap-1 text-xs uppercase', navButtonClassName)}
+        className={clsx(
+          'flex items-center gap-1 text-xs uppercase transition-opacity [&:disabled]:opacity-30',
+          navButtonClassName,
+        )}
+        disabled={isPending}
```
agurtovoy, line 844: Disable the trigger during the transition.

core/vibes/soul/primitives/navigation/index.tsx
```diff
@@ -892,6 +882,7 @@ function CurrencyForm({
   action: CurrencyAction;
   currencies: [Currency, ...Currency[]];
 }) {
+  const [isPending, startTransition] = useTransition();
```
agurtovoy, line 885: Apply the "transition in progress" affordance to the currency picker as well.

core/components/header/header-section.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+import { HeaderSection as VibesHeaderSection } from '@/vibes/soul/sections/header-section';
+import { usePathname, useRouter } from '~/i18n/routing';
+
+type Props = ComponentProps<typeof VibesHeaderSection>;
+
+const useSwitchLocale = () => {
+  const pathname = usePathname();
+  const router = useRouter();
+  const params = useParams();
```
migueloller, line 14: The `usePathname` and `useParams` hook require the component to be wrapped on a `Suspense` boundary otherwise the whole page will de-opt to client-side rendering.

In my work on PPR I'm having to rework components that use these hooks so that we can avoid this and have a proper fallback state. Introducing here at this level will be a bit problematic for those efforts 🙁

migueloller (commented):
The outcome of this change—preserving the current page when changing locales—is most definitely welcome. Unfortunately, I think that we need a different solution for a couple of reasons:
- Using `usePathname` at the top-level like this will make it difficult for us to finalize our move to PPR
- This would require the header section component to be a client component. And while it is a client component at the moment, there is no such requirement from the current API so we could always change it to an RSC later

Here's a suggestion: what if we continued to use a server action but passed in the pathname in the payload so that the `href` can preserve the current URL?

agurtovoy:
> what if we continued to use a server action but passed in the pathname in the payload so that the href can preserve the current URL?

@migueloller Even ignoring the home page redirect, our server action implementation is [buggy](https://makeswifthq.slack.com/archives/C04JY4R3YSY/p1740178839985139). Can we fix it if we spent enough time on it? Probably, but we're already using a battle-tested library that offers a recommended solution for this, and I think we should be able to do just that.

core/vibes/soul/primitives/navigation/index.tsx
```diff
@@ -824,26 +820,43 @@ function SearchResults({
...
+    (locale: string) =>
+      router.push(
+        // @ts-expect-error -- TypeScript will validate that only known `params`
+        // are used in combination with a given `pathname`. Since the two will
+        // always match for the current route, we can skip runtime checks.
+        { pathname, params },
+        { locale },
+      ),
+    [pathname, params, router],
+  );
```
agurtovoy, line 838: Use `next-intl`'s [recommended way](https://next-intl.dev/docs/routing/navigation#userouter) to switch between locales, with the exception of pushing the URL change to history (`router.push` vs `router.replace`).

bookernath (approved):

PR #1991, opened 2025-02-21
fix: store the cart id in a signed JWT

chanceaclark:
## What/Why?
Stores the cart id in a signed JWT that Catalyst can verify before reading. This improves the security implications around hijacking another customers session. This also provides backwards compatibility to reading the old cart id value on the first read, but resets the cart id value to the new signed JWT value so future consumption is now verified.

### Signed JWT value:
Screenshot 2025-02-21 at 10 18 13

### Decoded JWT value:
Screenshot 2025-02-21 at 10 18 28

### Verified JWT is read and cart information populates:
Screenshot 2025-02-21 at 10 23 03

ping @bigcommerce/security

### DIFF
--- a/.changeset/metal-tools-flash.md
+++ b/.changeset/metal-tools-flash.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": minor
+---
+
+Cart ID is now written to a signed JWT to improve security around reading cart IDs. Provides backwards compatibility to reading unsigned cart cookie but signs the value for the next read.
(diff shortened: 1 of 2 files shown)


bookernath (approved):

bookernath:
💚

chanceaclark:
Synced with @bc-evan-johnson about this. If we continue with this approach we need to remove the backwards compatibility logic as a follow up PR. It really is intended to prevent current merchants who update to the new code to let the current sessions keep going. However, it still allows fetching the cart information for any given cart id which is still not secure.

However, we will probably not go with this approach and am targeting fixing this through a single session. Closing for now.

PR #2006, opened 2025-03-03
fix(core): handle registration errors in form

jorgemoya:
## What/Why?
We were incorrectly looking for registration errors in the wrong object of the response. With this change, we actually show a correct error response and short-circuit the login flow.

## Testing
Error messages display from response correctly.


jordanarldt (approved):
LGTM ✅

PR #2012, opened 2025-03-06
fix: correctly handle switching between domains

agurtovoy:
## What/Why?

WIP


chanceaclark:
@agurtovoy can we close this PR?

agurtovoy:
@chanceaclark yep, it's unproven that this is needed, closing, will come back with a fresh eye if the issue resurfaces.

PR #2014, opened 2025-03-06
refactor: replace Makeswift middleware internals

arvinpoddar:
## What/Why?
Refactors the Makeswift middleware decorator to use an experimental utility from the Makeswift runtime to create a request with draft mode cookies. This avoid the decorator leaking Makeswift internals and from Catalyst having to maintain a bespoke draft mode route.

## Testing
Tested via local production build and Vercel deployment to ensure that builder interactivity works as expected.


core/middlewares/with-makeswift.ts
```diff
@@ -1,53 +1,32 @@
...
-      const prerenderBypassValue = cookies.find((c) => c.name === '__prerender_bypass')?.value;
-
-      if (!prerenderBypassValue) {
-        return middleware(request, event);
-      }
-
-      // https://github.com/vercel/next.js/issues/52967#issuecomment-1644675602
-      // if we don't pass request twice, headers are stripped
-      const proxiedRequest = new NextRequest(request, request);
+    strict(process.env.MAKESWIFT_SITE_API_KEY, 'MAKESWIFT_SITE_API_KEY is required');
```
migueloller, line 12: This isn't a pattern that we use in the Catalyst repo for verifying environment variables. Let's try and keep it consistent with what's already there

migueloller (commented):
Changes look good. I'm a bit concerned about using `unstable_` APIs for a 1.0 release but the expectation is that we're not even going to have this middleware for 1.0, right?

Also, there seems to be some linting errors that we need to address before we can merge

core/middlewares/with-makeswift.ts
arvinpoddar: We do use `strict` in two other places for the Makeswift Site API key:

- https://github.com/bigcommerce/catalyst/blob/7dba4cc6266adb87b242b0b9ec9aac36bc20ec53/core/app/api/makeswift/%5B...makeswift%5D/route.ts#L8
- https://github.com/bigcommerce/catalyst/blob/7dba4cc6266adb87b242b0b9ec9aac36bc20ec53/core/lib/makeswift/client.ts#L10

I've gone ahead and replaced this with a top-level nullish check that throws an error for now, which is what some other scripts do for BC related env variables. I'm not really sure what the pattern for verifying environment variables is.

arvinpoddar:
> I'm a bit concerned about using unstable_ APIs for a 1.0 release but the expectation is that we're not even going to have this middleware for 1.0, right?

@migueloller Correct. As we discussed, this PR is meant to be an incremental improvement because it stops the leaking of Makeswift internals into Catalyst. Now that we've seen all the faults with proxying attempts or patching draft mode, the real solution I want to pursue is native draft mode with a toolbar.

core/middlewares/with-makeswift.ts
migueloller: Ah, I did my global search on `canary`, not `integrations/makeswift`.

For this sort of stuff I like validating environment variables are present at build time to "fail fast" rather than having the problem show up later after deploying. Tools like https://github.com/t3-oss/t3-env are nice for this sort of stuff but there's many ways this can be done.

More generally, environment variables are a way to provide configuration to the application and ultimately configuration can come from multiple places. My approach there would be to create a configuration utility and then have that read from the necessary sources, like environment variables. Then in the application, use the configuration utility.

PR #2103, opened 2025-03-14
refactor: update signin behavior to use different providers

chanceaclark:
## What/Why?
Updates the sign in behavior to use different providers instead of relying on a single provider. This is prework to in order to properly handle session syncing.

### Additional changes
Also added some folders to be linted by `next lint` and fixed them: https://github.com/bigcommerce/catalyst/pull/2103/commits/77634f9c8c734337ff731417a839cbc8c83a4c2a

## Testing
Screenshot 2025-03-14 at 17 04 15

### DIFF
--- a/.changeset/cold-poets-refuse.md
+++ b/.changeset/cold-poets-refuse.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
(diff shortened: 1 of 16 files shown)


jordanarldt (approved):
LGTM

jorgemoya (approved):

PR #2108, opened 2025-03-18
feat: add session-sync endpoint

chanceaclark:
## What/Why?
Adds a session sync endpoint, used for when needing to sync the session between a "3rd party" session and your Catalyst session. In the case of this endpoint, we are primarily prepping for syncing across Catalyst <-> Stencil when navigating to the checkout. We still need some work to be done on the Stencil checkout piece, but this should enable the team to sync the session back and forth.

### Logging in
https://github.com/user-attachments/assets/c506aaae-31f6-4f90-85c7-45890ebe75c2

### Logging out
https://github.com/user-attachments/assets/f307ba23-0b87-4bc3-9517-7dfa08cccf93


core/auth/index.ts
```diff
@@ -148,6 +154,10 @@ async function loginWithJwt(credentials: unknown): Promise<User | null> {
   };
 }
 
+function loginWithSessionSync(credentials: unknown): User {
+  return SessionSyncCredentials.parse(credentials);
+}
```
chanceaclark, line 159: This is a little awkward... Usually you should handle the validation of the sign in here but because we are going to move in a direction where there will alway be a session, this `CredentialsProvider` will actually be removed and we will just update the session accordingly. We also need to handle things like the redirect and any API validation errors so it made more sense to do the validation in the route handler, rather than in the `signIn` function.

core/app/api/session-sync/[jwt]/route.ts
```diff
@@ -0,0 +1,102 @@
...
+      email: content.customer.email,
+      customerAccessToken: content.customerAccessToken.value,
+      redirect: false,
+    });
+  } else {
+    await signOut({ redirect: false });
+  }
+
+  // Note: redirectTo is going to include the full url, not a partial path
+  return NextResponse.redirect(content.redirectTo);
```
Comment (willPrattUPL) [inline core/app/api/session-sync/[jwt]/route.ts:98]: Do I have this right:

this redirect url here will basically be the same/ do the same thing as this?

image

So that this API endpoint acts as a middleware between the redirect to checkout -> managed checkout redirect?

Comment (chanceaclark) [reply core/app/api/session-sync/[jwt]/route.ts:98]: Yeah that `redirectedCheckoutUrl` uses a session-sync endpoint on the Stencil side of things, which is similar to what this endpoint does.

The big thing about this approach is it allows sessions to be sync'd when we are scoping our session cookies with the `_Host` prefix as they are scoped to subdomains. For example a cookie wouldn't be able to be reused from `example.com` (`www.` subdomain) to `checkout.example.com`.

We need to follow this general approach:
1. Redirect to endpoint on another/same domain, e.g. `example.com` -> `checkout.example.com`.
2. Response contains the updated session via the `Set-Cookie` header + `Location` header + 3xx status.
3. Browser consumes the `Location` header and redirect to the correct end destination (via `redirectTo`).

core/app/api/session-sync/[jwt]/route.ts
```diff
@@ -0,0 +1,117 @@
...
+      status: 204,
+      headers: {
+        'Access-Control-Allow-Origin': 'https://session-sync-example.vercel.app',
+        'Access-Control-Allow-Credentials': 'true',
+      },
+    });
+  }
+
+  // Note: redirectTo is going to include the full url, not a partial path
+  return NextResponse.redirect(content.redirectTo);
```
Comment (davidchin) [inline core/app/api/session-sync/[jwt]/route.ts:113]: We need to validate `redirectTo`, making sure it is an expected value. Otherwise, we would be vulnerable to [open redirect attack](https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html#dangerous-url-redirects) (e.g.: phishing). We could validate the host to make sure it matches the host of BC checkout of the same channel.

core/components/session-sync/provider.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export const SessionSyncProvider = () => {
+  const handleBrowserBackButtonEvent = useCallback((event: PageTransitionEvent) => {
+    if (event.persisted) {
+      window.location.reload();
+    }
+  }, []);
+
+  useEffect(() => {
+    window.addEventListener('pageshow', handleBrowserBackButtonEvent);
```
davidchin, line 13: Does the force reload happen on all pages when the back button is hit even if no syncing is needed? Or does it happen only when the user clicks on the back button coming from BC checkout?

willPrattUPL: Could we set a flag like "redirected_to_checkout" when users click the checkout button?

Then session sync provider deletes this cookie when it successfully reloads once.

core/app/api/session-sync/[jwt]/route.ts
Comment (chanceaclark) [reply core/app/api/session-sync/[jwt]/route.ts:113]: I'll add some validation here – I did some testing with some bad urls and I think we have some protection already as BigCommerce prepends the merchant-provided url with either the storefront or checkout domain. The only use-case I would be concerned about is if a merchant has the domain `https://foo.co` and bad actor has `https://foo.com`, they can set the `redirectTo` field to `m/bad-path` and it'll be redirected to `https://foo.com/bad-path`.

core/components/session-sync/provider.tsx
chanceaclark: I'll need to do some digging into here – I bet I can check the referrer before doing this.

core/app/api/session-sync/[jwt]/route.ts
Comment (davidchin) [reply core/app/api/session-sync/[jwt]/route.ts:113]: Thanks for checking, much appreciated! There's also the possibility that an attacker could construct the URL themselves (instead of BC) and trick users into navigating to it as part of a phishing attack. Therefore, even if we have existing protections on the BC side for storing the checkout domain, it's still important to have validation in place for the session-sync endpoint on the Catalyst side.

core/components/session-sync/provider.tsx
chanceaclark: Dug into this a bit and yeah `strict-origin-when-cross-origin` is mucking the waters here. And we probably can't add a cookie to this as we'll need some sort of validation that it was the other origin who set it - I could see a security issue where a bad actor could set the cookie and trigger infinite reloads, exhausting our resources in the process.

@bookernath from engineering perspective this might more effort than it's worth - just letting users reload the page and the session will update it. I think this is the best bet :(

PR #2110, opened 2025-03-18
feat: Sync Button and DynamicForm with VIBES

jordanarldt:
## What/Why?
Sync `Button` and `DynamicForm` component with VIBES
https://github.com/makeswift/vibes/pull/606

## Testing
Tested locally and build passes


core/vibes/soul/primitives/button/index.tsx
```diff
@@ -101,16 +107,21 @@ export function Button({
           variant === 'secondary' && 'mix-blend-difference',
         )}
       >
+        {srOnlyText !== '' && <span className="sr-only">{srOnlyText}</span>}
```
chanceaclark, line 110: Shouldn't we use `aria-label` instead? 🤔

jordanarldt: @chanceaclark This part in particular was just pulled from VIBES, so I'm not 100% certain of the intent. I think this prop is only for additional screen-reader text if the provided children aren't enough? You can always provide `aria-label` explicitly to the button, and the `children` prop will also work on a screen-reader if you provide it with a string.

cc @hunterbecton

chanceaclark: Yeah I am questioning the intent here, because if you need to provide extra context for screen readers, it's better to provide that via `aria-label` rather than adding `sr-only` text prefixed. It's also more flexible what you provide to screen readers using `aria-label`, e.g. you need to add postfix text to children rather than prefixed. Code-wise it's a little more verbose as you might need to duplicate what you are displaying via children, but it's the preferred way. Unless I am missing something here.

chanceaclark: If we have a delete button, I'll only be able to prefix the button:
```tsx
<Button srOnlyText="Product">Delete</Button> // Screen reader will read "Product Delete"
 ```

If I try to fix this by expanding out the text:
```tsx
<Button srOnlyText="Delete product">Delete</Button> // Screen reader will read "Delete product Delete"
```

If you use `aria-label`, it's solved:
```tsx
<Button aria-label="Delete product">Delete</Button> // Screen reader will read "Delete product"
```

In the case we need to use a button for navigation we can add additional context, while visual users can use surrounding context:
```tsx
<Button aria-label="Navigate to product creation page">Add</Button> // Screen reader will read "Navigate to product creation page"
```

In the case of duplicating what we are displaying via children, it's more around being verbose for example when we need to translate it we need an additional message:
```tsx
<Button aria-label={t('deleteProductButtonAria', product.name)}>{t('deleteProductButton')}</Button>
```

>Seems like could use aria-label. I wonder if that should be applied to the <button> directly or if it's ok for it to be a nested <span>?

It looks like it's already applied to the `<button>` via `{...props}`, so we should be good there. We should just remove this prop altogether as `aria-label` is better for accessibility.

jordanarldt: @chanceaclark Okay cool, yeah that's what I was going to say - you can just pass `aria-label` to the button directly like `<Button aria-label="myLabel">...</Button>` and the `srOnlyText` is just an additional prop if someone needs it.

If we want to remove `srOnlyText`, can we make a separate VIBES ticket for this in linear so that we can push this PR along?

jamesqquick: Ok, so the proposed solution would be to remove the screen reader only text and pass the built-in `aria-label` prop?

chanceaclark: `srOnlyText` is useless if we just use `aria-label`, so yeah just remove the prop and rely on `aria-label`.

chanceaclark (approved):

PR #2111, opened 2025-03-19
feat: add anonymous session with cart tracking

chanceaclark:
## What/Why?
This PR solves a two-fold:
1. Creates an anonymous session whenever a user navigates to a Catalyst page. This will allow us to store a bunch of information we need for shoppers and creates a consistent place to read customer-specific data from.
2. Moves reading/writing the cart id into the session. Regardless of if the session is an anonymous session or a logged-in shopper, the cart id will be tied to the session and can move freely between the two.

This is a better pattern to practice for security reasons. It makes it so attackers can't access the shoppers cart id without being validated by the Catalyst server.

### Anonymous session is created:
Screenshot 2025-03-19 at 16 50 20

### Cart is created and added onto the session:
https://github.com/user-attachments/assets/0e53a570-6d6f-4449-be0b-c1b153219d8a

### Cart is transferred over to logged-in session:
https://github.com/user-attachments/assets/f6e94ed4-51df-4fdf-a6b4-98600975362d

### Cart is cleared on session when logging out:
https://github.com/user-attachments/assets/71ab0f12-24e2-4d02-a903-6c10341a86c2

### DIFF
--- a/core/app/[locale]/(default)/(auth)/layout.tsx
+++ b/core/app/[locale]/(default)/(auth)/layout.tsx
@@ -1,6 +1,6 @@
 import { PropsWithChildren } from 'react';
 
-import { auth } from '~/auth';
+import { isLoggedIn } from '~/auth';
 import { redirect } from '~/i18n/routing';
 
 interface Props extends PropsWithChildren {
@@ -8,10 +8,10 @@ interface Props extends PropsWithChildren {
 }
 
 export default async function Layout({ children, params }: Props) {
-  const session = await auth();
+  const loggedIn = await isLoggedIn();
   const { locale } = await params;
 
-  if (session) {
+  if (loggedIn) {
     redirect({ href: '/account/orders', locale });
   }
 
--- a/.changeset/ten-pots-vanish.md
+++ b/.changeset/ten-pots-vanish.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": minor
+---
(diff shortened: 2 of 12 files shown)


jorgemoya (approved):

core/app/[locale]/(default)/(auth)/layout.tsx
```diff
@@ -1,17 +1,17 @@
...
   params: Promise<{ locale: string }>;
 }
 
 export default async function Layout({ children, params }: Props) {
-  const session = await auth();
+  const loggedIn = await isLoggedIn();
   const { locale } = await params;
 
-  if (session) {
+  if (loggedIn) {
```
Comment (migueloller) [inline core/app/[locale]/(default)/(auth)/layout.tsx:14]: I know this isn't really in scope for this PR, but I wanted to mention that Next.js docs recommend against authorization in layouts: https://nextjs.org/docs/app/building-your-application/authentication#layouts-and-auth-checks

migueloller (approved):
Nice find on the `router.refresh()` 👍🏻

core/app/[locale]/(default)/(auth)/layout.tsx
Comment (chanceaclark) [reply core/app/[locale]/(default)/(auth)/layout.tsx:14]: Yeah I've chatted with @jorgemoya about this, but I want us to move it back into middleware (`with-auth.ts`). It has already been causing some issues 😅

bc-evan-johnson (approved):

PR #2113, opened 2025-03-20
[WIP/RFC] Switch to node.js runtime for middleware to use Data Cache in with-routes; eliminate KV adapters

bookernath:
## What/Why?
Catalyst has KV adapters solely to improve the performance of the with-routes middlware, because Next.js' default Data Cache capabilites were unavailable in Middleware on Edge Runtime, and we wanted to ensure the data used by this middleware was cached and available very fast.

The new [Node.js Runtime support for Middleware](https://vercel.com/changelog/middleware-now-supports-node-js) offers the ability to use Data Cache in middleware, so for Catalyst customers this means that moving to Node.js Middleware and using native Data Cache provides these benefits:

- Fewer infrastructure dependencies --> easier setup, less cost
- Better performance
- Consistent caching policies & revalidation strategy

I'm putting this PR up for feedback, not for immediate merging. This feature is still experimental and we should allow some time for it to bake before committing to it.

## Testing
Middleware uses Data Cache:
image


core/middlewares/with-routes.ts
```diff
@@ -94,7 +90,7 @@ const getRawWebPageContent = async (id: string) => {
 };
 
 const GetStoreStatusQuery = graphql(`
-  query getStoreStatus {
+  query GetStoreStatus {
```
bookernath, line 93: fixing inconsistent casing

PR #2116, opened 2025-03-20
fix: flicker on rerendering `<ProductList />`

agurtovoy:
## What/Why?

Fix incidental rendering of `<ProductList />` elements, most visible when the page is edited in the builder.

### Before

https://github.com/user-attachments/assets/d4087405-12c9-4c37-9bd1-fefb807df35c

### After

https://github.com/user-attachments/assets/dc687f9b-0b94-4de0-81d4-e748928f9098

### DIFF
--- a/core/lib/makeswift/utils/use-products.ts
+++ b/core/lib/makeswift/utils/use-products.ts
@@ -1,3 +1,4 @@
+import { useMemo } from 'react';
 import useSWR from 'swr';
 import { z } from 'zod';
 
@@ -40,14 +41,20 @@ export function useProducts({ collection, collectionLimit = 20, additionalProduc
     fetcher,
   );
 
-  const combinedProducts = [
-    ...(collectionData?.products.slice(0, collectionLimit) ?? []),
-    ...(additionalData?.products ?? []),
-  ];
+  const combinedProducts = useMemo(
+    () => [
+      ...(collectionData?.products.slice(0, collectionLimit) ?? []),
+      ...(additionalData?.products ?? []),
+    ],
+    [collectionData, additionalData, collectionLimit],
+  );
 
   const isLoading = isCollectionLoading || isAdditionalLoading;
 
-  const products = isLoading ? null : combinedProducts.map(bcProductToVibesProduct);
+  const products = useMemo(
+    () => (isLoading ? null : combinedProducts.map(bcProductToVibesProduct)),
+    [isLoading, combinedProducts, bcProductToVibesProduct],
+  );
 
   return { products, isLoading };
 }
--- a/core/vibes/soul/lib/streamable.tsx
+++ b/core/vibes/soul/lib/streamable.tsx
@@ -53,6 +53,10 @@ function weakRefCache<K, T extends object>() {
 
 const promiseCache = weakRefCache<string, Promise<unknown>>();
 
+function isPromise<T>(value: Streamable<T>): value is Promise<T> {
+  return value instanceof Promise;
+}
+
 // eslint-disable-next-line valid-jsdoc
 /**
  * A suspense-friendly upgrade to `Promise.all`, guarantees stability of
@@ -61,6 +65,13 @@ const promiseCache = weakRefCache<string, Promise<unknown>>();
 function all<T extends readonly unknown[] | []>(
   streamables: T,
 ): Streamable<{ -readonly [P in keyof T]: Awaited<T[P]> }> {
+  // Avoid creating an unnecessary promise with the `Promise.all` call below
+  // if none of the streamables is a promise
+  if (!streamables.some(isPromise)) {
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+    return streamables as { -readonly [P in keyof T]: Awaited<T[P]> };
+  }
+
   const cacheKey = getCompositeKey(streamables);
 
   const cached = promiseCache.get(cacheKey);
@@ -80,7 +91,7 @@ export const Streamable = {
 };
 
 export function useStreamable<T>(streamable: Streamable<T>): T {
-  return streamable instanceof Promise ? use(streamable) : streamable;
+  return isPromise(streamable) ? use(streamable) : streamable;
 }
 
 function UseStreamable<T>({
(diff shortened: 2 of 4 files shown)


core/lib/makeswift/utils/use-products.ts
```diff
@@ -40,14 +41,20 @@ export function useProducts({ collection, collectionLimit = 20, additionalProduc
...
+    [collectionData, additionalData, collectionLimit],
+  );
 
   const isLoading = isCollectionLoading || isAdditionalLoading;
 
-  const products = isLoading ? null : combinedProducts.map(bcProductToVibesProduct);
+  const products = useMemo(
+    () => (isLoading ? null : combinedProducts.map(bcProductToVibesProduct)),
+    [isLoading, combinedProducts, bcProductToVibesProduct],
+  );
```
agurtovoy, line 57: Essentially, we were returning a new array of "combined" products every time, triggering downstream re-renders.

core/vibes/soul/lib/streamable.tsx
```diff
@@ -61,6 +65,12 @@ const promiseCache = weakRefCache<string, Promise<unknown>>();
 function all<T extends readonly unknown[] | []>(
   streamables: T,
 ): Streamable<{ -readonly [P in keyof T]: Awaited<T[P]> }> {
+  // Avoid creating an unnecessary promise if all values are already resolved
+  if (streamables.every(isResolved)) {
```
agurtovoy, line 69: If the `streamables` argument is stable, `Streamable.all` will correctly cache the (extra) promise below. However, when `streamables` doesn't have a stable identity, but contains only resolved promises, wrapping it with `Streamable.all` exacerbates incidental rerenders by adding a `Suspense`-induced flicker that wouldn't be otherwise present.

One can argue that this is a good thing, as it makes an existing performance issue much more visible, but, all things considering, I think we're better off fixing it.

agurtovoy: I'll follow up with a corresponding PR for this in VIBES.

core/vibes/soul/lib/streamable.tsx
```diff
@@ -53,6 +53,10 @@ function weakRefCache<K, T extends object>() {
 
 const promiseCache = weakRefCache<string, Promise<unknown>>();
 
+function isResolved<T>(value: Streamable<T>): value is T {
+  return !(value instanceof Promise);
+}
```
migueloller, line 58: The naming here is a bit confusing IMO given that promises have a "resolved" state. I first expected us to be checking that the promise was resolved as opposed to checking if something was a promise or not. Perhaps we can name this `isPromise` although honestly, this is the sort of function that I deem unnecessary when the alternative, `value instanceof Promise` vs `isPromise(value)` is the same in terms of mental overhead but without the additional abstraction/function that you have to go read to learn how it works

core/lib/makeswift/utils/use-products.ts
migueloller: Do we know why downstream re-renders were resulting in the loading state being displayed? I would've thought that only happens when the component re-mounts. Were re-renders resulting in re-mounts?

migueloller (approved):
Looks good, although this feels like a bandaid to me and the real underlying problem is in re-renders causing the flicker at all. `useMemo` and `useCallback` and stable values should really just be there as performance optimizations and shouldn't impact rendering like this

core/lib/makeswift/utils/use-products.ts
agurtovoy: > Do we know why downstream re-renders were resulting in the loading state being displayed? I would've thought that only happens when the component re-mounts. Were re-renders resulting in re-mounts?

Yes, `Streamable.all` was constructing a new promise for each new value. Either of these two fixes (making the product list stable or eliminating an unnecessary promise) eliminates the flicker. I chose to fix both.

agurtovoy:
@migueloller This PR contains two fixes, one for the unnecessary rerenders and one for the unnecessary promise that was causing the flicker in the presence of the unnecessary rerenders. Let me know if [this comment](https://github.com/bigcommerce/catalyst/pull/2116/files#r2006455817) makes it more clear what's happening here.

core/vibes/soul/lib/streamable.tsx
agurtovoy: I didn't want to spread `value instanceof Promise` through the implementation, considering that at some point we weren't even sure that `value instanceof Promise` works in all cases.

agurtovoy: > although honestly, this is the sort of function that I deem unnecessary when the alternative, value instanceof Promise vs isPromise(value) is the same in terms of mental overhead but without the additional abstraction/function that you have to go read to learn how it works

`isPromise(value)` tells you exactly what it does. `value instanceof Promise` is an implementation detail that (a) needs to be deconstructed every time its read to assign it the same meaning and (a) might change in the future. The distinction is subtle but non-negligible.

core/lib/makeswift/utils/use-products.ts
migueloller: Ah, makes sense. I would hope the pending state showing was a result of a new promise being passed in such that the stable values aren't necessary. Often adding `useMemo` and `useCallback` adds more overhead than it saves in terms of work on re-renders. Thanks for clarifying 🙏🏻

core/vibes/soul/lib/streamable.tsx
migueloller: It seems that my latter comments on my opinions on `isPromise` distracted from my main point: I think that `isResolved` is a confusing name and something like `isPromise`, which is more explicit, would be a tad better. In particular, when I read `isResolved` I'm assuming it takes in a promise and returns `true` if the promise has already resolved. Instead, what it does is check that something is not a promise. `!isPromise(value)` would be less confusing in that scenario, IMO.

agurtovoy: I agree, `isResolved` was a misleading name, I renamed the function to `isPromise`, sorry for not making it clear in the first place.

migueloller: No worries! Thanks for fixing 🙏🏻

core/vibes/soul/lib/streamable.tsx
agurtovoy: https://github.com/makeswift/vibes/pull/617

PR #2119, opened 2025-03-20
feat: Add wishlist translations, transformers, and core components

jordanarldt:
## What/Why?
- Add wishlist translations
- Add data transformers for wishlists
- Add `Modal` component wrapper to support wishlist functionality

## Testing
Tested locally


core/lib/user-agent.ts
chanceaclark, line 1: Are these functions actually being used?

chanceaclark (approved):

PR #2122, opened 2025-03-20
feat: Add wishlist pages

jordanarldt:
## What/Why?
Add wishlist pages:
- `/account/wishlists`
- `/account/wishlists/[id]`
- `/wishlist/[token]`

## Testing
Tested locally

image

image
image
image
image
image
image

https://github.com/user-attachments/assets/1a75a99d-566c-459b-86ce-96b2e5b63563

### DIFF
--- a/core/app/[locale]/(default)/account/wishlists/[id]/page.tsx
+++ b/core/app/[locale]/(default)/account/wishlists/[id]/page.tsx
@@ -1,0 +1,107 @@
+import { notFound } from 'next/navigation';
+import { getFormatter, getTranslations } from 'next-intl/server';
+import { SearchParams } from 'nuqs';
+import { createSearchParamsCache, parseAsInteger, parseAsString } from 'nuqs/server';
+
+import { CursorPaginationInfo } from '@/vibes/soul/primitives/cursor-pagination';
+import { Wishlist, WishlistDetails } from '@/vibes/soul/sections/wishlist-details';
+import { ExistingResultType } from '~/client/util';
+import { defaultPageInfo, pageInfoTransformer } from '~/data-transformers/page-info-transformer';
+import { wishlistDetailsTransformer } from '~/data-transformers/wishlists-transformer';
+import { isMobileUser } from '~/lib/user-agent';
+
+import { removeWishlistItem } from '../_actions/remove-wishlist-item';
+import { getDeleteWishlistModal, getRenameWishlistModal } from '../modals';
+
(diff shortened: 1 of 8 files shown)


core/app/[locale]/(default)/account/wishlists/[id]/page.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const entityId = Number(id);
+  const searchParamsParsed = searchParamsCache.parse(await searchParamsPromise);
+  const formatter = await getFormatter();
+  const wishlist = await getCustomerWishlist({ entityId, ...searchParamsParsed });
+
+  if (!wishlist) {
+    return notFound();
+  }
+
+  return wishlistDetailsTransformer(wishlist, t, pt, formatter);
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/account/wishlists/[id]/page.tsx:48]: I'm not sure how big of a fan I am of passing the translations getters to the transformer. Is there a big performance benefit over just calling these in the transformer directly?

jorgemoya (commented):
Looking good, just a couple of questions and translations and caching!

core/app/[locale]/(default)/account/wishlists/[id]/page.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const t = await getTranslations('Account.Wishlists');
+  const pt = await getTranslations('Product.ProductDetails');
+  const wishlistActions = (wishlist?: Wishlist) => {
+    if (!wishlist) {
+      return <WishlistActionsSkeleton />;
+    }
+
+    return (
+      <WishlistActions
+        isMobileUser={isMobileUser()}
```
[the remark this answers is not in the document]
Comment (jordanarldt) [reply core/app/[locale]/(default)/account/wishlists/[id]/page.tsx:73]: @jorgemoya We need this for the wishlist `Share` button. If a user is on mobile, we use the native share UI. If you're on desktop, we use the modal popup where you can copy to clipboard.

core/app/[locale]/(default)/account/wishlists/[id]/page.tsx
Comment (jordanarldt) [reply core/app/[locale]/(default)/account/wishlists/[id]/page.tsx:48]: @jorgemoya This is a similar pattern we follow for the Orders transformer. Since we've already fetched the translations in the page, I figured it made sense to just pass them down instead of turning the transformer into an `async` function.

Since transformers' sole purpose is to just convert model A into model B, I personally feel like they shouldn't be async at all 💭  We can discuss this more though!

jorgemoya (approved):

PR #2126, opened 2025-03-21
refactor(core): abstract cart creation and addition logic

chanceaclark:
## What/Why?
Abstracts the cart creation and addition logic to handle some shared logic across add to cart server actions.

## Testing

https://github.com/user-attachments/assets/dff8b2c8-5897-433a-933a-d3226edcdba6

https://github.com/user-attachments/assets/0f81bad0-60e5-4aed-a637-34c99a0a74a8


core/lib/cart/index.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export async function clearCartId(): Promise<void> {
+  const cookieStore = await cookies();
+
+  cookieStore.delete('cartId');
+}
+
+export const addToOrCreateCart = async (
+  data: CreateCartInput | AddCartLineItemsInput['data'],
+): Promise<void> => {
```
jordanarldt, line 39:
```suggestion
export async function addToOrCreateCart(
  data: CreateCartInput | AddCartLineItemsInput['data'],
): Promise<void> {
```

Can we use an `async function` here instead of a const to match the style of the rest of the file?

jordanarldt (approved):

jordanarldt (approved):

PR #2146, opened 2025-03-25
Add slots to Login and Cart

apledger:
## What/Why?
Added slots to Login and Cart.

### Cart
https://github.com/user-attachments/assets/d3e9fcbb-fb55-4d83-bc35-62a8dc15dc97

### Login
https://github.com/user-attachments/assets/7c53a2a9-2411-4a9b-8f41-4952804af53f

### DIFF
--- a/core/app/[locale]/(default)/(auth)/login/page.tsx
+++ b/core/app/[locale]/(default)/(auth)/login/page.tsx
@@ -4,6 +4,7 @@ import { getTranslations } from 'next-intl/server';
 import { ButtonLink } from '@/vibes/soul/primitives/button-link';
 import { SignInSection } from '@/vibes/soul/sections/sign-in-section';
 import { ForceRefresh } from '~/components/force-refresh';
+import { Slot } from '~/lib/makeswift/slot';
 
 import { login } from './_actions/login';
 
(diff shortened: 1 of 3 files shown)


core/app/[locale]/(default)/(auth)/login/page.tsx
```diff
@@ -28,20 +36,26 @@ export default async function Login() {
...
+                <li>{t('CreateAccount.ordersHistory')}</li>
+                <li>{t('CreateAccount.ordersTracking')}</li>
+                <li>{t('CreateAccount.wishlists')}</li>
+              </ul>
+              <ButtonLink href="/register" variant="secondary">
+                {t('CreateAccount.createLink')}
+              </ButtonLink>
+            </div>
+          }
+          label="Login sidebar content"
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/(auth)/login/page.tsx:56]: Do we need to translate this `label`? Same with the cart page?

core/app/[locale]/(default)/cart/page.tsx
```diff
@@ -19,19 +22,36 @@ export async function generateMetadata(): Promise<Metadata> {
...
-    return (
+  const contentSnapshotTop = await client.getComponentSnapshot('cart-top-content', {
+    siteVersion: await getSiteVersion(),
+    locale,
+  });
+
+  const contentSnapshotBottom = await client.getComponentSnapshot('cart-bottom-content', {
+    siteVersion: await getSiteVersion(),
+    locale,
+  });
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/cart/page.tsx:39]: Did we want to use the streamable pattern here? Maybe @migueloller can provide some insight here.

jordanarldt (approved):
LGTM

core/app/[locale]/(default)/(auth)/login/page.tsx
```diff
@@ -15,8 +18,13 @@ export async function generateMetadata(): Promise<Metadata> {
   };
 }
 
-export default async function Login() {
+export default async function Login(props: { params: Promise<{ locale: string }> }) {
+  const { locale } = await props.params;
   const t = await getTranslations('Login');
+  const sidebarContent = await client.getComponentSnapshot('login-sidebar-content', {
+    siteVersion: await getSiteVersion(),
```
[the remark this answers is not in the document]
Comment (apledger) [reply core/app/[locale]/(default)/(auth)/login/page.tsx:25]: Ah nice! I'll update to use that.

core/app/[locale]/(default)/(auth)/login/page.tsx
Comment (apledger) [reply core/app/[locale]/(default)/(auth)/login/page.tsx:56]: Hm, it's mainly for the publish dialog and builder. I don't think we're translating these labels anywhere else.

chanceaclark (approved):

core/app/[locale]/(default)/cart/page.tsx
Comment (migueloller) [reply core/app/[locale]/(default)/cart/page.tsx]: Yeah. I think that Catalyst right now overuses `Streamable` @chancellor so it's a valid question to ask. I'm hoping that once we're done with the refactor it will be used strategically for specific things. For example, price in the PDP since that is dynamic based on the currency cookie

PR #2182, opened 2025-04-01
feat(core): add the ability to redirect from login

chanceaclark:
## What/Why?
Add the ability to redirect from the login page. Developers can now append a relative path to the `?redirectTo=` query param on the `/login` page. When a shopper successfully logs in, it'll redirect them to the given relative path. Defaults to `/account/orders` to prevent a breaking change.

This functionality will be critical for redirecting a user back to checkout to ensure the session is synced between headless and checkout (stencil).

## Testing

https://github.com/user-attachments/assets/06586442-0d67-465c-9398-27574546851e

### DIFF
--- a/core/app/[locale]/(default)/(auth)/login/page.tsx
+++ b/core/app/[locale]/(default)/(auth)/login/page.tsx
@@ -1,14 +1,19 @@
+/* eslint-disable react/jsx-no-bind */
 import { Metadata } from 'next';
 import { getTranslations, setRequestLocale } from 'next-intl/server';
 
 import { ButtonLink } from '@/vibes/soul/primitives/button-link';
 import { SignInSection } from '@/vibes/soul/sections/sign-in-section';
+import { buildConfig } from '~/build-config/reader';
 import { ForceRefresh } from '~/components/force-refresh';
 
 import { login } from './_actions/login';
 
 interface Props {
   params: Promise<{ locale: string }>;
+  searchParams: Promise<{
+    redirectTo?: string;
+  }>;
 }
 
 export async function generateMetadata({ params }: Props): Promise<Metadata> {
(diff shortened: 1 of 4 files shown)


core/app/[locale]/(default)/(auth)/login/page.tsx
```diff
@@ -15,14 +23,18 @@ export async function generateMetadata(): Promise<Metadata> {
   };
 }
 
-export default async function Login() {
+export default async function Login({ searchParams }: Props) {
+  const { redirectTo = '/account/orders' } = await searchParams;
   const t = await getTranslations('Login');
 
+  const vanityUrl = buildConfig.get('urls').vanityUrl;
+  const redirectToPathname = new URL(redirectTo, vanityUrl).pathname;
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/(auth)/login/page.tsx:31]: @bc-donfran pinging you here to show that we are only accepting relative paths from a security standpoint. Can you see if there is anything missing from this validation?

core/app/[locale]/(default)/(auth)/login/page.tsx
```diff
@@ -15,14 +23,18 @@ export async function generateMetadata(): Promise<Metadata> {
...
+
   return (
     <>
       <ForceRefresh />
       <SignInSection
-        action={login}
+        action={login.bind(null, { redirectTo: redirectToPathname })}
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/(auth)/login/page.tsx:37]: This is a similar pattern to what we do in the `/change-password` page.

core/app/[locale]/(default)/checkout/route.ts
Comment (chanceaclark) [inline core/app/[locale]/(default)/checkout/route.ts:1]: Felt it was better to redirect to an empty cart page instead of showing a blank screen with JSON data, hence the changes here.

core/app/[locale]/(default)/(auth)/login/page.tsx
Comment (bc-donfran) [reply core/app/[locale]/(default)/(auth)/login/page.tsx:31]: This overall should be safe with just relative paths

bc-donfran (approved):

core/app/[locale]/(default)/(auth)/login/page.tsx
```diff
@@ -15,14 +23,18 @@ export async function generateMetadata(): Promise<Metadata> {
   };
 }
 
-export default async function Login() {
+export default async function Login({ searchParams }: Props) {
+  const { redirectTo = '/account/orders' } = await searchParams;
```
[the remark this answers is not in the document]
Comment (chanceaclark) [reply core/app/[locale]/(default)/(auth)/login/page.tsx:27]: Yep, and I think that's fine if we want to support redirecting.

jorgemoya (approved):

PR #2184, opened 2025-04-02
chore: use `params.locale` in all routes

migueloller:
## What/Why?

- `setRequestLocale` in all pages and layouts
- `locale` parameter for `getTranslations` in all `generateMetadata`

These changes are in preparation for enabling PPR and `dynamicIO` for all pages. We want to maximize the amount of HTML that is present in the static shell and minimize how much we're streaming in dynamically.

## Testing

TODO

### DIFF
--- a/core/app/[locale]/(default)/(auth)/change-password/page.tsx
+++ b/core/app/[locale]/(default)/(auth)/change-password/page.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable react/jsx-no-bind */
 import { Metadata } from 'next';
-import { getLocale, getTranslations } from 'next-intl/server';
+import { getTranslations, setRequestLocale } from 'next-intl/server';
 
 import { ResetPasswordSection } from '@/vibes/soul/sections/reset-password-section';
 import { redirect } from '~/i18n/routing';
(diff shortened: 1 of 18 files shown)


core/app/[locale]/(default)/compare/page.tsx
```diff
@@ -57,21 +57,28 @@ const getProducts = cache(async (productIds: number[] = []): Promise<CompareCard
...
 });
 
-export async function generateMetadata(): Promise<Metadata> {
-  const t = await getTranslations('Compare');
-
-  return {
-    title: t('title'),
-  };
-}
-
```
Comment (migueloller) [inline core/app/[locale]/(default)/compare/page.tsx:67]: This diff is a bit weird because I moved the `Props` interface to be before `generateMetadata` to be consistent with other files. GitHub thinks the entire `generateMetadata` function was remove, but it's down below with the `params.locale` change that has been present in other files in this PR.

core/app/[locale]/(default)/(auth)/change-password/page.tsx
```diff
@@ -1,31 +1,37 @@
...
 
-export default async function ChangePassword({ searchParams }: Props) {
+export default async function ChangePassword({ params, searchParams }: Props) {
+  const { locale } = await params;
+
+  setRequestLocale(locale);
+
   const { c: customerEntityId, t: token } = await searchParams;
   const t = await getTranslations('ChangePassword');
-  const locale = await getLocale();
```
Comment (migueloller) [inline core/app/[locale]/(default)/(auth)/change-password/page.tsx:28]: This was a weird case where we were using `getLocale` instead of `params.locale`. I looked at the history and it seems that @jorgemoya you added `useLocale` a while ago before this was an `async` component and then @chanceaclark you changed it from `useLocale` to `getLocale` after moving to Next.js v15 and having to turn it into an `async` function due to having to await `props.searchParams`.

Please let me know if there's a reason we should continue to use `getLocale` over `params.locale` 🙏🏻

jorgemoya (approved):

PR #2185, opened 2025-04-02
refactor: /change-password for dynamicIO

migueloller:
## What/Why?

This is a WIP to try and understand what it would take to adopt `dynamicIO`. In this particular case, we want the page to be dynamic since it's fundamental nature is to rely on URL search params. There's some weird ergonomics due to how `dynamicIO` works, though:
- Had to create a `loading.tsx` that returns `null` to have a Suspense boundary since the component accesses `searchParams`. Using `use cache` isn't an option since we are accessing a dynamic API (`searchParams`).
- The redirect done in the component isn't reflected in the HTTP status code because it's handled client-side due to streaming.
  - Ideally we'd be able to disable dynamic IO for this page and do vanilla dynamic rendering. Unfortunately we can't use the [`dynamic`](https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#dynamic) route segment config because it's not compatible with `dynamicIO`.
  - We could move the redirect to middleware but splitting up the business logic like this feels icky. Furthermore, we'd still need to validate the search params are present and handle that accordingly either by rendering an error or throwing one and providing an `error.tsx` (the former is probably more appropriate so we can render something specific—`error.tsx` is more intended for _unexpected_ errors since it masks error details).

## Testing

This is still WIP, so testing is TODO.


core/app/[locale]/(default)/(auth)/change-password/loading.tsx
```diff
@@ -0,0 +1,3 @@
+export default function Loading() {
+  return null;
+}
```
Comment (migueloller) [inline core/app/[locale]/(default)/(auth)/change-password/loading.tsx:3]: Not great. I'd prefer if the server simply blocked while the dynamic page is rendered instead of immediately returning an empty loading state followed by streaming in the dynamic page later.

core/app/[locale]/(default)/(auth)/change-password/page.tsx
```diff
@@ -22,17 +25,18 @@ export async function generateMetadata(): Promise<Metadata> {
...
-  const { c: customerEntityId, t: token } = await searchParams;
+export default async function ChangePassword({ params, searchParams }: Props) {
+  const { locale } = await params;
+  const { c: customerEntityId, t: token } = await loadSearchParams(searchParams);
   const t = await getTranslations('ChangePassword');
-  const locale = await getLocale();
 
   if (!customerEntityId || !token) {
     return redirect({ href: '/login', locale });
   }
```
Comment (migueloller) [inline core/app/[locale]/(default)/(auth)/change-password/page.tsx:35]: This could be moved to middleware but then we'd be splitting business logic in two places, which is not great.

PR #2186, opened 2025-04-02
feat(core): add the ability to redirect after logout

chanceaclark:
## What/Why?
Enables developers to redirect to a specific pathname after a user logs out. This enables us to session sync back to checkout when a user logs out.

## Testing

https://github.com/user-attachments/assets/5662a0b2-bc32-48ef-bf8c-aff54f813ffa

### DIFF
--- a/core/app/[locale]/(default)/(auth)/logout/route.ts
+++ b/core/app/[locale]/(default)/(auth)/logout/route.ts
@@ -1,14 +1,19 @@
-import { getLocale } from 'next-intl/server';
+import { NextRequest } from 'next/server';
 
 import { signOut } from '~/auth';
 import { redirect } from '~/i18n/routing';
 import { setForceRefreshCookie } from '~/lib/force-refresh';
 
-export const GET = async () => {
-  const locale = await getLocale();
+export const GET = async (
+  request: NextRequest,
+  { params }: { params: Promise<{ locale: string }> },
+) => {
+  const { locale } = await params;
+  const redirectTo = request.nextUrl.searchParams.get('redirectTo') ?? '/login';
+  const redirectToPathname = new URL(redirectTo, request.nextUrl.origin);
 
   await signOut({ redirect: false });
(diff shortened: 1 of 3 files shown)


jorgemoya (approved):

core/app/[locale]/(default)/(auth)/logout/route.ts
```diff
@@ -1,14 +1,19 @@
...
 import { setForceRefreshCookie } from '~/lib/force-refresh';
 
-export const GET = async () => {
-  const locale = await getLocale();
+export const GET = async (
+  request: NextRequest,
+  { params }: { params: Promise<{ locale: string }> },
+) => {
+  const { locale } = await params;
+  const redirectTo = request.nextUrl.searchParams.get('redirectTo') ?? '/login';
```
Comment (migueloller) [inline core/app/[locale]/(default)/(auth)/logout/route.ts:12]: Is there any security risk if we don't have an exclusive list of allowed redirect origins? Could an attacker attempt to craft a malicious URL by having logout redirect to a phishing site that looks otherwise like a login screen?

Comment (migueloller) [reply core/app/[locale]/(default)/(auth)/logout/route.ts:12]: Redirecting with URL search params is quite a common thing so I would imagine this isn't a concern but I wonder if usage of a nonce or some other piece of state is generally used to prevent spoofing and whether that is necessary in this context

Comment (chanceaclark) [reply core/app/[locale]/(default)/(auth)/logout/route.ts:12]: We are essentially doing the same thing on the `/login` path and had @bc-donfran review that path https://github.com/bigcommerce/catalyst/pull/2182#discussion_r2023549260 and give his blessings from a security standpoint. I assumed it was fine to take the same approach here and only accept a relative URL.

Now that I look more closely, I do see a miss on my end – I forgot to do `.pathname` after the `new URL` so I'll push a fix for that.

Comment (chanceaclark) [reply core/app/[locale]/(default)/(auth)/logout/route.ts:12]: https://github.com/bigcommerce/catalyst/pull/2198

PR #2190, opened 2025-04-03
feat: syncs and adds support for button and button-link in site theme

hunterbecton:
## What/Why?
- syncs `Button` and `ButtonLink` with Soul
- renames `Foreground` to `Text` in Site Theme

## Testing
https://github.com/user-attachments/assets/b35cbd19-158d-48f8-ace6-f6a6a0b60d6a


core/lib/makeswift/components/site-theme/client.tsx
```diff
@@ -17,6 +17,7 @@ export const SiteTheme = ({ fontTokens, ...theme }: TokensProps & ThemeProps) =>
       ${fontTokensToCssVars(fontTokens).join('\n')}
       ${themeToCssVars(theme).join('\n')}
       --font-family-mono: var(--font-family-accent);
+      --button-primary-text: var(--button-primary-foreground);
```
pvaladez, line 20: We could change all the `foreground` prop names to `text` in [core/lib/makeswift/components/site-theme/components/button.ts](https://github.com/bigcommerce/catalyst/blob/ce302f8871d086a5442fe90881017a5e8975f183/core/lib/makeswift/components/site-theme/components/button.ts#L23) to avoid needing to declare a new `--button-primary-text` css variable here. We're already building the css variables from the Makeswift control values via [themeToCssVars()](https://github.com/bigcommerce/catalyst/blob/ce302f8871d086a5442fe90881017a5e8975f183/core/lib/makeswift/components/site-theme/to-css.ts#L52-L55), so I would prefer to see that functionality not overridden if we can avoid the extra complexity.

PR #2192, opened 2025-04-03
Update wishlist transformer to account for updated schema

jordanarldt:
## What/Why?
Recent GraphQL schema changes made `product` field optional on the wishlist item type. This fixes the data transformer for this update.

## Testing
Tested locally, build passes

### DIFF
--- a/core/data-transformers/wishlists-transformer.ts
+++ b/core/data-transformers/wishlists-transformer.ts
@@ -8,7 +8,7 @@ import { ExistingResultType } from '~/client/util';
 import {
   PublicWishlistFragment,
   WishlistFragment,
-  WishlistItemFragment,
+  WishlistItemProductFragment,
   WishlistPaginatedItemsFragment,
   WishlistsFragment,
 } from '~/components/wishlist/fragment';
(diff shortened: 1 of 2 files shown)


core/data-transformers/wishlists-transformer.ts
```diff
@@ -55,18 +55,27 @@ function wishlistItemsTransformer(
   formatter: ExistingResultType<typeof getFormatter>,
   pt?: ExistingResultType<typeof getTranslations<'Product.ProductDetails'>>,
 ): WishlistItem[] {
-  return removeEdgesAndNodes(wishlistItems).map((item) => ({
```
chanceaclark, line 58: Can we just use `.filter().map()` here instead?

jordanarldt: @chanceaclark That's what I had originally, but for some reason typescript was having a hard time inferencing that `item.product` is not null after that operation, so I went with reduce since it's easier for TS to infer the type with an early return

chanceaclark: You can provide a type-guard to the `.filter` to narrow the returned type so it should be doable.

jordanarldt: @chanceaclark True, I sometimes forget those exist. It doesn't look pretty, but it works. Updated!

chanceaclark: It's fine if it doesn't look super pretty, we just prefer it as it's easier to comprehend `.filter.map` over `.reduce` most of the times. And we don't necessarily need to worry about "efficiency" here in the JS land as the compiler and engine will handle some of these refinements.

chanceaclark (approved):

PR #2195, opened 2025-04-04
feat(infra): SST AWS starter

zbaig:
## Testing

> **Note:** This is just a starter pack. Caching, secutiry, roles, and other optimizations can be easily added based on requirements.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,9 @@ bigcommerce.graphql
 bigcommerce-graphql.d.ts
 .DS_Store
 coverage/
+# sst
+.sst
+
+# open-next
+.open-next
+sst-env.d.ts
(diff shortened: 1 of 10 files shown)


bookernath:
Thanks for your contribution @zbaig . What did you have in mind for the final shape of this PR? Are you hoping to provide this as an example for how to deploy Catalyst on AWS? If that's the case, it might make more sense to put this in a branch or as a reference in our documentation rather than merge this PR (as it changes things in a way that would be disadvantageous for people who _aren't_ deploying to Vercel.

Happy to get you with our DevRel team to figure out the best shape of where to deploy this.

CC @jamesqquick

zbaig:
@bookernath - yes I wanted to share how we'd do lambda based deployment with catalyst. This was meant as an example - ref implementation and a starting point to ideate and configure this as a possible production grade hosting solution. No need to merge this branch.

PR #2203, opened 2025-04-07
fix: temp unregister of Customer Group Slot

drewpledger:
Admin API key was removed from Catalyst for security purposes. A higher level API key is required to query Customer Groups for the Customer Group Slot. We need to unregister this component while we find a solution to this.


bookernath (approved):
After further discussion, plan is to temporarily remove this component while we re-evaluate the approach to the builder experience and re-introduce it soon.

PR #2204, opened 2025-04-07
refactor(core): category for dynamicIO 

jorgemoya:
## New

This refactor optimizes category PLP for caching and the eventual use of dynamicIO. With these changes we leverage data caching to hit mostly cache data for guest shoppers in different locales and with different currencies.

## Key modifications include:

- We don't stream in Category page data, instead it's a blocking call that will redirect to `notFound` when category is not found. Same for metadata.

## Migration instructions:

- Update `/(facted)/category/[slug]/page.tsx`
  - For this page we are now doing a blocking request for category page data. Instead of having functions that each would read from props, we share streamable functions that can be passed to our UI components. We still stream in filter and product data.
- Update `/(facted)/category/[slug]/page-data.tsx`
  - Request now accept `customerAccessToken` as a prop instead of calling internally.
- Update`/(facted)/category/[slug]/fetch-compare-products.ts`
  - Request now accept `customerAccessToken` as a prop instead of calling internally.
- Update `/(faceted)/fetch-faceted-search.ts`
  - Request now accept `customerAccessToken` and `currencyCode` as a prop instead of calling internally.

## Testing

### DIFF
--- a/.changeset/salty-laws-hunt.md
+++ b/.changeset/salty-laws-hunt.md
@@ -2,4 +2,24 @@
 "@bigcommerce/catalyst-core": minor
 ---
 
-Refactor Category PLP in preparation for dynamic IO, by making sure there are no blocking calls other than static requests.
+## New
+
+This refactor optimizes category PLP for caching and the eventual use of dynamicIO. With these changes we leverage data caching to hit mostly cache data for guest shoppers in different locales and with different currencies.
+
+## Key modifications include:
+
+- We don't stream in Category page data, instead it's a blocking call that will redirect to `notFound` when category is not found. Same for metadata.
+- Our query functions now take in all params required for fetching, instead of accessing dynamic variables internally. This is important to serialize arguments if we want to eventually `use cache`.
+- Use `Streamable.from` to generate our streaming props that are passed to our UI components.
+- Remove use of nuqs' `createSearchParamsCache` in favor of nuqs' `createLoader`.
+
+## Migration instructions:
+
+- Update `/(facted)/category/[slug]/page.tsx`
+  - For this page we are now doing a blocking request for category page data. Instead of having functions that each would read from props, we share streamable functions that can be passed to our UI components. We still stream in filter and product data.
+- Update `/(facted)/category/[slug]/page-data.tsx`
+  - Request now accept `customerAccessToken` as a prop instead of calling internally.
+- Update`/(facted)/category/[slug]/fetch-compare-products.ts`
+  - Request now accept `customerAccessToken` as a prop instead of calling internally.
+- Update `/(faceted)/fetch-faceted-search.ts`
+  - Request now accept `customerAccessToken` and `currencyCode` as a prop instead of calling internally.
--- a/.changeset/salty-laws-hunt.md
+++ b/.changeset/salty-laws-hunt.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": minor
+---
+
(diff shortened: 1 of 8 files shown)


.changeset/salty-laws-hunt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": minor
+---
+
+Refactor Category PLP in preparation for dynamic IO, by making sure there are no blocking calls other than static requests.
```
chanceaclark, line 5: Can we think/provide some sort of guide to this changelog on how to migrate to this new functionality?

core/app/[locale]/(default)/(faceted)/category/[slug]/page.tsx
```diff
@@ -77,297 +49,219 @@ const createCategorySearchParamsCache = cache(async (props: Props) => {
...
+  const categoryId = Number(slug);
 
-async function getCompareProducts(props: Props) {
-  const compareIds = await cachedCompareProductIds(props);
+  const { category, settings, categoryTree } = await getCategoryPageData(categoryId);
 
-  const products = await getCompareProductsData(compareIds);
+  if (!category) {
+    return notFound();
+  }
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/(faceted)/category/[slug]/page.tsx:95]: If we are not using the customer access token for this query, a shopper might get a 404 page if they are logged in navigating to a category gated by a customer group. (looks like you removed it in `page-data.ts`)

core/app/[locale]/(default)/(faceted)/category/[slug]/page.tsx
```diff
@@ -77,297 +49,219 @@ const createCategorySearchParamsCache = cache(async (props: Props) => {
...
+    }
 
-interface Props {
-  params: Promise<{
-    slug: string;
-    locale: string;
-  }>;
-  searchParams: Promise<SearchParams>;
-}
+    const compareLoader = createCompareLoader();
```
Comment (migueloller) [inline core/app/[locale]/(default)/(faceted)/category/[slug]/page.tsx:204]: Should we create this outside the component? How come we have a function here vs exporting the loader?

migueloller (commented):
Looks like we're losing some customer group functionality here based on @chanceaclark's [comment](https://github.com/bigcommerce/catalyst/pull/2204#discussion_r2033632401). I propose we adopt the Flags SDK and use the [precompute pattern](https://flags-sdk.dev/frameworks/next/examples/suspense-fallbacks) with PPR so that we can switch Suspense fallbacks based on the user session, currency, etc.

core/app/[locale]/(default)/(faceted)/category/[slug]/page.tsx
Comment (chanceaclark) [reply core/app/[locale]/(default)/(faceted)/category/[slug]/page.tsx:95]: >would the result be different in the GraphQL response based on the customer access token?

Yeah it would. Let's say I have a Guest Shopper, a Logged in Shopper under customer group A, and a Logged in Shopper under customer group B. Guest Shoppers can see categories W and X; Customer Group A can see categories W, X, and Y; Customer Group B can see categories W, X, Y, and Z. Depending on the access each customer group has they can see different pages so really we need to precompute based on customer groups here.

Comment (migueloller) [reply core/app/[locale]/(default)/(faceted)/category/[slug]/page.tsx:95]: Heard.

> Depending on the access each customer group has they can see different pages so really we need to precompute based on customer groups here.

If I'm understanding what you're saying, we don't know the customer group until we send the request and its implicitly encoded in the CAT. So, the flag would probably be a "is anonymous"/"is authenticated" flag vs a "customer group" flag

migueloller (approved):

PR #2205, opened 2025-04-08
refactor(core): brand for dynamicIO

jorgemoya:
## New

This refactor optimizes brand PLP for caching and the eventual use of dynamicIO. With these changes we leverage data caching to hit mostly cache data for guest shoppers in different locales and with different currencies.

## Key modifications include:

- We don't stream in Brand page data, instead it's a blocking call that will redirect to `notFound` when brand is not found. Same for metadata.

## Migration instructions:

- Update `/(facted)/brand/[slug]/page.tsx`
  - For this page we are now doing a blocking request for brand page data. Instead of having functions that each would read from props, we share streamable functions that can be passed to our UI components. We still stream in filter and product data.
- Update `/(facted)/brand/[slug]/page-data.tsx`
  - Request now accept `customerAccessToken` as a prop instead of calling internally.

## Testing

### DIFF
--- a/core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx
+++ b/core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx
@@ -1,7 +1,7 @@
 import type { Metadata } from 'next';
 import { notFound } from 'next/navigation';
 import { getFormatter, getTranslations, setRequestLocale } from 'next-intl/server';
-import { createSearchParamsCache } from 'nuqs/server';
+import { createLoader, SearchParams } from 'nuqs/server';
 import { cache } from 'react';
 
 import { Streamable } from '@/vibes/soul/lib/streamable';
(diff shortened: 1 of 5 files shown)


core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx
```diff
@@ -80,229 +47,185 @@ const createBrandSearchParamsCache = cache(async (props: Props) => {
...
-  return t('maxCompareLimit');
-}
+  const streamableFacetedSearch = Streamable.from(async () => {
+    const searchParams = await props.searchParams;
+    const customerAccessToken = await getSessionCustomerAccessToken();
+    const currencyCode = await getPreferredCurrencyCode();
 
-async function getEmptyStateTitle(): Promise<string> {
-  const t = await getTranslations('Faceted.Brand.Empty');
+    const searchParamsCache = await createBrandSearchParamsCache(slug);
```
Comment (migueloller) [inline core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx:101]: Creating a search params cache like this seems worthless since we can't access this cache and thus read set values on other RSCs. If we're instead looking to parse the search params,`createLoeader` seems more appropriate

core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx
```diff
@@ -80,229 +47,185 @@ const createBrandSearchParamsCache = cache(async (props: Props) => {
...
-}
+    return pageInfoTransformer(search.products.pageInfo);
+  });
 
-async function getResetFiltersLabel(): Promise<string> {
-  const t = await getTranslations('Faceted.FacetedSearch');
+  const streamableFilters = Streamable.from(async () => {
+    const searchParams = await props.searchParams;
+    const searchParamsCache = await createBrandSearchParamsCache(slug);
+    const parsedSearchParams = searchParamsCache?.parse(searchParams) ?? {};
```
Comment (migueloller) [inline core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx:150]: Should we extract the parsed search params so we can reuse it both here and [above](https://github.com/bigcommerce/catalyst/blob/8bb203a71b9ea8e304bdb00c3890e5fab4f61ab9/core/app/%5Blocale%5D/(default)/(faceted)/brand/%5Bslug%5D/page.tsx#L101)?

migueloller (approved):
This looks good 👍🏻 

Something that comes to mind is how we set expectations for users pulling in changes. This is refactoring the entire file so not much we can do there in terms of merge conflicts but perhaps we can explain our rationale here—in particular how we're optimizing for:
- Better DX in terms of following what's happening
- Better performance by separating static data and dynamic data
- Better support for PPR and `dynamicIO` when that's stable

PR #2207, opened 2025-04-08
refactor(core): search for dynamicIO

jorgemoya:
## New

This refactor optimizes search PLP for caching and the eventual use of dynamicIO. With these changes we leverage data caching to hit mostly cache data for guest shoppers in different locales and with different currencies.

## Key modifications include:

- We don't stream in Search page data, instead it's a blocking call to get page data.

## Migration instructions:

- Update `/(facted)/search/page.tsx`
  - For this page we are now doing a blocking request for brand page data. Instead of having functions that each would read from props, we share streamable functions that can be passed to our UI components. We still stream in filter and product data.

## Testing

### DIFF
--- a/core/app/[locale]/(default)/(faceted)/search/page.tsx
+++ b/core/app/[locale]/(default)/(faceted)/search/page.tsx
@@ -5,287 +5,19 @@ import { cache } from 'react';
 
 import { Streamable } from '@/vibes/soul/lib/streamable';
 import { createCompareLoader } from '@/vibes/soul/primitives/compare-drawer/loader';
-import { CursorPaginationInfo } from '@/vibes/soul/primitives/cursor-pagination';
-import { Product } from '@/vibes/soul/primitives/product-card';
-import { Breadcrumb } from '@/vibes/soul/sections/breadcrumbs';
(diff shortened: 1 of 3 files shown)


core/app/[locale]/(default)/(faceted)/search/page.tsx
```diff
@@ -5,332 +5,261 @@ import { cache } from 'react';
...
+    // ```
+    // Error: [nuqs] Empty search params cache. Search params can't be accessed in Layouts.
+    //   See https://err.47ng.com/NUQS-500
+    // ```
+    if (Object.keys(filterParsers).length === 0) {
+      return null;
+    }
+
+    return createSearchParamsCache(filterParsers);
+  });
```
Comment (migueloller) [inline core/app/[locale]/(default)/(faceted)/search/page.tsx:79]: We should replace this with a loader and instead of a function, we could create a loader and parse the search params within a `Streamable.from` so that we have a `streamableParsedSearchParams` we use instead of `props.searchParam` throughout the component

migueloller (commented):
I would provide similar feedback in this PR as in https://github.com/bigcommerce/catalyst/pull/2204 and https://github.com/bigcommerce/catalyst/pull/2205

migueloller (approved):

PR #2212, opened 2025-04-09
fix: default slot QoL

drewpledger:
There is now a default slot that is used where conditions do not apply, or a user does not have a group id

Before, if a group was simulated that did not exist in the "Targeted customer groups" control, we would render a message saying that you need to add this group before it can render properly. However, I realized after documenting our instructions for this that this would present several issues when modifying your storefront customer groups.
* If you add a customer group, and assign customers to that group, every customer would see the error div until the MS user changes every instance of this component.
* There was no way to render what every _other_ group would see.

This PR aims to address those potential issues:

https://github.com/user-attachments/assets/a18d532c-54bf-41e9-ba79-45a23da1daac


core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
```diff
@@ -80,11 +76,11 @@ interface Props {
...
   simulatedGroup?: string;
-  noGroupSlot: ReactNode;
+  defaultSlot: ReactNode;
 }
 
-function CustomerGroupSlot({ className, slots, simulatedGroup = NO_GROUP_ID, noGroupSlot }: Props) {
-  const allSlots = slots?.concat({ group: NO_GROUP_ID, slot: noGroupSlot });
+function CustomerGroupSlot({ className, slots, simulatedGroup, defaultSlot }: Props) {
+  const targetedSlots = slots?.concat({ group: DEFAULT_GROUP_ID, slot: defaultSlot });
   const isInBuilder = useIsInBuilder();
```
jamesqquick, line 84: Was there a decision made on if this function was something that was ok to use?

drewpledger: Yup, Miguel approved the usage

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.tsx
```diff
@@ -132,18 +129,21 @@ runtime.registerComponent(CustomerGroupSlot, {
...
 
-                return data
-                  .map((d) => ({
-                    id: `${d.id}`,
-                    label: d.name,
-                    value: `${d.id}`,
-                  }))
-                  .filter((option) => option.label.toLowerCase().includes(query.toLowerCase()));
+                return [
+                  { id: NO_GROUP_ID, label: 'No-group', value: NO_GROUP_ID },
```
migueloller, line 133: Should the label be "No group", without the hyphen?

PR #2213, opened 2025-04-09
fix: deletion of `NEXT_LOCALE` cookie triggers page rerender

agurtovoy:
## What/Why?

This is a rework of https://github.com/bigcommerce/catalyst/pull/1984 that fixes the unintended page reload introduced by that PR. See [ENG-7924](https://linear.app/makeswift/issue/ENG-7924/unconditional-deletion-of-next-locale-cookie-triggers-page-rerender) for more details.

## Testing

https://github.com/user-attachments/assets/8caf3611-4b6d-40db-9e61-b880e1ade156

### DIFF
--- a/core/middlewares/with-makeswift.ts
+++ b/core/middlewares/with-makeswift.ts
@@ -1,4 +1,6 @@
 import { unstable_createMakeswiftDraftRequest } from '@makeswift/runtime/next/middleware';
+import { NextResponse } from 'next/server';
+import { splitCookiesString } from 'set-cookie-parser';
 
 import { routing } from '~/i18n/routing';
 
@@ -13,6 +15,32 @@ const MAKESWIFT_SITE_API_KEY = process.env.MAKESWIFT_SITE_API_KEY;
 const localeCookieName = ({ localeCookie }: { localeCookie?: boolean | { name?: string } }) =>
   (typeof localeCookie === 'object' ? localeCookie.name : undefined) ?? 'NEXT_LOCALE';
 
+const stripSetLocaleCookieHeaders = (response: NextResponse | Response) => {
+  const cookieName = localeCookieName(routing);
+  const isLocaleCookie = (value: string) => value.startsWith(`${cookieName}=`);
+
+  response.headers.forEach((value, name) => {
+    if (name === 'set-cookie') {
+      if (isLocaleCookie(value)) {
+        response.headers.delete(name);
+      }
+    }
+
+    if (name === 'x-middleware-set-cookie') {
+      // In contrast to the 'set-cookie' header, Next.js' 'x-middleware-set-cookie' header may
+      // contain multiple comma-separated cookie values in a single header--see
+      // https://github.com/vercel/next.js/blob/canary/packages/next/src/server/web/spec-extension/response.ts#L60
+      const values = splitCookiesString(value).filter((v) => !isLocaleCookie(v));
+
+      if (values.length > 0) {
+        response.headers.set(name, values.join(','));
+      } else {
+        response.headers.delete(name);
+      }
+    }
+  });
+};
+
 export const withMakeswift: MiddlewareFactory = (middleware) => {
   return async (request, event) => {
     const draftRequest = await unstable_createMakeswiftDraftRequest(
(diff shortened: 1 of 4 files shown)


core/middlewares/with-makeswift.ts
```diff
@@ -21,16 +34,16 @@ export const withMakeswift: MiddlewareFactory = (middleware) => {
...
     if (draftRequest != null) {
-      if (routing.localeCookie) {
-        // If the i18n middleware is configured to use a cookie, it will first try to derive the
-        // locale from the existing request cookie before attempting to match the URL against the
-        // locale routes. The locale switcher in the Makeswift Builder expects the host to always
-        // determine the locale from the URL, though, so we have to erase the cookie from the
-        // proxied request to force that behavior.
-        draftRequest.cookies.delete(localeCookieName(routing));
-      }
+      const response = await middleware(draftRequest, event);
```
agurtovoy, line 37: Best reviewed with "Hide whitespace" toggle on.

core/middlewares/with-makeswift.ts
```diff
@@ -13,6 +14,18 @@ const MAKESWIFT_SITE_API_KEY = process.env.MAKESWIFT_SITE_API_KEY;
...
+  const cookieName = localeCookieName(routing);
+
+  response.headers.forEach((value, name) => {
+    if (name === 'set-cookie' || name === 'x-middleware-set-cookie') {
+      if (value.startsWith(`${cookieName}=`)) {
+        response.headers.delete(name);
+      }
+    }
+  });
+};
```
arvinpoddar, line 27: It seems like this code deletes the entire cookie header if it contains `NEXT_LOCALE` anywhere within it. What if the response had other cookies in it? Would this wipe all of them? That would impact functionality such as logging in, etc.

core/middlewares/with-makeswift.ts
```diff
@@ -13,6 +14,18 @@ const MAKESWIFT_SITE_API_KEY = process.env.MAKESWIFT_SITE_API_KEY;
 const localeCookieName = ({ localeCookie }: { localeCookie?: boolean | { name?: string } }) =>
   (typeof localeCookie === 'object' ? localeCookie.name : undefined) ?? 'NEXT_LOCALE';
 
+const stripSetLocaleCookieHeaders = (response: NextResponse | Response) => {
```
arvinpoddar, line 17: nit: any reason to not have this as a function declaration?

agurtovoy: It's customary for the Catalyst codebase.

arvinpoddar: Got it, thanks

core/middlewares/with-makeswift.ts
arvinpoddar: I knew this was true for `set-cookie`, but is this true for `x-middleware-set-cookie`? I thought this behavior was specific to the `set-cookie` key, hence the presence of the `getSetCookie` method.

Let me ask this: have we tried setting a cookie on the response and seeing if this code still preserves that cookie while booting the `NEXT_LOCALE` cookie? That would alleviate my concerns.

agurtovoy: @arvinpoddar Hmm, it looks like you're right. I thought that all `x-middleware-[header]` headers mirror their original counterparts (if any), but I guess the wise guys at Next.js decided to extend the `set-cookie` format by [comma-separating multiple headers](https://github.com/vercel/next.js/blob/d55bdd711fab1106d47ab508a8c05ee681ac0cbd/packages/next/src/server/web/spec-extension/response.ts#L60). Note that this is neither the `set-cookie` nor the `cookie` header format, but rather a proprietary Next.js invention.

agurtovoy: > Note that this is neither the set-cookie nor the cookie header format, but rather a proprietary Next.js invention.

This is a side note, but [this comment](https://github.com/nfriedly/set-cookie-parser/blob/4b2a8b6a1ec3cef505228ef4363c9d9b7a55401f/lib/set-cookie.js#L138) in the `set-cookie-parser` source code made me to go on a goose chase for whether what I said above is true. I think [it is](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1):

set-cookie header grammar

agurtovoy: Also note that `set-cookie-parser`'s [`parse`](https://github.com/nfriedly/set-cookie-parser/blob/4b2a8b6a1ec3cef505228ef4363c9d9b7a55401f/lib/set-cookie.js#L81) doesn't handle the "multi-value" `set-cookie` strings; one has to explicitly call [`splitCookiesString`](https://github.com/nfriedly/set-cookie-parser/blob/4b2a8b6a1ec3cef505228ef4363c9d9b7a55401f/lib/set-cookie.js#L148) to separate those values.

agurtovoy: [Addressed](https://github.com/bigcommerce/catalyst/pull/2213/files#diff-6922f70d647c8f5698008a4cba3003ff645557f0a77ee1b2db8de3c7ca14162bR29); here's a demo:

https://github.com/user-attachments/assets/3943d9ca-f541-4bb6-becd-a53e0ba840ba

agurtovoy: Thanks for pointing out the `x-middleware-set-cookie` idiosyncrasy to me, @arvinpoddar!

PR #2216, opened 2025-04-10
Add wishlist button to PDP

jordanarldt:
## What/Why?
Add ability to add a product to a wishlist from the PDP

## Testing
- Locally/manual
- GraphQL functional tests

image

https://github.com/user-attachments/assets/ff43ca63-719c-415c-b502-fab3127d949d

## Migration
- Ensure `WishlistButton` component is passed to `additionalActions` prop on ProductDetail
- Ensure `WishlistButtonForm` is used on product page

[CATALYST-954]: https://bigcommercecloud.atlassian.net/browse/CATALYST-954?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+            name
+          }
+        }
+      }
+    }
+  }
+`);
+
+const getWishlistButtonData = cache(async (productId: number) => {
+  const customerAccessToken = await getSessionCustomerAccessToken();
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/index.tsx:56]: We need to pass all dynamic data as query params for eventual use of dynamicIO.

core/app/[locale]/(default)/product/[slug]/page.tsx
```diff
@@ -318,6 +330,13 @@ export default async function Product(props: Props) {
           </>
         )}
       </Stream>
+
+      <WishlistButtonForm
+        formId={detachedWishlistFormId}
+        productId={productId}
+        productSku={streamableProductSku}
+        searchParams={props.searchParams}
+      />
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/product/[slug]/page.tsx:339]: What is the reason for this form to not live inside `product-details-form`?

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
```diff
@@ -0,0 +1,304 @@
...
+          edges {
+            node {
+              entityId
+            }
+          }
+        }
+      }
+    }
+  }
+`);
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:35]: Should we fetch variantId from product page query instead?

jorgemoya (commented):
Did a quick pass and got some questions, but overall this looks great!

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
```diff
@@ -0,0 +1,304 @@
...
+        isPublic: false,
+        items: [{ productEntityId: productId, variantEntityId: variantId }],
+      },
+    },
+    customerAccessToken,
+    fetchOptions: { cache: 'no-store' },
+  });
+
+  if (!data.wishlist.createWishlist?.result) {
+    throw new Error('Failed to add item to default wishlist. Response was empty.');
```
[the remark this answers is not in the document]
Comment (jordanarldt) [reply core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:133]: @jorgemoya These aren't exposed to the user, just logged when we `catch`. It'll propagate to the user as the translated unexpected error 👍

core/app/[locale]/(default)/product/[slug]/page.tsx
Comment (jordanarldt) [reply core/app/[locale]/(default)/product/[slug]/page.tsx:339]: @jorgemoya Forms can't be nested, so having it all live inside of the form and doing everything from the `addToCart` action seems unnecessarily complex since we can just have a "decoupled" form where the inputs are attached to the form ID. Seems like it will be much easier to maintain this way 👍

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
Comment (jordanarldt) [reply core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:35]: @jorgemoya I considered it, but then we would be querying this additional data even in cases where it will go unused. The current implementation will only make the query when we need it (when adding a product to the wishlist), so it's slightly less load on our infrastructure 👍

Comment (jorgemoya) [reply core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:35]: Awesome, makes sense

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
```diff
@@ -0,0 +1,304 @@
...
+  wishlistItemId: number;
+}
+
+const getVariantIdFromSku = cache(
+  async (productId: number, sku: string, customerAccessToken?: string) => {
+    const { data } = await client.fetch({
+      document: VariantIdFromSkuQuery,
+      variables: { productId, sku },
+      customerAccessToken,
+      fetchOptions: { next: { revalidate } },
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:106]: Should we skip cache for logged in customer?

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
```diff
@@ -0,0 +1,304 @@
...
+  const locale = await getLocale();
+  const customerAccessToken = await getSessionCustomerAccessToken();
+  const t = await getTranslations('Product.ProductDetails.WishlistButton');
+  const submission = parseWithZod(payload, { schema });
+
+  if (submission.status !== 'success') {
+    await serverToast.error(t('Errors.unexpected'));
+
+    return;
+  }
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:182]: We usually just response with a `reply()` in these cases because it means the form was invalid. In this case, the form should never be invalid since we control the form, but still, we should probably propagate as a reply.

```
  if (submission.status !== 'success') {
    return { lastResult: submission.reply() };
  }
```

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
```diff
@@ -0,0 +1,304 @@
...
+  const locale = await getLocale();
+  const t = await getTranslations('Product.ProductDetails.WishlistButton');
+  const submission = parseWithZod(formData, { schema: addToNewWishlistSchema });
+
+  if (submission.status !== 'success') {
+    return {
+      ...prevState,
+      lastResult: submission.reply({ formErrors: [t('Errors.unexpected')] }),
+    };
+  }
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:277]: Same comment as above.

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
```diff
@@ -0,0 +1,304 @@
...
+    newWishlistFormData.append('wishlistItems[0].variantEntityId', variantId.toString());
+  }
+
+  if (!customerAccessToken) {
+    redirect({ href: getLoginRedirect(redirectTo), locale });
+
+    return { lastResult: null };
+  }
+
+  const result = await newWishlist(prevState, newWishlistFormData);
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:297]: I don't follow why we are appending the form fields, wouldn't we simply pass as props?

Comment (jorgemoya) [reply core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:297]: Oh I guess it is calling another action that we have defined before.

core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/add-to-new-wishlist-modal.tsx
```diff
@@ -0,0 +1,75 @@
+'use client';
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/add-to-new-wishlist-modal.tsx:1]: I wonder if `wishlist-button` (all components) should live under the `@/vibes/soul/section/product-detail` section.

core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/add-to-new-wishlist-modal.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+import { useSearchParams } from 'next/navigation';
+
+import { NewWishlistModal } from '~/app/[locale]/(default)/account/wishlists/_components/modals/new';
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/add-to-new-wishlist-modal.tsx:5]: Should this file be moved to `core/components` since it is reused.

core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/form.tsx
```diff
@@ -0,0 +1,50 @@
...
+// Since the wishlist button lives inside of the ProductDetailsForm, we need a separate/detached form
+// to handle the actions through. The form ID is provided to the 'form' attribute of the wishlist menu items.
+// The Modal also contains a form, which needs to be detached.
+export const WishlistButtonForm = async ({
+  formId,
+  productId,
+  productSku,
+  searchParams,
+}: Props) => {
+  const t = await getTranslations('Account.Wishlists');
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/form.tsx:27]: This is a 🍹 but since these labels are not for the account page, I would re-define them under Product.

core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+const getWishlistButtonData = cache(async (productId: number, customerAccessToken?: string) => {
+  if (!customerAccessToken) {
+    return null;
+  }
+
+  const { data } = await client.fetch({
+    document: WishlistButtonQuery,
+    variables: { productId, first: wishlistButtonLimit },
+    customerAccessToken,
+    fetchOptions: { next: { revalidate, tags: [TAGS.customer] } },
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/index.tsx:64]: Makes sense to me that we cache this, but we usually don't cache customer queries. Maybe @bookernath has an opinion for this?

core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+  wishlistItemId?: number;
+}
+
+interface WishlistButton {
+  isProductInWishlist: boolean;
+  wishlists: WishlistButtonWishlistInfo[];
+}
+
+async function getWishlistButton(
+  t: ExistingResultType<typeof getTranslations<'Product.ProductDetails.WishlistButton'>>,
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/index.tsx:82]: You're already awaiting for data, might as well await for translations inside the function too.

core/vibes/soul/sections/product-detail/product-detail-form.tsx
```diff
@@ -197,16 +200,19 @@ function FormField({
...
 
   const handleChange = useCallback(
     (value: string) => {
-      void setParams({ [field.name]: value });
-      controls.change(value);
+      // Ensure that if page is reached without a full reload, we are still setting the selection properly based on query params.
+      const fieldValue = value || String(params[field.name]);
+
+      void setParams({ [field.name]: fieldValue });
+      controls.change(fieldValue);
```
jorgemoya, line 213: Is this to fix the issue with the reload?

core/app/[locale]/(default)/product/[slug]/_components/search-params-router-refresh.tsx
```diff
@@ -0,0 +1,28 @@
+'use client';
+
+import { useSearchParams } from 'next/navigation';
+import { SearchParams } from 'nuqs';
+import { useEffect } from 'react';
+
+import { useRouter } from '~/i18n/routing';
+
+// Not-so-great workaround for https://github.com/vercel/next.js/issues/59407
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/product/[slug]/_components/search-params-router-refresh.tsx:9]: Honestly, maybe we should disable PPR since it flashes when there's a change. Maybe @bookernath and @migueloller can provide an opinion.

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
Comment (jordanarldt) [reply core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:297]: @jorgemoya Yeah, I wanted to reuse the existing `newWishlist` action instead of rolling a whole new one. Makes things simpler and keeps the "new wishlist" in one place

core/vibes/soul/sections/product-detail/product-detail-form.tsx
jordanarldt: @jorgemoya This is separate - this is a fix for login redirect back to the product page. If you attempt to add a product to a wishlist after selecting options while logged out, you'll go to `/login?redirectTo=/my-product?123=1234`, and then after logging in, you'll be redirected to the product page with the query params. This change here ensures that after the redirect, the product options are re-selected properly.

core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/add-to-new-wishlist-modal.tsx
Comment (jordanarldt) [reply core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/add-to-new-wishlist-modal.tsx:1]: @jorgemoya I had a brief discussion about this in DMs with Miguel. This particular wishlist implementation is very specific to BC, so we opted for a "slotted" pattern where we just pass in the additional component through the `additionalActions` prop on the ProductDetails.

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
Comment (jordanarldt) [reply core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:182]: @jorgemoya We don't use any sort of form states here. This is a void promise, so we don't need to return any sort of submission or result - all we need to do is fire off the toast messages

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
Comment (jordanarldt) [reply core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:277]: @jorgemoya If we don't reply with the `formErrors`, then there will be nothing on the UI indicating an error:

`return { lastResult: submission.reply() };`
image

### vs

`return { lastResult: submission.reply({ formErrors: [t('Errors.unexpected')] }) };`
image

core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/index.tsx
Comment (jordanarldt) [reply core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/index.tsx:82]: Good call

core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/add-to-new-wishlist-modal.tsx
Comment (jordanarldt) [reply core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/add-to-new-wishlist-modal.tsx:5]: @jorgemoya I was considering it as well, but this is likely going to be the only place that it gets reused 🤔

Comment (jordanarldt) [reply core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/add-to-new-wishlist-modal.tsx:5]: We could probably move all of the wishlist modal definitions to `core/components`, as well as the share button? 💭

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
Comment (jorgemoya) [reply core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:106]: Then we shouldn't pass the `customerAccessToken` no? I thought the CAT would be necessary for hidden variants. 

In terms of Vercel cache, we don't want to cache stuff that will be specific to a customer, since we don't want to spam the cache storage for each customer.

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
Comment (jorgemoya) [reply core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:182]: That makes sense, thanks!

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
Comment (jorgemoya) [reply core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:294]: Not the biggest fan, seems weird, but this will do for now!

core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/add-to-new-wishlist-modal.tsx
Comment (jorgemoya) [reply core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/add-to-new-wishlist-modal.tsx:1]: I understand but this kind of goes against the themeing work we want to implement. We need to stop thinking about `vibes/soul` and more like `ui`. cc @migueloller

core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/index.tsx
Comment (jorgemoya) [reply core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/index.tsx:63]: Same as above, I guess so far our understand is that we don't want to cache customer related queries because it can store a lot of cache if you have many customers.

core/app/[locale]/(default)/product/[slug]/page.tsx
Comment (jorgemoya) [reply core/app/[locale]/(default)/product/[slug]/page.tsx:339]: I still think this should live under `ProductDetail` (and ProductDetail should have both forms). 🤔

core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/add-to-new-wishlist-modal.tsx
Comment (jorgemoya) [reply core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/add-to-new-wishlist-modal.tsx]: Looks good to me!

core/vibes/soul/sections/product-detail/product-detail-form.tsx
jorgemoya: Nice!

core/app/[locale]/(default)/product/[slug]/page.tsx
Comment (jordanarldt) [reply core/app/[locale]/(default)/product/[slug]/page.tsx:339]: @jorgemoya I disagree, I still feel that we should keep it all decoupled from the VIBES component since Wishlists on the PDP is specific to BC functionality 💭  I discussed it with @migueloller already and we landed on taking the approach of having the `additionalActions` prop on `ProductDetails` so that any consumer of VIBES can pass a `ReactNode` to add stuff after the "add to cart" button. With this approach, we're able to keep everything at the top-level and decoupled from VIBES. 🤔

Comment (migueloller) [reply core/app/[locale]/(default)/product/[slug]/page.tsx:339]: Just wanted to chime in and say that we still think as VIBES being decoupled when being brought in to Catalyst so it should be perfectly fine to change the VIBES components. This is why we renamed the directory from `vibes/soul` to `ui`. That said, I do think that for VIBES it would be a good idea to have a prop that allows extending the product detail with additional actions by taking a `ReactNode`. A good use case for this, for example, is adding BNPL options without worrying about having to for the UI too much. I will admit, though, that these ideas mostly came up when we were thinking about theming from a VIBES perspective

core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/index.tsx
Comment (jordanarldt) [reply core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/index.tsx:63]: @jorgemoya Yeah for sure. I'm fine either way! If we remove caching we can always re-add it later if it's causing performance issues.

core/app/[locale]/(default)/product/[slug]/page.tsx
Comment (jordanarldt) [reply core/app/[locale]/(default)/product/[slug]/page.tsx:339]: Another quick thought - the current approach should also make it easier for developers to migrate and update to this wishlist PDP functionality. If we move it into the ProductDetail component, it would be a bit more complicated of a migration since I think developers are more likely to have modified that component rather than the root page 💭

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
Comment (chanceaclark) [reply core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:106]: Jorge is right here, we need to not cache this query if we are using a `customerAccessToken`. Next doesn't use headers as part of the cache key, so if Customer A has access to the product/variant and Customer B doesn't, whatever customer fetches this query first, the other would get the data from the other customer.

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
```diff
@@ -0,0 +1,301 @@
...
+const getVariantIdFromSku = cache(
+  async (productId: number, sku: string, customerAccessToken?: string) => {
+    const { data } = await client.fetch({
+      document: VariantIdFromSkuQuery,
+      variables: { productId, sku },
+      customerAccessToken,
+      fetchOptions: { next: { revalidate } },
+    });
+
+    return data.site.product?.variants.edges?.[0]?.node.entityId ?? undefined;
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:109]: Can we use the `removeEdgesAndNodes` util here?

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
```diff
@@ -0,0 +1,301 @@
...
+async function addToWishlist(customerAccessToken: string, variables: WishlistAddMutationVariables) {
+  const { data } = await client.fetch({
+    document: AddToWishlistMutation,
+    variables,
+    customerAccessToken,
+    fetchOptions: { cache: 'no-store' },
+  });
+
+  if (!data.wishlist.addWishlistItems?.result) {
+    throw new Error('Failed to add item to wishlist. Response was empty.');
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:146]: 🍹 Something to think about here, if we are throwing a bunch of generic errors with custom message it might make sense to create our own custom errors. Even if it seems like a lot right now, it allows consumers to extend functionality if they ever have the need, e.g. `error instanceof CustomError`. It helps not checking `error.message.includes('some text')` to see what the error type is.

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
```diff
@@ -0,0 +1,301 @@
...
+  if (!data.wishlist.deleteWishlistItems?.result) {
+    throw new Error('Failed to add item to wishlist. Response was empty.');
+  }
+}
+
+function getLoginRedirect(redirectTo: string) {
+  const loginParams = new URLSearchParams({ redirectTo });
+
+  return `/login?${loginParams.toString()}`;
+}
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:170]: For any sort of `redirectTo`, on the Catalyst side we can only accept relative pathnames to be passed. `URLSearchParams` I think only escapes the param, but we need to pass it into `new URL` and get the pathname (and maybe search params) only. It might make sense to do it right here.

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
Comment (chanceaclark) [reply core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:294]: Curious, since `formData` is already a `FormData`, could we just append to that one instead of creating a new one?

Comment (jordanarldt) [reply core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:294]: @chanceaclark They are both `FormData` types, but the schemas don't overlap. The `newWishlist` expected schema is this:

```
  z.object({
    wishlistName: z.string({ required_error }).trim().nonempty(),
    wishlistIsPublic: z.enum(['true', 'false']).optional(),
    wishlistItems: z.array(wishlistItemSchema).optional(),
  });
```

Additionally, we don't know the `variantEntityId` until after the `addToNewWishlist` function fetches it using the SKU. The only way would be to add that query on every page load, but it makes more sense to only fetch the variant ID from the SKU when we need it (during this mutation), rather than any time the wishlist button is rendered.

Constructing the new FormData may seem odd, but it lets us reuse the `newWishlist` function and also avoids an unnecessary query on every PDP load since we would need the variant ID 👍

.changeset/tiny-wombats-turn.md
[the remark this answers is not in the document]
jordanarldt: Good call!

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
Comment (jordanarldt) [reply core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:106]: Fair enough!

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
Comment (jordanarldt) [reply core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:170]: @chanceaclark It looks like we're already enforcing this pattern on the login page:
https://github.com/bigcommerce/catalyst/commit/a739f01d0b91c549dae15a1fc350a388c98bbe56#diff-2b5f4d2b2e234bc3346764b7e78954e330e60912ad21040b916b79df65d1fbb7R37-R39

Are we wanting to do this pattern everywhere that we're constructing something to pass to the redirectTo param? I think ultimately it should be fine here since we're explicitly setting a relative path on the formdata that gets passed into this function

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
Comment (chanceaclark) [reply core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts]: It's more for consistency across the whole repo.

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
Comment (chanceaclark) [reply core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:166]: We should be protecting against this at the boundary. The page is fine, but if someone else consumes this server action (which would be rare), they could bypass our protections.

core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts
Comment (jordanarldt) [reply core/app/[locale]/(default)/product/[slug]/_actions/wishlist-action.ts:290]: @chanceaclark Actually, I re-read this and realized I might have misunderstood the original request. I was also under the impression that the conform-to library would fail if there's additional fields that aren't part of the schema, but that's not the case 😄  we can just append to the existing formData and it works. I'll fix this

jorgemoya (approved):
Since wishlist isn't generic enough to be added to the vibes repo, I approve the current implementation of the wishlist button as a slot that can be passed in.

PR #2226, opened 2025-04-13
Add operation name and type to graphQL URL for easier logging

bookernath:
## What/Why?
Adds the name of the GraphQL operation as a query parameter on the GraphQL URL, such as:

`/graphql?operation=HomePageQuery&type=query`

This improves observability in server logs, making it easier to tell queries apart without parsing the body.

In the future we may add `locale` as well.

## Testing
Tested on preview deployment by observing logs.


migueloller (approved):
+1. We need better logging overall and this is a step in the right direction

PR #2231, opened 2025-04-14
feat(core): B2B-2562 cart id as optional search param in checkout route

bc-victor:
## What/Why?
Adding cart id as an optional search param in the checkout route - this will help with B2B payments of quotes and invoices since the buyer portal creates carts and checks out in the same submit event, this leads to issues with passing a cart id to the session in between. 

## Testing
Tested in B2B while paying a quote and redirecting to catalyst checkout route.

## Migration
No migration changes

### DIFF
--- a/core/app/[locale]/(default)/checkout/route.ts
+++ b/core/app/[locale]/(default)/checkout/route.ts
@@ -25,9 +25,12 @@ const CheckoutRedirectMutation = graphql(`
   }
 `);
 
-export async function GET(_: NextRequest, { params }: { params: Promise<{ locale: string }> }) {
+export async function GET(req: NextRequest, { params }: { params: Promise<{ locale: string }> }) {
   const { locale } = await params;
-  const cartId = await getCartId();
+  const url = new URL(req.url);
+  const searchParams = url.searchParams;
+  const paramsCartId = searchParams.get('cartId'); 
+  const cartId = paramsCartId || await getCartId();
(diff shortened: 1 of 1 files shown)


core/app/[locale]/(default)/checkout/route.ts
```diff
@@ -25,9 +25,12 @@ const CheckoutRedirectMutation = graphql(`
...
 `);
 
-export async function GET(_: NextRequest, { params }: { params: Promise<{ locale: string }> }) {
+export async function GET(req: NextRequest, { params }: { params: Promise<{ locale: string }> }) {
   const { locale } = await params;
-  const cartId = await getCartId();
+  const url = new URL(req.url);
+  const searchParams = url.searchParams;
+  const paramsCartId = searchParams.get('cartId'); 
+  const cartId = paramsCartId || await getCartId();
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/checkout/route.ts:33]: This can be simplified
```suggestion
  const cartId = req.nextUrl.searchParams.get('cartId') ?? await getCartId();
```

chanceaclark (commented):
For additional context about this change, @icatalina and I chatted that this is a native feature built into Stencil today that allows carts/checkouts to be created on the fly without changing the existing session. This is a valid feature and I thought it would be a good add to this route handler.

chanceaclark:
@bc-victor please run `pnpm lint --fix` locally to fix linting issues.

PR #2236, opened 2025-04-15
fix: remove featured products from 404 page, allowing the page to be static

bookernath:
## What/Why?
We'd like to move in the direction of a static 404 page with just a search box instead of a featured products panel, in order to make 404 pages inexpensive to serve. This is particularly important because of [this issue.](https://github.com/vercel/next.js/issues/67532#issuecomment-2688292808)

This is the first step of that change which removes the featured products panel, the search box can be added in a separate PR.

## Testing
Preview deployment

image

### DIFF
--- a/core/vibes/soul/sections/not-found/index.tsx
+++ b/core/vibes/soul/sections/not-found/index.tsx
@@ -8,7 +8,7 @@ export function NotFound({
   subtitle = "Take a look around if you're lost.",
 }: Props) {
   return (
-    <section className="@container">
+    <section className="@container pb-20">
       <div className="mx-auto max-w-3xl px-4 pt-10 @xl:px-6 @xl:pt-14 @4xl:px-8 @4xl:pt-20">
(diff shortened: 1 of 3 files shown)


core/vibes/soul/sections/not-found/index.tsx
```diff
@@ -8,7 +8,7 @@ export function NotFound({
   subtitle = "Take a look around if you're lost.",
 }: Props) {
   return (
-    <section className="@container">
+    <section className="@container pb-20">
```
migueloller, line 11: This change has a high risk of being blasted a way by pulling in updates from VIBES. We might want to handle it elsewhere unless you think that we should make this change upstream as well @bookernath?

migueloller (approved):
Changes look good to me. Have we confirmed that this does actually make the page static and we won't need to add anything else, like a `generateStaticParams`?

migueloller:
Looks like we're good

CleanShot 2025-04-15 at 10 57 31

jorgemoya (approved):
Always dreamed of this 🙏

PR #2238, opened 2025-04-15
fix(core): revert streamings fields in product form

jorgemoya:
## What/Why?
Fixes an issue of not being able to add product to cart when you redirect to product via navigation.

## Testing
After the fix, you can add product to cart.

## Migration
Remove changes to UI related to #2191 .

### DIFF
--- a/.changeset/clean-humans-wonder.md
+++ b/.changeset/clean-humans-wonder.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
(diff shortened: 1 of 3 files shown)


migueloller (approved):
@jorgemoya, what was the root cause of the problem?

jorgemoya:
> @jorgemoya, what was the root cause of the problem?

Undiscovered by now, I was just able to fix it removing the streaming of fields. I will need more time to debug and reproduce in a clean environment to see if it is an issue with our form library or something else.

PR #2241, opened 2025-04-16
Detect FinalizationRegistry and WeakRef features and fallback if they are not available

bookernath:
## What/Why?
Per title

Not all environments support FinalizationRegistry and WeakRef; this change to the Streamable pattern uses feature detection and a fallback in this case. This has the tradeoff of potentially leaking memory, so we use an LRU approach to evict the cache naively as it grows large.

## Testing
In addition to the preview deployment on this PR, I created an arbitrary deployment where I forced LRU on and set `maxSize` to `2` to test exhaustion behavior, and I observed no breakage.

https://catalyst-canary-p74486ofb-bigcommerce-platform.vercel.app/

### DIFF
--- a/.changeset/chatty-keys-cover.md
+++ b/.changeset/chatty-keys-cover.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Provide fallback behavior for environments where FinalizationRegistry and WeakRef are not available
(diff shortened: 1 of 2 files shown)


migueloller (approved):
LGTM! 👍🏻

@bookernath, do you mind contributing back to https://github.com/makeswift/vibes?

PR #2245, opened 2025-04-17
refactor(core): home, header and footer for dynamicIO

jorgemoya:
## New

This refactor optimizes home page, header, and footer for caching and the eventual use of dynamicIO. With these changes we leverage data caching to hit mostly cache data for guest shoppers in different locales and with different currencies.

## Key modifications include:

- Header and Footer now have a blocking request for the shared data that is the same for all users.
- Data that can change for logged in users is now a separate request.
- Dynamic fetches (using customerAccessToken or preferred currency) are now all streaming queries.
- Update Header UI component to allow streaming in of currencies data.

## Migration instructions:

- Renamed `/app/[locale]/(default)/query.ts` to `/app/[locale]/(default)/page-data.ts`, include page query on this page.
- Updated `/app/[locale]/(default)/page.ts` to use `Streamable.from` pattern.
- Split data that can vary by user from `core/components/footer/fragment.ts` and `core/components/header/fragment.ts`
- Updated `core/components/header/index.tsx` and `core/components/footer/index.tsx` to fetch shared data in a blocking request and pass data that varies by customer as streamable data. Updated to use the new `Streamable.from` pattern.

## Testing
https://www.webpagetest.org/video/compare.php?tests=250417_AiDcJ9_97Z,250417_BiDc20_8XM

Screenshot 2025-04-17 at 3 20 40 PM


core/components/header/index.tsx
```diff
@@ -143,16 +144,17 @@ export const Header = async () => {
...
         mobileMenuTriggerLabel: t('toggleNavigation'),
         openSearchPopupLabel: t('Icons.search'),
         logoLabel: t('home'),
-        cartCount: Streamable.from(getCartCount),
+        cartCount: streamableCartCount,
         activeLocaleId: locale,
         locales,
         currencies,
-        activeCurrencyId,
+        // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
```
migueloller, line 156: Why not update the component so that it accepts a streamable for `activeCurrencyId`?

migueloller (approved):
LGTM 👍🏻 

Just have 1 minor comment

core/components/header/index.tsx
jorgemoya: Good catch! Missed this.

PR #2246, opened 2025-04-17
feat(core): add new analytics provider

chanceaclark:
## What/Why?
Adds a new analytics provider meant to replace the other provider. This provider is built being framework agnostic but exposes a react provider to use within context. The initial implementation comes with a Google Analytics provider with some basic events to get started. We need to add some other events around starting checkout, banners, consent loading, and search.

**Upcoming PRs:**
- Consuming the new analytics provider and firing events
- Removing the old analytics provider

## Testing
Sample test from removing a product from the cart.

Screenshot 2025-04-17 at 16 08 46

## Migration
This is an additive change so no migration is needed.


jorgemoya (approved):

core/lib/analytics/react/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+const AnalyticsContext = createContext<Analytics | null>(null);
+
+interface AnalyticsProviderProps {
+  analytics: Analytics | null;
+  children: React.ReactNode;
+}
+
+export const AnalyticsProvider = ({ children, analytics }: AnalyticsProviderProps) => {
+  useEffect(() => {
+    analytics?.initialize();
```
migueloller, line 16: Should we have `initialize` return a cleanup function or is that something that most analytics providers don't provide? I'm guessing that it being a singleton covers any reinitialization issue that could occur?

migueloller (approved):
Looks great! Did we already have a provider set up or is the plan to add this in a separate PR?

chanceaclark:
>Did we already have a provider set up or is the plan to add this in a separate PR?

I'm going to add it all in a separate PR. It was a lot for one PR: https://github.com/bigcommerce/catalyst/commit/8f2d4923d1164922f543357570eb380857bca3c9

core/lib/analytics/react/index.tsx
chanceaclark: I didn't find a need so left it out. If we do see a situation where a cleanup is need I would add it, but yeah the singleton + browsers having to remount analytics solves most of our use-cases.

PR #2252, opened 2025-04-21
feat: bump makeswift runtime to 0.24.0, simplify implementation

arvinpoddar:
## What/Why?

Upgrades `@makeswift/runtime`, which resolves an issue with editing SSGd Catalyst pages in Makeswift.

## Testing

Manually tested locally and via Vercel deployments

### DIFF
--- a/core/middlewares/with-makeswift.ts
+++ b/core/middlewares/with-makeswift.ts
@@ -1,27 +1,18 @@
-import { unstable_createMakeswiftDraftRequest } from '@makeswift/runtime/next/middleware';
+import { unstable_isDraftModeRequest } from '@makeswift/runtime/next/middleware';
 
 import { MiddlewareFactory } from './compose-middlewares';
 
-if (!process.env.MAKESWIFT_SITE_API_KEY) {
-  throw new Error('MAKESWIFT_SITE_API_KEY is not set');
-}
-
-const MAKESWIFT_SITE_API_KEY = process.env.MAKESWIFT_SITE_API_KEY;
-
 export const withMakeswift: MiddlewareFactory = (middleware) => {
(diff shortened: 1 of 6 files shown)


core/middlewares/with-makeswift.ts
migueloller, line 1: Since we aren't creating a new request where we strip out the headers anymore, couldn't we completely remove this middleware and then in the localization middleware handle this logic?

core/next.config.ts
```diff
@@ -8,7 +8,7 @@ import { client } from './client';
 import { graphql } from './client/graphql';
 import { cspHeader } from './lib/content-security-policy';
 
-const withMakeswift = createWithMakeswift({ previewMode: false });
+const withMakeswift = createWithMakeswift({ previewMode: true });
```
migueloller, line 11: Should we remove the option altogether given it's optional?

arvinpoddar: Sure!

core/middlewares/with-makeswift.ts
arvinpoddar: Good point. My instincts are to lean away from having our changes bleed into middleware that isn't ours, but that ship has already sailed. I'll update to move this into `with-intl`.

migueloller: Yeah, I can see the benefit of keeping concerns separate there. No need to move it there if you feel like it's better to have a separation like this

arvinpoddar: Cool, if it has your cosign then I'd like to keep it. It's been helpful in the past to have this decorator, and it may be helpful in the future.

bookernath (approved):

PR #2259, opened 2025-04-22
Update GQL client and auth middleware to handle invalid tokens and invalidate session

jordanarldt:
## What/Why?
- Update `withAuth` middleware to redirect protected routes to `/login` page when user is unauthorized
- Add customer access token validation on protected routes
- Minor updates to end-user error messages on unauthorized GQL requests to stop returning raw GQL error and use our translated error message instead.
- Add `onError` event to GraphQL client
- Use the `onError` callback to invalidate user session when GQL returns an invalid or missing token error

## Testing

https://github.com/user-attachments/assets/d3c519dd-1655-4a15-8801-cd0ef368711d

## Migration

1. Copy all changes from the `/core/client` directory and the `/packages/client` directory
2. Copy logic from withAuth middleware to handle the `?invalidate-session` query param
3. Copy translation values
4. Copy all changes from the `/core/app/[locale]/(default)/account/` directory server actions
5. Copy all changes from the `/core/app/[locale]/(default)/checkout/route.ts` file

### DIFF
--- a/core/app/[locale]/(default)/account/orders/page-data.ts
+++ b/core/app/[locale]/(default)/account/orders/page-data.ts
@@ -94,7 +94,7 @@ export const getCustomerOrders = cache(
       variables: { ...paginationArgs, ...filtersArgs },
       customerAccessToken,
       fetchOptions: { cache: 'no-store', next: { tags: [TAGS.customer] } },
-      errorPolicy: 'ignore',
+      errorPolicy: 'auth',
     });
 
     const orders = response.data.customer?.orders;
--- a/core/app/[locale]/(default)/account/wishlists/_actions/change-wishlist-visibility.ts
+++ b/core/app/[locale]/(default)/account/wishlists/_actions/change-wishlist-visibility.ts
@@ -1,6 +1,6 @@
 'use server';
 
-import { BigCommerceAPIError, BigCommerceGQLError } from '@bigcommerce/catalyst-client';
+import { BigCommerceAuthError } from '@bigcommerce/catalyst-client';
 import { SubmissionResult } from '@conform-to/react';
 import { parseWithZod } from '@conform-to/zod';
 import { revalidateTag } from 'next/cache';
@@ -73,28 +73,22 @@ export async function toggleWishlistVisibility(
     // eslint-disable-next-line no-console
     console.error(error);
 
-    if (error instanceof BigCommerceGQLError) {
+    if (error instanceof BigCommerceAuthError) {
+      const authErrorMessage = t('Errors.unauthorized');
+
       return {
         ...prevState,
-        lastResult: submission.reply({ formErrors: [t('Errors.unexpected')] }),
-        errorMessage: error.message.includes('Please sign in')
-          ? t('Errors.unauthorized')
-          : t('Errors.unexpected'),
+        lastResult: submission.reply({ formErrors: [authErrorMessage] }),
+        errorMessage: authErrorMessage,
       };
     }
 
-    if (error instanceof BigCommerceAPIError) {
-      return {
-        ...prevState,
-        lastResult: submission.reply({ formErrors: [t('Errors.unexpected')] }),
-        errorMessage: t('Errors.unexpected'),
-      };
-    }
+    const errorMessage = t('Errors.unexpected');
 
     return {
       ...prevState,
-      lastResult: submission.reply({ formErrors: [t('Errors.unexpected')] }),
-      errorMessage: t('Errors.unexpected'),
+      lastResult: submission.reply({ formErrors: [errorMessage] }),
+      errorMessage,
     };
   }
 }
--- a/core/middlewares/with-auth.ts
+++ b/core/middlewares/with-auth.ts
@@ -1,20 +1,51 @@
-import { auth, signIn } from '~/auth';
+import { NextResponse, URLPattern } from 'next/server';
+
+import { auth, signIn, signOut } from '~/auth';
 
 import { type MiddlewareFactory } from './compose-middlewares';
+
+// Path matcher for any routes that require authentication
+const protectedPathPattern = new URLPattern({ pathname: '/(account)/*' });
+const sessionInvalidateParam = 'invalidate-session';
+
+function redirectToLogin(url: string) {
+  return NextResponse.redirect(new URL('/login', url), { status: 302 });
+}
 
 export const withAuth: MiddlewareFactory = (next) => {
   return async (request, event) => {
     // @ts-expect-error: The `auth` function doesn't have the correct type to support it as a MiddlewareFactory.
     const authWithCallback = auth(async (req) => {
+      const isProtectedRoute = protectedPathPattern.test(req.nextUrl.toString().toLowerCase());
+      const isGetRequest = req.method === 'GET';
+
       if (!req.auth) {
         await signIn('anonymous', { redirect: false });
 
+        if (isProtectedRoute && isGetRequest) {
+          return redirectToLogin(req.url);
+        }
+
         return next(req, event);
+      }
+
+      const { customerAccessToken } = req.auth.user ?? {};
+
+      if (isProtectedRoute && isGetRequest && !customerAccessToken) {
+        return redirectToLogin(req.url);
       }
 
       // Continue the middleware chain
       return next(req, event);
     });
+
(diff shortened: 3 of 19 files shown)


core/middlewares/with-auth.ts
```diff
@@ -1,17 +1,68 @@
...
+function redirectToLogin(url: string) {
+  return NextResponse.redirect(new URL('/login', url), { status: 302 });
+}
+
 export const withAuth: MiddlewareFactory = (next) => {
   return async (request, event) => {
     // @ts-expect-error: The `auth` function doesn't have the correct type to support it as a MiddlewareFactory.
     const authWithCallback = auth(async (req) => {
+      const isProtectedRoute = protectedPathRegExp.test(req.nextUrl.pathname);
+      const isGetRequest = req.method === 'GET';
```
jordanarldt, line 37: We only redirect on `GET` requests to avoid breaking any server actions.

core/middlewares/with-auth.ts
```diff
@@ -1,17 +1,68 @@
...
+          return redirectToLogin(req.url);
+        }
+
         return next(req, event);
       }
 
+      const { email, customerAccessToken } = req.auth.user ?? {};
+
+      if (customerAccessToken && isProtectedRoute) {
+        // To avoid unnecessary expense, only check token validity on routes that require authentication
```
chanceaclark, line 52: Technically this doesn't just apply to account related routes but also products and categories. Since we don't want to add additional overhead fetches, and we do some fetching in the `with-routes` middleware, I am tempted that we try to handle an invalid token there 🤔

core/app/[locale]/(default)/account/addresses/_actions/create-address.ts
```diff
@@ -249,11 +249,13 @@ export async function createAddress(prevState: Awaited<State>, formData: FormDat
...
+      const formErrors = error.message.includes('Please sign in')
+        ? [t('unauthorized')]
+        : error.errors.map(({ message }) => message);
+
       return {
         ...prevState,
-        lastResult: submission.reply({
-          formErrors: error.errors.map(({ message }) => message),
-        }),
+        lastResult: submission.reply({ formErrors }),
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/account/addresses/_actions/create-address.ts:258]: All these changes to check to see if the message includes `Please sign in` indicates to me that we probably want to return an error from the client instead, e.g. `BigCommerceAuthorizationError`. Can you explore that a bit more?

core/middlewares/with-auth.ts
```diff
@@ -1,17 +1,68 @@
...
+
+function redirectToLogin(url: string) {
+  return NextResponse.redirect(new URL('/login', url), { status: 302 });
+}
+
 export const withAuth: MiddlewareFactory = (next) => {
   return async (request, event) => {
     // @ts-expect-error: The `auth` function doesn't have the correct type to support it as a MiddlewareFactory.
     const authWithCallback = auth(async (req) => {
+      const isProtectedRoute = protectedPathRegExp.test(req.nextUrl.pathname);
```
chanceaclark, line 36: Instead of using regex, could we try using [`URLPattern`](https://developer.mozilla.org/en-US/docs/Web/API/URLPattern/URLPattern) instead? If that is a bit cumbersome here (since you'd need to pass in a hostname when constructing the object), then maybe we can just use hardcoded paths in an array.

core/app/[locale]/(default)/account/addresses/_actions/create-address.ts
Comment (jordanarldt) [reply core/app/[locale]/(default)/account/addresses/_actions/create-address.ts:258]: @chanceaclark Oh definitely, good call there. Having it handled by the client would be much better. I'll explore this 👍

core/middlewares/with-auth.ts
chanceaclark: After chatting, we should be doing this check at the DAL and middleware should be specific to protecting the routes if it's authed or not.

jordanarldt: @chanceaclark For sure. What concerns me there is that it's possible for the `customerAccessToken` to exist, even though it's invalid. This would allow access to protected routes and you would just end up at a 404 page again like before.

In order to avoid that, we would still need the token validation here so that we can redirect to the login page if your current token isn't valid 🤔 Unless we are able to remove the customer access token at the DAL before the request reaches middleware, but I'm not sure if that would be possible.

chanceaclark: I think the DAL should be the one to handle if a user has the ability to access data, if not log them out to clear the session. Middleware should be the one to basically check if the session exists or not and allow users to navigate there or not. Like mentioned in standup, I think our Storefront GQL should return an error (with partial data) if using an invalid customer access token with the request. I think it's a good improvement to consumers of our API too here.

packages/client/src/client.ts
```diff
@@ -74,6 +85,7 @@ class Client<FetcherRequestInit extends RequestInit = RequestInit> {
     fetchOptions?: FetcherRequestInit;
     channelId?: string;
     errorPolicy?: 'none' | 'all' | 'ignore';
+    validateCustomerAccessToken?: boolean;
```
jorgemoya, line 88: Is there a reason why this is optional?

packages/client/src/client.ts
```diff
@@ -23,6 +25,10 @@ interface Config<FetcherRequestInit extends RequestInit = RequestInit> {
   beforeRequest?: (
     fetchOptions?: FetcherRequestInit,
   ) => Promise<Partial<FetcherRequestInit> | undefined> | Partial<FetcherRequestInit> | undefined;
+  onError?: (
+    error: BigCommerceGQLError,
+    queryType: 'query' | 'mutation' | 'subscription',
+  ) => Promise<void> | void;
```
jorgemoya, line 31: Is this an existing pattern or something we built for this use case?

core/app/[locale]/(default)/account/wishlists/_actions/change-wishlist-visibility.ts
```diff
@@ -73,28 +73,22 @@ export async function toggleWishlistVisibility(
...
-    if (error instanceof BigCommerceGQLError) {
-      return {
-        ...prevState,
-        lastResult: submission.reply({ formErrors: [t('Errors.unexpected')] }),
-        errorMessage: error.message.includes('Please sign in')
-          ? t('Errors.unauthorized')
-          : t('Errors.unexpected'),
-      };
-    }
+    if (error instanceof BigCommerceAuthError) {
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/account/wishlists/_actions/change-wishlist-visibility.ts:76]: Are we just expecting an error of auth to be present here? GQL will never return a custom error?

packages/client/src/client.ts
jordanarldt: @jorgemoya Yep, it's optional so that we don't have to update every single `client.fetch` in Catalyst to get this behavior. It's optional, but defaults to `true`, so if we want to override the behavior somewhere, we can.

packages/client/src/client.ts
jordanarldt: It's built for this use-case. As @chanceaclark  was saying, we need the auth failure handling to be part of the DAL, which will require having this `onError` handler built into the client. 👍

core/app/[locale]/(default)/account/wishlists/_actions/change-wishlist-visibility.ts
Comment (jordanarldt) [reply core/app/[locale]/(default)/account/wishlists/_actions/change-wishlist-visibility.ts:76]: Correct. Changing the wishlist visibility is an action that requires 0 user input, so the only point of failures are auth, or an internal error

core/app/[locale]/(default)/account/orders/page-data.ts
```diff
@@ -94,7 +94,6 @@ export const getCustomerOrders = cache(
       variables: { ...paginationArgs, ...filtersArgs },
       customerAccessToken,
       fetchOptions: { cache: 'no-store', next: { tags: [TAGS.customer] } },
-      errorPolicy: 'ignore',
```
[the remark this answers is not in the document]
Comment (jordanarldt) [reply core/app/[locale]/(default)/account/orders/page-data.ts:97]: @jorgemoya Ahh I see. I wasn't sure why that was there, but that makes sense. I'll fix this.

packages/client/src/client.ts
jorgemoya: Yeah I guess my question was more in the sense of why would we want to override it but having the optional override shouldn't hurt.

jorgemoya (approved):

PR #2268, opened 2025-04-24
Make parity with BigCommerce 301 redirect behavior

jordanarldt:
## What/Why?
Ensure Catalyst 301 redirects has parity with Stencil 301 redirects. Currently, 301 redirects do not pass query params in Catalyst, but they should.

Query params will carry with a 301 redirect under these conditions:
- If the target URL is **not** an external link
- If the Old URL does not already match a query param

### Redirect setup for all examples:
image

### Query params are passed for manual redirects:

https://github.com/user-attachments/assets/566d9679-105b-44a3-bb30-b46f7ac4af36

### 301 redirects for specific query params are matched:

https://github.com/user-attachments/assets/b6710fce-446b-4967-a227-863f6b5c69f2

### Dynamic 301 redirects pass query params:

https://github.com/user-attachments/assets/78aa6570-a713-49af-80a9-0af1e6c01e7c

### External redirects do **not** pass query params:

https://github.com/user-attachments/assets/e7add212-f94f-43b7-b42e-01857d43631f

## Migration
1. Copy changes to GraphQL query in `with-routes.ts`
2. Make the necessary zod schema updates for the modified schema fields in the `RedirectSchema` variable
3. Copy the redirect logic


core/middlewares/with-routes.ts
```diff
@@ -282,21 +289,36 @@ export const withRoutes: MiddlewareFactory = () => {
     };
 
     if (route?.redirect) {
+      // Only carry over query params if the fromPath does not have any, as Bigcommerce 301 redirects support matching by specific query params.
+      const fromPathHasParams = new URL(route.redirect.fromPath, request.url).search.length > 0;
+      const searchParams = fromPathHasParams ? '' : request.nextUrl.search;
```
chanceaclark, line 294: I try to lean on boolean variables name as little as possible. I get they make hard to read logic more readable, but they generally tend to only be used once or twice.
```suggestion
      // Only carry over query params if the fromPath does not have any, as Bigcommerce 301 redirects support matching by specific query params.
      const fromPathSearchParams = new URL(route.redirect.fromPath, request.url).search;
      const searchParams = fromPathSearchParams.length > 0 ? '' : request.nextUrl.search;
```

core/middlewares/with-routes.ts
```diff
@@ -282,21 +289,36 @@ export const withRoutes: MiddlewareFactory = () => {
...
+        }
+
+        case 'ManualRedirect': {
+          // For manual redirects, to.url will be a relative path if it is an internal redirect and an absolute URL if it is an external redirect.
+          // URL constructor will correctly handle both cases.
+          // If the manual redirect is an external URL, we should not carry query params.
+          const isExternalRedirect =
+            new URL(route.redirect.to.url, request.url).origin !== request.nextUrl.origin;
+          const manualSearchParams = isExternalRedirect ? '' : searchParams;
+          const redirectUrl = new URL(route.redirect.to.url + manualSearchParams, request.url);
```
chanceaclark, line 315: Similar thing here, but just to make it more concise.
```suggestion
          const redirectUrl = new URL(route.redirect.to.url, request.url);

          if (redirectUrl.origin === request.nextUrl.origin) {
            redirectUrl.search = searchParams;
          }
```

core/middlewares/with-routes.ts
```diff
@@ -221,10 +226,12 @@ const getRouteInfo = async (request: NextRequest, event: NextFetchEvent) => {
   const channelId = request.headers.get('x-bc-channel-id') ?? '';
 
   try {
-    const pathname = clearLocaleFromPath(request.nextUrl.pathname, locale);
+    // For route resolution parity, we need to also include query params, otherwise certain redirects will not work.
+    const { pathname, search } = request.nextUrl;
+    const path = clearLocaleFromPath(pathname + search, locale);
```
chanceaclark, line 231: Just so we can limit the changes in this file, we can probably just inline this.
```suggestion
    // For route resolution parity, we need to also include query params, otherwise certain redirects will not work.
    const pathname = clearLocaleFromPath(request.nextUrl.pathname + request.nextUrl.search, locale);
```

PR #2269, opened 2025-04-25
fix: toast & cart updates on "add to cart"

fikrikarim:
## What/Why?

React / Next.js bug remounts a memoised client component rendered directly by an RSC. This caused the `ProductDetail` to remount before it can show the toast and update the cart.

https://github.com/vercel/next.js/issues/44901
https://github.com/vercel/next.js/issues/73507

### Fix
Introduce a non-memo shims around `MakeswiftComponent` and `MakeswiftPage` that prevent the unwanted remount.

## Testing

https://github.com/user-attachments/assets/23831b75-2d5e-45ab-83be-2dbe67434d9f

Test on the preview branch vs the `integrations/makeswift` branch:

https://catalyst-canary-git-fikri-eng-8075-3c0f42-bigcommerce-platform.vercel.app/smith-journal-13/

https://integrations-makeswift.catalyst-canary.store/spray-bottle/

### DIFF
--- a/.changeset/seven-banks-bow.md
+++ b/.changeset/seven-banks-bow.md
@@ -1,0 +1,8 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Fix: missing **Add-to-cart** feedback on `integrations/makeswift` branch:
+
+- Toast success notification now appears when **Add to cart** is clicked.
+- Cart button badge in the header now updates to show `1` when the first item is added.
--- a/core/lib/makeswift/component.tsx
+++ b/core/lib/makeswift/component.tsx
@@ -1,6 +1,6 @@
-import { MakeswiftComponent } from '@makeswift/runtime/next';
-
 import { getComponentSnapshot } from '~/lib/makeswift/client';
+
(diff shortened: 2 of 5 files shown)


chanceaclark (approved):
Not a fan that this is the fix for the issue, but I guess it fine. Thanks for digging into this and finding the issue!

fikrikarim:
> Not a fan that this is the fix for the issue, but I guess it fine. Thanks for digging into this and finding the issue!

The proper fix is to fix the issue on the Next.js/React codebase 😄

migueloller:
> Not a fan that this is the fix for the issue, but I guess it fine. Thanks for digging into this and finding the issue!

I'm not a fan either 😞

Seems like something we might want to internalize it in `@makeswift/runtime`

migueloller:
Right, but we can expect users to use `MakeswiftComponent` with RSCs. I think it would be a good shim to add if it doesn't add much overhead. We're also planning to have an RSC version of `MakeswiftComponent` that internally calls `getComponentSnapshot`, perhaps we could use the shim there. Regardless, I think that if we know what causes the issue and we can solve it for the user without a bunch of negative consequences, it might well be worth it

PR #2275, opened 2025-04-29
Revert "refactor(core): move auth related route handlers under the (auth) group (#2179)

bookernath:
This reverts commit c09566368830b6a024d17c464fe8a594c20c5507.

## What/Why?
This re-organization of code broke `/login/token` by moving it under the locale path, which has an undesirable interplay with the i18n middleware. Reverting for now to fix the customer login API route.

## Testing
Preview deployment

## Migration
N/A


migueloller (approved):
LGTM 👍🏻 

Only weirdness is how the changeset is also being reverted vs it being a new changeset with the stated revert reasons.

jordanarldt (approved):

migueloller (approved):

PR #2282, opened 2025-05-01
fix(core): check for error.type for login auth error

jorgemoya:
## What/Why?
Check for `error.type` instead of `error.name` auth error in Login, since `error.name` gets minified in production and the check never returns `true`. Additionally, add a check for the `cause.err` to be of type `BigcommerceGQLError`.

## Testing
In production:
Screenshot 2025-05-01 at 11 31 21 AM

## Migration
Change `error.name === 'CallbackRouteError'` to `error.type === 'CallbackRouteError'` check in the error handling of the login action and include `error.cause.err instanceof BigCommerceGQLError`.

### DIFF
--- a/core/app/[locale]/(default)/(auth)/login/_actions/login.ts
+++ b/core/app/[locale]/(default)/(auth)/login/_actions/login.ts
@@ -45,6 +45,7 @@ export const login = async (
 
     if (
       error instanceof AuthError &&
+      error.type === 'CallbackRouteError' &&
       error.cause &&
       error.cause.err?.message.includes('Invalid credentials')
     ) {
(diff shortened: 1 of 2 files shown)


core/app/[locale]/(default)/(auth)/login/_actions/login.ts
```diff
@@ -45,7 +45,6 @@ export const login = async (
 
     if (
       error instanceof AuthError &&
-      error.name === 'CallbackRouteError' &&
```
Comment (migueloller) [inline core/app/[locale]/(default)/(auth)/login/_actions/login.ts:48]: I inspected the code a bit and the appropriate thing to do here is this:

```ts
error.type === 'CallbackRouteError'
```

migueloller (changes requested):

migueloller (approved):

PR #2286, opened 2025-05-01
feat: translate bigcommerce product data in makeswift editor

migueloller:
Jira: [CATALYST-1134](https://bigcommercecloud.atlassian.net/browse/CATALYST-1134)

## What/Why?

Users noticed that while working within the Makeswift visual page builder, storefront content fetched from BigCommerce, such as product data for the product carousel, was not translated.

Ultimately, the root cause of the user seeing non-translated data while in the builder came down to the fact that while using the Makeswift visual page builder, we disable the [automatic locale detection feature](https://next-intl.dev/docs/routing/middleware#locale-detection) provided by `next-intl`:
https://github.com/bigcommerce/catalyst/blob/74eae822d70ad1f6f205c822e4b94fb12dd9af16/core/middlewares/with-makeswift.ts#L10-L13

With that setting disabled, making a request to `fetch("/api/products")` will never be automatically translated, because if middleware were to intercept the request, it wouldn't be able to rely on an existing `NEXT_LOCALE` cookie, or `accept-language` header. Therefore, you would need to manually modify the URL you are fetching to `fetch(\`/${locale}/api/products\`)`. 

Since we have to modify the URL manually anyway, we opted to follow the `next-intl` [recommendation](https://next-intl.dev/docs/environments/actions-metadata-route-handlers#route-handlers) and attach `locale` as a search parameter for client component API requests.

## Testing

Before:

https://github.com/user-attachments/assets/03f8e24e-0b3f-4df3-ba9f-6586cf224ee4

After: 

https://github.com/user-attachments/assets/42d0c0b6-b07d-4d04-8702-28b4b18313e7

## Migration

1. Collect all `route.ts` files that serve responses to requests that originate from Makeswift components. In the default `integrations/makeswift` branch, that list is:
   a. `core/app/api/products/[entityId]/route.ts`
   b. `core/app/api/products/group/[group]/route.ts`
   c. `core/app/api/products/ids/route.ts`
2. Add the code to each `route.ts` file required to parse and validate that locale. That code typically looks like:
   ```
   // ...
   import { hasLocale } from 'next-intl';
   import { routing } from '~/i18n/routing';

   // Then, inside of the route hander:
     const searchParams = request.nextUrl.searchParams;
     const locale = searchParams.get('locale') ?? routing.defaultLocale;

     if (!hasLocale(routing.locales, locale)) {
       return NextResponse.json(
         { status: 'error', error: 'Invalid locale parameter' },
         { status: 400 },
       )
     }
   ```
3. At this point, you now have the locale available in your API route to do whatever you'd like with; in this PR, we pass locale to `client.fetch` and add some conditional logic inside of `fetchOptions` to set the `"Accept-Language"` header based on locale.
4. Finally, don't forget to pass locale as a search parameter from client `fetch()` calls!

[CATALYST-1134]: https://bigcommercecloud.atlassian.net/browse/CATALYST-1134?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/app/api/products/route.ts
chanceaclark, line 1: Was this unused?

chanceaclark (approved):

core/app/api/products/route.ts
matthewvolk: Converted to server action 😄

chanceaclark: Ooh nice, didn't realize that.

PR #2293, opened 2025-05-07
feat(core): accept a list of CDNs for image loader

jorgemoya:
## What/Why?
Allow a list of CDN hostnames for cases when there can be more than one CDN available for image loader.

## Testing
Locally, I run two different CDNs and get no image loader errors.

## Migration
- Update `build-config` schema to make `cndUrls` an array of strings.
- Update `next.config.ts` to set `cdnUrls` as an array, and set multiple preconnected Link headers (one per CDN).
- `shouldUseLoaderProp` function now reads from array.

### DIFF
--- a/core/next.config.ts
+++ b/core/next.config.ts
@@ -38,7 +38,9 @@ async function writeSettingsToBuildConfig() {
     locales: data.site.settings?.locales,
     urls: {
       ...data.site.settings?.url,
-      cdnUrl: process.env.NEXT_PUBLIC_BIGCOMMERCE_CDN_HOSTNAME ?? data.site.settings?.url.cdnUrl,
+      cdnUrls: process.env.NEXT_PUBLIC_BIGCOMMERCE_CDN_HOSTNAME
+        ? process.env.NEXT_PUBLIC_BIGCOMMERCE_CDN_HOSTNAME.split(',').map((s) => s.trim())
+        : [data.site.settings?.url.cdnUrl],
     },
   });
 }
@@ -76,6 +78,11 @@ export default async (): Promise<NextConfig> => {
     trailingSlash: process.env.TRAILING_SLASH !== 'false',
     // eslint-disable-next-line @typescript-eslint/require-await
     async headers() {
+      const cdnLinks = settings.urls.cdnUrls.map((url) => ({
+        key: 'Link',
+        value: `<https://${url}>; rel=preconnect`,
+      }));
+
       return [
         {
           source: '/(.*)',
@@ -84,10 +91,7 @@ export default async (): Promise<NextConfig> => {
               key: 'Content-Security-Policy',
               value: cspHeader.replace(/\n/g, ''),
             },
-            {
-              key: 'Link',
-              value: `<https://${settings.urls.cdnUrl}>; rel=preconnect`,
-            },
+            ...cdnLinks,
           ],
         },
       ];
--- a/.changeset/dull-chairs-invite.md
+++ b/.changeset/dull-chairs-invite.md
@@ -1,0 +1,11 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Allow a list of CDN hostnames for cases when there can be more than one CDN available for image loader.
+
+Migration:
+
+- Update `build-config` schema to make `cndUrls` an array of strings.
+- Update `next.config.ts` to set `cdnUrls` as an array, and set multiple preconnected Link headers (one per CDN).
+- `shouldUseLoaderProp` function now reads from array.
--- a/core/build-config/schema.ts
+++ b/core/build-config/schema.ts
@@ -9,7 +9,7 @@ export const buildConfigSchema = z.object({
   ),
   urls: z.object({
     vanityUrl: z.string(),
-    cdnUrl: z.string().default('cdn11.bigcommerce.com'),
+    cdnUrls: z.array(z.string()).default(['cdn11.bigcommerce.com']),
     checkoutUrl: z.string(),
   }),
 });
--- a/core/components/image/index.tsx
+++ b/core/components/image/index.tsx
@@ -7,10 +7,11 @@ import { buildConfig } from '~/build-config/reader';
 import bcCdnImageLoader from '~/lib/cdn-image-loader';
 
 function shouldUseLoaderProp(props: ImageProps): boolean {
-  return (
-    typeof props.src === 'string' &&
-    props.src.startsWith(`https://${buildConfig.get('urls').cdnUrl}`)
-  );
+  if (typeof props.src !== 'string') return false;
+
+  const { src } = props;
+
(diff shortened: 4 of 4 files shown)


core/lib/store-assets.ts
```diff
@@ -13,7 +13,7 @@ const storeHash = process.env.BIGCOMMERCE_STORE_HASH ?? '';
  * @returns {string} The CDN image URL.
  */
 const cdnImageUrlBuilder = (sizeSegment: string, source: string, path: string): string => {
-  return `https://${buildConfig.get('urls').cdnUrl}/s-${storeHash}/images/stencil/${sizeSegment}/${source}/${path}`;
+  return `https://${buildConfig.get('urls').cdnUrls[0]}/s-${storeHash}/images/stencil/${sizeSegment}/${source}/${path}`;
```
migueloller, line 16: I think we should `at` instead of array index access and that will force us to handle the case where this array is empty.

core/lib/store-assets.ts
```diff
@@ -26,7 +26,7 @@ const cdnImageUrlBuilder = (sizeSegment: string, source: string, path: string):
  * @returns {string} The full URL to the content asset.
  */
 export const contentAssetUrl = (path: string): string => {
-  return `https://${buildConfig.get('urls').cdnUrl}/s-${storeHash}/content/${path}`;
+  return `https://${buildConfig.get('urls').cdnUrls[0]}/s-${storeHash}/content/${path}`;
```
migueloller, line 29: As mentioned in previous comment, we should handle the empty case

core/next.config.ts
```diff
@@ -38,7 +38,9 @@ async function writeSettingsToBuildConfig() {
     locales: data.site.settings?.locales,
     urls: {
       ...data.site.settings?.url,
-      cdnUrl: process.env.NEXT_PUBLIC_BIGCOMMERCE_CDN_HOSTNAME ?? data.site.settings?.url.cdnUrl,
+      cdnUrls: process.env.NEXT_PUBLIC_BIGCOMMERCE_CDN_HOSTNAME
+        ? process.env.NEXT_PUBLIC_BIGCOMMERCE_CDN_HOSTNAME.split(',').map((s) => s.trim())
+        : [data.site.settings?.url.cdnUrl],
```
migueloller, line 43: Should we filter out nullish values so we can guarantee this will be an array of strings?

core/lib/store-assets.ts
jorgemoya: Do we need to handle the empty case if there is a default value being set by us?

jorgemoya: Changed to use `.at`

core/lib/store-assets.ts
jorgemoya: Updated to use `.at`

core/next.config.ts
jorgemoya: Added filtering out nullish values

core/lib/store-assets.ts
migueloller: In that case, we can type it as `[string, ...string[]]` which will then make sure we always have a default. 

Or a better option is to have the build config have a method that returns the default so that the fact we always have one is enforced by the interface we've modeled

core/lib/store-assets.ts
```diff
@@ -13,7 +13,7 @@ const storeHash = process.env.BIGCOMMERCE_STORE_HASH ?? '';
  * @returns {string} The CDN image URL.
  */
 const cdnImageUrlBuilder = (sizeSegment: string, source: string, path: string): string => {
-  return `https://${buildConfig.get('urls').cdnUrl}/s-${storeHash}/images/stencil/${sizeSegment}/${source}/${path}`;
+  return `https://${buildConfig.get('urls').cdnUrls.at(0)}/s-${storeHash}/images/stencil/${sizeSegment}/${source}/${path}`;
```
migueloller, line 16: The point of using `at` here was so that we would handle the zero-length array case. 

If the build config guarantees a non-zero length array, I guess that's fine, although IMO it would be better to model that in the interface

core/next.config.ts
```diff
@@ -38,7 +39,10 @@ async function writeSettingsToBuildConfig() {
     locales: data.site.settings?.locales,
     urls: {
       ...data.site.settings?.url,
-      cdnUrl: process.env.NEXT_PUBLIC_BIGCOMMERCE_CDN_HOSTNAME ?? data.site.settings?.url.cdnUrl,
+      cdnUrls: (process.env.NEXT_PUBLIC_BIGCOMMERCE_CDN_HOSTNAME
+        ? process.env.NEXT_PUBLIC_BIGCOMMERCE_CDN_HOSTNAME.split(',').map((s) => s.trim())
+        : [data.site.settings?.url.cdnUrl]
+      ).filter((url): url is string => !!url),
```
migueloller, line 45: This line here means that if `settings` is null we could get an empty array for URLs so probably a good idea to handle that case?

migueloller (approved):
Have a couple of comments on handling empty URL arrays and how we (or our users) could avoid out of bounds access as the code changes but I'll leave the final decision up to you!

core/lib/store-assets.ts
chanceaclark: Does `source` contain the full url? If so, can we use `.find` to find the appropriate url?

core/next.config.ts
chanceaclark, line 45: 🍹 Just to make this a bit cleaner, can we extract transforming the `NEXT_PUBLIC_BIGCOMMERCE_CDN_HOSTNAME` before the return so it's something like `manualCdnUrls.length ? manualCdnUrls : [data.site.settings?.url.cdnUrl]`?

jorgemoya: Yes but I think that is fine. If all are empty (which is rare), I think it is fine that the server errors.

core/lib/store-assets.ts
jorgemoya: > Does `source` contain the full url? If so, can we use `.find` to find the appropriate url?

`@param {string} source - The source of the image. Can be either `content` or `image-manager`.`

Doesn't seems like?

chanceaclark: Ah okay.

jorgemoya: > The point of using `at` here was so that we would handle the zero-length array case.
> 
> If the build config guarantees a non-zero length array, I guess that's fine, although IMO it would be better to model that in the interface

Where exactly do you mean?

core/next.config.ts
migueloller: But the error would happen somewhere else, in a very confusing way, instead of us explicitly handling the error here with a clear error message. 

If the expectation is we have the URL present, then we should enforce it instead of letting the server error later because the value is missing

jorgemoya: Got it! I added an error when there are no CDNs set.

PR #2298, opened 2025-05-09
PoC HWB

bc-nick:
## What/Why?

PoC of Headless Wallet Buttons feature

PoC related to checkout-sdk part: https://github.com/bigcommerce/checkout-sdk-js/pull/2855

Implementation includes:

- New wallet infrastructure: `core/lib/wallet-buttons/{index.ts,error.ts,types.ts}` — core logic/types for wallet buttons

- Cart UI wiring: `core/app/[locale]/(default)/cart/page-data.ts`, `page.tsx` and a new `_components/client-wallet-buttons/index.tsx` component to render the wallet buttons, also touches `core/vibes/soul/sections/cart/client.tsx`

Important: whitelisting mechanism will not be included in this implementation, as the corresponding approach has already been implemented in this [PR](https://github.com/bigcommerce/catalyst/pull/2825).

### DIFF
--- a/core/app/[locale]/(default)/cart/page.tsx
+++ b/core/app/[locale]/(default)/cart/page.tsx
@@ -13,7 +13,13 @@ import { updateLineItem } from './_actions/update-line-item';
 import { updateShippingInfo } from './_actions/update-shipping-info';
 import { CartViewed } from './_components/cart-viewed';
 import { CheckoutPreconnect } from './_components/checkout-preconnect';
-import { getCart, getShippingCountries } from './page-data';
+import {
+  getCart,
+  getCurrencyData,
+  getPaymentWallets,
+  getPaymentWalletWithInitializationData,
+  getShippingCountries,
+} from './page-data';
 
 interface Props {
   params: Promise<{ locale: string }>;
@@ -28,6 +34,40 @@ export async function generateMetadata({ params }: Props): Promise<Metadata> {
     title: t('title'),
   };
 }
+
+const createWalletButtonsInitOptions = async (
+  walletButtons: string[],
+  cart: {
+    entityId: string;
+    currencyCode: string;
+    amount: {
+      value: number;
+    };
+  },
+) => {
+  const currencyData = await getCurrencyData(cart.currencyCode);
+
+  return Streamable.all(
+    walletButtons.map(async (entityId) => {
+      const initData = await getPaymentWalletWithInitializationData(entityId, cart.entityId);
+      const methodId = entityId.split('.').join('');
+
+      return {
+        methodId,
+        containerId: `${methodId}-button`,
+        [methodId]: {
+          cartId: cart.entityId,
+          currency: {
+            code: currencyData?.code,
+            decimalPlaces: currencyData?.display.decimalPlaces,
+          },
+          amount: cart.amount.value,
+          ...initData,
+        },
+      };
+    }),
+  );
+};
 
 const getAnalyticsData = async (cartId: string) => {
   const data = await getCart({ cartId });
@@ -88,6 +128,16 @@ export default async function Cart({ params }: Props) {
       />
     );
   }
+
+  const walletButtons = await getPaymentWallets({
+    filters: {
+      cartEntityId: cartId,
+    },
+  });
+
+  const walletButtonsInitOptions = Streamable.from(() =>
+    createWalletButtonsInitOptions(walletButtons, cart),
+  );
 
   const lineItems = [...cart.lineItems.physicalItems, ...cart.lineItems.digitalItems];
 
@@ -282,6 +332,7 @@ export default async function Cart({ params }: Props) {
           }}
           summaryTitle={t('CheckoutSummary.title')}
           title={t('title')}
+          walletButtonsInitOptions={walletButtonsInitOptions}
         />
       </CartAnalyticsProvider>
       <CartViewed
--- a/core/app/api/wallets/graphql/route.ts
+++ b/core/app/api/wallets/graphql/route.ts
@@ -1,0 +1,143 @@
+import { OperationDefinitionNode, parse, SelectionNode } from 'graphql';
+import { NextRequest } from 'next/server';
+import { z } from 'zod';
+
+import { client } from '~/client';
+import { graphql } from '~/client/graphql';
+
+const ENABLE_LIST = [
+  {
+    name: 'PaymentWalletWithInitializationDataQuery',
+    type: 'query',
+    allowedFields: {
+      site: {
+        paymentWalletWithInitializationData: true, // allow any subfields of paymentWallets
+      },
+    },
+  },
+  {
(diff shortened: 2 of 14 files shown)


core/middlewares/with-wallet-buttons-proxy-requests.ts
chanceaclark, line 1: For this PoC, this is fine, just wanted to make sure we make this a route handler and whitelist the queries and mutations via a route handler instead. This should all be working within https://github.com/bigcommerce/catalyst/compare/canary...spike/gql-enablelist#diff-1e657e3f6573e741ec9de194cf6f83b13cda623925400917c480315d2cc4d905 for both queries and mutations cc. @davidchin

core/lib/wallet-buttons/index.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  private async initCheckoutButtonInitializer() {
+    if (!window.checkoutKitLoader) {
+      throw new InitializationError();
+    }
+
+    const checkoutButtonModule = await window.checkoutKitLoader.load('headless-checkout-wallet');
+
+    return checkoutButtonModule.createHeadlessCheckoutWalletInitializer();
+  }
```
chanceaclark, line 50: One thing that @davidchin and I chatted about was being able to pass in the endpoint location into checkout-sdk in case other headless solutions need to change the path, but keep the logic consistent. We should add that as part of the final solution. For example:
```js
window.checkoutKitLoader.setWalletGqlEndpoint("/wallets/graphql")
```

core/app/[locale]/(default)/cart/page.tsx
```diff
@@ -88,6 +93,14 @@ export default async function Cart({ params }: Props) {
     );
   }
 
+  const walletButtons = await getPaymentWallets({
+    filters: {
+      cartEntityId: cartId,
+    },
+  });
+
+  const walletButtonsInitOptions = await createWalletButtonsInitOptions(walletButtons, cart);
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/cart/page.tsx:102]: Instead of these being waterfall requests, can you see about using our `Streamable` pattern here?

core/middlewares/with-wallet-buttons-proxy-requests.ts
davidchin: Looking at the whitelist POC, I believe it's sufficient to whitelist based on the name of the query or mutation. Otherwise, the filtering logic seems overly complicated as we need to traverse the query.

I recall we previously discussed delegating the filtering logic to the server by passing a header. The distinction between browser and server client already exists on the API side. We already limit what a browser client can do, so it would be ideal if we can leverage that existing logic instead of introducing a separate filtering layer on the Catalyst side. WDYT?

chanceaclark: We want to prevent other scripts/integrations from utilizing this endpoint by passing different query information. For example, you can send this query with the enabled query name:
```ts
    const query = graphql(`
      query PaymentWalletWithInitializationDataQuery {
        site {
          settings {
            storeName
          }
        }
      }
    `);
```

Passing the header will only prevent sensitive data from coming through, which we do want, but we also don't want this API handler to be open to all queries, just payment wallet specific queries.

davidchin: Yes, I suppose my point of view is that, from a security standpoint, enforcement should really happen at the SF API level. The SF API already distinguishes between browser and server-side clients and limits the fields that browser clients can access. What’s currently missing is the ability to recognise another type of client - those that use a server-side proxy to access the API via the browser.

If that distinction is made, then the filter we apply at the proxy layer isn’t really for security purposes, but rather to further narrow what queries/mutations are exposed through Catalyst. Therefore, it doesn’t need to be very granular. It’s sufficient to specify just the top-level nodes like cart, checkout, currency, and payment etc…

That said, you’re right in saying that if there are no restrictions at the SF API level for proxy clients, then we do need to be more granular to avoid accidental leakage of sensitive data. We probably don’t need to group the enabled list by operation names to avoid repetition, since what really matters are the query and mutation field names.

chanceaclark: Only reason why I would leave the operation name is so we could have a consistent set of names we can use for identifying queries (e.g. through otel and our other logging systems) since this is a proxy. If we don't enforce an operation name, you can provide an anonymous query and we won't have any sort of insight what the query is. For example:
```gql
query {
  site {
    settings {
      storeName
    }
  }
}
```

 I don't mind if we remove it, but I do see it as a nice-to-have to enforce consistency + a narrower enablement of what is proxied.

davidchin: I see, it's mainly to enforce best practices. For that, we just need to ensure that clients provide _an_ operation name, not specific operation names, as we're not aiming to enforce a particular naming convention.

Ideally, we would enforce this validation at the SF GQL API layer, but I suppose we can’t do that anymore since it would be a backward incompatible change.

That said, I still believe the more restrictions we impose at the proxy layer, the harder it will be to roll out changes in the future. For example, if the client code (Checkout SDK) requires a new field in a few months, merchants would need to manually update the Catalyst codebase themselves to modify the whitelist.

We really have two options here to get around this problem:
1. Option 1: Use the header approach and delegate the filtering on the server side, minimising the possible changes we have to do on the proxy side in the future.
2. Option 2: Go with the library approach, where we ask the merchants to periodically pull updates for the client code and the server code.

core/app/[locale]/(default)/cart/page.tsx
```diff
@@ -28,6 +34,36 @@ export async function generateMetadata({ params }: Props): Promise<Metadata> {
...
 }
 
+const createWalletButtonsInitOptions = async (
+  walletButtons: string[],
+  cart: {
+    entityId: string;
+    currencyCode: string;
+  },
+) => {
+  const currencyData = await getCurrencyData(cart.currencyCode);
```
Comment (davidchin) [inline core/app/[locale]/(default)/cart/page.tsx:48]: It seems this request can be made concurrently, as the requests below don't depend on the result of this request.

core/app/api/wallets/graphql/route.ts
```diff
@@ -0,0 +1,143 @@
+import { OperationDefinitionNode, parse, SelectionNode } from 'graphql';
+import { NextRequest } from 'next/server';
+import { z } from 'zod';
+
+import { client } from '~/client';
+import { graphql } from '~/client/graphql';
+
+const ENABLE_LIST = [
+  {
+    name: 'PaymentWalletWithInitializationDataQuery',
```
davidchin, line 10: As mentioned earlier (https://github.com/bigcommerce/catalyst/pull/2298#discussion_r2094733087), I think we should not enforce specific operation names. Otherwise, the exact operation names become a part of the API contract and we're not free to change in the future. Also, ideally I want to move the filtering logic out of Catalyst to the API layer so the clients won't be coupled to a specific version of Catalyst.

PR #2317, opened 2025-05-15
fix: `useCompareDrawer` does not throw on missing context

agurtovoy:
## What/Why?

The `ProductCard` component, when used on a standalone Makeswift page, causes a fatal error due to the `useCompareDrawer` hook failing in the absence of `CompareDrawerProvider`. This renders the page unusable, and the problematic `ProductCard` element cannot be removed, effectively locking the page.

Generally speaking, we want Makeswift components to gracefully handle missing context.

product card error

## Testing

product card after


core/vibes/soul/primitives/compare-drawer/index.tsx
```diff
@@ -30,7 +30,11 @@ interface CompareDrawerContext {
   maxItems?: number;
 }
 
-export const CompareDrawerContext = createContext<CompareDrawerContext | undefined>(undefined);
+export const CompareDrawerContext = createContext<CompareDrawerContext>({
+  optimisticItems: [],
+  setOptimisticItems: () => {},
+  maxItems: 0,
```
agurtovoy, line 36: `maxItems: 0` signals that the compare functionality is not available.

fikrikarim (approved):

chanceaclark (approved):

jorgemoya (approved):

agurtovoy:
I realized this is not Makeswift-specific per se and should be fixed in `canary`. Rebased against `canary`, here's the corresponding before and after demos:

### Before

https://github.com/user-attachments/assets/ff834030-7aa8-4293-98c6-1d53952a12d8

### After

https://github.com/user-attachments/assets/18590de1-b5fe-4ddf-9260-2271780f86f5

PR #2322, opened 2025-05-19
feat(core): update ProductListSection totalCount prop to string

jorgemoya:
## What/Why?
Update ProductListSection's and ReviewsSection's `totalCount` prop to string.

## Testing
`totalCount` still displays count number.

## Migration
Since the prop has changed for these components, you need to update the type of the prop passed down to these sections.

### DIFF
--- a/.changeset/free-chicken-hammer.md
+++ b/.changeset/free-chicken-hammer.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Update ProductListSection's and ReviewsSection's `totalCount` prop to string.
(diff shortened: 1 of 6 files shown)


matthewvolk (approved):

migueloller:
I’m chiming in here a bit late but we seem to be missing the main point here of making the prop a string, which was to also make it translatable. Are we planning on setting up translations for this field? Otherwise the change from a number to a string is useless

jorgemoya:
> I’m chiming in here a bit late but we seem to be missing the main point here of making the prop a string, which was to also make it translatable. Are we planning on setting up translations for this field? Otherwise the change from a number to a string is useless

You're right 😅 !

https://github.com/bigcommerce/catalyst/pull/2330

PR #2325, opened 2025-05-21
Customer Access Token was missing in the getRoute Request

thebigrick:
## What/Why?
When requesting route information, Catalyst does not send the Customer Access Token. If - for example - a category were hidden from the user, they would still see it.

## Testing
- Create a customer group
- Assign a customer to it
- Mark a category as hidden for the customer group
- Browse the category as a user belonging to another group: You should see it
- Browse the category as the customer belonging to the group you created: You should not see it

### DIFF
--- a/.changeset/vast-cooks-kneel.md
+++ b/.changeset/vast-cooks-kneel.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+FIX missing support for routes permissions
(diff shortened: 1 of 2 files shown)


.changeset/vast-cooks-kneel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+FIX missing support for routes permissions
```
chanceaclark, line 5:
```suggestion
Add `customerAccessToken` when fetching the route url.
```

chanceaclark (approved):
Just a little changeset adjustment, but thanks for catching this!

PR #2329, opened 2025-05-21
feat: add analytics cookies

solofeed:
## What/Why?
Add analytics cookies to support native BigCommerce Analytics

## Testing
Added e2e test to verify cookies are set as expected

## Migration
Since this is an add-only change, no migration is needed beside pulling in the new code.


core/middleware.ts
```diff
@@ -1,10 +1,17 @@
...
 import { withRoutes } from './middlewares/with-routes';
 
-export const middleware = composeMiddlewares(withAuth, withIntl, withChannelId, withRoutes);
+export const middleware = composeMiddlewares(
+  withAnalyticsCookies,
+  withAuth,
+  withIntl,
+  withChannelId,
+  withRoutes,
+);
```
chanceaclark, line 14: Since middleware runs in order and the `withIntl` stuff does some interesting rewrites/redirects depending on the config, can we add this middleware either directly before or after `withChannelId`? It shouldn't matter before or after, just that it's after `withIntl`.

core/lib/analytics/bigcommerce/index.ts
```diff
@@ -0,0 +1,61 @@
+import { cookies } from 'next/headers';
+import { validate as isUuid, v4 as uuidv4 } from 'uuid';
+
+const VISITOR_COOKIE_NAME = 'visitorId';
+const VISIT_COOKIE_NAME = 'visitId';
```
chanceaclark, line 5: Something I have been wanting to do is prefix cookies that Catalyst sets for better visibility what sets the cookie? Can we follow auth.js' footsteps here, e.g. `__Secure-authjs.session-token` and do something like `catalyst.visitId`.

core/middlewares/with-analytics-cookies.ts
```diff
@@ -0,0 +1,18 @@
+import { setVisitIdCookie, setVisitorIdCookie } from '~/lib/analytics/bigcommerce';
+
+import { MiddlewareFactory } from './compose-middlewares';
+
+export const withAnalyticsCookies: MiddlewareFactory = (next) => {
+  return async (request, event) => {
+    await setVisitorIdCookie();
+    await setVisitIdCookie(onNewVisit);
+
```
chanceaclark, line 9: 🍹 Instead of a callback, can we return a new visit id value and handle that within the middleware request? It'll be easier to use `event.waitUntil` within the middleware body than having to pass it as an argument.

For example, you can trigger an API request without blocking the initial render using `event.waitUntil`:
```ts
async (request, event) => {
  await setVisitorIdCookie();
  
  const newVisitId = await setVisitIdCookie();

  if (newVisitId) {
    // middleware will continue and won't end the process until after the event is triggered
    event.waitUntil(() => {
      await sendShopperVisitEvent(newVisitId);
    });
  }
}
```

core/tests/ui/e2e/analytics-session.spec.ts
```diff
@@ -0,0 +1,75 @@
+import { expect, test } from '@playwright/test';
+import { z } from 'zod';
+
+const uuidMatcher = /[0-9a-fA-F-]{16,}/;
```
chanceaclark, line 4: Can we extend expect and use the uuid validate function? Something like this?
```ts
import { validate as isUuid } from 'uuid';

expect.extend({
  toBeUuid(received) {
    const pass = isUuid(received);

    if (pass) {
      return {
        message: () =>`expected ${received} not to be a uuid`,
        pass: true,
      }
    }

    return {
      message: () =>`expected ${received} to be a uuid`,
      pass: false,
    }
  }
})

File: core/middlewares/with-analytics-cookies.ts [hunk 5]
```diff
@@ -0,0 +1,24 @@
...
+    return next(request, event);
+  };
+};
+
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+async function sendShopperVisitEvent(request: NextRequest, visitorId: string, visitId: string) {
+  // fetch from request user agent, url and referer url
+  // TODO: Send shopper visit event to BigCommerce
+}
```
BC-krasnoshapka, line 24: Alternatively can [after](https://nextjs.org/docs/app/api-reference/functions/after) method be used for sending analytic events in non-blocking way, instead of middleware?

solofeed: We will consider this when will be implementing GQL request

chanceaclark: `waitUntil` has more deployment provider support than `after` does, so something to think about here.

core/middlewares/with-analytics-cookies.ts
```diff
@@ -0,0 +1,24 @@
...
+export const withAnalyticsCookies: MiddlewareFactory = (next) => {
+  return async (request, event) => {
+    const visitorId = await setVisitorIdCookie();
+    const { id: visitId, isNew: isVisitNew } = await setVisitIdCookie();
+
+    if (isVisitNew) {
+      event.waitUntil(sendShopperVisitEvent(request, visitorId, visitId));
+    }
+
+    return next(request, event);
```
chanceaclark, line 16: The more I think about this, I think the `set*` functions shouldn't contain side effects 🤔 Can we move the logic into the middleware instead of magically handling it inside the fns?
```suggestion
    let visitorId = await getVisitorIdCookie();
    let visitId = await getVisitIdCookie();
    
    if (!visitorId) {
      visitorId = await setVisitorIdCookie();
    }

    if (!visitId) {
      visitId = await setVisitIdCookie();

      event.waitUntil(sendShopperVisitEvent(request, visitorId, visitId));
    }

    return next(request, event);
```

jorgemoya (approved):

PR #2351, opened 2025-05-30
feat: pass analytics cookies to external checkout

solofeed:
## What/Why?
Pass analytics cookies to checkout mutation, so whenever shopper redirected to the external checkout we preserve the analytics session

## Testing
Added e2e test case

Also, manually
Catalyst cart page with analytics cookies
image

BigCommerce checkout page, analytics cookies has been correctly transferred and preserved
`short_athena_visit_id` matches with `catalyst.visitId` and `fornax_anonymousId` matches with `visitorId`
image

## Migration
None

### DIFF
--- a/.changeset/rotten-badgers-spend.md
+++ b/.changeset/rotten-badgers-spend.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": minor
+---
+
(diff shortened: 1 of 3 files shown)


solofeed:
Added changeset

PR #2352, opened 2025-05-30
Add test config, README, and setup files

jordanarldt:
## What/Why?

This is a prerequisite to my upcoming test changes

- Update playwright configuration
- Add eslint rules for test imports
- Update `.gitignore` for new test reporting directory
- Add `README` for basic testing outline and instructions
- Upgrade zod, add t3-oss/env-core for type-safe environment parsing

## Testing
Tested locally

## Migration
Copy `/core/tests` directory from this PR

### DIFF
--- a/core/.env.test.example
+++ b/core/.env.test.example
@@ -1,0 +1,36 @@
+# This file should hold any environment variables specific to test execution.
+# Any overlapping variables in here will overwrite variables from the other .env files.
+# For example, if BIGCOMMERCE_CHANNEL_ID=1 in .env.local, but BIGCOMMERCE_CHANNEL_ID=5 in this file, the test environment will use BIGCOMMERCE_CHANNEL_ID=5
+
+# Access token client ID and client secret are required for JWT login tests. If not provided, tests that require these will be skipped.
+BIGCOMMERCE_CLIENT_ID=
+BIGCOMMERCE_CLIENT_SECRET=
+
+# Base URL for all tests. This should always be set. It is recommended to run tests against a production build.
+# If not set or undefined, it will default to 'http://localhost:3000'
+PLAYWRIGHT_TEST_BASE_URL=http://localhost:3000
+
+# If true, playwright will automatically start the Catalyst NextJS server before running tests.
+PLAYWRIGHT_START_WEBSERVER=false
+PLAYWRIGHT_START_WEBSERVER_COMMAND=pnpm start
+
+# If true, no tests will execute that require modifying/deleting any data on your store.
+# This is particularly useful if you want to run the tests against a live store and avoid test data being visible to any customers.
+# If this is false, you should only run tests against a staging store where test data is okay.
+TESTS_READ_ONLY=false
+
+# Locale to use during test execution
+TESTS_LOCALE=en
+
+# The fallback locale to use in the occurence of missing translation keys in the selected TESTS_LOCALE
+# This should match the default locale you have set for your storefront
+TESTS_FALLBACK_LOCALE=en
+
+# An existing customer account to use for all tests - use this if you don't want tests to create a new customer account.
+# If TESTS_READ_ONLY is `true`, then any test requiring a customer will be skipped if these values are not set.
+TEST_CUSTOMER_ID=1
+TEST_CUSTOMER_EMAIL=example@test.com
+TEST_CUSTOMER_PASSWORD=Password1
+
+# The default product ID to use for tests. This needs to be a product without variants (simple)
+DEFAULT_PRODUCT_ID=111
--- a/core/playwright.config.ts
+++ b/core/playwright.config.ts
@@ -1,24 +1,29 @@
 import { defineConfig, devices } from '@playwright/test';
-import { config } from 'dotenv';
 
-config();
+import { testEnv } from '~/tests/environment';
 
 export default defineConfig({
   testDir: './tests',
+  outputDir: './.tests/test-results',
+  workers: 1, // TODO: Implement parallel workers in the future
   expect: {
     toHaveScreenshot: {
       maxDiffPixelRatio: 0.02,
     },
   },
-  reporter: [['list'], ['html']],
+  reporter: [
+    ['list', { outputFolder: './.tests/reports/list' }],
+    ['html', { outputFolder: './.tests/reports/html' }],
+  ],
   use: {
-    baseURL: process.env.PLAYWRIGHT_TEST_BASE_URL,
+    locale: testEnv.TESTS_LOCALE,
+    baseURL: testEnv.PLAYWRIGHT_TEST_BASE_URL,
     screenshot: 'only-on-failure',
     video: 'retain-on-failure',
     trace: 'retain-on-failure',
     extraHTTPHeaders: {
-      'x-vercel-protection-bypass': process.env.VERCEL_PROTECTION_BYPASS || '',
-      'x-vercel-set-bypass-cookie': process.env.CI ? 'true' : 'false',
+      'x-vercel-protection-bypass': testEnv.VERCEL_PROTECTION_BYPASS,
+      'x-vercel-set-bypass-cookie': testEnv.CI.toString(),
     },
   },
   projects: [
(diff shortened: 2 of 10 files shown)


core/.env.test.example
```diff
@@ -0,0 +1,41 @@
...
+PLAYWRIGHT_TEST_BASE_URL=http://localhost:3000
+
+# The maximum amount of time that a test can be running for in milliseconds. Default is 30000 (30 seconds)
+PLAYWRIGHT_TEST_TIMEOUT=30000
+
+# The timeout to use for test expectations/assertions in milliseconds. Default is 30000 (30 seconds)
+PLAYWRIGHT_TEST_EXPECT_TIMEOUT=30000
+
+# If true, playwright will automatically start the Catalyst NextJS server before running tests.
+PLAYWRIGHT_START_WEBSERVER=false
```
matthewvolk, line 20: 🍹 Maybe add a recommendation for when someone might want to set this to `true`

core/tests/README.md
```diff
@@ -0,0 +1,66 @@
...
+| `TEST_CUSTOMER_PASSWORD`             |                             | The password for your dedicated testing customer account.                                                                                                                                                                                                                                                                                                                  |
+| `DEFAULT_PRODUCT_ID`                 |                             | The default product ID to use for tests that require a product. If this is not provided, any tests that require a product will fail.                                                                                                                                                                                                                                       |
```
matthewvolk, line 37: 🍹 Historically when we've tried to maintain a table with detailed descriptions of environment variables, they easily become out of sync if someone updates `.env.test.example` and forgets to update this README. How do you feel about moving the descriptions to `.env.test.example` as comments, where defaults are already set?

core/.env.test.example
```diff
@@ -0,0 +1,41 @@
...
+TESTS_READ_ONLY=false
+
+# Locale to use during test execution
+TESTS_LOCALE=en
+
+# The fallback locale to use in the occurence of missing translation keys in the selected TESTS_LOCALE
+# This should match the default locale you have set for your storefront
+TESTS_FALLBACK_LOCALE=en
+
+# Testing customer account - use this if you don't want tests to create a new customer account.
```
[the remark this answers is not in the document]
jordanarldt: @matthewvolk Good call, could be worth clarifying that. Cause yeah, these credentials would need to be for an existing customer.

matthewvolk (approved):
Looks good, I would just want to confirm that comment I left RE: using `npx` 🙂

core/tests/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Running tests
+
+When running tests locally, ensure that you have an active build for your Catalyst storefront by running `pnpm build`. This will ensure that the latest changes are included in the tests. Alternatively, you can use `pnpm dev` to run the development server, but this is not recommended.
+
+When running tests against a live deployment, ensure that you have set the `PLAYWRIGHT_TEST_BASE_URL` environment variable to the URL of your live deployment storefront, and ensure that `TESTS_READ_ONLY` is set accordingly for your needs.
+
+**NOTE:** The examples use `pnpm exec`, but you can also use `npx` if you prefer. Additionally, you can leverage the Playwright VSCode extension to run your tests.
```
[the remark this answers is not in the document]
jordanarldt: @matthewvolk I've been using `npx` the whole time writing these tests without realizing there was an option to run them with `pnpm` 🤣 it's been working fine, so I think we can assume that both are supported.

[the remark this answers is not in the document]
chanceaclark: pnpm is moving away from `pnpx` or `pnpm exec`. Instead you should use `pnpm dlx`

core/tests/environment.ts
chanceaclark, line 1: This might be a good opportunity to use t3/env https://env.t3.gg/docs/introduction which I've been wanting to use for a while.

🍹 Also, I noticed that zod/v4 now has the `stringbool()` type, useful for environment variables. Might be worth an upgrade before this PR 🤔

jordanarldt: Done and done!

core/playwright.config.ts
```diff
@@ -1,24 +1,31 @@
...
-config();
+import { testEnv } from '~/tests/environment';
 
 export default defineConfig({
   testDir: './tests',
+  outputDir: './.tests/test-results',
+  timeout: testEnv.PLAYWRIGHT_TEST_TIMEOUT,
+  workers: 1, // TODO: Implement parallel workers in the future
   expect: {
+    timeout: testEnv.PLAYWRIGHT_TEST_EXPECT_TIMEOUT,
```
chanceaclark: I am very hesitant on adding these timeouts configs. Most tests should be quick and if they are not, then it usually is a smell that we need to rewrite the test more efficiently, you need to configure parallelism better, or the test runner is not powerful enough. If a test does need a longer timeout, thinking of longer e2e tests, I think we can rely on configuring it on a test-by-test basis (there should be a configuration for this).

For us, I don't see the need of using environment variables to configure this. For other consumers of Catalyst, I feel like this should be something they can add themselves as it should be rarely used.

```suggestion
  workers: 1, // TODO: Implement parallel workers in the future
  expect: {
```

core/playwright.config.ts
```diff
@@ -1,24 +1,31 @@
...
-    baseURL: process.env.PLAYWRIGHT_TEST_BASE_URL,
+    locale: testEnv.TESTS_LOCALE,
+    baseURL: testEnv.PLAYWRIGHT_TEST_BASE_URL,
     screenshot: 'only-on-failure',
     video: 'retain-on-failure',
     trace: 'retain-on-failure',
     extraHTTPHeaders: {
-      'x-vercel-protection-bypass': process.env.VERCEL_PROTECTION_BYPASS || '',
-      'x-vercel-set-bypass-cookie': process.env.CI ? 'true' : 'false',
+      'x-vercel-protection-bypass': testEnv.VERCEL_PROTECTION_BYPASS ?? '',
```
chanceaclark: Can we default this env var instead of using `??`?

core/tests/environment.ts
```diff
@@ -0,0 +1,52 @@
...
+        code: 'custom',
+        message: `${v} is not a valid value for TESTS_LOCALE environment variable. TESTS_LOCALE must be one of: ${locales.join(', ')}`,
+      });
+
+      return z.NEVER;
+    }
+
+    return v;
+  }),
+};
```
chanceaclark: This seems like an unnecessary abstraction 🤔  Can we just inline the `string`, `bool` and `number` types but define the locale as its own?

For locale, I think we want to use `.refine` here instead of transform as `.transform` is more for handling data after parsing, while we want to handle the validation during the parsing phase.
```suggestion
const LocaleSchema = z.string().refine((val: string) => !locales.includes(val), {
  message: `${val} is not a valid locale. Should be one of: ${locales.join(', ')}"
})};
```

core/tests/environment.ts
```diff
@@ -0,0 +1,52 @@
...
+import { z } from 'zod/v4';
+
+import { defaultLocale, locales } from '~/i18n/locales';
+
+dotenvConfig({ path: ['.env', '.env.local', '.env.test'], override: true });
+
+const types = {
+  string: z.string().optional(),
+  bool: z.stringbool().optional(),
+  number: z.string().transform(Number).optional(),
```
chanceaclark: I think the `.coerce` utility is better here instead of `.transform`
```suggestion
  number: z.coerce.number().optional(),
```

jordanarldt: @chanceaclark Ahh good call, I was using `z.number({ coerce: true })` previously, but it wasn't available anymore after upgrading so I thought it was removed. I didn't realize `z.coerce` was a thing still 👍

core/playwright.config.ts
jordanarldt: @chanceaclark Fair enough!

core/tests/README.md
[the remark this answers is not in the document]
jordanarldt: @chanceaclark Hmm, using `pnpm dlx playwright test` doesn't seem to work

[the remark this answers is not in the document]
chanceaclark: If the binary is installed locally, I think you can omit the `dlx` part 🤔 
```bash
pnpm playwright test
```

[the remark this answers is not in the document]
jordanarldt: @chanceaclark For sure, updated!

PR #2353, opened 2025-05-30
Add new fixtures to power e2e tests, remove replaced tests

jordanarldt:
## What/Why?
- Create new API client to be used by fixtures in conjunction with Zod
- Create `customer` fixture and schemas
- Create `catalog` fixture and schemas
- Create `order` fixture and schemas
- Update `fixtures.ts` to reference the new fixtures
- Create overrides for playwright `page` type and `browser` types to support localization.
- Remove all tests that will be replaced in subsequent PR

## Testing
Tested locally

## Migration
Copy all changes from `/core/tests` directory


jorgemoya (approved):
LGTM!

PR #2358, opened 2025-06-02
feat: revert project to Tailwind 3 for improved browser support

hunterbecton:
## What/Why?
- reverts project to Tailwind 3 for improved browser support
- adds polyfills using `postcss-preset-env` and `postcss-aspect-ratio-polyfill` PostCSS plugins
- adds container-query polyfill that conditionally renders on unsupported browsers

## Testing

_The tests below were on Safari 14.1._

### Catalyst on Tailwind 4:

https://github.com/user-attachments/assets/a6bb0e85-8fec-4fb0-b2c5-517eccc08d67

### Catalyst on Tailwind 3 (no polyfills):

https://github.com/user-attachments/assets/56848aab-0737-4c96-9af7-ab00bb2b343b

### Catalyst on Tailwind 3 (with polyfills):

https://github.com/user-attachments/assets/b972863a-ba7e-47e7-8a04-2f26a4812e7e


packages/eslint-config-catalyst/base.js
chanceaclark, line 1: Hmmm the files in `packages/eslint-config-catalyst` shouldn't have changed 🤔

.github/workflows/lighthouseCommentMaker.js
chanceaclark, line 1: You might need to rebase to remove these files.

packages/eslint-config-catalyst/base.js
hunterbecton: I think I ran prettier write on the entire project to fix prettier issues without realizing these file also changed. Fixed now!

.github/workflows/lighthouseCommentMaker.js
hunterbecton: Rebased 👍

packages/eslint-config-catalyst/base.js
chanceaclark: Okay cool! Yeah we have prettier running through eslint so we don't need to run prettier globally like with the makeswift repos.

core/app/[locale]/layout.tsx
```diff
@@ -107,23 +108,26 @@ export default async function RootLayout({ params, children }: Props) {
...
-                {children}
-              </Providers>
-            </AnalyticsProvider>
-          </NuqsAdapter>
-        </NextIntlClientProvider>
-        <VercelComponents />
-      </body>
-    </html>
+    <>
+      <ContainerQueryPolyfill />
```
Comment (hunterbecton) [inline core/app/[locale]/layout.tsx]: @chanceaclark I know there's not a benefit of putting the `ContainerQueryPolyfill` here, but I liked that it felt removed from the rest of the code and providers below.

core/components/polyfills/container-query/index.tsx
```diff
@@ -0,0 +1,70 @@
+import Script from 'next/dist/client/script';
```
hunterbecton: @chanceaclark New component that conditionally renders container query polyfills.

core/postcss.config.js
```diff
@@ -1,5 +1,19 @@
 module.exports = {
   plugins: {
-    '@tailwindcss/postcss': {},
+    tailwindcss: {},
+    autoprefixer: {},
```
hunterbecton, line 4: @chanceaclark I added new PostCSS polyfills here to support some more modern CSS features for older browsers.

core/components/polyfills/container-query/index.tsx
chanceaclark: Should the import be this instead?
```suggestion
import Script from 'next/script';
```

core/components/polyfills/container-query/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+      return versionNumber <= 109;
+
+    default:
+      return false;
+  }
+}
+
+export async function ContainerQueryPolyfill() {
+  const headersList = await headers();
+  const userAgent = headersList.get('user-agent') || '';
```
chanceaclark: Can we use the `userAgent` function instead?

https://nextjs.org/docs/app/api-reference/functions/userAgent

core/app/[locale]/layout.tsx
Comment (chanceaclark) [reply core/app/[locale]/layout.tsx]: Let's move this underneath the `<VercelComponents />` component so it's within the `<body>` element at least.

core/components/polyfills/container-query/index.tsx
hunterbecton: I thought you had to pass the entire request into this function, but had no idea I could do `const { browser } = userAgent({ headers: headersList });` super cool

slimboyfats:
Why did you do this PR? Seems like a really lazy variant of fixing the actual issues. We are going to have to update to Tailwind 4 sooner than later so this is just a quick fix and results into doing the exact same thing again in a couple of weeks?

migueloller:
> Why did you do this PR? Seems like a quick fix variant instead of fixing the actual issues? We are going to have to update to Tailwind 4 sooner than later, (as the original PR) so this is just a quick fix and results into doing the exact same thing again in a couple of weeks?

@slimboyfats, we tried very hard to get Tailwind 4 working with Catalyst in such a way that it wouldn't negatively impact browser support and were unsuccessful. We looked at polyfills, PostCSS plugins, and more and given that Tailwind 4 has made the decision to compile down to CSS that is only supported by modern browsers, it was not possible to get something workable without an impact to the site's performance.

While we want to use the best tools that are available when possible, we can't just adopt the latest and the greatest without considering the impact it will have to Catalyst and its users. Specifically, less than 2% of browsers used is not acceptable as that is a meaningful amount of revenue when it comes to ecommerce traffic. We had customers using Catalyst that reported the Tailwind 4 upgrade would have a meaningful impact on their top-line if they pulled in those changes.

After evaluating all of our options, we decided that this would be the best way to move forward. If you want to move your Catalyst fork to Tailwind 4, you're welcome to do so. Also, if you have any thoughts or alternatives as to how we could approach this, we're all ears!

slimboyfats:
Thanks for the answer, so in that case i guess the decision is made to not update to tailwind 4 until Safari 14.1 falls below the threshold if I understand you correct? 

What was the biggest things that Safari didn't support in Tailwind 3 vs 4? Would be a good thing to read up for me and see the biggest issues with the upgrade.

hunterbecton:
> Thanks for the answer, so in that case i guess the decision is made to not update to tailwind 4 until Safari 14.1 falls below the threshold if I understand you correct?
> 

Yes, we won't support Tailwind until Safari falls below the threshold, or when Tailwind better supports older browsers. Tailwind moved to Lightning CSS in Tailwind 4, which has options for improving browser support compatibility, but Tailwind doesn't expose any of those options in its configuration.  

> What was the biggest things that Safari didn't support in Tailwind 3 vs 4? Would be a good thing to read up for me and see the biggest issues with the upgrade.

Tailwind covers some of the new features supported by Tailwind 4 in its [upgrade guide](https://tailwindcss.com/docs/upgrade-guide). I'd suggest you read through this to get an idea of what's changed, and you'll want to follow along if you're upgrading your project to Tailwind 4.

It's also worth noting that Tailwind generates and optimizes the CSS; we have little control over that outside of what version we're using. In the PR video, you can see what Tailwind 3 vs Tailwind 4 browser support looks like with zero configurations. So even though we're still using polyfills to support features like container queries, color-mix, and more, Tailwind 3 gives us a _way_ better starting point to support those features than Tailwind 4.

One last thing is that there's not an easy way to downgrade from Tailwind 4 to Tailwind 3, but upgrading is much easier with the [Tailwind 4 upgrade guide](https://tailwindcss.com/docs/upgrade-guide) and upgrade tool. This is a better customer experience because customers who want to move to Tailwind 4 have the resources to do so much more easily than customers needing to downgrade to Tailwind 3.

PR #2361, opened 2025-06-03
feat: add locale override parameter to client.fetch

matthewvolk:
## What/Why?
Allows passing a `locale` that will be used to set the outgoing `"Accept-Language"` header sent to the BigCommerce Storefront GraphQL API. This additional, optional parameter gives you more explicit control over what language you'd like responses to be translated into, [leveraging BigCommerce's resolved locale feature](https://developer.bigcommerce.com/docs/storefront/graphql/multi-language-support#returned-data).

> [!NOTE]
> This PR originated from #2286 

## Testing
N/A - Additive change

## Migration
* For those consuming the package from the NPM registry, this feature will be available in version `0.16` or higher, so a simple version bump should suffice. 
* For those using the package via `pnpm` workspaces, simply cherry-pick the additive diff from this PR into your branch.

### DIFF
--- a/.changeset/metal-owls-travel.md
+++ b/.changeset/metal-owls-travel.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-client": minor
+---
+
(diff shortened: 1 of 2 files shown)


matthewvolk:
Converted to draft; discussed with the team, and we should either more appropriately name this parameter based on what it is actually doing (overriding `Accept-Language` header), or find a way to deliver #2286 without a change to the client. Going to think a bit on the best path forward, but don't want to give the impression this PR is final.

matthewvolk:
I'm going to go the route of not modifying the client. Feels aggressive to extend the client's API just for the one case needed in #2286

PR #2370, opened 2025-06-04
refactor: replace fetchShippingZones with supportedShippingDestinations

matthewvolk:
## What/Why?

* Added a new query to fetch supported shipping destinations directly from the GraphQL API.
* Updated the cart page to utilize the new shipping countries retrieval method.
* Removed the unused getShippingZones function and related code to simplify the client structure and remove dependence on `BIGCOMMERCE_ACCESS_TOKEN` environment variable

## Testing

Tested locally after changes:

https://github.com/user-attachments/assets/990dd8ac-d0c5-4d54-a4aa-be8600b75cc6

## Migration

1. The return type of `getShippingCountries` has the same shape as the `Country` BigCommerce GraphQL type, so you should be able to copy the graphql query from `core/app/[locale]/(default)/cart/page-data.ts` into your project and replace the existing `getShippingCountries` method in there.
2. Remove the argument `data.geography` from the `getShippingCountries` invocation in `core/app/[locale]/(default)/cart/page.tsx`
3. Finally, you should be able to delete the file `core/client/management/get-shipping-zones.ts` assuming it is no longer referenced anywhere in `core/`
4. If you are using the version of the client published to NPM, simply ensure you are using at least `@bigcommerce/catalyst-client@0.16.0` or higher.
5. If you are using the client in your pnpm workspace, simply remove the `xAuthToken` references in `packages/client/src/client.ts` as well as the `fetchShippingZones` method.
6. Remove the reference to `xAuthToken` in `core/client/index.ts`


core/app/[locale]/(default)/cart/page-data.ts
```diff
@@ -262,18 +264,34 @@ export const getCart = async (variables: Variables) => {
...
 
-    return isCountryInTheList || !hasAccessToken;
+export const getShippingCountries = cache(async () => {
+  const { data } = await client.fetch({
+    document: SupportedShippingDestinationsQuery,
+    fetchOptions: { next: { revalidate } },
   });
-};
+
+  return data.site.settings?.shipping?.supportedShippingDestinations.countries ?? [];
```
Comment (matthewvolk) [inline core/app/[locale]/(default)/cart/page-data.ts:296]: Is this the right fallback (empty array)?

core/app/[locale]/(default)/cart/page-data.ts
```diff
@@ -262,18 +264,34 @@ export const getCart = async (variables: Variables) => {
...
+      }
+    }
+  }
+`);
 
-    return isCountryInTheList || !hasAccessToken;
+export const getShippingCountries = cache(async () => {
+  const { data } = await client.fetch({
+    document: SupportedShippingDestinationsQuery,
+    fetchOptions: { next: { revalidate } },
```
Comment (matthewvolk) [inline core/app/[locale]/(default)/cart/page-data.ts:293]: Is this the right caching strategy? I wasn't sure if CAT influences the returned list of countries 🤔

Comment (jorgemoya) [reply core/app/[locale]/(default)/cart/page-data.ts:293]: This look good to me IF customers don't have special shipping countries available to them (which I don't think so).

core/app/[locale]/(default)/cart/page-data.ts
Comment (jorgemoya) [reply core/app/[locale]/(default)/cart/page-data.ts:296]: 👍

PR #2371, opened 2025-06-04
Update fixture schemas, abstract test fixtures API client

jordanarldt:
## What/Why?
 - Use defined "create" schemas for API resource models
 - Abstract test fixtures API client to allow simple swapping in the future (if we wanted to switch to Admin GQL instead of the Management API)

## Testing

**Account tests**
image

**Auth tests**
image

## Migration
Copy all changes from `/core/tests` directory


core/tests/fixtures/customer/index.ts
```diff
@@ -1,18 +1,63 @@
...
+  Wishlist,
+} from '~/tests/fixtures/utils/api/customers';
 import { getTranslations } from '~/tests/lib/i18n';
 
-import { Address } from './address';
-import { Customer } from './customer';
 import { customerSessionStore } from './session';
-import { Wishlist } from './wishlist';
+
+function fakeCreateAddressData({
```
chanceaclark: 🍹 I bet we can just make this a private method of the class below.

core/tests/fixtures/fixture.ts
```diff
@@ -1,15 +1,22 @@
...
     readonly page: Page,
     readonly test: TestInfo,
   ) {
-    this.api = new TestApiClient(test);
+    this.api = httpApiClient;
+  }
+
+  protected skipIfReadonly(): void {
+    if (testEnv.TESTS_READ_ONLY) {
+      this.test.skip(true, 'Tests are running in read-only mode.');
```
chanceaclark: Can we see an example of this being skipped? I'm just curious how the test message will look because I though providing `tests are running in read-only mode.` overrides the test name 🤔

chanceaclark (approved):

core/tests/fixtures/fixture.ts
jordanarldt: It doesn't override the test name 👍  It is only used to give more information about why the test was skipped in the reporting.
Here's the run result:
image

And the reporting result:
image

PR #2372, opened 2025-06-04
feat(core): render address fields for customer registration

chanceaclark:
## What/Why?

As part of the update with VIBES, we decided to slim down the account creation from. Part of this we added a custom hardcoded list of form fields we rendered within the form based on the form fields provided from the API. We'd then filter out the form fields provided by the API with this hardcoded list. Where this broke down is when you set a field on the address fields, e.g. `Phone number`, as required and tried to submit the form, you'd get API errors because the form field was not present on the form submission. A developer would need to go in and manually add the form field as part of the hard coded list.

As a solution to this problem, we are now taking all the form fields from the API and dynamically rendering them (still) with the `DynamicForm` component. This also aligns the logic to be the same as the customer address creation/update logic (was copied over). We can probably abstract this logic a bit as it's quite duplicative, but for the interim we can leave it separate in case functionality diverges in the future.

The alternative to this solution would be to create a new API to handle separating the customer form fields and the address field collection. However, that would require a lot more work. Alternatively, the following solution follows what Stencil follows which aligns with "feature parity".

## Testing
screencapture-localhost-3000-register-2025-06-04-16_49_30

https://github.com/user-attachments/assets/7d0aa499-198d-41c8-8b6e-5a3a0b204767

## Migration

Unfortunately the migration for the work is not straight-forward if the customer registrations logic has been customized. Otherwise, it's mostly additive changes, so rebasing should pull in the changes needed.

### DIFF
--- a/core/app/[locale]/(default)/(auth)/register/page.tsx
+++ b/core/app/[locale]/(default)/(auth)/register/page.tsx
@@ -6,13 +6,18 @@ import { getTranslations, setRequestLocale } from 'next-intl/server';
 // import { bypassReCaptcha } from '~/lib/bypass-recaptcha';
 
 import { DynamicFormSection } from '@/vibes/soul/sections/dynamic-form-section';
-import { formFieldTransformer } from '~/data-transformers/form-field-transformer';
+import {
+  formFieldTransformer,
+  injectCountryCodeOptions,
+} from '~/data-transformers/form-field-transformer';
 import {
   CUSTOMER_FIELDS_TO_EXCLUDE,
-  FULL_NAME_FIELDS,
+  REGISTER_CUSTOMER_FORM_LAYOUT,
+  transformFieldsToLayout,
 } from '~/data-transformers/form-field-transformer/utils';
 import { exists } from '~/lib/utils';
 
+import { ADDRESS_FIELDS_NAME_PREFIX, CUSTOMER_FIELDS_NAME_PREFIX } from './_actions/prefixes';
 import { registerCustomer } from './_actions/register-customer';
 import { getRegisterCustomerQuery } from './page-data';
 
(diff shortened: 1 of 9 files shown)


core/app/[locale]/(default)/(auth)/register/page.tsx
```diff
@@ -46,22 +51,55 @@ export default async function Register({ params }: Props) {
...
+  const { addressFields, customerFields, countries } = registerCustomerData;
   // const reCaptcha = await bypassReCaptcha(reCaptchaSettings);
 
+  const fields = transformFieldsToLayout(
+    [
+      ...addressFields.map((field) => {
+        if (!field.isBuiltIn) {
+          return {
+            ...field,
+            name: `${ADDRESS_FIELDS_NAME_PREFIX}${field.label}`,
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/(auth)/register/page.tsx:63]: We need to prefix the fields in order to split them out inside the server action.

core/data-transformers/form-field-transformer/index.ts
```diff
@@ -116,17 +119,6 @@ export const formFieldTransformer = (
   }
 };
 
-// TODO: See if we can merge this is with the above function
-// Will require testing and refactoring of register customer functionality
```
chanceaclark, line 120: I took care of this in this PR.

core/data-transformers/form-field-transformer/utils.ts
```diff
@@ -57,29 +58,6 @@ export const ADDRESS_FORM_LAYOUT = [
   [FieldNameToFieldId.postalCode, FieldNameToFieldId.countryCode],
 ];
 
-export const createFieldName = (field: FormField, fieldOrigin: 'customer' | 'address') => {
```
chanceaclark, line 60: Unused fn.

core/tests/ui/e2e/auth/register.spec.ts
```diff
@@ -24,6 +29,15 @@ test('Registration works as expected', { tag: [TAGS.writesData] }, async ({ page
...
   await page.getByLabel('Password', { exact: true }).fill(password);
   await page.getByLabel('Confirm Password').fill(password);
+  await page.getByLabel('Phone').fill(phone);
+  await page.getByLabel('Address Line 1').fill(streetAddress);
+  await page.getByLabel('Suburb/City').fill(city);
+  await page.getByLabel('State/Province').fill(state);
+  await page.getByLabel('Zip/Postcode').fill(postalCode);
+  await page.getByRole('combobox', { name: 'Country' }).click();
+  await page.keyboard.type('United States');
+  await page.keyboard.press('Enter');
```
chanceaclark, line 39: Working with a long list in a `Select` was very difficult. This was the best way I could do it.

jordanarldt: I think that's fine!

jordanarldt: @chanceaclark This is how I did it in the account-settings tests:
```
await page.getByLabel('Country').click();
await page.getByRole('option', { name: country }).first().click();
```

It seems to work there, might be worth a shot to see if it works in this case as well, but no big deal since either way works and represents how someone might actually attempt it

chanceaclark: The issue I was having with that approach was if an option was either non-existent or you had to scroll down the list it wouldn't work.

PR #2375, opened 2025-06-05
chore: update and sync .env.example for makeswift branch

matthewvolk:
## What/Why?

It was reported that the `CustomerGroupSlot` component relies on the `BIGCOMMERCE_ACCESS_TOKEN` environment variable which is not documented anywhere in that branch.

This PR documents the environment variable in `.env.example` and syncs `.env.example` with `core/.env.example`

## Testing

N/A, simply add the environment variable to your `.env.local` file.

## Migration

N/A, simply add the environment variable to your `.env.local` file.


agurtovoy (approved):
🙏

PR #2379, opened 2025-06-06
fix: separate component registration from (client) component definitions

agurtovoy:
## What/Why?

As a rule of thumb, we tell users that custom component registrations should be universal -- that is, capable of running on both the client and the server -- since we currently need the corresponding component definition to properly introspect page and component snapshots in order to preload associated resources.

However, it was pointed out to us that we do not follow this rule ourselves in Catalyst: the `CustomerGroupSlot`, `ProductCard`, `ProductCarousel`, and `ProductsList` component registrations are, in fact, client-only. While in these particular cases there are no practical consequences to violating the rule (none of the components in question can embed Makeswift resources), it's still best to follow our own recommendations to avoid confusion and prevent potential future bugs as these components evolve.

## Testing

Testing of the affected components:

https://github.com/user-attachments/assets/bdbf3570-b5e2-41d2-9cae-f66c509e76b6


core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.ts
```diff
@@ -0,0 +1,93 @@
+import { Combobox, Group, List, Slot, Style } from '@makeswift/runtime/controls';
```
agurtovoy, line 1: Here and below, I'm simply splitting up `<component>.makeswift.tsx` into  `<component>.makeswift.ts` and  (client-side) `<component>.tsx`, but for some reason GitHub doesn't recognize the rename and shows it as deletion + addition.

There are no other semantic changes.

arvinpoddar (approved):

PR #2381, opened 2025-06-06
fix(core): prefix anonymous cookie with __Secure for https://

chanceaclark:
## What/Why?
This adds the `__Secure-` cookie prefix to the anonymous session cookie to add extra security measures. This logic follows the same behavior as Auth.js does for prefixing it's cookies with the `__Secure-` prefix.

#### Local dev
Screenshot 2025-06-06 at 13 41 10

#### Production
Screenshot 2025-06-06 at 13 40 49

## Migration
If you haven't merge the anonymous session code, pulling it in should just work ™️.

### DIFF
--- a/.changeset/silver-items-tie.md
+++ b/.changeset/silver-items-tie.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Adds the `__Secure-` prefix to the add additional broswer security policies around this cookie.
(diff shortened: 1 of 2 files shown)


jkanive (commented):
Would this cahnge break existing in-session activity after deployment?

chanceaclark:
Since I merged the anonymous session stuff 2 days ago and we haven't tagged a release yet, I wouldn't be too worried about it breaking any sessions. Theoretically yeah it could wipe an existing cart away if someone already pulled that code in.

jkanive (approved):

PR #2387, opened 2025-06-06
fix(core): use tailwind aspect-ratio plugin

chanceaclark:
## What/Why?
Utilizes the tailwind plugin to support aspect ratios in older browsers like Safari 14. 

## Testing
Screenshot 2025-06-06 at 15 47 30

## Migration
All the old arbitrary classes will need to define the height and width classes:
```
// before
.aspect-[4/5]

//after
.aspect-w-4 .aspect-h-5
```

### DIFF
--- a/core/vibes/soul/sections/blog-post-content/index.tsx
+++ b/core/vibes/soul/sections/blog-post-content/index.tsx
@@ -81,7 +81,7 @@ export function BlogPostContent({
                 {image?.src != null && image.src !== '' && (
                   <Image
                     alt={image.alt}
-                    className="mb-8 aspect-video w-full rounded-2xl bg-contrast-100 object-cover @2xl:mb-12 @4xl:mb-16"
+                    className="aspect-h-9 aspect-w-16 mb-8 w-full rounded-2xl bg-contrast-100 object-cover @2xl:mb-12 @4xl:mb-16"
                     height={780}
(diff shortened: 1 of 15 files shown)


jordanarldt (approved):

core/vibes/soul/sections/blog-post-content/index.tsx
```diff
@@ -81,7 +81,7 @@ export function BlogPostContent({
                 {image?.src != null && image.src !== '' && (
                   <Image
                     alt={image.alt}
-                    className="mb-8 aspect-video w-full rounded-2xl bg-contrast-100 object-cover @2xl:mb-12 @4xl:mb-16"
+                    className="mb-8 aspect-w-1 aspect-h-9 w-full rounded-2xl bg-contrast-100 object-cover @2xl:mb-12 @4xl:mb-16"
```
hunterbecton: Video should `aspect-w-16 aspect-h-9`

hunterbecton (commented):
Looks good, I just had one quick comment to fix. Thank you!

core/vibes/soul/sections/blog-post-content/index.tsx
chanceaclark: Good catch, missed adding the `6` for `16` when doing a global replace here 🤦

PR #2388, opened 2025-06-06
Add product fixtures, add wishlist details tests, minor change to wishlist details page

jordanarldt:
## What/Why?
- Add fixtures for creating products
- Make `/accounts/wishlists/[id]` redirect to `/accounts/wishlists/` if the wishlist is not found
- Add e2e tests for wishlist details page

## Migration
Copy all changes from `/core/tests`


core/app/[locale]/(default)/account/wishlists/[id]/_components/wishlist-actions.tsx
Comment (chanceaclark) [inline core/app/[locale]/(default)/account/wishlists/[id]/_components/wishlist-actions.tsx:1]: Any changes to `core/app/` should probably be decoupled from the test changes and in it's own PR to make migration documentation a little easier to follow.

core/tests/ui/e2e/account/wishlist-details.spec.ts
```diff
@@ -0,0 +1,241 @@
+import { expect, test } from '~/tests/fixtures';
+import { getTranslations } from '~/tests/lib/i18n';
+import { TAGS } from '~/tests/tags';
+
+test.describe('Wishlist Details', () => {
```
chanceaclark: 🍹 We already get context in the test runs which file we are running tests against so it's a little redundant to wrap tests in a `.describe`. I tend to lean on `.describe` blocks to segment testing edge-cases or different branches of logic like you do with the actions below.

jordanarldt: @chanceaclark Yeah, I'm not super opinionated about this - the main reason I've been doing it this way is it makes it easier to run the whole group of tests using the playwright VSCode extension if we wrap the whole suite in a `test.describe`, just a small quality of life improvement. Otherwise, you have to manually browse to the test explorer, find the file, and then run it that way instead of running it directly from the code.

jordanarldt: I think it's generally a good practice though, because if you happened to need to add tags to a whole suite, you do it in one place. Otherwise you'd have to add tags to every single test in the file. For example:
```
test.describe('Some test', { tag: ['my-test-tag', 'other-tag'] }, () => ... );
```

vs

```
test('Some test 1', { tag: ['my-test-tag', 'other-tag'] }, () => ... );

test('Some test 2', { tag: ['my-test-tag', 'other-tag'] }, () => ... );

test('Some test 3', { tag: ['my-test-tag', 'other-tag'] }, () => ... );

test('Some test 4', { tag: ['my-test-tag', 'other-tag'] }, () => ... );
```

jordanarldt: This would be particularly useful if we ended up needing domain tags to allow BC services to run specific Catalyst e2e tests during their CI, instead of annotating every test in a suite, we just annotate the root `test.describe`

PR #2395, opened 2025-06-09
Add aria labels to currency selector and PDP wishlist favorite button

jordanarldt:
## What/Why?
Add title to currency selector and PDP wishlist buttons. This is good for accessibility but also allows e2e tests to find the elements.

## Testing
Build passes locally

## Migration
1. Copy all changes from the `/messages/en.json` file to get updated translation keys
2. Add the `title` prop to the `Heart` component in `/core/vibes/soul/primitives/favorite/heart.tsx`
3. Add the `title` prop to the `Favorite` component in `/core/vibes/soul/primitives/favorite/index.tsx` and pass it to the `Heart` component
4. Copy all changes in the `/core/vibes/soul/navigation/index.tsx` file to add the `switchCurrencyLabel` property
5. Update `/core/components/header/index.tsx` file to pass the `switchCurrencyLabel` to the `HeaderSection` component
6. Update `/core/app/[locale]/(default)/product/[slug]/_components/wishlist-button/index.tsx` to pass the `label` prop to the `Favorite` component

### DIFF
--- a/core/vibes/soul/primitives/favorite/index.tsx
+++ b/core/vibes/soul/primitives/favorite/index.tsx
@@ -3,6 +3,7 @@ import * as Toggle from '@radix-ui/react-toggle';
 import { Heart } from '@/vibes/soul/primitives/favorite/heart';
 
 export interface FavoriteProps {
+  label?: string;
   checked?: boolean;
   setChecked?: (liked: boolean) => void;
 }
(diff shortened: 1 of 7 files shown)


core/vibes/soul/primitives/favorite/index.tsx
```diff
@@ -22,9 +23,10 @@ export interface FavoriteProps {
  * }
  * ```
  */
-export const Favorite = ({ checked = false, setChecked }: FavoriteProps) => {
+export const Favorite = ({ checked = false, setChecked, label = 'Favorite' }: FavoriteProps) => {
   return (
     <Toggle.Root
+      aria-label={label}
```
chanceaclark: The [first rule of `aria`](https://www.w3.org/TR/using-aria/#firstrule) is to use built-in behavior before we use `aria-` attributes. We should be able to utilize the svg's `<title>` tag here instead:
```tsx
<Heart filled={checked}>
  <title>{label}</title>
</Heart>
```

core/vibes/soul/primitives/navigation/index.tsx
```diff
@@ -916,6 +921,7 @@ function CurrencyForm({
   return (
     <DropdownMenu.Root>
       <DropdownMenu.Trigger
+        aria-label={switchCurrencyLabel}
```
chanceaclark: I think we can do the same here to.

jordanarldt: Ah right, good call!

core/vibes/soul/primitives/favorite/index.tsx
```diff
@@ -3,6 +3,7 @@ import * as Toggle from '@radix-ui/react-toggle';
 import { Heart } from '@/vibes/soul/primitives/favorite/heart';
 
 export interface FavoriteProps {
+  title?: string;
```
chanceaclark: Hmmm... at first I thought this prop was the native `title` attribute when looking at the consumption. Could we change this to `label` or something else instead so it's not to be confused with native html attributes?

jordanarldt: Done 👍

jorgemoya (approved):

PR #2396, opened 2025-06-09
Add cart page e2e tests

jordanarldt:
## What/Why?
Rewrite the e2e tests for the cart page

## Testing
image

## Migration
Copy all changes from `/core/tests`


jorgemoya (approved):
This looks good to me as is, but there are certainly other steps missing from this, like all the shipping flow and checkout. We should track adding that in a different PR.

PR #2397, opened 2025-06-10
feat: re-enable Customer Group Slot component

agurtovoy:
## What/Why?

We initially unregistered this component because fetching customer groups requires the BigCommerce authentication token, which we don't expose in the default OCC configuration. However, removing the component entirely reduced discoverability and led to more confusion about its availability.

This PR re-enables the component by default, while providing a notice to inform users that it's not fully functional and requires additional configuration.

## Testing

https://github.com/user-attachments/assets/af8303e5-2797-45bc-839f-a62a4310ab47

### DIFF
--- a/core/app/api/customer/groups/route.ts
+++ b/core/app/api/customer/groups/route.ts
@@ -1,8 +1,10 @@
 import { NextResponse } from 'next/server';
 
 export async function GET(): Promise<NextResponse> {
-  if (!process.env.BIGCOMMERCE_ACCESS_TOKEN) {
-    throw new Error('BIGCOMMERCE_ACCESS_TOKEN is not set');
+  const authToken = process.env.BIGCOMMERCE_ACCESS_TOKEN;
+
+  if (!authToken) {
+    return NextResponse.json(null);
   }
 
   const response = await fetch(
@@ -12,13 +14,18 @@ export async function GET(): Promise<NextResponse> {
       headers: {
         Accept: 'application/json',
         'Content-Type': 'application/json',
-        'X-Auth-Token': process.env.BIGCOMMERCE_ACCESS_TOKEN,
+        'X-Auth-Token': authToken,
       },
     },
   );
 
   if (!response.ok) {
     throw new Error(`Failed to fetch customer groups: ${response.statusText}`);
+  }
+
+  // `/v2/customer_groups` endpoint returns a `204 No Content` response if there are no customer groups
+  if (response.status === 204) {
+    return NextResponse.json([]);
   }
 
   const jsonResponse: unknown = await response.json();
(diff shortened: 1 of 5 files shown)


core/app/api/customer/groups/route.ts
```diff
@@ -21,6 +23,11 @@ export async function GET(): Promise<NextResponse> {
     throw new Error(`Failed to fetch customer groups: ${response.statusText}`);
   }
 
+  // `/v2/customer_groups` endpoint returns a `204 No Content` response if there are no customer groups
+  if (response.status === 204) {
```
agurtovoy, line 32: Added a fix for the empty list of groups.

core/app/api/customer/groups/route.ts
```diff
@@ -1,8 +1,10 @@
 import { NextResponse } from 'next/server';
 
 export async function GET(): Promise<NextResponse> {
-  if (!process.env.BIGCOMMERCE_ACCESS_TOKEN) {
-    throw new Error('BIGCOMMERCE_ACCESS_TOKEN is not set');
+  const authToken = process.env.BIGCOMMERCE_ACCESS_TOKEN;
+
+  if (!authToken) {
```
matthewvolk, line 6: 🍹Should we add a log letting the customer know they aren't getting data in their response because they are missing a token?

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.ts
```diff
@@ -64,7 +64,16 @@ runtime.registerComponent(CustomerGroupSlot, {
         try {
           const data = await getAllCustomerGroups();
 
-          if (!data) return [];
+          if (!data) {
+            // nullish data means we don't have access to the customer groups API
+            return [
+              {
+                id: NO_GROUP_ID,
+                label: 'Setup needed. See docs.',
```
chanceaclark, line 77: Do we have a link to the docs we can provide?

core/app/api/customer/groups/route.ts
chanceaclark: This or a status code.

migueloller: 🍹 +1 for both a warning log as well as more context in the HTTP response. Here we could provide links to docs, etc.

core/lib/makeswift/components/customer-group-slot/customer-group-slot.makeswift.ts
migueloller: 🍹  Given that the "builder user" is non-technical, should the message be something more akin to "Talk to your developer"?

agurtovoy: @migueloller The character space for this is quite limited, and I think "See docs" is more general. The docs can give them more context and tell them to talk to the developer as needed.

agurtovoy: @chanceaclark We don’t have a public link to the docs yet, and even if we did, including it in the value dropdown wouldn’t be very helpful since it wouldn’t be fully visible or clickable. That said, we have a builder feature in the works that should provide a better experience here, I’ll follow up with another PR once it’s available.

PR #2399, opened 2025-06-10
feat: add disabled styles to checkbox input and label

matthewvolk:
## What/Why?

Adds Tailwind classes used to style the checkbox input and label based on the disabled state of the checkbox. 

## Testing

Default:

https://github.com/user-attachments/assets/e4fdfed8-930b-4ea5-8038-1ea71c63f6e5

Disabled:

https://github.com/user-attachments/assets/1a044ff6-1923-478e-ae36-07ca858afdf3

Checked + disabled:

https://github.com/user-attachments/assets/c17e4e3b-cecb-494b-b0eb-66ef47748159

## Migration

Since this is a one-file change, you should be able to simply grab the diff from this PR. The main changes to note are that we are [adding a `peer` class](https://v3.tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-sibling-state) to the CheckboxPrimitive.Root, explicitly styling the `enabled` pseudoclass, and only applying hover styles when the checkbox is enabled.

### DIFF
--- a/.changeset/common-brooms-dress.md
+++ b/.changeset/common-brooms-dress.md
@@ -1,0 +1,8 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Adds Tailwind classes used to style the checkbox input and label based on the disabled state of the checkbox.
+
+Migration:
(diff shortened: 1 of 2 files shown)


jorgemoya (approved):
Would you mind pushing this to vibes if you have time?

PR #2400, opened 2025-06-10
feat: drop CSS support for Safari < 15

chanceaclark:
## What/Why?
Drops CSS support for Safari < 15 due to the [global usage being 0.09%](https://browsersl.ist/#q=Safari+%3C+15). This should enable us to utilize the native CSS `aspect-ratio` as it is semantically different than utilizing [Aspect Ratio Boxes](https://css-tricks.com/aspect-ratio-boxes/). Aspect Ratio Boxes falls apart when utilizing it within a flex container, having use to change markup and semantics.

## Testing
Screenshot 2025-06-10 at 13 20 35
Screenshot 2025-06-10 at 13 21 26
Screenshot 2025-06-10 at 13 21 42

## Migration
- Pull in the revert commit (git should be able to handle this)
- Remove the `@tailwindcss/aspect-ratio` package in `package.json`
- Remove the PossCSS plugins around aspect ratios


jorgemoya (approved):
Thanks for doing this

PR #2402, opened 2025-06-10
Fix breadcrumb keys

jordanarldt:
## What/Why?
Alter `Breadcrumb` component to use `index` instead of `href` for the key. When using large breadcrumbs for nested pages or categories, we truncate the breadcrumbs to put a `...` (e.g. `Home > Page 1 > ... > Page 5`), Where the `...` and `Page 5` will both use `#` for the href property.

This means that if you have a deeply nested page, you will get a duplicate key error.

This was previously fixed, but had a regression when we sync'd the Breadcrumbs component from VIBES.

## Testing
Tested locally

## Migration
Update `core/vibes/soul/sections/breadcrumbs/index.tsx` to use `index` as the `key` property instead of `href`

### DIFF
--- a/.changeset/busy-snakes-agree.md
+++ b/.changeset/busy-snakes-agree.md
@@ -1,0 +1,9 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Fix possibility of duplicate `key` error in Breadcrumbs component for truncated breadcrumbs.
+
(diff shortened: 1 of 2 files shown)


jkanive (approved):
LGTM 👍

jorgemoya (approved):

PR #2404, opened 2025-06-11
feat: Add support for Buyer Portal in Catalyst

icatalina:


core/auth/index.ts
```diff
@@ -279,6 +318,13 @@ const config = {
       },
       authorize: loginWithPassword,
     }),
+    CredentialsProvider({
+      id: 'anonymous',
+      credentials: {
+        cartId: { type: 'text' },
+      },
+      authorize: loginWithAnonymous,
+    }),
```
chanceaclark: Probably want to give this another rebase as we moved the anonymous session into its own cookie instead of reusing the next-auth one. This is to account for persistent cart to work correctly. See these PRs:
- https://github.com/bigcommerce/catalyst/pull/2350
- https://github.com/bigcommerce/catalyst/pull/2368
- https://github.com/bigcommerce/catalyst/pull/2381

icatalina: another rebase given, can you have a look 🙇

chanceaclark: You shouldn't need this `CredentialsProvider` if you've pull in the code from those 3 PRs.

chanceaclark: I went ahead and removed the code and did another merge for you. This should have all the code you need.

PR #2413, opened 2025-06-12
Integrations/b2b buyer portal-footer-poc

arshiyakhair-2118118:


core/components/footer/index.tsx
arshiyakhair-2118118, line 1: git status
git add
git commit

arshiyakhair-2118118, line 1: footrt-poc

PR #2415, opened 2025-06-12
fix: usability/terminology fix, Accordions -> Accordion

agurtovoy:
## What/Why?

Rename `Accordions` to `Accordion` to avoid confusion and align with [standard UX terminology](https://chatgpt.com/share/676758f0-d0b0-8012-968f-4b0949cc6826). Note that the underlying Soul component is already correctly named, this change only affects the Makeswift component and builder naming.

Aside from the terminology update, this change is fully backward compatible and does not affect user-visible component behavior.

## Testing

https://github.com/user-attachments/assets/65a1d834-50b9-43f3-a755-78ed5a19d0d0


core/lib/makeswift/components/accordion/accordion.makeswift.tsx
```diff
@@ -41,21 +41,21 @@ runtime.registerComponent(
   },
   {
     type: 'primitive-accordions',
```
agurtovoy, line 43: Keeping the component type as-is for backward compatibility.

jorgemoya (approved):

PR #2418, opened 2025-06-12
chore: unify Makeswift component registrations

agurtovoy:
## What/Why?

For historical reasons, our codebase contains a mix of different Makeswift component registration patterns, which fails us on multiple fronts:

- We send mixed signals about which pattern users should follow when registering their own components.

- The prevailing pattern — defining a React component inline and registering it with Makeswift in the same file, `component.makeswift.tsx` — only works if the component doesn’t rely on client-only functionality (e.g., hooks or browser APIs). This constraint is implicit and not enforced by tooling, making the setup fragile for both our users and [ourselves](https://github.com/bigcommerce/catalyst/pull/2379).

- More broadly, the `component.makeswift.tsx` convention was originally intended for registering preexisting user components in Makeswift, with the goal of clearly delineating Makeswift-specific code. In our case, however, all components in the `lib/makeswift/components` folder are inherently Makeswift-specific, and all files within each component directory contribute to their Makeswift bindings. In this context, singling out one file with a `.makeswift` suffix misrepresents the intent and structure of our file organization. 
 
- In [multi-file registration cases](https://github.com/bigcommerce/catalyst/blob/integrations/makeswift/core/lib/makeswift/components/site-footer) that follow the `component.makeswift.tsx` convention, the naming of individual files is inconsistent with the rest of the codebase, where we generally [use `index` files](https://github.com/bigcommerce/catalyst/blob/integrations/makeswift/core/components/header) and [avoid prefixing filenames with the component name](https://github.com/bigcommerce/catalyst/blob/integrations/makeswift/core/components/analytics) inside its directory.

This PR reconciles the issues outlined above by unifying all Makeswift component registrations under the following pattern:

```shell
component/
├── register.ts     # component registration
├── client.tsx      # (client) React component definition, if any
└── index.tsx       # RSC entry point, if any
```

There are no other semantic changes.

<table>
<tr><th>Before</th> <th>After</th></tr>
<tr>
<td>
before
</td>
<td>
after
</td>
<table>

## Testing

https://github.com/user-attachments/assets/a55a5a6c-26a5-4d85-99c0-ea9141d38e62

## Migration

Please see above.


core/lib/makeswift/components.ts
```diff
@@ -1,20 +1,19 @@
...
 import './components/product-detail/register';
-
+import './components/products-carousel/register';
+import './components/products-list/register';
+import './components/section/register';
+import './components/site-footer/register';
+import './components/site-header/register';
 import './components/site-theme/register';
+import './components/slideshow/register';
+import './components/sticky-sidebar/register';
```
agurtovoy, line 16: Reordered the imports alphabetically.

fikrikarim (approved):

migueloller (approved):
Thanks!

PR #2424, opened 2025-06-13
Add e2e checkout test, fix some flaky add to cart actions

jordanarldt:
## What/Why?
- Rewrite e2e test for checkout flow
- Fix some other tests to wait for the toast message when adding product to the cart. I noticed that sometimes the `networkidle` is flaky in playwright, so it's more reliable to wait for the toast.

## Testing
image

## Migration
Copy changes from `/core/tests` directory


jordanarldt:
> Would you mind fixing the add-to-cart fix in the compare.spec.ts as well?

Sure, I'll do that now!

PR #2425, opened 2025-06-14
fix: remove `await connection()` workaround on 404 pages

agurtovoy:
## What/Why?

This was originally put in place to work around the issue of non-published pages not being editable in the builder. That issue has long since been resolved, so the workaround is no longer necessary.

## Testing

As an extra test, here's a demo of a 404 page with an editable Makeswift component on it:

https://github.com/user-attachments/assets/21a264d7-00e4-4f90-b676-8739da43fc7f

### DIFF
--- a/.changeset/evil-dogs-shine.md
+++ b/.changeset/evil-dogs-shine.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Remove `await connection()` workaround on 404 pages
--- a/core/lib/makeswift/page.tsx
+++ b/core/lib/makeswift/page.tsx
@@ -1,5 +1,4 @@
 import { notFound } from 'next/navigation';
-import { connection } from 'next/server';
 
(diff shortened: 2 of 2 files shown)


ferrata (approved):

fikrikarim (changes requested):
From the original [PR](https://github.com/bigcommerce/catalyst/pull/1889), the 404 pages not being editable only happened when the catch-all page is PPR (◐) instead of server-rendered (ƒ).

I checked the [preview deployment](https://vercel.com/bigcommerce-platform/catalyst-canary/2cqUQ1cF8aVZ6FXR4n4oLkNaWo7Q) from this branch, and the catch-all page is server-rendered.
```
ƒ /[locale]/[...rest]
```
This is probably because we're calling functions that force the page to be server-rendered.

I'm worried that this issue will resurface again in the future if the catch-all page is updated to be properly rendered with PPR. 

Perhaps we can test if the 404 page is still editable when the page is rendered with PPR?

PR #2441, opened 2025-06-18
fix: manually prepend 1.0.0 entry in core/CHANGELOG.md

matthewvolk:
## What/Why?

We may have overloaded the Changesets Action with the number of `.changeset/*.md` entries included in #1892. This PR collected all of  the entries in that PR and re-wrote them to the top of the `core/CHANGELOG.md` file. 

## Testing

N/A

## Migration

N/A

### DIFF
--- a/core/CHANGELOG.md
+++ b/core/CHANGELOG.md
@@ -1,5 +1,897 @@
 # Changelog
 
+## 1.0.0
+
+### Major Changes
+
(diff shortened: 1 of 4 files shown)


chanceaclark (changes requested):
With changesets, the changes are scoped to each package. I am not sure we want to have a global `CHANGELOG.md` here as we'll need to manually update it.

chanceaclark (approved):
My bad, I didn't noticed it was scoped to `core/` 🤦‍♂️

jorgemoya (approved):

PR #2446, opened 2025-06-25
(feat) Add launch.json example for vscode debugger

bc-0dp:
## What/Why?
Adds `.vscode/launch.json` to provide a shared debug configuration for VS Code users.
This streamlines local development by allowing anyone on the team to start debugging without manual setup, improving consistency and developer experience.

## Testing
1. From the project root execute `cp .vscode/launch.json.example .vscode/launch.json`
1. Open VScode debugger `command+D`, and press `F5` or chose a debug profile and run

Screenshot 2025-06-25 at 15 30 38


.vscode/launch.example.json
```diff
@@ -0,0 +1,39 @@
...
+            "env": {
+                "NODE_ENV": "development"
+            },
+            "serverReadyAction": {
+                "action": "openExternally",
+                "killOnServerStop": true,
+                "pattern": "- Local:.+(https?://.+)",
+                "uriFormat": "%s"
+            },
+            "preLaunchTask": "generate"
```
chanceaclark: Doesn't our `dev` task already run the `generate` command?
```json
"dev": "npm run generate && next dev",
```

bc-0dp: `Catalyst: debug server-side` will run `dev` but `Catalyst: debug full stack` does not.
So I think it's needed.

But alternatively we could just have `Catalyst: debug server-side` and scrap `Catalyst: debug full stack`
They almost do the same except `Catalyst: debug full stack` runs a full debugger to my knowledge.

PR #2461, opened 2025-07-02
chore: update @bigcommerce/catalyst deps

matthewvolk:
## What/Why?

Update dependencies - includes major updates for `commander`. 

## Testing

Vitest:
```
vitest run

 RUN  v3.2.4 /Users/matt.volk/code/catalyst/packages/cli

stdout | tests/index.spec.ts

◢ @bigcommerce/catalyst v0.1.0


 ✓ tests/commands/version.spec.ts (2 tests) 3ms
 ✓ tests/index.spec.ts (2 tests) 2ms

 Test Files  2 passed (2)
      Tests  4 passed (4)
   Start at  19:15:07
   Duration  278ms (transform 50ms, setup 0ms, collect 69ms, tests 5ms, environment 0ms, prepare 106ms)
```

## Migration

N/A

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -21,20 +21,20 @@
     "node": "^20.0.0 || ^22.0.0"
   },
   "dependencies": {
-    "@commander-js/extra-typings": "^12.1.0",
+    "@commander-js/extra-typings": "^14.0.0",
     "chalk": "^5.3.0",
-    "commander": "^12.1.0"
+    "commander": "^14.0.0"
   },
   "devDependencies": {
     "@bigcommerce/eslint-config": "^2.11.0",
     "@bigcommerce/eslint-config-catalyst": "workspace:^",
     "@types/node": "^22.15.30",
-    "@vitest/coverage-v8": "^3.2.2",
-    "@vitest/ui": "^3.2.2",
+    "@vitest/coverage-v8": "^3.2.4",
+    "@vitest/ui": "^3.2.4",
     "eslint": "^8.57.1",
-    "prettier": "^3.5.3",
+    "prettier": "^3.6.2",
     "tsup": "^8.5.0",
     "typescript": "^5.8.3",
-    "vitest": "^3.2.2"
+    "vitest": "^3.2.4"
   }
 }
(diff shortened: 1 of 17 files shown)


packages/cli/package.json
```diff
@@ -21,20 +21,20 @@
     "node": "^20.0.0 || ^22.0.0"
   },
   "dependencies": {
-    "@commander-js/extra-typings": "^12.1.0",
+    "@commander-js/extra-typings": "^14.0.0",
     "chalk": "^5.3.0",
-    "commander": "^12.1.0"
+    "commander": "^14.0.0"
```
chanceaclark, line 26: 🍹 [callout-only] Only thing I see as a big thing is this [breaking change](https://github.com/tj/commander.js/releases/tag/v13.0.0#:~:text=Breaking%3A%20excess%20command%2Darguments%20cause%20an%20error%20by%20default%2C%20see%20migration%20tips). Only if we update `commander` in the `create-catalyst` package will we need to thoroughly test the OCC CLI commands.

packages/cli/package.json
```diff
@@ -21,20 +21,20 @@
...
-    "@vitest/ui": "^3.2.2",
+    "@vitest/coverage-v8": "^3.2.4",
+    "@vitest/ui": "^3.2.4",
     "eslint": "^8.57.1",
-    "prettier": "^3.5.3",
+    "prettier": "^3.6.2",
     "tsup": "^8.5.0",
     "typescript": "^5.8.3",
-    "vitest": "^3.2.2"
+    "vitest": "^3.2.4"
```
chanceaclark, line 38: Some of these updates we probably want to do globally so we don't have separate versions for other packages (thinking about `prettier` here). I recommend running `pnpm up -ir --latest` (or no `--latest`) in the root of the repo to update most of these (minus the commander ones). Running that along with `pnpm dedupe` would be good.

chanceaclark: Should be good now

matthewvolk: Thank you sir! So interesting that `prettier` decided to wake up for those files now that it's a dependency in the workspace root

chanceaclark (approved):

Issue #2109, opened 2025-03-18
JWT login is not working due to unhandled error


MattyyttaM:
**Describe the bug**
When I go to the route `/login/token/[token]` to use the customer login with JWT, the request errors.

The request to the GraphQL API for the JWT login is successful, however a call is made to `handleCartLogin` within the `loginWithJwt` function inside `core/auth` which errors, which causes the auth to not be successful

**To Reproduce**

1. Create a customer SSO JWT
2. Go to `/login/token/[token]` replacing `[token]` with the JWT
3. This should error and take you to the login, however the JWT login has actually been succesfull.

**Expected behavior**
Should login successfully with JWT.

**Screenshots**

Showing that the GraphQL API returned successfully
![Image](https://github.com/user-attachments/assets/e02b84a0-3125-4009-a3bb-01fa1fa56f59)

A couple of steps through where after it errors
![Image](https://github.com/user-attachments/assets/881db5d6-3294-4117-9926-8dccbcf32842)

PR #2464, opened 2025-07-02
chore(core): remove runtime edge declarations

jorgemoya:
## What/Why?
Remove edge runtime declarations to be able to run Catalyst with OpenNext.

We had previously removed these declarations, unsure if there's a specific reason these routes had these declared.

## Testing
Sitemap redirect works
Screenshot 2025-07-03 at 11 37 00 AM

Auth flow works
Screenshot 2025-07-03 at 11 37 37 AM

Admin route redirects to bigcommerce
Screenshot 2025-07-03 at 11 45 43 AM

## Migration
Remove `export const runtime = 'edge';` from all files.

### DIFF
--- a/.changeset/silent-ideas-flash.md
+++ b/.changeset/silent-ideas-flash.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Remove edge runtime declarations to be able to run Catalyst with OpenNext.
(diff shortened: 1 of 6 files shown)


chanceaclark:
@jorgemoya awesome, thank you!

I know we have #2107 and #2109 still open so just want to confirm that it still works.

chanceaclark (approved):

PR #2474, opened 2025-07-10
Use min and max purchase quantity to set bounds of quantity number input

bookernath:
## What/Why?
Respect the min and max purchase quantity of the product, as returned by GraphQL.

## Testing
Before:
image

After:
image

## Migration
N/A


migueloller (approved):
LGTM 👍🏻

PR #2475, opened 2025-07-10
Replace Vercel KV adapter with Vercel Runtime Cache API

bookernath:
## What/Why?
Implement Vercel Runtime Cache API as replacement for Vercel KV

Remove Vercel KV and BC KV adapters as they are no longer needed.

As a next step, we could refactor the KV adapters to look more like simple fetch cache adapters, as the only purpose of these adapters is to cache a fetch in middleware.


core/package.json
```diff
@@ -50,7 +50,7 @@
     "lodash.debounce": "^4.0.8",
     "lru-cache": "^11.1.0",
     "lucide-react": "^0.474.0",
-    "next": "15.4.2-canary.10",
+    "next": "15.5.1-canary.4",
```
migueloller, line 53: Was upgrading Next.js required to leverage the new Runtime Cache?

core/lib/kv/adapters/upstash.ts
```diff
@@ -2,23 +2,14 @@ import { Redis } from '@upstash/redis';
 
 import { KvAdapter, SetCommandOptions } from '../types';
 
-import { MemoryKvAdapter } from './memory';
-
-const memoryKv = new MemoryKvAdapter();
```
migueloller, line 7: How come we removed the memory adapter from the Upstash adapter? This seems unrelated to adding Runtime Cache.

Also, should we have deleted the `./memory.ts` file as well?

core/lib/kv/adapters/vercel-runtime-cache.ts
```diff
@@ -0,0 +1,78 @@
...
+
+      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+      return values as Array<Data | null>;
+    } catch (error) {
+      this.logger(
+        `RUNTIME_CACHE ACCESS ERROR - Returning null values: ${error instanceof Error ? error.message : 'Unknown error'}`,
+      );
+
+      // Return null for all keys if Runtime Cache is unavailable
+      return keys.map(() => null);
```
migueloller, line 47: Is the idea here that returning `null` behaves as if the cache entries don't exist?

migueloller (commented):
LGTM 👍🏻 

Just had a couple of questions.

core/package.json
bookernath: No, this was initially desirable when I thought we would move to Node.js middleware as part of this change, so I wanted to be on a version where that feature was considered "stable". Upon review I probably should have decoupled the version bump from this PR once it was clear it was not necessary.

core/lib/kv/adapters/upstash.ts
bookernath: Memory KV is now handled centrally outside of the individual adapters in the kv/index.ts.

core/lib/kv/adapters/vercel-runtime-cache.ts
bookernath: Yes, I want the application to treat this as a cache miss and gracefully continue to load the page even if the cache is degraded. Memory KV will still provide some offload in most scenarios.

core/package.json
migueloller: Yeah, I would suggested we do that if we hadn't merged already. Thanks for the explanation 🙏🏻

PR #2477, opened 2025-07-14
Render scripts from Script Manager

bookernath:
## What/Why?
Checks for scripts in Script Manager tied to the Catalyst storefront's channel and renders them in the root layout using [next/script](https://nextjs.org/docs/app/api-reference/components/script).

Related docs PR: https://github.com/bigcommerce/docs/pull/1014

⚠️ This makes a best-effort attempt to map [Scripts API](https://developer.bigcommerce.com/docs/integrations/scripts) properties to the optimal settings in a [next/script](https://nextjs.org/docs/app/api-reference/components/script) context, but it does prefer performance over parity for purposes of rendering scripts on Catalyst ⚠️ 

⚠️ Scripts that worked on Stencil are NOT guaranteed to work on Catalyst, and may need to be re-written, especially if they depend on the DOM or need to fire on every navigation ⚠️ 

⚠️ Overall this is not the recommended way of integrating apps into Catalyst, but it may be appropriate for some use cases such as analytics pixels and popups ⚠️ 

⚠️ This does NOT yet handle cookie consent for scripts, all scripts will render all the time until we add a consent manager ⚠️ 

Future work will add a consent manager into Catalyst which will apply to both scripts as well as analytics calls, respecting cookie consent automatically.

## Testing
Preview deployment

- [x] Inline script works
- [x] CDN script works
- [x] SRI works

## Migration
N/A

### DIFF
--- a/core/components/scripts/README.md
+++ b/core/components/scripts/README.md
@@ -1,0 +1,66 @@
+# Script Components
+
+This directory contains components for rendering BigCommerce scripts using Next.js Script component.
+
+## Components
+
+### ScriptManagerScripts
+A single component that renders scripts with configurable strategy. Location filtering is now handled at the GraphQL level, simplifying the component logic.
+
+## Usage
+
+```tsx
+import { ScriptManagerScripts } from '~/components/scripts';
+
+// In your layout component:
+<head>
+  <ScriptManagerScripts 
+    scripts={data.site.content.headerScripts} 
+    strategy="afterInteractive" 
+  />
+</head>
+<body>
+  {children}
+  <ScriptManagerScripts 
+    scripts={data.site.content.footerScripts} 
+    strategy="lazyOnload" 
+  />
+</body>
+```
+
+## Architecture
+
+The component accepts:
+- `scripts`: Pre-filtered GraphQL scripts data (headerScripts or footerScripts)
+- `strategy`: 'afterInteractive', 'lazyOnload', or other Next.js Script strategies
+
+## Script Processing
+
+The component handles both types of BigCommerce scripts:
+
+### External Scripts (SrcScript)
+```tsx
+<Script 
+  id="bc-script-123" 
+  src="https://example.com/script.js" 
+  strategy="afterInteractive"
+  integrity="sha256-abc123..."
+  crossOrigin="anonymous"
+/>
+```
+
+### Inline Scripts (InlineScript)
+Inline scripts are rendered using `dangerouslySetInnerHTML` for reliable execution:
+```tsx
+<Script 
+  id="bc-script-456" 
+  strategy="lazyOnload"
+  dangerouslySetInnerHTML={{ __html: scriptContent }}
+/>
+```
+
+## Performance & Security
+
+- Header scripts use `afterInteractive` strategy for critical functionality  
+- Footer scripts use `lazyOnload` strategy for better performance
+- Integrity hashes are included when available for security
--- a/core/components/scripts/fragment.ts
+++ b/core/components/scripts/fragment.ts
@@ -1,0 +1,39 @@
+import { graphql } from '~/client/graphql';
+
+export const ScriptsFragment = graphql(`
+  fragment ScriptsFragment on Content {
+    headerScripts: scripts(
+      first: 50
+      filters: { visibilities: [ALL_PAGES, STOREFRONT], location: HEAD }
+    ) {
+      ...ScriptTypeConnectionFragment
+    }
+    footerScripts: scripts(
+      first: 50
+      filters: { visibilities: [ALL_PAGES, STOREFRONT], location: FOOTER }
+    ) {
+      ...ScriptTypeConnectionFragment
+    }
+  }
+
+  fragment ScriptTypeConnectionFragment on ScriptTypeConnection {
+    edges {
+      node {
+        __typename
+        integrityHashes {
+          hash
+        }
+        entityId
+        consentCategory
+        location
+        visibility
+        ... on InlineScript {
+          scriptTag
+        }
+        ... on SrcScript {
+          src
+        }
+      }
+    }
+  }
+`);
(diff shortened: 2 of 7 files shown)


core/components/scripts/scripts.tsx
```diff
@@ -0,0 +1,121 @@
...
+interface ScriptProps {
+  src?: string;
+  dangerouslySetInnerHTML?: {
+    __html: string;
+  };
+  strategy: 'afterInteractive' | 'lazyOnload';
+  integrity?: string;
+  id?: string;
+  crossorigin?: string;
+}
```
chanceaclark: A lot of these props can be computed via the `FragmentOf` util exported from the client + Next.js `Script` props. Can we update this to utilize those instead?

core/components/scripts/scripts.tsx
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * Renders scripts that should be placed in the <head> tag
+ * Uses afterInteractive strategy for critical functionality
+ * @param {ScriptsProps} props - The props containing scripts data
+ * @param {ScriptsData | null} props.scripts - The scripts data from GraphQL
+ * @returns {JSX.Element | null} JSX element containing header scripts
+ */
+export const HeaderScripts = ({ scripts }: ScriptsProps) => {
+  return <ScriptRenderer location="head" scripts={scripts} strategy="afterInteractive" />;
+};
```
chanceaclark: I think this might not be the right abstraction. Can we just utilize the server to filter the scripts and just have one component we utilize: `ScriptRenderer`? See https://github.com/bigcommerce/catalyst/pull/2477/files#r2205218637

core/components/scripts/README.md
```diff
@@ -0,0 +1,91 @@
+# Script Components
+
+This directory contains components for rendering BigCommerce scripts using Next.js Script component.
+
+## Components
+
+Both `HeaderScripts` and `FooterScripts` are exported from a single `scripts.tsx` file that follows DRY principles by sharing common logic through an internal `ScriptRenderer` component.
```
chanceaclark: Per https://github.com/bigcommerce/catalyst/pull/2477/files#r2205200995, I don't think we should splitting it up into `HeaderScripts` or `FooterScripts`, it should probably be just `ScriptRenderer` (or a better name) that is used. This is probably not the right abstraction, thus I think we shouldn't add this comment, or anything that mentioned DRY for that matter.
```suggestion
```

core/components/scripts/fragment.ts
```diff
@@ -0,0 +1,26 @@
...
+            scriptTag
+          }
+          ... on SrcScript {
+            src
+          }
+        }
+      }
+    }
+  }
+`);
```
chanceaclark: We probably should actually be doing the filtering on the server rather than in the app:
```suggestion
export const ScriptsFragment = graphql(`
  fragment ScriptsFragment on Content {
    headScripts: scripts(first: 50, filters: { visibilities: [ALL_PAGES, STOREFRONT], location: [HEAD] }) {
      edges {
        node {
          ...ScriptFragment
        }
      }
    }
    footerScripts: scripts(first: 50, filters: { visibilities: [ALL_PAGES, STOREFRONT], location: [FOOTER] }) {
      edges {
        node {
          ...ScriptFragment
        }
      }
    }
  }
`, [ScriptFragment]);

const ScriptFragment = graphql(`
  __typename
  integrityHashes {
    hash
  }
  entityId
  consentCategory
  location
  visibility
  ... on InlineScript {
    scriptTag
  }
  ... on SrcScript {
    src
  }
`)
```

bookernath: Here I am pre-filtering for just any scripts relevant to Catalyst; scripts that render on the checkout page or the order confirmation page are irrelevant to Catalyst because it does not render those pages, so doing this filter at the API layer makes sense.

chanceaclark: >on the server

I'm referencing the BC servers/API layer here since I added `location: [FOOTER]` or `location: [HEAD]` as part of this suggestion

bookernath: I see what you mean - my priority here was doing everything in 1 API request/1 thing to cache instead of breaking this out into separate requests. Do you view the benefits of breaking up the requests as worth the tradeoffs? I could also structure the query to do this with aliases for e.g. `headScripts` vs `footerScripts`

chanceaclark: >I could also structure the query to do this with aliases for e.g. headScripts vs footerScripts

That's what my code suggestion does 🤔

core/components/scripts/scripts.tsx
```diff
@@ -0,0 +1,59 @@
...
+        // Handle inline scripts (InlineScript)
+        if (script.__typename === 'InlineScript' && script.scriptTag) {
+          const scriptMatch = /<script[^>]*>([\s\S]*?)<\/script>/i.exec(script.scriptTag);
+
+          if (scriptMatch?.[1]) {
+            scriptProps.dangerouslySetInnerHTML = {
+              __html: scriptMatch[1],
+            };
+          }
+        }
```
chanceaclark: I think we should have `dom-purify` here in the repo that you might want to use instead of a regex 🤔

core/components/scripts/scripts.tsx
```diff
@@ -0,0 +1,59 @@
...
+}
+
+export const ScriptManagerScripts = ({ scripts, strategy }: ScriptRendererProps) => {
+  if (!scripts?.edges) return null;
+
+  const scriptNodes = removeEdgesAndNodes(scripts);
+
+  return (
+    <>
+      {scriptNodes.map((script) => {
```
chanceaclark: 🍹 Only if you want to update this... but a better practice, which might seem weird, is to do two `.map`'s here. One to build the props, then another one to render. It allows each `.map` to have separate concerns.
```tsx
{scriptNodes
  .map((script) => {
    // ...
    return {
      src: "...",
      // ....
    };
  })
  .map((script) => {
    return (
      <Script key={script.id} {...script} />
    );
  })
}
```

chanceaclark (commented):
The `dom-purify` is the only major thing, otherwise looks good to me.

chanceaclark (approved):

PR #2479, opened 2025-07-16
feat: recursively copy cwd into temp folder

matthewvolk:
## What/Why?

* Adds [`consola`](https://github.com/unjs/consola) for logging
* Runs CLI tests in all popular OS's to match the different environments where our customers run the CLI 
* Recursively copy a Catalyst project to a temporary directory, OS-agnostic API's from Node.js are being used
* Includes tests

## Testing

Includes unit tests

## Migration

N/A

### DIFF
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -1,0 +1,76 @@
+import { Command } from 'commander';
+import consola from 'consola';
+import { cp, mkdtemp, rm } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { join, relative, sep } from 'node:path';
+
+const SKIP_DIRS = new Set([
+  'node_modules',
+  '.bigcommerce',
+  '.git',
+  '.turbo',
+  '.next',
+  '.vscode',
+  '.github',
+  '.changeset',
+  'dist',
+]);
+
+export function createFilter(root: string, skipDirs: Set<string>) {
+  return (src: string) => {
+    const rel = relative(root, src);
+    const parts = rel.split(sep);
+
+    return !parts.some((part) => skipDirs.has(part));
+  };
+}
+
+export async function mkTempDir() {
+  const prefix = join(tmpdir(), 'catalyst-build-');
+  const path = await mkdtemp(prefix);
+
+  consola.info(`Created temporary directory: ${path}`);
+
+  return [
+    path,
+    async () => {
+      try {
+        consola.info(`Cleaning up temporary directory: ${path}`);
+        await rm(path, { recursive: true, force: true });
+        consola.success('Cleanup complete');
+      } catch (error) {
+        consola.warn(`Failed to clean up temporary directory: ${path}`, error);
+      }
+    },
+  ] as const;
+}
+
+export const build = new Command('build')
+  .option('--keep-temp-dir', 'Keep the temporary directory after the build')
+  .action(async (options) => {
+    const [tmpDir, rmTempDir] = await mkTempDir();
+
+    try {
+      consola.start('Copying project to temp directory...');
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/build.ts
```diff
@@ -0,0 +1,76 @@
...
+        consola.success('Cleanup complete');
+      } catch (error) {
+        consola.warn(`Failed to clean up temporary directory: ${path}`, error);
+      }
+    },
+  ] as const;
+}
+
+export const build = new Command('build')
+  .option('--keep-temp-dir', 'Keep the temporary directory after the build')
```
jorgemoya, line 49: Should we set the pattern to use `kebab case` for all options? In preference over `camel case`?

packages/cli/src/commands/build.ts
```diff
@@ -0,0 +1,76 @@
+import { Command } from 'commander';
+import consola from 'consola';
+import { cp, mkdtemp, rm } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { join, relative, sep } from 'node:path';
```
jorgemoya, line 5: Curious to understand why we use these libs instead of using `path` or `fs`?

packages/cli/tests/commands/build.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+    await mkdir(join(path, 'subdir'));
+    await writeFile(join(path, 'file.txt'), 'test content');
+    await writeFile(join(path, 'subdir', 'nested.txt'), 'nested content');
+
+    await cleanup();
+
+    await expect(access(path)).rejects.toThrow();
+  });
+
+  test('cleanup is idempotent', async () => {
```
jorgemoya, line 40: `idempotent` TIL

packages/cli/tests/commands/build.spec.ts
```diff
@@ -0,0 +1,67 @@
+import { access, mkdir, stat, writeFile } from 'node:fs/promises';
+import { join } from 'node:path';
+import { describe, expect, test } from 'vitest';
+
+import { createFilter, mkTempDir } from '../../src/commands/build';
+
+describe('mkTempDir', () => {
```
jorgemoya, line 7: Should we wrap all describes under the `build command` describe?

jorgemoya (approved):
Looks good to me, just have some questions.

packages/cli/src/commands/build.ts
matthewvolk: Not required, I just do it to explicitly indicate a built-in Node.js module and avoid ambiguity between built-ins and potentially conflicting userland modules (like https://www.npmjs.com/package/test vs `node:test`)

packages/cli/src/commands/build.ts
matthewvolk: Yeah, POSIX [standardized](https://www.gnu.org/prep/standards/html_node/Option-Table.html) on `--kebab-case` convention, so it's what people are used to IMO. Commander [automatically transforms those to camel case too](https://github.com/tj/commander.js?tab=readme-ov-file#options):
> Multi-word options such as "--template-engine" are camel-cased, becoming `program.opts().templateEngine` etc.

packages/cli/tests/commands/build.spec.ts
matthewvolk: Yeah, in other words, `cleanup()` should not fail on the second call

packages/cli/tests/commands/build.spec.ts
matthewvolk: Chatted offline, will leave as is until it gets unwieldy

PR #2481, opened 2025-07-16
feat(cli): scaffold deploy command and generate bundle zip

jorgemoya:
## What/Why?

- Scaffolds `deploy` command
- `deploy` builds the zip file and stores in `.bigcommerce/dist/bundle.zip`

## Testing
Added unit tests

## Migration
N/A

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -21,7 +21,6 @@
     "node": "^20.0.0 || ^22.0.0"
   },
   "dependencies": {
-    "@commander-js/extra-typings": "^12.1.0",
     "adm-zip": "^0.5.16",
     "chalk": "^5.3.0",
     "commander": "^14.0.0",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -21,6 +21,9 @@
     "node": "^20.0.0 || ^22.0.0"
   },
   "dependencies": {
+    "@commander-js/extra-typings": "^12.1.0",
+    "adm-zip": "^0.5.16",
+    "archiver": "^5.3.2",
     "chalk": "^5.3.0",
     "commander": "^14.0.0",
     "consola": "^3.4.2"
@@ -29,6 +32,8 @@
     "@bigcommerce/eslint-config": "^2.11.0",
     "@bigcommerce/eslint-config-catalyst": "workspace:^",
     "@commander-js/extra-typings": "^14.0.0",
+    "@types/adm-zip": "^0.5.7",
+    "@types/archiver": "^5.1.10",
     "@types/node": "^22.15.30",
     "@vitest/coverage-v8": "^3.2.4",
     "@vitest/ui": "^3.2.4",
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -23,7 +23,6 @@
   "dependencies": {
     "@commander-js/extra-typings": "^12.1.0",
     "adm-zip": "^0.5.16",
-    "archiver": "^5.3.2",
     "chalk": "^5.3.0",
     "commander": "^14.0.0",
     "consola": "^3.4.2"
@@ -33,7 +32,6 @@
     "@bigcommerce/eslint-config-catalyst": "workspace:^",
     "@commander-js/extra-typings": "^14.0.0",
     "@types/adm-zip": "^0.5.7",
-    "@types/archiver": "^5.1.10",
     "@types/node": "^22.15.30",
     "@vitest/coverage-v8": "^3.2.4",
     "@vitest/ui": "^3.2.4",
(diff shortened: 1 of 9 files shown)


packages/cli/src/commands/deploy.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  const outputZip = join(distDir, 'bundle.zip');
+
+  const output = createWriteStream(outputZip);
+  const archive = archiver('zip', { zlib: { level: 9 } });
+
+  output.on('close', () => {
+    consola.success(`Created ${outputZip} (${archive.pointer()} total bytes)`);
+  });
```
jorgemoya: Not sure if necessary 🤔

packages/cli/src/commands/deploy.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  try {
+    await access(distDir);
+  } catch {
+    await mkdir(distDir, { recursive: true });
+  }
+
+  const outputZip = join(distDir, 'bundle.zip');
+
+  const output = createWriteStream(outputZip);
```
jorgemoya: This is a stream, not sure if ideal.

packages/cli/tests/commands/deploy.spec.ts
```diff
@@ -0,0 +1,48 @@
+import { access, mkdir, writeFile, stat } from 'node:fs/promises';
+import { tmpdir } from 'node:os';
+import { dirname, join } from 'node:path';
+import { beforeAll, describe, expect, test } from 'vitest';
+import AdmZip from 'adm-zip';
+
+import { generateBundleZip } from '../../src/commands/deploy';
+
+const tmpDir = tmpdir();
```
jorgemoya: Should I do cleanup when using `tmpdir`? 🤔

packages/cli/package.json
```diff
@@ -21,6 +21,9 @@
     "node": "^20.0.0 || ^22.0.0"
   },
   "dependencies": {
+    "@commander-js/extra-typings": "^12.1.0",
+    "adm-zip": "^0.5.16",
+    "archiver": "^5.3.2",
```
chanceaclark: Seems like we are using two zip libraries. Is there any way we can get away with just one?

jorgemoya: Good call out. I was using `archiver` first, which can't read, so I added `adm-zip` to read, but turns out `adm-zip` can do both. Replaced.

packages/cli/package.json
```diff
@@ -21,6 +21,8 @@
     "node": "^20.0.0 || ^22.0.0"
   },
   "dependencies": {
+    "@commander-js/extra-typings": "^12.1.0",
```
jorgemoya: I'm not sure we need this, checking.

packages/cli/tests/commands/deploy.spec.ts
chanceaclark: Nah, locally it should get garbage collected over time, and when running on CI the fs gets nuked after the run.

chanceaclark (approved):

packages/cli/tests/commands/deploy.spec.ts
matthewvolk: @jorgemoya I have a little utility that makes tempdir's, if you want to use it for this:
https://github.com/bigcommerce/catalyst/blob/778bd214004e31383c3db5c6bfc7b52d425b8f7c/packages/cli/src/commands/build.ts#L28-L46

We can move it to a `lib` folder or something. It can be refactored to take a prefix so you can prefix test tmpdir's differently than my build one if you'd like

packages/cli/src/commands/deploy.ts
```diff
@@ -0,0 +1,58 @@
...
+import { Command } from 'commander';
+import { consola } from 'consola';
+import { access, mkdir, readdir } from 'node:fs/promises';
+import { join } from 'node:path';
+
+export const generateBundleZip = async (rootDir: string) => {
+  consola.info('Generating bundle.zip');
+
+  const buildDir = join(rootDir, '.open-next');
+  const distDir = join(rootDir, '.bigcommerce/dist');
```
matthewvolk: Sorry my `build` command isn't quite finished, because it would be easier to illustrate this—the `build` command will actually build an output folder that looks like:
```
.
└── .bigcommerce/
    ├── dist/
    │   ├── worker.js
    │   ├── worker.map.js
    │   └── assets/
    └── project.json
```

So I don't think you'll need to manually build this. 

With that said, since my `build` command isn't finished yet, I'm fine merging this so we can test and then changing it when `build` is finished.

core/.open-next.zip
matthewvolk, line 1: You didn't mean to commit this, did you?

packages/cli/src/commands/deploy.ts
jorgemoya: Ah, gotcha! I can make the changes now to adjust for this.

PR #2484, opened 2025-07-18
feat(cli): generate upload signature and upload bundle.zip

jorgemoya:
## What/Why?
Next steps for the deploy command:

- Add `storeHash` and `accessToken` variables to deploy command
- Generate upload signature
- Upload bundle to signed url
- Add `msw` to intercept requests for tests

## Testing
Unit tests

Tested locally
Screenshot 2025-07-18 at 1 59 11 PM

## Migration
N/A


packages/cli/src/commands/deploy.ts
```diff
@@ -33,19 +35,99 @@ export const generateBundleZip = async (rootDir: string) => {
...
 
 interface DeployOptions {
   rootDir: string;
+  apiHost: string;
 }
 
 export const deploy = new Command('deploy')
   .description('Deploy the application to Cloudflare')
+  .argument('<storeHash>')
+  .argument('<accessToken>')
```
jorgemoya: I would make these`options` if/when we're reading from .env. Should I already be reading from env? 🤔

matthewvolk: Yeah, reading from env will be especially useful in a CI context. Let's change to something like:
```suggestion
  .addOption(
    new Option(
      '--store-hash <hash>',
      'BigCommerce store hash. Can be found in the URL of your store Control Panel.',
    ).env('BIGCOMMERCE_STORE_HASH'),
  )
  .addOption(
    new Option(
      '--access-token <token>',
      'BigCommerce access token. Can be found after creating a store-level API account.',
    ).env('BIGCOMMERCE_ACCESS_TOKEN'),
  )
```

packages/cli/src/commands/deploy.ts
```diff
@@ -33,19 +35,99 @@ export const generateBundleZip = async (rootDir: string) => {
...
+    if (!response.ok) {
+      consola.error(`Failed to fetch upload signature: ${response.status} ${response.statusText}`);
+      process.exit(1);
+    }
+
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+    const { data } = (await response.json()) as BigCommerceResponse<{
+      upload_url: string;
+      upload_uuid: string;
+    }>;
```
matthewvolk: 🍹Should we take the opportunity to install `zod`?

matthewvolk (approved):
The only required change I'd make is for the store hash/access token to read from env, otherwise this lgtm! 👍

packages/cli/src/commands/deploy.ts
jorgemoya: I thought `zod` was used to validate input? 🤔 

I'm open to it, just need to understand how we want to use it.

chanceaclark: `zod` should be used at boundaries where you don't know the returned type, so this is a perfect place to use it.

jorgemoya: Ok looking into it 👀

jorgemoya: Does it make sense that I still set a type for the `response.json()`?

```ts
// eslint-disable-next-line @typescript-eslint/consistent-type-assertions
const json = (await response.json()) as BigCommerceResponse<{
  upload_url: string;
  upload_uuid: string;
}>;

const { data } = UploadSignatureSchema.parse(json);
```

matthewvolk: The pattern I usually follow to get rid of the `any` assignment is:
```ts
const parsed = UploadSignatureSchema.parse(await response.json())
```

jorgemoya: 🤦 😆

jorgemoya: Following this pattern @chanceaclark suggested to be able to interact with the `res` if needed.

```
    const res: unknown = await response.json();
    const { data } = UploadSignatureSchema.parse(res);
```

PR #2485, opened 2025-07-18
feat: debounce quantity form per cart line item

### LINKED ISSUE #2476: Many requests when changing quantity in cart
**Describe the bug**
When you change the quantity in the cart using the plus and minus buttons, a request is sent to the /cart/ with every single click. This call isn’t particularly fast to begin with, and it becomes even more problematic when a large number of requests are sent in quick succession (see screenshot). Is this something that could be improved?

**To Reproduce**
1. Add a product to the cart
2. Open the cart page with the network inspector attached
3. Click 10 times in quick succession on the plus-button to increase the quantity to 11
4. Check the network inspector: there are 10 requests to /cart/

Also reproduceable on https://catalyst-demo.site/.

**Expected behavior**
When clicking in quick succession on the plus of minus button, I don't expect a direct request for each click. Instead, I expect a debounce, and only one request after some time (let's say 300-500ms).

**Screenshots**
Image

matthewvolk:
## What/Why?

The `updateLineItem` server action can be spammed pretty easily due to the fact that we are optimistically updating the cart line items in response to increment, decrement, and delete actions. Quantity updates were a good candidate for debouncing, which is exactly what this PR adds. With that said, this PR does not fix the fact that users can still spam delete actions on line items. That will need to be a fix for another PR. 

> [!IMPORTANT]
> This PR changes the CounterForm component pretty drastically. For one, it removes the HTML `<form>` entirely. When reviewing this PR, we should weigh the impact of decisions like that, I'd be interested in hearing everyone's thoughts. 

Fixes #2476 

## Testing

1. Navigate to Preview URL
2. Add a product to cart
3. Navigate to Cart
4. Open network panel
5. Rapidly increase the quantity of a cart line item by 10
6. Notice only one POST request to /cart is logged in the network panel.

## Migration

🚧 TODO

### DIFF
--- a/core/app/[locale]/(default)/cart/_actions/update-line-item.ts
+++ b/core/app/[locale]/(default)/cart/_actions/update-line-item.ts
@@ -53,7 +53,7 @@ export const updateLineItem = async (
   }
 
   switch (submission.value.intent) {
-    case 'increment': {
+    case 'update': {
       const parsedSelectedOptions = cartLineItem.selectedOptions.reduce<CartSelectedOptionsInput>(
         (accum, option) => {
           let multipleChoicesOptionInput;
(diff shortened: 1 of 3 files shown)


jorgemoya:
Is this something we still want to merge? Can we rebase if so?

chanceaclark:
Closing, superseded by #3103 (TRAC-284), which shipped the cart quantity/delete coalescing this PR was after.

PR #2486, opened 2025-07-21
feat: install and build all deps required to bundle catalyst

matthewvolk:
## What/Why?

Leverages [`nypm`](https://github.com/unjs/nypm) to install project dependencies, add the `@opennextjs/cloudflare` dependency to `core/`, and build the GraphQL API client in `packages/client`. These three steps are necessary in order to build the Catalyst application in `core/`. 

## Testing

I don't think unit tests are the best kind of test for this functionality, largely because `nypm` is [already tested](https://github.com/unjs/nypm/tree/main/test) and we aren't adding any additional logic on top like conditionally choosing a package manager, or constructing installation arguments. We could mock `nypm` and confirm we are calling it, but that feels like a test that's very prescriptive of the implementation. We should wait until we are adding logic on top of `nypm` before writing unit tests, otherwise we can later test this in integration/e2e tests. 

## Migration

N/A

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -24,7 +24,8 @@
     "adm-zip": "^0.5.16",
     "chalk": "^5.3.0",
     "commander": "^14.0.0",
-    "consola": "^3.4.2"
+    "consola": "^3.4.2",
+    "nypm": "^0.6.0"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


matthewvolk:
@jorgemoya we definitely should, I just figured those would be hard to test if we're mocking methods like `addDevDependency`—I was thinking we should test for those things in integration tests

PR #2489, opened 2025-07-21
feat: copy templates from cli to catalyst build tmpdir

matthewvolk:
## What/Why?

There are at least two templates that do not exist in the `core/` folder that need to exist in order for `opennextjs-cloudflare build` to succeed. 

The `opennextjs-cloudflare` CLI does create these files for you [if they do not exist](https://github.com/opennextjs/opennextjs-cloudflare/blob/54e65c5092d21aee3731c105f915ff17bef8fae0/packages/cloudflare/src/cli/index.ts#L32), however you must answer an interactive prompt. Rather than trying to answer the prompt in our CLI, we can just manually copy the templates from the CLI module in NPM. This also gives us more control over the contents of the templates in case we want to tweak any settings. 

## Testing

Pull down this branch, install deps (`pnpm i`), build the CLI (`pnpm -F "./packages/cli" build`), and run the build command in Catalyst root (`node packages/cli/dist/index.js build --keep-temp-dir`). You should see the files copied:
Screenshot 2025-07-21 at 1 14 05 PM

## Migration

N/A

### DIFF
--- a/packages/cli/src/lib/get-module-cli-path.ts
+++ b/packages/cli/src/lib/get-module-cli-path.ts
@@ -1,0 +1,11 @@
+/* eslint-disable no-underscore-dangle */
+
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+export function getModuleCliPath() {
+  const __filename = fileURLToPath(import.meta.url);
+  const __dirname = dirname(__filename);
+
+  return join(__dirname, '..');
+}
--- a/packages/cli/templates/open-next.config.ts
+++ b/packages/cli/templates/open-next.config.ts
@@ -1,0 +1,29 @@
+// @ts-nocheck
+import { defineCloudflareConfig } from '@opennextjs/cloudflare';
+import { purgeCache } from '@opennextjs/cloudflare/overrides/cache-purge/index';
+import kvIncrementalCache from '@opennextjs/cloudflare/overrides/incremental-cache/kv-incremental-cache';
+import doQueue from '@opennextjs/cloudflare/overrides/queue/do-queue';
+import queueCache from '@opennextjs/cloudflare/overrides/queue/queue-cache';
+import doShardedTagCache from '@opennextjs/cloudflare/overrides/tag-cache/do-sharded-tag-cache';
+
+export default defineCloudflareConfig({
+  incrementalCache: kvIncrementalCache,
+  queue: queueCache(doQueue, {
+    regionalCacheTtlSec: 5,
+  }),
+  routePreloadingBehavior: 'withWaitUntil',
+  tagCache: doShardedTagCache({
+    baseShardSize: 12,
+    regionalCache: true,
+    regionalCacheTtlSec: 5,
+    shardReplication: {
+      numberOfSoftReplicas: 4,
+      numberOfHardReplicas: 2,
+      regionalReplication: {
+        defaultRegion: 'enam',
+      },
+    },
+  }),
+  enableCacheInterception: false,
+  cachePurge: purgeCache({ type: 'durableObject' }),
+});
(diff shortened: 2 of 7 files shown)


packages/cli/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "display": "Node",
   "compilerOptions": {
     "strict": true,
-    "target": "es6",
+    "target": "es2020",
```
matthewvolk, line 6: Needed to bump this since "import.meta" is not available in es6 (see: https://github.com/evanw/esbuild/blob/8c71947edbe5a158fec3a6d1cbfea1e8d5cdee70/CHANGELOG-2020.md?plain=1#L2698)

According to https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping, es2020 has support back to Node 14. Since we're requiring at least Node 20, this change should be fine.

packages/cli/templates/open-next.config.ts
```diff
@@ -0,0 +1,8 @@
+// @ts-nocheck
+// default open-next.config.ts file created by @opennextjs/cloudflare
+import { defineCloudflareConfig } from "@opennextjs/cloudflare/config";
+import r2IncrementalCache from "@opennextjs/cloudflare/overrides/incremental-cache/r2-incremental-cache";
+
+export default defineCloudflareConfig({
+	incrementalCache: r2IncrementalCache,
```
matthewvolk: @chanceaclark I can't remember, did we change this to `kvIncrementalCache` just for experimentation? Or should I also change it here?

chanceaclark: I have been playing around with it, and I found that `kvIncrementalCache` is faster and probably better suited for our needs. I'm still looking at it, but for ease of use for now, just us KV.

packages/cli/src/lib/get-module-cli-path.ts
matthewvolk, line 1: Open to suggestions on how to unit test this... it's tricky because `import.meta.url` returns the _resolved_ path... so when `vitest` executes this file, `import.meta.url` is `/Users/.../catalyst/packages/cli/src/lib/get-module-cli-path.ts`. But when you actually execute this file from NPM, `import.meta.url` points to the bundle and will look something like `/Users/.../catalyst/packages/cli/dist/index.js`,

packages/cli/templates/open-next.config.ts
chanceaclark, line 1: This should probably be your whole file for now:
```ts
import { defineCloudflareConfig } from '@opennextjs/cloudflare';
import { purgeCache } from '@opennextjs/cloudflare/overrides/cache-purge/index';
import kvIncrementalCache from '@opennextjs/cloudflare/overrides/incremental-cache/kv-incremental-cache';
import doQueue from '@opennextjs/cloudflare/overrides/queue/do-queue';
import queueCache from '@opennextjs/cloudflare/overrides/queue/queue-cache';
import doShardedTagCache from '@opennextjs/cloudflare/overrides/tag-cache/do-sharded-tag-cache';

export default defineCloudflareConfig({
  incrementalCache: kvIncrementalCache,
  queue: queueCache(doQueue, {
    regionalCacheTtlSec: 5,
  }),
  routePreloadingBehavior: 'withWaitUntil',
  tagCache: doShardedTagCache({
    baseShardSize: 12,
    regionalCache: true,
    regionalCacheTtlSec: 5,
    shardReplication: {
      numberOfSoftReplicas: 4,
      numberOfHardReplicas: 2,
      regionalReplication: {
        defaultRegion: 'enam',
      },
    },
  }),
  enableCacheInterception: false,
  cachePurge: purgeCache({ type: 'durableObject' }),
});
```

jorgemoya (approved):

PR #2491, opened 2025-07-22
feat(cli): create deployment

jorgemoya:
## What/Why?

- Create a deployment with `project_uuid` and `upload_uuid`. 
- Reads project id from options, env vars, or .bigcommerce/project.json
- Updated logic to throw errors so we can test for errors

Still missing: poll after deploy

## Testing
Unit tests + locally
Screenshot 2025-07-22 at 2 47 36 PM

## Migration
N/A

### DIFF
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -166,16 +166,16 @@ export const deploy = new Command('deploy')
   )
   .addOption(
     new Option(
-      '--project-id <id>',
-      'BigCommerce headless project ID. Can be found via the BigCommerce API (GET /v3/headless/projects).',
-    ).env('BIGCOMMERCE_PROJECT_ID'),
+      '--project-uuid <uuid>',
+      'BigCommerce headless project UUID. Can be found via the BigCommerce API (GET /v3/headless/projects).',
+    ).env('BIGCOMMERCE_PROJECT_UUID'),
   )
   .option('--root-dir <rootDir>', 'Root directory to deploy from.', process.cwd())
   .action(async (opts) => {
     const config = new ProjectConfig(opts.rootDir);
 
     try {
-      const projectUuid = opts.projectId ?? config.get('projectId');
+      const projectUuid = opts.projectUuid ?? config.get('projectUuid');
 
       await generateBundleZip(opts.rootDir);
 
--- a/packages/cli/src/lib/project-config.ts
+++ b/packages/cli/src/lib/project-config.ts
@@ -2,7 +2,7 @@ import Conf from 'conf';
 import { join } from 'path';
 
 export interface ProjectConfigSchema {
-  projectId: string;
+  projectUuid: string;
 }
 
 export class ProjectConfig {
@@ -14,7 +14,7 @@ export class ProjectConfig {
       projectSuffix: '',
       configName: 'project',
       schema: {
-        projectId: { type: 'string', format: 'uuid' },
+        projectUuid: { type: 'string', format: 'uuid' },
       },
     });
   }
--- a/packages/cli/tests/commands/deploy.spec.ts
+++ b/packages/cli/tests/commands/deploy.spec.ts
@@ -15,7 +15,7 @@ let tmpDir: string;
 let cleanup: () => Promise<void>;
 let outputZip: string;
 
-const projectId = 'a23f5785-fd99-4a94-9fb3-945551623923';
+const projectUuid = 'a23f5785-fd99-4a94-9fb3-945551623923';
 const storeHash = 'test-store';
 const accessToken = 'test-token';
 const apiHost = 'api.bigcommerce.com';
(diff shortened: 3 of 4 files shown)


packages/cli/src/commands/deploy.ts
```diff
@@ -5,13 +5,47 @@ import { access, readdir, readFile } from 'node:fs/promises';
...
+  let parsed;
+
+  try {
+    parsed = ProjectJsonSchema.parse(JSON.parse(file));
+  } catch {
+    throw new Error('Failed to read project ID from .bigcommerce/project.json.');
+  }
+
+  return parsed.project_id;
+};
```
matthewvolk: 🍹 This is fine for now—in the future, I think we should use [`conf`](https://github.com/sindresorhus/conf) to read/write from `.bigcommerce/project.json`. Here's an example of using it in the link command POC: https://github.com/bigcommerce/catalyst/tree/cli-reference-link-command: https://github.com/bigcommerce/catalyst/blob/3816e88a8a44307071216ac6e72e96a2cf585c74/packages/cli/src/commands/link.ts#L7-L11

packages/cli/src/commands/deploy.ts
```diff
@@ -21,16 +55,14 @@ export const generateBundleZip = async (rootDir: string) => {
   try {
     await access(distDir);
   } catch {
-    consola.error(`Dist directory not found: ${distDir}`);
-    process.exit(1);
+    throw new Error(`Dist directory not found: ${distDir}`);
```
matthewvolk, line 32: Nice, thanks for switching to this! I like this pattern where we `throw` errors in individual functions (which makes them easier to test as units), let the errors bubble up to the action callback, and then display them with `consola.error` in the action callback's `catch`.

packages/cli/tests/commands/deploy.spec.ts
```diff
@@ -33,7 +42,37 @@ afterAll(async () => {
   await cleanup();
 });
 
-describe('bundle zip generation', () => {
+describe('loadProjectId', () => {
```
matthewvolk: 🍹 Might be able to ditch this set of tests if we switch to `conf`

matthewvolk (approved):
A couple of notes to consider for the future, but this looks good. Thanks for switching to the try/catch pattern to bubble up errors 🙏

packages/cli/src/commands/deploy.ts
jorgemoya: Ah, good idea. I will modify now so that we can streamline the process. Only wish I asked before I wrote the implementation and tests. 😆

packages/cli/src/commands/deploy.ts
```diff
@@ -136,18 +164,44 @@ export const deploy = new Command('deploy')
...
+  )
   .option('--root-dir <rootDir>', 'Root directory to deploy from.', process.cwd())
   .action(async (opts) => {
-    await generateBundleZip(opts.rootDir);
+    const config = new ProjectConfig(opts.rootDir);
+
+    try {
+      let projectUuid = opts.projectId;
+
+      projectUuid ??= config.getProjectId();
```
matthewvolk: 🍹 Any reason we don't just do `const projectUuid = opts.projectId ?? config.getProjectId();`? I only ask because I don't see us mutating `projectUuid` anywhere else

packages/cli/src/lib/project-config.ts
```diff
@@ -0,0 +1,37 @@
...
+    const projectId = this.config.get('projectId');
+
+    if (!projectId) {
+      throw new Error(
+        'No `projectId` found in .bigcommerce/project.json. Please add a valid `projectId` (UUID) to your bigcommerce configuration file.',
+      );
+    }
+
+    return projectId;
+  }
```
matthewvolk: 🍹I wonder if we'd be able to do something like `get(field: keyof ProjectConfigSchema)` for a future where we might want to store more keys in `project.json`. Not necessary right now though!

chanceaclark: >Please add a valid `projectId` (UUID) to your bigcommerce configuration file.

Can we just name the field `projectUuid` so we don't need to mention it's a uuid?

jorgemoya: I added generic get/set. I have also swapped the error to be more generic.

But now that you mentioned it, I'm fine with renaming `projectUuid`, should we use this throughout the CLI tho? Including command options, etc?

packages/cli/src/commands/deploy.ts
jorgemoya: Good callout.

packages/cli/src/lib/project-config.ts
matthewvolk: Good catch - I'm down to switch the options to `--project-uuid`, and anywhere else it pops up

PR #2498, opened 2025-07-24
feat(cli): add dev command

jorgemoya:
## What/Why?
This is just a wrapper for the `dev` command in the `core/package.json`. Allows changing port and root directory.

Independently of the framework chosen, it will just run `next dev` behind the scenes.

## Testing
Locally it runs.

Not sure how to unit test this, any ideas? 🤔 

## Migration
N/A


packages/cli/src/commands/dev.ts
```diff
@@ -0,0 +1,16 @@
...
+import consola from 'consola';
+import { execa } from 'execa';
+
+export const dev = new Command('dev')
+  .description('Start the development server')
+  .option('-p, --port <number>', 'Port to run the development server on', '3000')
+  .option('--root-dir <rootDir>', 'Root directory to run development server from.', process.cwd())
+  .action(async (opts) => {
+    try {
+      await execa('npm', ['run', 'dev'], { stdio: 'inherit', cwd: opts.rootDir });
```
chanceaclark: I think this command should look for the local `next` binary in node_modules to run. What do you think?

jorgemoya: If so, I assume we should also run the `generate` script before running dev?

chanceaclark: hmmm yeah

we probably want to abstract that generate script out at some point too 🤔

jorgemoya: Updated 👍

jordanarldt (approved):
Seems pretty straightforward. LGTM

PR #2499, opened 2025-07-25
feat: catalyst toast to window object

bc-victor:
## What/Why?
Adding the toast object to the window instance.
Also updating buyer portal url to microapps instance.

## Testing

https://github.com/user-attachments/assets/162d2b46-e875-442b-8b5b-201953f15c79

## Migration
not a breaking change

### DIFF
--- a/core/b2b/script-production.tsx
+++ b/core/b2b/script-production.tsx
@@ -35,7 +35,7 @@ export function ScriptProduction({ cartId, storeHash, channelId, token, environm
         data-channelid={channelId}
         data-environment={environment}
         data-storehash={storeHash}
-        src={`https://cdn.bundleb2b.net/b2b/${environment}/storefront/headless.js`}
+        src={'https://microapps.bigcommerce.com/b2b-buyer-portal/headless.js'}
         type="module"
       />
     </>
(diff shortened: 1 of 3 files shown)


core/b2b/script-production.tsx
```diff
@@ -35,7 +35,7 @@ export function ScriptProduction({ cartId, storeHash, channelId, token, environm
         data-channelid={channelId}
         data-environment={environment}
         data-storehash={storeHash}
-        src={`https://cdn.bundleb2b.net/b2b/${environment}/storefront/headless.js`}
+        src={'https://microapps.bigcommerce.com/b2b-buyer-portal/headless.js'}
```
bc-micah, line 38: Thanks for fixing this.

bc-micah (commented):
Thanks for including the testing video.

I would like to see more information why we are adding it to the window object (include a PR link for the Buyer Portal and the rationale behind it), and include some description of the exposed interface (include the link to the toaster component from vibes). I want anyone looking at this PR to be able to understand the rationale behind the changes.

I have context about why because I got linked here from the buyer portal PR, but catalyst team won't necessarily have that without those being explained.

bc-micah (approved):

PR #2501, opened 2025-07-28
feat(cli): add start command

jorgemoya:
## What/Why?
This is the first pass of the `start` command. A simple wrapper of the `next start` command. Next will be to conditionally run `next start` or `opennextjs-cloudflare preview` depending on selected `framework`.

## Testing
Locally runs.

## Migration
N/A

### DIFF
--- a/packages/cli/tests/index.spec.ts
+++ b/packages/cli/tests/index.spec.ts
@@ -18,5 +18,6 @@ describe('CLI program', () => {
     expect(commands).toContain('build');
     expect(commands).toContain('deploy');
     expect(commands).toContain('dev');
+    expect(commands).toContain('start');
   });
 });


packages/cli/src/commands/start.ts
```diff
@@ -0,0 +1,23 @@
...
+  .option('--root-dir <rootDir>', 'Root directory of your Catalyst project.', process.cwd())
+  .action(async (opts) => {
+    try {
+      const nextBin = join(opts.rootDir, 'node_modules', '.bin', 'next');
+
+      // @todo conditionally run `next start` or `opennextjs-cloudflare preview` based on the framework type
+      await execa({
+        stdio: 'inherit',
+        cwd: opts.rootDir,
+      })`${nextBin} start`;
```
jorgemoya: My only question is, should we build before we run this command?

PR #2508, opened 2025-08-01
feat: adding a safeguard to the B2B API host

bc-victor:
## What/Why?
B2B_API_HOST environment variable will break the B2B reference implementation when supplied with an extra trailing slash. This PR adds a validation step before making calls to the B2B API. If the env is dev, it will throw an error and the request won't be made. If the env is prod, we'll replace this value with the actual api host.

## Testing
https://github.com/user-attachments/assets/6c146ddf-04b7-4a21-a710-c2f5a507a19d

## Migration
NA

### DIFF
--- a/core/b2b/client.ts
+++ b/core/b2b/client.ts
@@ -14,6 +14,7 @@ const ENV = z
       B2B_API_TOKEN: z.string(),
       BIGCOMMERCE_CHANNEL_ID: z.string(),
       B2B_API_HOST: z.string().default('https://api-b2b.bigcommerce.com'),
+      NODE_ENV: z.enum(['development', 'production']).default('production'),
     }),
   })
   .transform(({ env }) => env);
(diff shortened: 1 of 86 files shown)


core/b2b/client.ts
```diff
@@ -29,9 +32,11 @@ const B2BTokenResponseSchema = z.object({
 });
 
 export async function loginWithB2B({ customerId, customerAccessToken }: LoginWithB2BParams) {
-  const { B2B_API_HOST, B2B_API_TOKEN, BIGCOMMERCE_CHANNEL_ID } = ENV.parse(process);
+  const { B2B_API_HOST, B2B_API_TOKEN, BIGCOMMERCE_CHANNEL_ID, NODE_ENV } = ENV.parse(process);
+
+  const apiHost = NODE_ENV === 'production' ? PROD_B2B_URL : B2B_API_HOST;
```
icatalina: It might be worth moving all this (including the warnings) into their own file:

```tsx
// ./getAPIHostname.ts

const DEFAULT_API_HOST = 'https://api.example.com';

const ENV = z.object({
  env: z.object({
    API_HOST: z.string().default(DEFAULT_API_HOST),
    NODE_ENV: z.enum(['development', 'production']).default('production'),
  })
})
.transform(({ env }) => env);

export const getAPIHostname = (): string => {}
```

Keeping the warn/use and everything in one place makes it easier to understand the reasons for it. Also, not a huge deal to parse `process` twice 🤷 

```tsx
const apiHost = getAPIHostname();
```

core/next.config.ts
```diff
@@ -1,12 +1,24 @@
 import bundleAnalyzer from '@next/bundle-analyzer';
 import type { NextConfig } from 'next';
 import createNextIntlPlugin from 'next-intl/plugin';
+import chalk from 'chalk';
```
icatalina: chalk seems unused

PR #2514, opened 2025-08-06
feat: introduce telemetry to catalyst cli

matthewvolk:
## What/Why?

Introduces telemetry logic to the new Catalyst CLI. This code is largely ported over from the `create-catalyst` CLI.
Changes include:
1. Installing the `@segment/analytics-node` package
2. A new `telemetry` command that allows developers to opt-in/out of data collection
3. Telemetry lifecycle hooks that track usage of each command run by a user and optionally the flags a command was run with (if they are whitelisted)
4. A telemetry service class that persists user data collection preferences and exposes methods to track and identify users and their activity
5. Adds a new compile-time environment variable to our `tsup.config.ts`

## Testing

Demo:

https://github.com/user-attachments/assets/1c390362-d5f8-42a1-984e-234b05923377

`identify` is also captured, though I'll omit the video to prevent exposing sensitive secrets. Here is a screenshot of running `link` with `--store-hash` and `--access-token`:

Screenshot 2025-08-06 at 12 06 27 PM

## Migration

N/A

### DIFF
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -7,6 +7,9 @@ import yoctoSpinner from 'yocto-spinner';
 import { z } from 'zod';
 
 import { ProjectConfig } from '../lib/project-config';
+import { Telemetry } from '../lib/telemetry';
+
+const telemetry = new Telemetry();
 
 const stepsEnum = z.enum([
   'initializing',
@@ -289,6 +292,8 @@ export const deploy = new Command('deploy')
   .action(async (opts) => {
     try {
       const config = new ProjectConfig(opts.rootDir);
+
+      await telemetry.identify(opts.storeHash);
 
       const projectUuid = opts.projectUuid ?? config.get('projectUuid');
 
--- a/packages/cli/src/commands/telemetry.ts
+++ b/packages/cli/src/commands/telemetry.ts
@@ -1,0 +1,43 @@
+import { Argument, Command, Option } from 'commander';
+import consola from 'consola';
+import { colorize } from 'consola/utils';
+
+import { Telemetry } from '../lib/telemetry';
+
+const telemetryService = new Telemetry();
+let isEnabled = telemetryService.isEnabled();
+
+export const telemetry = new Command('telemetry')
+  .addArgument(new Argument('[arg]').choices(['disable', 'enable', 'status']))
+  .addOption(new Option('--enable', `Enables CLI telemetry collection.`).conflicts('disable'))
+  .option('--disable', `Disables CLI telemetry collection.`)
+  .action((arg, options) => {
+    if (options.enable || arg === 'enable') {
+      telemetryService.setEnabled(true);
+      isEnabled = true;
+
+      consola.success('Success!\n');
+    } else if (options.disable || arg === 'disable') {
+      telemetryService.setEnabled(false);
+
+      if (isEnabled) {
+        consola.success('Your preference has been saved to .bigcommerce/project.json');
+      } else {
+        consola.info(`Catalyst CLI telemetry collection is already disabled.`);
+      }
+
+      isEnabled = false;
+    } else {
+      consola.info('Catalyst CLI Telemetry\n');
+    }
+
+    consola.info(
+      `Status: ${colorize('bold', isEnabled ? colorize('green', 'Enabled') : colorize('red', 'Disabled'))}`,
+    );
+
+    if (!isEnabled) {
+      consola.info(
+        `You have opted-out of Catalyst CLI telemetry. No data will be collected from your machine.`,
+      );
+    }
+  });
--- a/packages/cli/src/lib/telemetry.ts
+++ b/packages/cli/src/lib/telemetry.ts
@@ -1,0 +1,106 @@
+import { Analytics } from '@segment/analytics-node';
+import { randomBytes } from 'node:crypto';
+
+import PACKAGE_INFO from '../../package.json';
+
+import { ProjectConfig } from './project-config';
+
+export class Telemetry {
+  readonly sessionId: string;
+  readonly analytics: Analytics;
+
+  private projectConfig: ProjectConfig;
+  private CATALYST_TELEMETRY_DISABLED: string | undefined;
+
+  private readonly projectName = 'catalyst-cli';
+  private readonly projectVersion = PACKAGE_INFO.version;
+
+  constructor() {
+    this.CATALYST_TELEMETRY_DISABLED = process.env.CATALYST_TELEMETRY_DISABLED;
+
+    this.projectConfig = new ProjectConfig();
+
+    this.sessionId = randomBytes(32).toString('hex');
+    this.analytics = new Analytics({
+      writeKey: process.env.CLI_SEGMENT_WRITE_KEY ?? 'not-a-valid-segment-write-key',
+    });
+  }
+
+  async track(eventName: string, payload: Record<string, unknown>) {
+    if (!this.isEnabled()) {
+      return Promise.resolve(undefined);
+    }
+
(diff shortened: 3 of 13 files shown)


packages/cli/src/lib/telemetry.ts
```diff
@@ -0,0 +1,106 @@
...
+    } catch {
+      return false;
+    }
+  }
+
+  private getAnonymousId(): string {
+    let anonymousId;
+
+    try {
+      anonymousId = this.projectConfig.get('telemetry').anonymousId;
```
matthewvolk, line 89: @jorgemoya just an FYI, we might want to refactor the `ProjectConfig` class so that it does not wrap `get` and `set` methods. For example, the [`get`](https://github.com/sindresorhus/conf?tab=readme-ov-file#getkey-defaultvalue) method from Conf accepts a second parameter for a `defaultValue`, but we don't allow that. 

I'm thinking maybe we change the implementation to something like:
```ts
// lib/get-project-config.ts

export function getProjectConfig(rootDir = process.cwd()) {
  return new Conf<ProjectConfigSchema>({
    cwd: join(rootDir, '.bigcommerce'),
    projectSuffix: '',
    configName: 'project',
    schema: {
      projectUuid: { type: 'string', format: 'uuid' },
      framework: {
        type: 'string',
        enum: ['catalyst', 'nextjs'],
        default: 'catalyst',
      },
      telemetry: {
        type: 'object',
        properties: {
          enabled: { type: 'boolean' },
          anonymousId: { type: 'string' },
        },
      },
    },
  });
}
```

So that we just use the native `get` and `set` methods instead of trying to wrap them. I think the main job of the `ProjectConfig` class as we wrote it was to reduce the boilerplate required to instantiate `Conf` with `cwd`,  `projectSuffix`, `configName`, etc. options, and the function above does just that. 

Not for this PR, but maybe in the future.

jorgemoya: I agree. If it's not a heavy lift, we should do it sooner rather than later.

packages/cli/src/commands/telemetry.ts
```diff
@@ -0,0 +1,43 @@
...
+      if (isEnabled) {
+        consola.success('Your preference has been saved to .bigcommerce/project.json\n');
+      } else {
+        consola.info(`Catalyst CLI telemetry collection is already disabled.\n`);
+      }
+
+      isEnabled = false;
+    } else {
+      consola.info('Catalyst CLI Telemetry\n');
+    }
```
jorgemoya: Wonder if we can use a switch instead:

`const action = options.enable ? 'enable' : options.disable ? 'disable' : arg;`

packages/cli/src/commands/telemetry.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    consola.info(
+      `Status: ${colorize('bold', isEnabled ? colorize('green', 'Enabled\n') : colorize('red', 'Disabled\n'))}`,
+    );
+
+    if (!isEnabled) {
+      consola.info(
+        `You have opted-out of Catalyst CLI telemetry. No data will be collected from your machine.\n`,
```
jorgemoya: Do you have to add the new line `\n`?

packages/cli/tsup.config.ts
```diff
@@ -5,5 +5,8 @@ export default defineConfig((options: Options) => ({
   format: ['esm'],
   clean: !options.watch,
   sourcemap: true,
+  env: {
+    CLI_SEGMENT_WRITE_KEY: process.env.CLI_SEGMENT_WRITE_KEY ?? 'not-a-valid-segment-write-key',
```
jorgemoya, line 9: Why do we set this erroneous default?

packages/cli/src/lib/telemetry.ts
```diff
@@ -0,0 +1,106 @@
...
+  async identify(storeHash?: string) {
+    if (!this.isEnabled()) {
+      return Promise.resolve(undefined);
+    }
+
+    if (!storeHash) {
+      return Promise.resolve(undefined);
+    }
+
+    this.analytics.identify({
```
jorgemoya, line 59: What does `identify` do behind the scenes? Do you need to call this before using `track`?

packages/cli/src/commands/deploy.ts
```diff
@@ -175,6 +178,8 @@ export const deploy = new Command('deploy')
     const config = new ProjectConfig(opts.rootDir);
 
     try {
+      await telemetry.identify(opts.storeHash);
```
jorgemoya, line 296: We should add tests to make sure this is being called with the right storeHash. Same for any use case in the tests.

jorgemoya (commented):
Looking good, just got some comments.

packages/cli/src/program.ts
```diff
@@ -24,4 +26,7 @@ program
   .addCommand(start)
   .addCommand(build)
   .addCommand(deploy)
-  .addCommand(link);
+  .addCommand(link)
+  .addCommand(telemetry)
+  .hook('preAction', telemetryPreHook)
+  .hook('postAction', telemetryPostHook);
```
jorgemoya, line 32: Wonder we can test that the `preAction` is indeed calling `telemtryPreHook` and same for `postAction`, to be sure we don't remove these accidentally.

packages/cli/src/lib/telemetry.ts
chanceaclark: I don't think we should conflate project config with CLI telemetry config here. They should be two separate configurations. Reason being is that telemetry collection enable/disablement should be globally set vs. the project config is per project.

packages/cli/src/lib/telemetry.ts
chanceaclark: If we fire events all willy-nilly, it eats up our distinct user event allotment. Triggering an `identify` ties the event to a "user", thus reducing how much we consume.

packages/cli/src/program.ts
matthewvolk: Added!

packages/cli/src/commands/telemetry.ts
matthewvolk: It was mostly for readability, but I just tried with and without and there's hardly a difference. Got rid of the `\n`!

packages/cli/tsup.config.ts
matthewvolk: You're right that we can probably avoid it here, but it's ultimately because `writeKey` must not be `undefined` when instantiating `Analytics`

matthewvolk: It's because `env` is typed as:
```
env?: {
    [k: string]: string;
} | undefined
```

packages/cli/src/commands/telemetry.ts
matthewvolk: Definitely opportunities to refactor, but my preference is to keep the logic largely the same as `create-catalyst` for this PR just to ensure things work 1:1

packages/cli/src/lib/telemetry.ts
matthewvolk: This is definitely something we want to do in the future, especially if we come back to OAuth and want a global login state for the CLI; we'll need to add a global `conf` configuration for that in the future, which might be a good opportunity to tackle this.

chanceaclark: I would also think about our partners in Europe with GDPR laws, where they may be working on `N` different projects. Having to disable telemetry for each project would be pretty cumbersome.

FWIW, the existing telemetry already stores it in a global config: https://github.com/bigcommerce/catalyst/blob/canary/packages/create-catalyst/src/utils/telemetry/telemetry.ts#L34-L37

PR #2516, opened 2025-08-11
refactor: unwrap get and set instance methods of conf

matthewvolk:
## What/Why?

As discussed in https://github.com/bigcommerce/catalyst/pull/2514#discussion_r2257806601, this PR refactors the `ProjectConfig` class so that it does not wrap `get` and `set` methods. For example, the [`get`](https://github.com/sindresorhus/conf?tab=readme-ov-file#getkey-defaultvalue) method from Conf accepts a second parameter for a `defaultValue`, but we don't allow that. 

The original idea behind creating the `ProjectConfig` class was to reduce the boilerplate required to instantiate `Conf` with `cwd`,  `projectSuffix`, `configName`, etc. options; this refactor still accomplishes that goal, while making the `get` and `set` calls more flexible.

## Testing

Test suite passes

## Migration

N/A

### DIFF
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -6,7 +6,7 @@ import { join } from 'node:path';
 import yoctoSpinner from 'yocto-spinner';
 import { z } from 'zod';
 
-import { ProjectConfig } from '../lib/project-config';
+import { getProjectConfig } from '../lib/project-config';
 import { Telemetry } from '../lib/telemetry';
 
 const telemetry = new Telemetry();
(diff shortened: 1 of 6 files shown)


packages/cli/src/commands/deploy.ts
```diff
@@ -291,12 +291,18 @@ export const deploy = new Command('deploy')
...
 
       await telemetry.identify(opts.storeHash);
 
       const projectUuid = opts.projectUuid ?? config.get('projectUuid');
 
+      if (!projectUuid) {
+        throw new Error(
+          'Project UUID is required. Please run either `bigcommerce link` or this command again with --project-uuid <uuid>.',
+        );
+      }
```
jorgemoya, line 304: We should make it option mandatory, but I can fix this in a different PR if that works better.

jorgemoya (approved):

packages/cli/src/commands/deploy.ts
matthewvolk: Do you mean make the `--project-uuid` option mandatory? I was actually thinking the same, but if the option is mandatory, there is no opportunity to read it from `Conf`:
```
const projectUuid = opts.projectUuid ?? config.get('projectUuid');
```

In other words, if the option is mandatory, the left-side of the expression will never be nullish, right?

jorgemoya: Ah, that is correct. This is the way to implement it.

PR #2517, opened 2025-08-11
refactor: run build directly on catalyst core application

matthewvolk:
## What/Why?

Previously, `build` was running inside of a temporary directory in an effort to abstract away Cloudflare-related configuration files (`wrangler.jsonc`, `open-next.config.ts`) and dependencies (`@opennextjs/cloudflare`). The only artifact that we copied from this temporary directory back into the original Catalyst project was a single `worker.js` file.

Unfortunately, we discovered that the `opennextjs-cloudflare preview` command needs access to some of the intermediary files that were being deleted when the temporary folder was deleted at the end of `build`. So, we had to change course. We are now installing Cloudflare-related dependencies in the CLI (instead of core) and using an `esbuild` plugin to resolve imports pointing to these dependencies. To make this work, we had to resort to patching `@opennextjs/cloudflare`, but have started a conversation with Cloudflare to see if we can more reliable support.

## Testing

Updated test suite

## Migration

N/A

JIRA: [CATALYST-1421](https://bigcommercecloud.atlassian.net/browse/CATALYST-1421)

[CATALYST-1421]: https://bigcommercecloud.atlassian.net/browse/CATALYST-1421?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/cli/src/commands/build/opennext-build.ts
matthewvolk, line 1: @jorgemoya This is the file still in need of tests

packages/cli/package.json
```diff
@@ -22,11 +24,13 @@
     "node": "^20.0.0 || ^22.0.0"
   },
   "dependencies": {
+    "@opennextjs/cloudflare": "1.6.5",
```
matthewvolk, line 27: Dependency pinned due to patch

packages/cli/src/commands/build/action.ts
```diff
@@ -0,0 +1,116 @@
...
+  const exec = createExec(options);
+
+  const framework = resolveFramework(options, getProjectConfig, logger);
+
+  logger.info(`Building application with framework: ${framework}`);
+
+  if (framework === 'nextjs') {
+    logger.start('Generating gql.tada types...');
+    await generateGqlTadaTypes(exec);
+    logger.success('gql.tada types generated successfully');
```
jorgemoya: I had discussed this with @chanceaclark before, but we don't need to generate gql.tada here, I think we can keep that separate in `core/package.json`.

core/package.json
```diff
@@ -6,13 +6,14 @@
   "scripts": {
     "dev": "npm run generate && next dev",
     "generate": "dotenv -e .env.local -- node ./scripts/generate.cjs",
-    "build": "npm run generate && next build",
+    "build": "catalyst build",
```
jorgemoya:
```suggestion
    "build": "npm run generate && catalyst build",
```

packages/cli/src/commands/build/action.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    if (options.framework) {
+      logger.debug(`Using framework from --framework option: ${options.framework}`);
+    } else {
+      logger.debug(`Using framework from project configuration: ${config.get('framework')}`);
+    }
+  }
+
+  return options.framework ?? config.get('framework');
+}
```
jorgemoya: Can we add this to lib?

packages/cli/src/commands/build/action.ts
```diff
@@ -0,0 +1,116 @@
...
+
+export async function buildNextjs(
+  exec: Exec,
+  resolveBinary: (bin: string) => string = resolveLocalBinaryPath,
+) {
+  const nextBin = resolveBinary('next');
+
+  await exec(nextBin, ['build']);
+}
```
jorgemoya, line 44: Why we are we abstracting this?

packages/cli/src/commands/build/action.ts
```diff
@@ -0,0 +1,116 @@
...
+    } else {
+      logger.debug(`Using framework from project configuration: ${config.get('framework')}`);
+    }
+  }
+
+  return options.framework ?? config.get('framework');
+}
+
+export function createLogger(options: BuildCommandOptions) {
+  return createConsola({ level: options.verbose ? LogLevels.verbose : LogLevels.info });
```
jorgemoya, line 20: TIL `level`, but how is this used? How does consola know what is a verbose log?

packages/cli/src/commands/build/opennext-build.ts
```diff
@@ -0,0 +1,227 @@
...
+              defaultRegion: 'enam',
+            },
+          },
+        }),
+        enableCacheInterception: false,
+        cachePurge: purgeCache({ type: 'durableObject' }),
+      }),
+    };
+
+    const wranglerConfig = {
```
jorgemoya, line 71: Are we moving away from templates then? For `preview`, I need to create a copy of `wranger.jsonc` too and want to make sure we have one source of truth.

packages/cli/src/commands/build/opennext-build.ts
```diff
@@ -0,0 +1,227 @@
...
+      await rm(join(process.cwd(), '.bigcommerce', '.open-next'), { recursive: true });
+    } catch {
+      // ignore if does not exist
+    }
+
+    await cp(join(process.cwd(), '.open-next'), join(process.cwd(), '.bigcommerce', '.open-next'), {
+      recursive: true,
+      force: true,
+    });
+    await rm(join(process.cwd(), '.open-next'), { recursive: true });
```
jorgemoya: I guess we can remove this.

packages/cli/src/commands/build/action.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+export function createExec(options: BuildCommandOptions, exec: Exec = _execa) {
```
jorgemoya, line 23: Not sure if these abstraction are worth it, can you provide your reasoning? Maybe they need to be lib utilities?

packages/cli/src/commands/build/action.ts
matthewvolk: Interesting, I can remove it

packages/cli/src/commands/build/action.ts
matthewvolk: This abstraction was solely for purposes of unit testing

packages/cli/src/commands/build/action.ts
matthewvolk: Basically [this](https://github.com/unjs/consola/blob/512d570b3f99a27b6766a885cae2730aa3dc0ed4/src/constants.ts#L16-L38) - the number corresponds to the minimum log level that Consola will log, so if `level === LogLevels.info`, then `consola.debug` will not log anything but `consola.info` will

packages/cli/src/commands/build/opennext-build.ts
matthewvolk: I'm not opinionated on this, and I agree that my only requirement is that we have a single source of truth. If we want, I can read `wrangler.jsonc` from `fs`, and pass that to the `build` function from OpenNext... keep in mind, we also need to modify certain fields in `wrangler.jsonc`, so maybe we make a lib function to handle reading/writing/etc. the file. The file can have the config as an object, and then methods to write to `wrangler.jsonc` using `fs`, both using the same object as a source of truth?

packages/cli/src/commands/build/action.ts
matthewvolk: I'm cool with lib utils - I'm trying to follow some of the principles in [this](https://www.epicweb.dev/the-golden-rule-of-assertions) article, so this abstraction does a couple things:
1. Helps asserts behavior (“verbose uses inherit”), not which process lib (execa, child_process, etc.) we called.
2. It also encodes the `--verbose` policy on `exec` in a single place. Other commands can use `createExec` and get `--verbose` logging options for free. No drift across different call sites.

If you still think it's too early of an abstraction, let's discuss; I'm not set in stone by any means, and open to inlining it if you think it's better

packages/cli/src/commands/build/opennext-build.ts
jorgemoya: It's fine either way by me, just feel like we should delete the other templates if we're not using them?

packages/cli/src/lib/resolve-framework.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+type Logger = Pick<ConsolaInstance, 'debug' | 'start' | 'success' | 'fail' | 'info' | 'error'>;
+type Config = () => {
+  get: (key: 'framework') => 'catalyst' | 'nextjs';
+  path: string;
+};
+
+export function resolveFramework(options: Options, getConfig: Config, logger: Logger) {
+  const config = getConfig();
```
jorgemoya, line 15: Missing `rootDir` being passed to this function.

Which brings up another talking point, do we care about the `rootDir` option for commands? I know some of them have it, others don't. For my tests, adding this `rootDir` option was valuable.

packages/cli/src/commands/build/action.ts
```diff
@@ -0,0 +1,83 @@
+import { createConsola, LogLevels } from 'consola';
+import { execa as _execa, Options as ExecaOptions } from 'execa';
+import { existsSync } from 'node:fs';
+import { join } from 'node:path';
+
+import { getProjectConfig } from '../../lib/project-config';
+import { resolveFramework } from '../../lib/resolve-framework';
+
+import { opennextBuild } from './opennext-build';
```
jorgemoya, line 9: `openNextBuild` or `opennextBuild`? 🤔

packages/cli/src/commands/build/action.ts
jorgemoya: Been thinking about this all day. 

1. How are you asserting the behavior? Just curious, seems to be it would look the same with or without the abstraction. I understand we don't want to test implementation details, only outputs, but how do we test the output of the command?
2. I agree, but thinking about `start` as an example, we wouldn't really want to hide it away, thinking this is only required here for the moment?

packages/cli/src/commands/build/opennext-build.ts
```diff
@@ -0,0 +1,216 @@
...
+    b.onResolve({ filter: /^node:/ }, () => ({ external: true }));
+  },
+};
+
+export async function opennextBuild() {
+  const [tmpdir, cleanup] = await mkTempDir('catalyst-build-');
+
+  try {
+    const openNextConfig = {
+      buildCommand: 'dotenv -e .env.local -- node ./scripts/generate.cjs && next build',
```
jorgemoya:
```suggestion
      buildCommand: 'next build',
```

packages/cli/src/commands/build/opennext-build.ts
```diff
@@ -0,0 +1,216 @@
...
+    );
+
+    await execa(
+      'pnpm',
+      [
+        'dlx',
+        `wrangler@${WRANGLER_VERSION}`,
+        'deploy',
+        '--config',
+        join(process.cwd(), '.bigcommerce', 'wrangler.jsonc'),
```
jorgemoya, line 189: Is `process.cwd()` necessary since we define the `cwd` below?

matthewvolk:
Fixed by #2527

PR #2518, opened 2025-08-12
feat(cli): run opennext-cloudflare preview for catalyst framework

jorgemoya:
Builds on top of #2517 

## What/Why?
Runs either `nextjs start` or `opennextjs-cloudflare preview` when running `catalyst start`, depending on the selected framework.

## Testing
Locally it runs both versions + unit tests.

```
❯ pnpm run start

> @bigcommerce/catalyst-core@1.1.0 start /Users/jorge.moya/dev/catalyst/core
> catalyst start

◢ @bigcommerce/catalyst v0.1.0                                                                    3:24:47 PM


┌───────────────────────────────┐
│ OpenNext — Cloudflare preview │
└───────────────────────────────┘

Monorepo detected at /Users/jorge.moya/dev/catalyst

Populating KV incremental cache...
▲ [WARNING]                             You have defined bindings to the following internal Durable Objects:

                                - {"name":"NEXT_CACHE_DO_QUEUE","class_name":"DOQueueHandler"}
  - {"name":"NEXT_TAG_CACHE_DO_SHARDED","class_name":"DOShardedTagCache"}
  - {"name":"NEXT_CACHE_DO_PURGE","class_name":"BucketCachePurge"}
                                These will not work in local development, but they should work in production.

                                If you want to develop these locally, you can define your DO in a separate Worker, with a
  separate configuration file.
                                For detailed instructions, refer to the Durable Objects section here:
  https://developers.cloudflare.com/workers/wrangler/api#supported-bindings


workerd/server/server.c++:1874: warning: A DurableObjectNamespace in the config referenced the class "DOQueueHandler", but no such Durable Object class is exported from the worker. Please make sure the class name matches, it is exported, and the class extends 'DurableObject'. Attempts to call to this Durable Object class will fail at runtime, but historically this was not a startup-time error. Future versions of workerd may make this a startup-time error.
workerd/server/server.c++:1874: warning: A DurableObjectNamespace in the config referenced the class "DOShardedTagCache", but no such Durable Object class is exported from the worker. Please make sure the class name matches, it is exported, and the class extends 'DurableObject'. Attempts to call to this Durable Object class will fail at runtime, but historically this was not a startup-time error. Future versions of workerd may make this a startup-time error.
workerd/server/server.c++:1874: warning: A DurableObjectNamespace in the config referenced the class "BucketCachePurge", but no such Durable Object class is exported from the worker. Please make sure the class name matches, it is exported, and the class extends 'DurableObject'. Attempts to call to this Durable Object class will fail at runtime, but historically this was not a startup-time error. Future versions of workerd may make this a startup-time error.
Inserting 54 assets to KV in chunks of 25
100%|█████████████████████████████████████████████| 3/3 [00:01:<00:00:, 1.79it/s]
Successfully populated cache with 54 assets
Tag cache does not need populating

 ⛅️ wrangler 4.29.0
───────────────────
Your Worker has access to the following bindings:
Binding                                                      Resource            Mode
env.NEXT_CACHE_DO_QUEUE (DOQueueHandler)                     Durable Object      local
env.NEXT_TAG_CACHE_DO_SHARDED (DOShardedTagCache)            Durable Object      local
env.NEXT_CACHE_DO_PURGE (BucketCachePurge)                   Durable Object      local
env.NEXT_INC_CACHE_KV (placeholder-kv-id)                    KV Namespace        local
env.WORKER_SELF_REFERENCE (store-hash-project-uuid)          Worker              local [not connected]
env.ASSETS                                                   Assets              local


Service bindings, Durable Object bindings, and Tail consumers connect to other `wrangler dev` processes running locally, with their connection status indicated by [connected] or [not connected]. For more details, refer to https://developers.cloudflare.com/workers/runtime-apis/bindings/service-bindings/#local-development

╭──────────────────────────────────────────────────────────────────────╮
│  [b] open a browser [d] open devtools [c] clear console [x] to exit  │
╰──────────────────────────────────────────────────────────────────────╯
⎔ Starting local server...
[wrangler:info] Ready on http://localhost:3000
```

## Migration
N/A

### DIFF
--- a/packages/cli/src/lib/wrangler-config.ts
+++ b/packages/cli/src/lib/wrangler-config.ts
@@ -1,7 +1,7 @@
 export function getWranglerConfig(projectUuid: string, kvNamespaceId: string) {
   return {
     $schema: 'node_modules/wrangler/config-schema.json',
-    main: '.open-next/worker.js',
+    main: '../.open-next/worker.js',
     name: `project-${projectUuid}`,
     compatibility_date: '2025-07-15',
     compatibility_flags: ['nodejs_compat', 'global_fetch_strictly_public'],
@@ -15,7 +15,7 @@ export function getWranglerConfig(projectUuid: string, kvNamespaceId: string) {
       },
     },
     assets: {
-      directory: '.open-next/assets',
+      directory: '../.open-next/assets',
       binding: 'ASSETS',
     },
     services: [
--- a/packages/cli/tests/commands/start.spec.ts
+++ b/packages/cli/tests/commands/start.spec.ts
@@ -1,6 +1,7 @@
 import { Command } from 'commander';
+import consola from 'consola';
 import { execa } from 'execa';
-import { expect, test, vi } from 'vitest';
+import { afterEach, beforeAll, beforeEach, expect, test, vi } from 'vitest';
 
 import { start } from '../../src/commands/start';
 import { program } from '../../src/program';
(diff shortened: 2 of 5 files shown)


packages/cli/src/lib/wrangler-config.ts
```diff
@@ -1,7 +1,7 @@
 export function getWranglerConfig(projectUuid: string, kvNamespaceId: string) {
   return {
     $schema: 'node_modules/wrangler/config-schema.json',
-    main: '.open-next/worker.js',
+    main: '../.open-next/worker.js',
```
matthewvolk, line 4: 🍹 Def doesn't have to be figured out to merge this PR, but I think a really useful automated test in the future would test that this path always correctly points to where `.open-next` is.

packages/cli/tests/commands/start.spec.ts
```diff
@@ -14,18 +15,87 @@ vi.mock('execa', () => ({
...
 test('properly configured Command instance', () => {
   expect(start).toBeInstanceOf(Command);
   expect(start.name()).toBe('start');
   expect(start.description()).toBe('Start your Catalyst storefront in optimized production mode.');
+  expect(start.options).toEqual(
+    expect.arrayContaining([
+      expect.objectContaining({
+        flags: '--framework <framework>',
+        argChoices: ['catalyst', 'nextjs'],
+      }),
```
matthewvolk, line 51: 🍹 Again not for this PR, we should agree on a pattern. For example, I used `long` instead of `flags` and `argChoices`. 

Separately, I still wonder if these tests are worth writing, doesn't Commander kind of guarantee that options are attached to the Command as part of their own internal tests? (https://github.com/tj/commander.js/blob/master/tests/command.parseOptions.test.js)

matthewvolk (approved):

PR #2521, opened 2025-08-13
Preconnect to checkout origin on cart page

bookernath:
## What/Why?
On the cart page, add a preconnect tag for the checkout URL, to make sure DNS and TLS connection is already done by the time the user needs to redirect to checkout, thus improving checkout load performance for the redirected checkout.

https://nextjs.org/docs/app/api-reference/functions/generate-metadata#resource-hints

## Testing
image

### DIFF
--- a/core/app/[locale]/(default)/cart/_components/checkout-preconnect.tsx
+++ b/core/app/[locale]/(default)/cart/_components/checkout-preconnect.tsx
@@ -1,0 +1,9 @@
+'use client';
+
+import { preconnect } from 'react-dom';
+
+export function CheckoutPreconnect({ url }: { url: string }) {
+  preconnect(url);
+
+  return null;
+}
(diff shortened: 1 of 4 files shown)


core/app/[locale]/(default)/cart/_components/checkout-preconnect.tsx
Comment (matthewvolk) [inline core/app/[locale]/(default)/cart/_components/checkout-preconnect.tsx:1]: 🍹 Would it be beneficial to use this anywhere else? Wondering if we generalize it `CheckoutPreconnect` → `PreconnectResource` and add an optional param for `options: PreconnectOptions`

Comment (bookernath) [reply core/app/[locale]/(default)/cart/_components/checkout-preconnect.tsx:1]: I thought about that, and I don't *think* so - Next.js already has great defaults for this in the rest of the application, our situation here is special because we are redirecting externally. The other place where this might be useful is in areas where you're likely to proceed directly to checkout, but this is the only place in the Catalyst codebase where this is relevant right now.

PR #2533, opened 2025-08-19
fix: do not add trailing commas to jsonc files

matthewvolk:
## What/Why?

The `jsonc` spec does not allow trailing commas: https://jsonc.org/#trailing-commas

## Testing

N/A

## Migration

N/A


matthewvolk:
Closing, I think it's actually better to generate this file at runtime to make it easier to inject dynamic values for `name`, etc.

PR #2551, opened 2025-08-24
chore(common): Fixes maintenance page with issue

jkanive:
## What/Why?
Fixes an issue with the maintenance page where the content was not visible because of the width.

### Before
https://github.com/user-attachments/assets/da4d187f-20b3-4550-9997-fb9f325590b2

### After

https://github.com/user-attachments/assets/bf9e27c1-c447-47d3-ba6a-97ce81fa8276


jorgemoya:
@jkanive Is this good to merge?

PR #2552, opened 2025-08-25
fix(core): Implement maintenance mode using app router route handler

jkanive:
## What/Why?
There is a known issue with `NextResponse.rewrite` https://github.com/vercel/next.js/issues/50155
### Case: 
Store is InActive (maintenance mode)
### Expected Behavior: 
We get a 503 status code and the page displays `We are down for maintenance` message as defined by `Maintenance` component.
### Actual Behavior:
- Local Dev Mode:
Actual behavior matches the Expected behavior.
- Production:
We get 503 status code but it fails to display Maintenance component, instead display general `SERVICE_UNAVAILABLE` message screen
Screenshot 2025-08-24 at 6 53 13 PM

### Local Dev Mode

Screenshot 2025-08-24 at 6 55 45 PM

### Production
To Be Added...


core/app/[locale]/maintenance/route.ts
Comment (chanceaclark) [inline core/app/[locale]/maintenance/route.ts:1]: I don't think this is the right fix to this issue 🤔 That's a lot of html embedded into the route and seems like a _maintenance_ nightmare for us. We should look into more why the rewrite isn't working correctly.

PR #2558, opened 2025-08-26
feat(cli): add env option to deploy command

jorgemoya:
## What/Why?
Parses `env.options` to send to create deployment.

Will also read from env variables if the corresponding key was not passed in `env.options`.

Variables we're looking for:
`BIGCOMMERCE_STORE_HASH`
`BIGCOMMERCE_STOREFRONT_TOKEN`
`BIGCOMMERCE_CHANNEL_ID`

## Testing
Locally and unit tests.

## Migration
N/A


matthewvolk (approved):
Re-approving for variadic option fix

PR #2564, opened 2025-08-29
feat(core): add visual queues  when cart update is pending

jorgemoya:
## What/Why?
Add visual queues when the cart state is being updated in the Cart page. Will also warn about pending state when trying to navigate away from page.

https://github.com/user-attachments/assets/a7867ed9-1f96-47a7-8e59-cb1a1e1730fe

## Testing
Locally.

## Migration
1. Update `/core/vibes/soul/sections/cart/client.tsx` to include latest changes:
  - Use `isLineItemActionPending` to track when we need to disable checkout button and add a loading state.
  - Add skeletons to checkout summary fields that will update when the pending state is complete.
  - Add side effects to handle when a user `beforeunload` and when user tries to navigate using a link.
  - Add prop to `lineItemActionPendingLabel` to be able to pass in a translatable label to the window alert.
2. Add label to dictionary of choice.


core/vibes/soul/sections/cart/client.tsx
```diff
@@ -190,6 +194,86 @@ export function CartClient<LineItem extends CartLineItem>({
...
+
+          if (!shouldNavigate) {
+            event.preventDefault();
+            event.stopPropagation();
+          }
+        }
+      }
+    };
+
+    if (isLineItemActionPending) {
```
matthewvolk, line 266: 🍹Is this needed if the handlers also check for it?

matthewvolk (approved):
Everything works in the preview build as I'd expect it to. I can't think of any off the top of my head, but if there are any other interactions throughout the storefront that need this `beforeunload` behavior, maybe we consider turning this into a library to extract away some of the logic out of the primary cart logic.

core/vibes/soul/sections/cart/client.tsx
jorgemoya: Sorry, I missed this comment. I think so, we only want to add the event listeners if the action is pending. Now, this listener will only cleanup on the unmount, so we still need to check if the action is pending inside the callback.

PR #2568, opened 2025-09-02
feat: support target prop in Footer links

hunterbecton:
## What/Why?
- adds support for `target` in `Footer` links

## Testing

https://github.com/user-attachments/assets/896692bf-60e4-4dbf-ad08-3f099bdcc367


matthewvolk (approved):
Looks good, when `target` is `undefined` it will simply fall back to not opening a new tab.

PR #2569, opened 2025-09-02
chore: fix performance.now to 2 decimal places (#2565)

mxplantae:


matthewvolk (changes requested):
@mxplantae is there a chance this PR was opened on accident?

PR #2574, opened 2025-09-04
chore(core): removes unused dependencies

jorgemoya:
## What/Why?
Removes unused dependencies.

## Testing
N/A

## Migration
Remove `sharp` and `react-google-recaptcha` from core.


core/package.json
```diff
@@ -58,11 +58,9 @@
     "react": "^19.1.0",
     "react-day-picker": "^9.7.0",
     "react-dom": "^19.1.0",
-    "react-google-recaptcha": "^3.1.0",
```
chanceaclark, line 61: 🍹 Non-blocking, but should we also remove all the recaptcha related code too then? Haven't really seen people complain that recaptcha is not working. We can also add it back later if we need to.

jorgemoya: Good call, will add 👍

PR #2579, opened 2025-09-04
chore(core): removes unused export types from UI components

jorgemoya:
## What/Why?
Remove unused export types from UI components.

## Testing
N/A

## Migration
Remove `export` declaration from types in UI components.


chanceaclark (commented):
A part of me thinks that we should keep these as even though they are not being used by us, consumers of these components might.

jorgemoya:
Closing since they may be used in other branches.

PR #2584, opened 2025-09-05
feat(cli): use custom opennext config path in new opennextjs cli release

matthewvolk:
## What/Why?

Cloudflare's OpenNext team recently released `@opennextjs/cloudflare` 1.8.0 which includes a feature request we discussed [here](https://github.com/opennextjs/opennextjs-cloudflare/pull/853#issuecomment-3205095480). This PR makes use of that new flag... thanks Cloudflare!

## Testing

✅ CI Passes

Manual Test (local): 
Screenshot 2025-09-05 at 9 31 43 AM

## Migration

N/A

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -49,6 +49,6 @@
     "vitest": "^3.2.4"
   },
   "peerDependencies": {
-    "@opennextjs/cloudflare": "^1.6.5"
+    "@opennextjs/cloudflare": "^1.8.0"
   }
 }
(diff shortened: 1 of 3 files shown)


matthewvolk:
@jorgemoya does this allow us to move the entire `.open-next` folder into `.bigcommerce` as well? 🤔 I can't remember if these were tied together. If so, we should probably write tests that test across both `build` and other commands that need this folder.

jorgemoya:
@matthewvolk Problem I had was running `preview`: it would look for `.open-next` where called. I haven't looked to see if that is still the case in the new version.

PR #2605, opened 2025-09-11
feat(cli): general UX improvements

chanceaclark:
## What/Why?
- Provides a better user experience by utilizing `dotenv` to pull the users `.env` or `.env.local` file into `process.env` so they don't need to pass any arguments if their repo is configured correctly (going through the create-catalyst CLI to inject the env vars).
- Creates a logger instance to provide better logging support in dev vs. prod. Log level is `3` by default in `consola` which logs things like `Fetching projects`. We might not want that in production as it seems like an implementation detail. Setting it to log level `2` with the built CLI will provide just enough information needed.

## Testing
See the docs.

## Migration
N/A

### DIFF
--- a/packages/cli/src/lib/logger.ts
+++ b/packages/cli/src/lib/logger.ts
@@ -1,0 +1,5 @@
+import { createConsola } from 'consola';
+
+export const consola = createConsola({
+  level: process.env.CONSOLA_LEVEL ? parseInt(process.env.CONSOLA_LEVEL, 10) : 3,
+});
--- a/packages/cli/src/program.ts
+++ b/packages/cli/src/program.ts
@@ -18,8 +18,13 @@ import { consola } from './lib/logger';
 export const program = new Command();
 
 config({
-  // Should we add any other .env paths here?
-  path: [resolve(process.cwd(), '.env'), resolve(process.cwd(), '.env.local')],
+  path: [
+    resolve(process.cwd(), '.env'),
+    resolve(process.cwd(), '.env.local'),
+    // Assumes the parent directory is the monorepo root:
+    resolve(process.cwd(), '..', '.env'),
+    resolve(process.cwd(), '..', '.env.local'),
+  ],
 });
 
 consola.log(colorize('cyanBright', `◢ ${PACKAGE_INFO.name} v${PACKAGE_INFO.version}\n`));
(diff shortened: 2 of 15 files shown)


packages/cli/src/program.ts
```diff
@@ -12,9 +13,15 @@ import { start } from './commands/start';
...
 import { telemetryPostHook, telemetryPreHook } from './hooks/telemetry';
+import { consola } from './lib/logger';
 
 export const program = new Command();
 
+config({
+  // Should we add any other .env paths here?
+  path: [resolve(process.cwd(), '.env'), resolve(process.cwd(), '.env.local')],
+});
+
```
matthewvolk: 🍹 Can we traverse upwards to find files at the root? I'm sure there's a reason that's not a great idea, but I think there's a large number of folks who manage their `.env` at project root... maybe the solution is just documentation to `ln` it to `core`

packages/cli/src/lib/logger.ts
```diff
@@ -0,0 +1,5 @@
+import { createConsola } from 'consola';
+
+export const consola = createConsola({
+  level: process.env.CONSOLA_LEVEL ? parseInt(process.env.CONSOLA_LEVEL, 10) : 3,
```
matthewvolk, line 4: 🍹 Do we want to expose this to developers who would like to enable verbose logs on their machine? That would probably be helpful for debugging. If so, should we decouple the env var name from the implementation detail (`consola`)? I know in their docs, consola [specifically mentions](https://github.com/unjs/consola/tree/main?tab=readme-ov-file#log-level) naming it `CONSOLA_LEVEL`, but I think since you're setting it via the `level` option, I don't think it needs to be named `CONSOLA_LEVEL`. 

TL;DR: Should we name it `process.env.LOG_LEVEL`?

packages/cli/src/program.ts
chanceaclark: I'll add a manual path for now. I would say the symlink way is a little advanced. Good callout though.

packages/cli/src/lib/logger.ts
chanceaclark: tldr, it doesn't matter

 The build will take it and inject it into this constructor, but you can also pass it in the CLI command and get detailed logging:
Screenshot 2025-09-15 at 11 58 53

matthewvolk (approved):

PR #2611, opened 2025-09-17
chore: reorganize cli package for future extension 

chanceaclark:
## What/Why?
Rearrange the cli files in order to extend the `@bigcommerce/catalyst` package.

## Testing
See CI

Tested the built CLI locally:
Screenshot 2025-09-17 at 14 16 51

## Migration
This package shouldn't be used. If pulling in changes it should just be rebases on the incoming changes.


packages/catalyst/templates/open-next.config.ts
matthewvolk, line 1: 🍹 Is there a reason we don't colocate `templates` inside `src/cli`?

packages/catalyst/.eslintrc.cjs
matthewvolk, line 1: 🍹 Do we want to take the opportunity to scope some of the `rules` like `'no-console': 'off'` to just the `src/cli` directory?

matthewvolk (approved):
LGTM!

packages/catalyst/templates/open-next.config.ts
chanceaclark: We need to export that folder when we publish the package so leaving it top level is easier. Also, I didn't want to touch the logic part of the code to account for moving it.

matthewvolk: That's fair, I'm cool with that!

packages/catalyst/.eslintrc.cjs
chanceaclark: Done, added a new commit to clean it up!

matthewvolk: Nice, good catch on the `no-named-as-default`. Thank you!

PR #2619, opened 2025-09-22
Add stock level message to product page based on inventory settings

Tharaae:
## What/Why?
As part of backorders project, we want to give the merchant an option to display a backorders prompt to the shopper on the product details page for products that have backordering available. This need to be built on the stock level message feature (as in cornerstone) that is not implemented as a catalyst default feature.

This PR is to implement the stock level message on the product details page based on the store inventory settings.

## Testing
Screenshots

**When the product is in stock and "Show stock levels" is selected:**
Screenshot 2025-09-22 at 12 05 04 pm
Screenshot 2025-09-22 at 12 10 39 pm
Screenshot 2025-09-22 at 12 10 01 pm

**When the product is in stock and "Don's show stock levels" is selected:**
Screenshot 2025-09-22 at 12 05 04 pm
Screenshot 2025-09-22 at 12 12 41 pm
Screenshot 2025-09-22 at 12 13 24 pm

**When the product is in stock but below or equal warning level and "Only show stock levels when stock is low" is selected:**
Screenshot 2025-09-22 at 10 28 01 am
Screenshot 2025-09-22 at 10 29 04 am
Screenshot 2025-09-22 at 10 29 43 am

**When the product is in stock and above warning level and "Only show stock levels when stock is low" is selected:**
Screenshot 2025-09-22 at 12 05 04 pm
Screenshot 2025-09-22 at 10 29 04 am
Screenshot 2025-09-22 at 12 04 38 pm

**When the product is not in stock and "Out of stock message" is set:**
Screenshot 2025-09-22 at 12 16 57 pm
Screenshot 2025-09-22 at 12 14 49 pm
Screenshot 2025-09-22 at 12 16 35 pm

## Migration
None


core/app/[locale]/(default)/product/[slug]/page.tsx
```diff
@@ -196,6 +197,36 @@ export default async function Product({ params, searchParams }: Props) {
...
+    const { showOutOfStockMessage, stockLevelDisplay, defaultOutOfStockMessage } = inventorySetting;
+
+    const product = await streamableProduct;
+
+    if (!product.inventory.isInStock) {
+      return showOutOfStockMessage ? defaultOutOfStockMessage : null;
+    }
+
+    if (
+      stockLevelDisplay === 'DONT_SHOW' ||
```
Comment (animesh1987) [inline core/app/[locale]/(default)/product/[slug]/page.tsx]: Can we type `'DONT_SHOW'` and other values in an enum?

core/app/[locale]/(default)/product/[slug]/page.tsx
```diff
@@ -196,6 +197,36 @@ export default async function Product({ params, searchParams }: Props) {
...
+      (stockLevelDisplay === 'SHOW_WHEN_LOW' &&
+        (!product.inventory.aggregated?.warningLevel ||
+          product.inventory.aggregated.availableToSell <=
+            product.inventory.aggregated.warningLevel))
+    ) {
+      return null;
+    }
+
+    return t('ProductDetails.currentStock', {
```
Comment (animesh1987) [inline core/app/[locale]/(default)/product/[slug]/page.tsx:236]: What is `t` here?

core/app/[locale]/(default)/product/[slug]/page.tsx
```diff
@@ -196,6 +197,36 @@ export default async function Product({ params, searchParams }: Props) {
...
+
+    const product = await streamableProduct;
+
+    if (!product.inventory.isInStock) {
+      return showOutOfStockMessage ? defaultOutOfStockMessage : null;
+    }
+
+    if (
+      stockLevelDisplay === 'DONT_SHOW' ||
+      (stockLevelDisplay === 'SHOW_WHEN_LOW' &&
```
Comment (animesh1987) [inline core/app/[locale]/(default)/product/[slug]/page.tsx]: This condition feels a bit overloaded and hard to read. Can this be simplified? Maybe we can use a util method?

core/app/[locale]/(default)/product/[slug]/page.tsx
Comment (Tharaae) [reply core/app/[locale]/(default)/product/[slug]/page.tsx]: The type of `stockLevelDisplay` is already defined by bigcommerce graphql data types as a set of values. This is common for graphql fields in Catalyst.
Screenshot 2025-09-23 at 7 01 03 am

core/app/[locale]/(default)/product/[slug]/page.tsx
Comment (Tharaae) [reply core/app/[locale]/(default)/product/[slug]/page.tsx:236]: It's defined on [this line](https://github.com/bigcommerce/catalyst/blob/b7085753680cb4b61557f92263dc51a59f17bf6b/core/app/%5Blocale%5D/(default)/product/%5Bslug%5D/page.tsx#L74) where translations related to "Product" page are fetched.

core/app/[locale]/(default)/product/[slug]/page.tsx
Comment (Tharaae) [reply core/app/[locale]/(default)/product/[slug]/page.tsx]: I'll decompose it to be more clear like:
```
    if (stockLevelDisplay === 'DONT_SHOW') {
      return null;
    }

    if (stockLevelDisplay === 'SHOW_WHEN_LOW') {
      const { availableToSell, warningLevel } = product.inventory.aggregated ?? {};

      if (!warningLevel) {
        return null;
      }

      if (availableToSell && availableToSell > warningLevel) {
        return null;
      }
    }
```

core/vibes/soul/sections/product-detail/index.tsx
```diff
@@ -119,6 +120,11 @@ export function ProductDetail<F extends Field>({
                       )}
                     </Stream>
                   </div>
+                  <div className="group/product-stock-level">
+                    {Boolean(product.stockLevelMessage) && (
+                      <p className="prose prose-sm">{product.stockLevelMessage}</p>
+                    )}
```
chanceaclark: Since you are a `Streamable`, we should wrap it in a `<Stream>` component to actually stream this data in on page load otherwise it'll be a blocking request. Additionally, we should be using the CSS variables for the stylings here so we get the benefit of Makeswift theming.
```suggestion
                    <Stream fallback={<ProductStockSkeleton />} value={product.stockLevelMessage}>
                      {(stockLevelMessage) =>
                        Boolean(stockLevelMessage) && (
                          <p className="text-[var(--product-detail-secondary-text,hsl(var(--contrast-500)))]">
                            {stockLevelMessage}
                          </p>
                        )
                      }
                    </Stream>
```

PR #2621, opened 2025-09-23
fix: locale switcher correctly works w/ localized Makeswift paths

agurtovoy:
## What/Why?

Fix locale switcher on localized Makeswift pages with different paths.

## Testing

https://github.com/user-attachments/assets/4658ebc5-088a-463c-804d-b0f80ae27c6f

## Migration

N/A

### DIFF
--- a/core/vibes/soul/primitives/navigation/index.tsx
+++ b/core/vibes/soul/primitives/navigation/index.tsx
@@ -30,6 +30,8 @@ import { Link } from '~/components/link';
 import { usePathname, useRouter } from '~/i18n/routing';
 import { useSearch } from '~/lib/search';
 
+import { getLocalizedPathname } from './_actions/localized-pathname';
+
 interface Link {
   label: string;
   href: string;
@@ -862,22 +864,36 @@ function SearchResults({
   );
 }
 
-const useSwitchLocale = () => {
+const useSwitchLocale = ({ activeLocale }: { activeLocale: Locale | undefined }) => {
   const pathname = usePathname();
   const router = useRouter();
   const params = useParams();
   const searchParams = useSearchParams();
 
   return useCallback(
-    (locale: string) =>
+    async (locale: string) => {
+      const localizedPathname = await getLocalizedPathname({
+        pathname,
+        activeLocale: activeLocale?.id,
+        targetLocale: locale,
+      });
+
+      // the Next.js App Router guarantees a `startTransition` call on `router.push`,
+      // so we don’t need to wrap it in an explicit nested call as set out in
+      // https://react.dev/reference/react/useTransition#react-doesnt-treat-my-state-update-after-await-as-a-transition
       router.push(
-        // @ts-expect-error -- TypeScript will validate that only known `params`
-        // are used in combination with a given `pathname`. Since the two will
-        // always match for the current route, we can skip runtime checks.
-        { pathname, params, query: Object.fromEntries(searchParams.entries()) },
+        {
+          pathname: localizedPathname,
+          // @ts-expect-error -- TypeScript will validate that only known `params`
+          // are used in combination with a given `pathname`. Since the two will
+          // always match for the current route, we can skip runtime checks.
+          params,
+          query: Object.fromEntries(searchParams.entries()),
+        },
         { locale },
-      ),
-    [pathname, params, router, searchParams],
+      );
+    },
+    [pathname, activeLocale?.id, params, router, searchParams],
   );
 };
 
@@ -892,7 +908,7 @@ function LocaleSwitcher({
 }) {
   const activeLocale = locales.find((locale) => locale.id === activeLocaleId);
   const [isPending, startTransition] = useTransition();
-  const switchLocale = useSwitchLocale();
+  const switchLocale = useSwitchLocale({ activeLocale });
 
   return (
     <div className={className}>
--- a/.changeset/twenty-ants-sip.md
+++ b/.changeset/twenty-ants-sip.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-makeswift": patch
+---
+
+Fix locale switcher on localized Makeswift pages with different paths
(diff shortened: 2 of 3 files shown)


core/vibes/soul/primitives/navigation/index.tsx
```diff
@@ -862,21 +864,35 @@ function SearchResults({
...
+  const [_, startTransition] = useTransition();
 
   return useCallback(
-    (locale: string) =>
-      router.push(
-        // @ts-expect-error -- TypeScript will validate that only known `params`
-        // are used in combination with a given `pathname`. Since the two will
-        // always match for the current route, we can skip runtime checks.
-        { pathname, params, query: Object.fromEntries(searchParams.entries()) },
-        { locale },
```
agurtovoy, line 878: The canonical way to handle this in `next-intl` is to use the [`pathnames` configuration](https://next-intl.dev/docs/routing/configuration#pathnames), but it's too inflexible for our purposes here:

- If we kept it as a build-time configuration, it would require a site redeploy after each Makeswift publish that updates page paths.
- A build-time configuration would also be incorrect for previews.
- If we converted it into a run-time configuration, we’d have to rework our code to retrieve the derived navigation components from a context and then cache and revalidate them at runtime.

The approach in this PR is more localized and arguably more performant, as we don't have to retrieve the whole sitemap.

core/vibes/soul/primitives/navigation/index.tsx
```diff
@@ -862,21 +864,35 @@ function SearchResults({
...
+            pathname: localizedPathname,
+            // @ts-expect-error -- TypeScript will validate that only known `params`
+            // are used in combination with a given `pathname`. Since the two will
+            // always match for the current route, we can skip runtime checks.
+            params,
+            query: Object.fromEntries(searchParams.entries()),
+          },
+          { locale },
+        ),
+      );
```
migueloller, line 894: My understanding is that route changes already trigger transitions. I'm curious why we're using it here.

Perhaps instead we meant to do this:

```ts
return useCallback(startTransition(async () => {
  const localizedPathname = await getLocalizedPathname({})

  router.push()
}))
```

Note how the async transition will mark the transition as pending when the action itself is triggered. This pattern is actually what a "React Action" is, [by definition](https://react.dev/reference/react/useTransition#functions-called-in-starttransition-are-called-actions).

core/vibes/soul/primitives/navigation/index.tsx
```diff
@@ -862,21 +864,35 @@ function SearchResults({
   );
 }
 
-const useSwitchLocale = () => {
+const useSwitchLocale = ({ activeLocale }: { activeLocale: Locale | undefined }) => {
   const pathname = usePathname();
   const router = useRouter();
   const params = useParams();
   const searchParams = useSearchParams();
+  const [_, startTransition] = useTransition();
```
migueloller: Should we return the pending state so we can handle it appropriately as the transition is happening?

core/vibes/soul/primitives/navigation/index.tsx
agurtovoy: There is already a top-level `startTransition` call wrapping this callback [here](https://github.com/bigcommerce/catalyst/pull/2621/files#diff-646160177a2f06865e322e9f5a52f7c6e5b2df315d03827475a82ac7c7c6029eL926); the nested `startTransition` call follows the pattern explicitly set out for async transitions here: https://react.dev/reference/react/useTransition#react-doesnt-treat-my-state-update-after-await-as-a-transition. 

That said, I didn't know that using `router.push` guarantees a `startTransition` call, which does make the explicit `startTransition` call unnecessary.

core/vibes/soul/primitives/navigation/index.tsx
agurtovoy: It's already done at a higher level: https://github.com/bigcommerce/catalyst/pull/2621/files#diff-646160177a2f06865e322e9f5a52f7c6e5b2df315d03827475a82ac7c7c6029eL894

core/vibes/soul/primitives/navigation/index.tsx
agurtovoy: Added a clarifying comment.

migueloller: Ah, TIL. Thanks!

migueloller: A separate thought is: should the transition be handles outside of the hook, where it's not known whether there's a state update, or should it be handled inside as in my suggestion comment, so that the state update and transition are colocated 🤔

agurtovoy: I don't think there's a significant semantic difference in this case; you could argue it either way. I originally left it out of this particular hook and kept it in the `LocaleSwitcher` because, conceptually, switching the locale involves navigation, and that should be a transition, regardless of how the navigation is triggered.

agurtovoy: You could argue that *because* switching the locale involves navigation, the hook should encapsulate that and return the pending state together with the `switchLocale` callback. I'm sympathetic to that argument, but not enough to go rework this, haha.

migueloller: Yeah, I would argue the latter. And also, APIs like `router.push` encapsulate the transition as well.

That said, I'm not expecting you to rework this. Thanks for the fix!

agurtovoy: > APIs like `router.push` encapsulate the transition as well.

But it doesn't return a pending state, so you have to wrap it in a `startTransition` call if you want that anyway.  Similarly, the `useSwitchLocale` hook doesn't return a pending state because it was conceived as a lower-level primitive, and the pending state handling was left to the `LocaleSwitcher` component itself. 

Now, if it occurred to me at the time, I'd likely go with the design you're advocating, I'm just pointing out that either of them is reasonable.

PR #2625, opened 2025-09-29
Add support for deployed custom B2B Buyer Portal based on env vars

CNanninga:
## What/Why?

Using the B2B Buyer Portal in a headless storefront can involve loading the Buyer Portal assets from one of three sources:
1. The official hosted Buyer Portal
2. The Open Source Buyer Portal running in a local environment
3. The Open Source Buyer Portal deployed to a remote host

Each has a different set of scripts.

Catalyst currently supports #1 and #2, conditionally including the right loader scripts based on the presence or absence of `LOCAL_BUYER_PORTAL_HOST`.  What's needed is support for #3.

When loading a deployed custom Buyer Portal, the base URL must be known, and there might also be hashes included in the URL for each of the three "top-level" JS files.  

This PR introduces env vars for the base URL and the three build hashes and introduces a third loader style that incorporates these.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -50,3 +50,17 @@ B2B_API_TOKEN=
 
 # URL for the local buyer portal instance. Uncomment if developing locally.
 # LOCAL_BUYER_PORTAL_HOST=http://localhost:3001
+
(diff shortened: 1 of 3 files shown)


CNanninga:
Closing this in favor of separating two concerns into two PRs

PR #2626, opened 2025-09-30
Add support for deployed custom B2B Buyer Portal based on env vars

CNanninga:
## What/Why?

Using the B2B Buyer Portal in a headless storefront can involve loading the Buyer Portal assets from one of three sources:

1. The official hosted Buyer Portal
2. The Open Source Buyer Portal running in a local environment
3. The Open Source Buyer Portal deployed to a remote host

Each has a different set of scripts.

Catalyst currently 1 and 2, conditionally including the right loader scripts based on the presence or absence of `LOCAL_BUYER_PORTAL_HOST`. What's needed is support for 3.

This PR introduces an env var for the base URL of the deployed custom Buyer Portal and introduces a third loader style that incorporates this.

### DIFF
--- a/core/b2b/loader.tsx
+++ b/core/b2b/loader.tsx
@@ -10,7 +10,7 @@ const EnvironmentSchema = z.object({
   BIGCOMMERCE_STORE_HASH: z.string({ message: 'BIGCOMMERCE_STORE_HASH is required' }),
   BIGCOMMERCE_CHANNEL_ID: z.string({ message: 'BIGCOMMERCE_CHANNEL_ID is required' }),
   LOCAL_BUYER_PORTAL_HOST: z.string().url().optional(),
-  PROD_BUYER_PORTAL_HOST: z.string().url().optional(),
+  PROD_BUYER_PORTAL_BASE_URL: z.string().url().optional(),
   STAGING_B2B_CDN_ORIGIN: z.string().optional(),
 });
 
@@ -19,7 +19,7 @@ export async function B2BLoader() {
     BIGCOMMERCE_STORE_HASH,
     BIGCOMMERCE_CHANNEL_ID,
     LOCAL_BUYER_PORTAL_HOST,
-    PROD_BUYER_PORTAL_HOST,
+    PROD_BUYER_PORTAL_BASE_URL,
     STAGING_B2B_CDN_ORIGIN,
   } = EnvironmentSchema.parse(process.env);
 
@@ -37,14 +37,14 @@ export async function B2BLoader() {
     );
   }
   
-  if (PROD_BUYER_PORTAL_HOST) {
+  if (PROD_BUYER_PORTAL_BASE_URL) {
     return (
       <ScriptProductionCustom
         cartId={session?.user?.cartId}
         channelId={BIGCOMMERCE_CHANNEL_ID}
         storeHash={BIGCOMMERCE_STORE_HASH}
         token={session?.b2bToken}
-        prodUrl={PROD_BUYER_PORTAL_HOST}
+        prodUrl={PROD_BUYER_PORTAL_BASE_URL}
       />
     );
   }
--- a/.env.example
+++ b/.env.example
@@ -52,5 +52,5 @@ B2B_API_TOKEN=
 # LOCAL_BUYER_PORTAL_HOST=http://localhost:3001
 
 # Base URL for a production build of Buyer Portal. Uncomment if connecting to a deployed custom Buyer Portal.
-#   Example URL format: ${PROD_BUYER_PORTAL_URL}/index.js
-# PROD_BUYER_PORTAL_HOST=https://my-b2b-catalyst.com/b2b/20260101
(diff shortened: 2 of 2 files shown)


core/b2b/script-production-custom.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+import { useB2BAuth } from './use-b2b-auth';
+import { useB2BCart } from './use-b2b-cart';
+
+/**
+ * Use these vars if using build hashes in B2B Buyer Portal files.
+ */
+const hashIndex = undefined;
+const hashIndexLegacy = undefined;
+const hashPolyfills = undefined;
```
matthewvolk: Let's prefer `null` here if it doesn't break anything,

core/b2b/loader.tsx
```diff
@@ -4,11 +4,13 @@ import { auth } from '~/auth';
 
 import { ScriptDev } from './script-dev';
 import { ScriptProduction } from './script-production';
+import { ScriptProductionCustom } from './script-production-custom';
 
 const EnvironmentSchema = z.object({
   BIGCOMMERCE_STORE_HASH: z.string({ message: 'BIGCOMMERCE_STORE_HASH is required' }),
   BIGCOMMERCE_CHANNEL_ID: z.string({ message: 'BIGCOMMERCE_CHANNEL_ID is required' }),
   LOCAL_BUYER_PORTAL_HOST: z.string().url().optional(),
+  PROD_BUYER_PORTAL_URL: z.string().url().optional(),
```
matthewvolk: nitpick, but are we cool with calling these LOCAL_BUYER_PORTAL_**_HOST_** and PROD_BUYER_PORTAL_**_URL_**, even though they both get validated as URLs? Should we just default to the existing pattern of calling it a HOST?

matthewvolk (commented):
Just one comment, and then one smaller nitpick to consider. After that, this will look good from my end 👍

core/b2b/loader.tsx
CNanninga: I'm cool with that.

CNanninga:
Ready for re-review

matthewvolk (approved):

bc-victor (approved):

core/b2b/script-production-custom.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+import { useB2BAuth } from './use-b2b-auth';
+import { useB2BCart } from './use-b2b-cart';
+
+/**
+ * Use these vars if using build hashes in B2B Buyer Portal files.
+ */
+const hashIndex = null;
+const hashIndexLegacy = null;
+const hashPolyfills = null;
```
bc-micah, line 13: Is the idea that the hashes are hard coded - so that they can be versioned through the code?

I guess that would make it easier if you open a PR - you can test out the custom build without having to change env vars for that branch alone.

```suggestion
const hashIndex: string | null = null;
const hashIndexLegacy: string | null = null;
const hashPolyfills: string | null = null;
```

core/b2b/loader.tsx
```diff
@@ -4,11 +4,13 @@ import { auth } from '~/auth';
 
 import { ScriptDev } from './script-dev';
 import { ScriptProduction } from './script-production';
+import { ScriptProductionCustom } from './script-production-custom';
 
 const EnvironmentSchema = z.object({
   BIGCOMMERCE_STORE_HASH: z.string({ message: 'BIGCOMMERCE_STORE_HASH is required' }),
   BIGCOMMERCE_CHANNEL_ID: z.string({ message: 'BIGCOMMERCE_CHANNEL_ID is required' }),
   LOCAL_BUYER_PORTAL_HOST: z.string().url().optional(),
+  PROD_BUYER_PORTAL_HOST: z.string().url().optional(),
```
bc-micah: I saw matt push for `HOST` for consistency - but I think the `PROD_BUYER_PORTAL_BASE_URL` makes more sense as it can be a sub directory.

`HOST` usually stands for hostname - which is usually a domain / ip for the device - and does not really indicate that you can have subdirectories etc.

core/b2b/script-production-custom.tsx
CNanninga: Matt and I had some discussion about this, and for now it didn't seem like it was worthwhile to represent these as environment vars.  What we expect to be a common workflow is for devs to deploy their custom Buyer Portal build files _with_ Catalyst (as static files in `public`), so a new code deployment is commonly needed anyway.  (And having the hashes in env vars actually means a code re-deploy and then another re-deploy after updating vars.)

What I opted to do instead was just separate them out very conspicuously in this file.

core/b2b/loader.tsx
CNanninga: @matthewvolk Can you comment on this?

matthewvolk: I probably should have been less prescriptive in that comment... the main point I was driving at was consistency. If you feel these two variables behave differently and that's intentional, then consistency in naming probably doesn't need to apply.

PR #2631, opened 2025-10-07
fix: B2B-3725 add required b2b headers

bc-victor:
## What/Why?
After the release of PROJECT-6952, developers using a BC store-level oauth token with the new B2B Edition oauth scope need to send additional headers for the endpoint to respond properly.

Adding env variable `BIGCOMMERCE_ACCESS_TOKEN` to differentiate from the type of headers we need to send to the B2B APIs

## Testing
Login success ( only place where we use B2B APIs) 

With B2B token

https://github.com/user-attachments/assets/b2e19b82-9c20-47fb-8867-d0f3e04adb65

Without BC token

https://github.com/user-attachments/assets/d3c79391-8f74-47c4-a561-300085badda2

## Migration
No migrations

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -44,9 +44,11 @@ DEFAULT_REVALIDATE_TARGET=3600
 # URL for the B2B API. This is used to connect to the B2B API for features like customer impersonation and B2B-specific data
 B2B_API_HOST=https://api-b2b.bigcommerce.com
 
-# The B2B API Token is used to authenticate requests to the B2B API.
-# It can be generated in the B2B control panel Settings > API Accounts > Create API Account. 
-B2B_API_TOKEN=
+# A store-level API account token used for REST API actions. Optional by default, but required in
+# the sign-in logic that interacts with the buyer portal. This integration requires a 
+# `BIGCOMMERCE_ACCESS_TOKEN` with `modify` scope on B2B Edition.
+# See https://support.bigcommerce.com/s/article/Store-API-Accounts?language=en_US
+BIGCOMMERCE_ACCESS_TOKEN=
 
 # URL for the local buyer portal instance. Uncomment if developing locally.
 # LOCAL_BUYER_PORTAL_HOST=http://localhost:3001
--- a/core/b2b/client.ts
+++ b/core/b2b/client.ts
@@ -11,10 +11,17 @@ interface LoginWithB2BParams {
 
 const ENV = z
   .object({
-    env: z.object({
-      B2B_API_TOKEN: z.string(),
-      BIGCOMMERCE_CHANNEL_ID: z.string(),
-    }),
+    env: z.union([
+      z.object({
+        BIGCOMMERCE_CHANNEL_ID: z.string(),
+        B2B_API_TOKEN: z.string(),
+      }),
+      z.object({
+        BIGCOMMERCE_CHANNEL_ID: z.string(),
+        BIGCOMMERCE_STORE_HASH: z.string(),
+        BIGCOMMERCE_ACCESS_TOKEN: z.string(),
+      }),
+    ]),
   })
   .transform(({ env }) => env);
 
@@ -29,17 +36,27 @@ const B2BTokenResponseSchema = z.object({
 });
 
 export async function loginWithB2B({ customerId, customerAccessToken }: LoginWithB2BParams) {
-  const { B2B_API_TOKEN, BIGCOMMERCE_CHANNEL_ID } = ENV.parse(process);
+  const env = ENV.parse(process);
+  const BIGCOMMERCE_CHANNEL_ID = env.BIGCOMMERCE_CHANNEL_ID;
+  const headers: HeadersInit = {
+    Accept: 'application/json',
+    'Content-Type': 'application/json',
+  };
+
+  if ('BIGCOMMERCE_ACCESS_TOKEN' in env) {
+    headers['X-Auth-Token'] = env.BIGCOMMERCE_ACCESS_TOKEN;
+    headers['X-Store-Hash'] = env.BIGCOMMERCE_STORE_HASH;
+  } else if ('B2B_API_TOKEN' in env) {
+    headers['authToken'] = env.B2B_API_TOKEN;
+    console.warn('This is deprecated in favour or BIGCOMMERCE_ACCESS_TOKEN, read https://support.bigcommerce.com/s/article/Store-API-Accounts?language=en_US')
+  } else {
+    throw new Error('No B2B API token or BigCommerce token found in environment variables.');
+  } 
 
   const apiHost = getAPIHostname();
-
   const response = await fetch(`${apiHost}/api/io/auth/customers/storefront`, {
     method: 'POST',
-    headers: {
-      Accept: 'application/json',
-      'Content-Type': 'application/json',
-      authToken: B2B_API_TOKEN,
-    },
+    headers,
     body: JSON.stringify({
       channelId: BIGCOMMERCE_CHANNEL_ID,
       customerId,


bc-sylvialin:
Hey @bc-victor , this is not a bug. It is intentional to use new headers.

See below articles:
https://developer.bigcommerce.com/b2b-edition/docs/authentication
https://developer.bigcommerce.com/resource-hub/authenticate-bigcommerce-and-b2b-edition-apis-with-one-api-account

core/b2b/client.ts
```diff
@@ -14,6 +14,7 @@ const ENV = z
     env: z.object({
       B2B_API_TOKEN: z.string(),
       BIGCOMMERCE_CHANNEL_ID: z.string(),
+      BIGCOMMERCE_STORE_HASH: z.string(),
```
bc-micah: Did we update the docs that explain what a `B2B_API_TOKEN` is as well?

core/b2b/client.ts
```diff
@@ -39,6 +40,8 @@ export async function loginWithB2B({ customerId, customerAccessToken }: LoginWit
       Accept: 'application/json',
       'Content-Type': 'application/json',
       authToken: B2B_API_TOKEN,
+      'X-Store-Hash': BIGCOMMERCE_STORE_HASH,
+      'X-Auth-Token': B2B_API_TOKEN,
```
bc-micah: Can we include full tests of both tokens (old b2b only s2s token + bigcommerce api account token)?

I would like to see the full flow tested for both scenarios and include showing / setting the env var + verifying the login is working appropriately (including opening the buyer portal as that shows the storefront is logged in as well as the b2b functions)

bc-micah (commented):
I want to see the docs updated + including tests for both the old and new tokens - comments have details.

core/b2b/client.ts
bc-victor: PR here: https://github.com/bigcommerce/docs/pull/1104

core/b2b/client.ts
bc-victor: thanks for the review. After testing realized we can't just send `authToken`,  `X-Auth-Token` and `X-Store-Hash` together, we need to know which token is being used. I added a new variable `BIGCOMMERCE_TOKEN_WITH_B2B_SCOPE` to differntiatte from scopes

.env.example
```diff
@@ -44,9 +44,13 @@ DEFAULT_REVALIDATE_TARGET=3600
...
 B2B_API_HOST=https://api-b2b.bigcommerce.com
 
+# BigCommerce API token with B2B Edition scopes used to authenticate requests to the B2B API 
+BIGCOMMERCE_TOKEN_WITH_B2B_SCOPE=
+
+# DEPRECATED - Please prefer using the above variable
 # The B2B API Token is used to authenticate requests to the B2B API.
 # It can be generated in the B2B control panel Settings > API Accounts > Create API Account. 
-B2B_API_TOKEN=
+# B2B_API_TOKEN=
```
bc-victor: maybe lets remove this entirely? also from the docs

icatalina: can we do it directly?

maybe add a check, if B2B_API_TOKEN is present, throw and show an error message?

```tsx
z.object({
  B2B_API_TOKEN: z.undefined({ message: 'This is deprecated in favour or B2B_API_TOKEN'}).optional(),
  BIGCOMMERCE_TOKEN: z.string()
})
```

.env.example
```diff
@@ -44,9 +44,13 @@ DEFAULT_REVALIDATE_TARGET=3600
 # URL for the B2B API. This is used to connect to the B2B API for features like customer impersonation and B2B-specific data
 B2B_API_HOST=https://api-b2b.bigcommerce.com
 
+# BigCommerce API token with B2B Edition scopes used to authenticate requests to the B2B API 
+BIGCOMMERCE_TOKEN_WITH_B2B_SCOPE=
```
icatalina:
```suggestion
BIGCOMMERCE_TOKEN=
```

if we require new scopes in the future, are we gonna rename the variable to:

```
BIGCOMMERCE_TOKEN_WITH_B2B_AND_SOME_OTHER_SCOPE=
```

?

bc-victor: Thats a good point, I'll add this

CNanninga: "With x scope" seems like an awkward naming convention in general to me.

Just for reference, the optional BigCommerce REST token that can be included in the environment config for, e.g., supporting the customer groups lookup in Makeswift, is called `BIGCOMMERCE_ACCESS_TOKEN`.  We might want to conform the naming convention here to more closely match that.

Or is it possibly appropriate to just _use_ `BIGCOMMERCE_ACCESS_TOKEN` and give clear instructions that the token provided there needs to have the B2B scope?  I think that might make more sense than having separate BigCommerce REST tokens for separate operations.

core/b2b/client.ts
```diff
@@ -29,17 +31,26 @@ const B2BTokenResponseSchema = z.object({
...
+  };
+
+  if (B2B_API_TOKEN) {
+    headers['authToken'] = B2B_API_TOKEN;
+  }
+
+  if (BIGCOMMERCE_TOKEN_WITH_B2B_SCOPE) {
+    headers['X-Auth-Token'] = BIGCOMMERCE_TOKEN_WITH_B2B_SCOPE;
+    headers['X-Store-Hash'] = BIGCOMMERCE_STORE_HASH;
+  }
```
bc-micah:
```suggestion
  if (BIGCOMMERCE_TOKEN_WITH_B2B_SCOPE) {
    headers['X-Auth-Token'] = BIGCOMMERCE_TOKEN_WITH_B2B_SCOPE;
    headers['X-Store-Hash'] = BIGCOMMERCE_STORE_HASH;
  } else  if (B2B_API_TOKEN) {
    headers['authToken'] = B2B_API_TOKEN;
  }
```

They both should not be used at the same time, we should prefer the `BIGCOMMERCE_TOKEN` if possible.

.env.example
bc-micah: If it's already being used in makeswift, let's stick to the existing name to reduce the number of things the user has to setup.

We can then just add that `B2B` scope is required in the setup docs.

core/b2b/client.ts
```diff
@@ -29,17 +31,24 @@ const B2BTokenResponseSchema = z.object({
...
+    Accept: 'application/json',
+    'Content-Type': 'application/json',
+  };
+
+  if (BIGCOMMERCE_ACCESS_TOKEN) {
+    headers['X-Auth-Token'] = BIGCOMMERCE_ACCESS_TOKEN;
+    headers['X-Store-Hash'] = BIGCOMMERCE_STORE_HASH;
+  } else {
+    throw new Error(`No B2B API token or BigCommerce token found in environment variables.`);
+  }
```
bc-micah: This is only using the new `BIGCOMMERCE_ACCESS_TOKEN`, will break when existing customers upgrade to this version.

core/b2b/client.ts
```diff
@@ -12,8 +12,10 @@ interface LoginWithB2BParams {
 const ENV = z
   .object({
     env: z.object({
-      B2B_API_TOKEN: z.string(),
       BIGCOMMERCE_CHANNEL_ID: z.string(),
+      BIGCOMMERCE_STORE_HASH: z.string(),
+      BIGCOMMERCE_ACCESS_TOKEN: z.string().optional(),
+      B2B_API_TOKEN: z.string().optional().describe('This is deprecated in favour or BIGCOMMERCE_ACCESS_TOKEN, read https://support.bigcommerce.com/s/article/Store-API-Accounts?language=en_US'),
```
icatalina: 💅 Not sure if we need the `describe` here. 

We might want to define a union in this case:

```tsx
z.union([
  z.object({
    BIGCOMMERCE_CHANNEL_ID: z.string(),
    B2B_API_TOKEN: z.string(),
  }),
  z.object({
    BIGCOMMERCE_CHANNEL_ID: z.string(),
    BIGCOMMERCE_STORE_HASH: z.string(),
    BIGCOMMERCE_ACCESS_TOKEN: z.string(),
  }),
])
```

The current approach makes `BIGCOMMERCE_STORE_HASH` mandatory, even for old consumers where it is not needed, and makes both `B2B_API_TOKEN` and `BIGCOMMERCE_ACCESS_TOKEN` which mean they can potentially be both missing.

PR #2635, opened 2025-10-14
feat(core): Add link to Gift Certificates page in header/footer

jordanarldt:
## What/Why?
Add a link in the header and footer to route to `/gift-certificates`. This link is only visible when the active currency has Gift Certificates enabled.

Adding them to both will make it easier if consumers want to customize it and remove it from just the header or the footer.

## Testing

image

image

### Update GraphQL Query

In `/app/[locale]/(default)/page-data.ts`, update the `LayoutQuery` to accept a `$currencyCode` variable and include the new giftCertificates field.

### Update Header Fragment

In `/components/header/fragment.ts`, add the `giftCertificates(currencyCode: $currencyCode)` field to the header fragment, including its properties.

### Update Header Data Fetching

In `/components/header/index.tsx`, pass the preferred currency code when fetching header data and add new props for gift certificates (label, link, enabled flag) to header links.

### Add Localization

In /messages/en.json, add a new entry for "giftCertificates": "Gift Certificates" under the "Icons" section.

### Update Navigation Component

In `/vibes/soul/primitives/navigation/index.tsx`:
Import `GiftIcon` from `lucide-react`.

Add new props for the gift certificates label, link, and enabled flag.

Render the new "Gift Certificates" navigation link and icon, conditionally shown when enabled.

### DIFF
--- a/core/components/footer/index.tsx
+++ b/core/components/footer/index.tsx
@@ -16,7 +16,9 @@ import { getSessionCustomerAccessToken } from '~/auth';
 import { client } from '~/client';
 import { readFragment } from '~/client/graphql';
 import { revalidate } from '~/client/revalidate-target';
+import { CurrencyCode } from '~/components/header/fragment';
 import { logoTransformer } from '~/data-transformers/logo-transformer';
+import { getPreferredCurrencyCode } from '~/lib/currency';
 
 import { FooterFragment, FooterSectionsFragment } from './fragment';
 import { AmazonIcon } from './payment-icons/amazon';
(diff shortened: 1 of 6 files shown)


core/vibes/soul/primitives/navigation/index.tsx
```diff
@@ -595,6 +609,14 @@ export const Navigation = forwardRef(function Navigation<S extends SearchResult>
             </Stream>
           </Link>
 
+          <Link
+            aria-label={giftCertificatesLabel}
+            className={navButtonClassName}
+            href={giftCertificatesHref}
+          >
+            {giftCertificatesEnabled ? <GiftIcon size={20} strokeWidth={1} /> : null}
```
matthewvolk: Sorry, I think I forgot to click save on my comment 🤦 

I could be wrong but I think this leaves a clickable link in the DOM when `giftCertificatesEnabled` is falsy. I would double check to make sure.

jordanarldt: @matthewvolk Good catch!

core/components/footer/index.tsx
```diff
@@ -68,8 +71,8 @@ const getFooterData = cache(async () => {
 
 export const Footer = async () => {
   const t = await getTranslations('Components.Footer');
-
-  const data = await getFooterData();
+  const currencyCode = await getPreferredCurrencyCode();
+  const data = await getFooterData(currencyCode);
```
jorgemoya: This is an unfortunate side effect, but fetching `getPreferredCurrencyCode` dynamic data will prevent Header/Footer from being pre-rendered I believe? But, as always, I don't see a work around unless we stream in this specific section/link somehow?

jordanarldt: @jorgemoya Hmm, streaming in the link would require making a separate GQL call right? I know we want to pre-render as much as possible, but I'm wondering if it would be worth it to make a duplicate call just to stream a single link in. Won't the request already be cached by the `currencyCode` argument we pass to the data fetcher? 🤔

jorgemoya: I could be wrong (I'm rusty++), but although the request will be cached, the layout will be dynamic and never prerendered since it looks for dynamic data. 

Yeah, one solution would be to stream in the one link, I believe.

PR #2636, opened 2025-10-14
feat(core): Add /gift-certificates page on enabled currencies, add gift certificate ui components

jordanarldt:
## What/Why?
- Add the `/gift-certificates` root page
- Add `GiftCertificateCard` and `GiftCertificateSection` components

## Testing

image

image

### UI Components

Copy the `GiftCertificateCard` and `GiftCertificateSection` components from this PR.

### Add translations

Copy the translations from this PR into your translations file (e.g. `en.json`)
```
  "GiftCertificates": {
    "title": "Gift certificates",
    "description": "Give the perfect gift that never goes out of style. Let friends and loved ones choose exactly what they want from our entire collection.",
    "purchaseLabel": "Shop now",
    "checkBalanceLabel": "Check balance"
  }
```

### Add the page route

Copy all the files in the `core/app/[locale]/(default)/gift-certificates` from this branch/PR

### DIFF
--- a/core/app/[locale]/(default)/gift-certificates/page-data.ts
+++ b/core/app/[locale]/(default)/gift-certificates/page-data.ts
@@ -1,0 +1,37 @@
+import { cache } from 'react';
+
+import { client } from '~/client';
+import { graphql } from '~/client/graphql';
+import { revalidate } from '~/client/revalidate-target';
+import { CurrencyCode } from '~/components/header/fragment';
+import { StoreLogoFragment } from '~/components/store-logo/fragment';
+import { logoTransformer } from '~/data-transformers/logo-transformer';
+
+const GiftCertificatesRootQuery = graphql(
+  `
+    query GiftCertificatesRootQuery($currencyCode: currencyCode) {
+      site {
+        settings {
+          giftCertificates(currencyCode: $currencyCode) {
+            isEnabled
+          }
+          ...StoreLogoFragment
+        }
+      }
+    }
+  `,
+  [StoreLogoFragment],
+);
+
+export const getGiftCertificatesData = cache(async (currencyCode?: CurrencyCode) => {
+  const response = await client.fetch({
+    document: GiftCertificatesRootQuery,
+    variables: { currencyCode },
+    fetchOptions: { next: { revalidate } },
+  });
+
+  return {
+    giftCertificatesEnabled: response.data.site.settings?.giftCertificates?.isEnabled ?? false,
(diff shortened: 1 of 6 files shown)


core/app/[locale]/(default)/gift-certificates/page-data.ts
```diff
@@ -0,0 +1,37 @@
...
+    document: GiftCertificatesRootQuery,
+    variables: { currencyCode },
+    fetchOptions: { next: { revalidate } },
+  });
+
+  return {
+    giftCertificatesEnabled: response.data.site.settings?.giftCertificates?.isEnabled ?? false,
+    logo: response.data.site.settings ? logoTransformer(response.data.site.settings) : '',
+  };
+});
```
Comment (matthewvolk) [inline core/app/[locale]/(default)/gift-certificates/page-data.ts:37]: In #2635 we're doing the currency code fetch inside the memoized `getXYZData` fn. Here, we accept currency code as an argument; should we choose one for consistency? The cache is invalidated on [each new server request](https://react.dev/reference/react/cache#caveats), so I don't think there's much of a difference at runtime... but I think it's worth being consistent if possible

core/app/[locale]/(default)/gift-certificates/page.tsx
```diff
@@ -0,0 +1,55 @@
...
+  setRequestLocale(locale);
+
+  const t = await getTranslations('GiftCertificates');
+  const format = await getFormatter();
+  const currencyCode = await getPreferredCurrencyCode();
+  const data = await getGiftCertificatesData(currencyCode);
+
+  if (!data.giftCertificatesEnabled) {
+    return redirect({ href: '/', locale });
+  }
```
Comment (matthewvolk) [inline core/app/[locale]/(default)/gift-certificates/page.tsx:36]: 🍹 This looks sus to me... I was about to comment to ask if we should move this check to middleware so it runs on the edge for better performance... but I see it depends on the graphql response with the `data.giftCertificatesEnabled` field. Is this something we should cache similar to the pattern we have in `with-routes.ts`?

core/app/[locale]/(default)/gift-certificates/page-data.ts
Comment (jordanarldt) [reply core/app/[locale]/(default)/gift-certificates/page-data.ts:37]: Done! Updated the other PR to take `currencyCode` as an argument on the getFooterData function

core/app/[locale]/(default)/gift-certificates/page.tsx
Comment (jorgemoya) [reply core/app/[locale]/(default)/gift-certificates/page.tsx:36]: Should we redirect to `notFound` instead?

PR #2643, opened 2025-10-21
feat(scripts): add consent manager provider using c15t

matthewvolk:
## What/Why?
Adds a new consent manager provider using the `@c15t/nextjs` package to handle cookie consent management with support for multiple consent categories and cookie-based persistence.

The consent manager includes:
- A React provider component that wraps the C15T consent manager
- Custom endpoint handlers for showing consent banners, setting consent preferences, and verifying consent status via cookies

> [!NOTE]
> Future PR's will include UI components for the cookie banner, dialog, and consent customization widget.

## Testing
This implementation uses a "custom" mode with predefined consent categories, allowing for flexible consent management without requiring external API calls. In order to test, you'll need to first wrap the RootLayout in the provider, then import the unstyled `<CookieBanner />` and `<ConsentManagerDialog />` components from `@c15t/nextjs`.

## Migration
**For developers pulling in these changes:**

- **New dependency**: The `@c15t/nextjs` package has been added to `core/package.json`. Run `pnpm install` to install it.
- **New files added**:
   - `core/lib/consent-manager/handlers.ts` - Contains the consent management logic and API handlers
   - `core/lib/consent-manager/index.ts` - Exports the provider component
   - `core/lib/consent-manager/provider.tsx` - The main React provider component
- **No breaking changes**: This is a purely additive change that doesn't modify existing functionality. The consent manager operates independently and can be integrated gradually.

### DIFF
--- a/core/lib/consent-manager/handlers.ts
+++ b/core/lib/consent-manager/handlers.ts
@@ -1,0 +1,58 @@
+import { getConsentCookie, setConsentCookie } from './cookies/client';
+
+const ok = <T>(data: T | null = null) => ({
+  data,
+  error: null,
+  ok: true,
+  response: null,
+});
+
+export function showConsentBanner() {
+  let show = true;
+  let language = 'en';
+
+  if (typeof document !== 'undefined') {
+    language = document.documentElement.lang;
+  }
+
+  try {
+    const consent = getConsentCookie();
+
+    show = !consent;
+  } catch {
+    show = false;
+  }
+
+  return ok({
+    showConsentBanner: show,
+    translations: {
+      language,
+    },
+    branding: 'none',
+  });
+}
+
+export function setConsent(options?: { body?: { preferences?: Record<string, boolean> } }) {
+  const { preferences } = options?.body ?? {};
+
+  setConsentCookie({
+    preferences: preferences ?? {},
+    timestamp: new Date().toISOString(),
+  });
+
+  return ok();
+}
+
+export function verifyConsent() {
+  const consent = getConsentCookie();
+
+  if (!consent) {
+    return ok({
+      isValid: false,
+    });
+  }
+
+  return ok({
+    isValid: true,
+  });
+}
--- a/.changeset/full-bees-worry.md
+++ b/.changeset/full-bees-worry.md
@@ -1,0 +1,5 @@
+---
+'@bigcommerce/catalyst-core': minor
+---
+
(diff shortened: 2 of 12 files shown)


core/lib/consent-manager/cookies.ts
```diff
@@ -0,0 +1,74 @@
...
+    },
+    set: (consent: ConsentCookie) => {
+      if (typeof document === 'undefined') return;
+      document.cookie = serialize(encode(consent));
+    },
+  };
+};
+
+export const consentCookieServer = async () => {
+  const { cookies } = await import('next/headers');
```
chanceaclark: This might import the whole `next/headers` export and probably won't be tree shaked. Maybe we need to change the file structure a bit. What do you think about this?
- `lib/consent-manager/schema.ts`
- `lib/consent-manager/encoder.ts`
- `lib/consent-manager/decoder.ts`
- `lib/consent-manager/cookies/client.ts`
- `lib/consent-manager/cookies/server.ts`

jorgemoya: +1

core/lib/consent-manager/handlers.ts
```diff
@@ -0,0 +1,58 @@
+import { consentCookieClient } from './cookies';
+
+const ok = <T>(data: T | null = null) => ({
```
jorgemoya, line 3: What is the purpose of this `ok`? Is this an expected object for c15t?

matthewvolk: Correct! It's just a helper that follows the interface they define here: https://github.com/c15t/c15t/blob/f4411d824c14b5a59984cc726ecc8d54ae42e44b/packages/core/src/client/types.ts#L247-L303 

I wish there was a better way to import that type somehow (they don't expose it from the package from what I've seen) so that we can be more explicit... I just couldn't figure out exactly how to do that.

core/lib/consent-manager/cookies.ts
matthewvolk: I like it! Working on it now

matthewvolk: Done ✅

jorgemoya: Should `encoder.ts` and `decoder.ts` be under `consent-manager/cookies`? 🍹

chanceaclark: Might be good enough to combine into a `lib/consent-manager/serializer.ts` file.

PR #2645, opened 2025-10-22
feat(core): Add gift certificates to orders list and order details

jordanarldt:
## What/Why?
Add ability to see Gift Certificates in the order details of previous orders, and ensure that they are displayed in the orders list as well.

## Testing

image

image

### DIFF
--- a/core/data-transformers/orders-transformer.ts
+++ b/core/data-transformers/orders-transformer.ts
@@ -9,6 +9,53 @@ export const ordersTransformer = (
   format: ExistingResultType<typeof getFormatter>,
 ): Order[] => {
   return orders.map((order) => {
+    const lineItems =
+      order.consignments.shipping?.flatMap((consignment) => {
+        return consignment.lineItems.map((lineItem) => {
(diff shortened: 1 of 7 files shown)


core/data-transformers/orders-transformer.ts
```diff
@@ -17,38 +64,7 @@ export const ordersTransformer = (
...
-              image: lineItem.image
-                ? {
-                    src: lineItem.image.url,
-                    alt: lineItem.image.altText,
-                  }
-                : undefined,
-            };
-          });
-        }) ?? [],
+      lineItems: [...lineItems, ...giftCertificates].flat(),
```
jorgemoya, line 67: Did we ever show `digitalLineItems`? 🤔

jordanarldt: @jorgemoya Nope! That's currently a Catalyst gap that we need to address as soon as we can. We never implemented digital items in the order details.

jorgemoya: Completely forgot, good to know we're keeping track!

PR #2650, opened 2025-10-27
feat(scripts): add consent manager ui and internationalization

matthewvolk:
> [!WARNING]
> Depends on #2643 

## What/Why?
Adds a consent manager UI and internationalization by integrating the C15T `CookieBanner` and `ConsentManagerDialog` with Catalyst components. The update includes:
- A cookie banner for first-time visitors
- A full-screen dialog for managing preferences based off C15T's native component
- i18n via next-intl

## Testing
- First visit: banner offers accept, reject, or customize
- Accept/reject: consent persists; closes dialog
- Customize: opens dialog with accordion list; checkbox per category
- Saved preferences persist and the banner no longer appears
- Translations use the new "ConsentManager" i18n keys

Confirm the `ConsentManagerProvider` from `@/lib/consent-manager` is added to your root layout.

## Migration
1. New files:
   - `core/components/consent-manager/consent-manager-dialog.tsx` — dialog and widget
   - `core/components/consent-manager/cookie-banner.tsx` — banner component

2. Layout updates:
   - Add `ConsentManagerProvider` around app content
   - Render `CookieBanner`
   - Render `ConsentManagerDialog`

3. i18n updates:
   - Add "ConsentManager" translations under `Components.ConsentManager.Common`, `.CookieBanner`, `.Dialog`, and `.ConsentTypes`

4. Breaking changes:
   - None; add `ConsentManagerProvider`, `CookieBanner`, `ConsentManagerDialog`, and translations to your layout

5. Integration:
```tsx
import { ConsentManagerProvider } from '@/lib/consent-manager';
import { CookieBanner } from '@/components/consent-manager/cookie-banner';
import { ConsentManagerDialog } from '@/components/consent-manager/consent-manager-dialog';

// In your layout, ensure this is wrapped by NextIntlClientProvider
<ConsentManagerProvider>
  <YourOtherProviders>
    <CookieBanner />
    <ConsentManagerDialog />
    {/* rest of your app */}
  </YourOtherProviders>
</ConsentManagerProvider>
```

6. i18n keys to add:
   - `Components.ConsentManager.Common.*`
   - `Components.ConsentManager.CookieBanner.*`
   - `Components.ConsentManager.Dialog.*`
   - `Components.ConsentManager.ConsentTypes.*`

---

PR #2651, opened 2025-10-27
feat(scripts): filter script manager gql response based on cookie consent

matthewvolk:
> [!WARNING]
> Depends on #2650

## What/Why?
Filters BigCommerce scripts based on cookie consent. Reads consent from `c15t-consent` on the server, maps C15T consent categories (which cannot be changed, unfortunately) to BigCommerce Script Consent Categories as follows:
* C15T `measurement` → BigCommerce `ANALYTICS`
* C15T `necessary` → BigCommerce `ESSENTIAL`
* C15T `functionality` → BigCommerce `FUNCTIONAL`
* C15T `experience` → BigCommerce `FUNCTIONAL` (BigCommerce does not have a fifth category, C15T does)
* C15T `marketing` → BigCommerce `TARGETING`

These categories are passed to the GraphQL query filter, which does not fetch scripts from BC until the user consents.

## Testing
1. Created a script for each category of consent BigCommerce offers (even "unknown", which is the default for scripts created via API that do not specify "consent_category". For these scripts, our API states Scripts with an unknown consent category do not display on stores with customer cookie consent banners enabled, so I followed that and only display "unknown" scripts when full consent is provided)
2. Shows storefront loading only ESSENTIAL scripts when consent has not been provided
3. Shows storefront loading only ESSENTIAL scripts when "Reject All" (only ESSENTIAL) consent has been provided
4. Shows storefront loading ALL scripts when "Accept All" consent has been provided (including UNKNOWN)
5. Shows storefront loading selective scripts based on customized consent settings (and notably, not loading UNKNOWN in the case where some consent categories have been rejected)

Demo of the testing above:
https://github.com/user-attachments/assets/be44cb40-9077-40a0-9775-0b5170de68f8

## Migration
1. New files:
   - `core/lib/consent-manager/action.ts` - Server action for consent updates
2. Modified files:
   - `core/lib/consent-manager/cookies.ts` - Adds `experience`, exports `ConsentStateSchema`, adds `getConsentCategoriesFromCookie` mapping
   - `core/lib/consent-manager/handlers.ts` - `setConsent` is async; uses server action with revalidation
   - `core/app/[locale]/layout.tsx` - Reads cookie, maps categories, passes to GraphQL
   - `core/components/scripts/fragment.ts` - Adds `consentCategories` filter to header/footer queries
3. Breaking changes:
   - `setConsent` is async; update callers to await it
   - Adds `experience` consent type; UI components already support it
4. Category mapping:
   - `measurement` → `ANALYTICS`
   - `necessary` → `ESSENTIAL`
   - `functionality`, `experience` → `FUNCTIONAL`
   - `marketing` → `TARGETING`
   - If all categories accepted, includes `UNKNOWN`
   - No consent defaults to `ESSENTIAL` only
5. Cache revalidation:
   - Uses `revalidateTag('layout')` so layout data updates after consent changes
7. GraphQL variables:
   - `RootLayoutMetadataQuery` now requires `$consentCategories`
   - Layout reads `c15t-consent` and derives categories before the query

### DIFF
--- a/core/app/[locale]/layout.tsx
+++ b/core/app/[locale]/layout.tsx
@@ -10,7 +10,7 @@ import { cache, PropsWithChildren } from 'react';
 
 import '../../globals.css';
 
-import { Streamable } from '@/vibes/soul/lib/streamable';
+import { Stream, Streamable } from '@/vibes/soul/lib/streamable';
 import { fonts } from '~/app/fonts';
 import { CookieNotifications } from '~/app/notifications';
 import { Providers } from '~/app/providers';
@@ -25,6 +25,8 @@ import { ContainerQueryPolyfill } from '~/components/polyfills/container-query';
 import { ScriptManagerScripts, ScriptsFragment } from '~/components/scripts';
 import { routing } from '~/i18n/routing';
 import { ConsentManagerProvider } from '~/lib/consent-manager';
+import { getConsentCookie } from '~/lib/consent-manager/cookies/server';
+import { getConsentCategoriesFromCookie } from '~/lib/consent-manager/cookies/utils';
 import { getToastNotification } from '~/lib/server-toast';
 
 const RootLayoutMetadataQuery = graphql(
@@ -40,16 +42,26 @@ const RootLayoutMetadataQuery = graphql(
           }
           ...WebAnalyticsFragment
         }
-        content {
-          ...ScriptsFragment
-        }
       }
       channel {
         entityId
       }
     }
   `,
-  [WebAnalyticsFragment, ScriptsFragment],
+  [WebAnalyticsFragment],
+);
+
+const ScriptsQuery = graphql(
+  `
+    query ScriptsQuery($consentCategories: [ScriptConsentCategory!]) {
+      site {
+        content {
+          ...ScriptsFragment
+        }
+      }
+    }
+  `,
+  [ScriptsFragment],
 );
 
 const fetchRootLayoutMetadata = cache(async () => {
@@ -104,7 +116,6 @@ interface Props extends PropsWithChildren {
 export default async function RootLayout({ params, children }: Props) {
   const { locale } = await params;
 
-  const rootData = await fetchRootLayoutMetadata();
   const toastNotificationCookieData = await getToastNotification();
 
   if (!routing.locales.includes(locale)) {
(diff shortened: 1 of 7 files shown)


core/lib/consent-manager/cookies.ts
```diff
@@ -72,3 +87,38 @@ export const consentCookieServer = async () => {
...
+    marketing: 'TARGETING',
+  };
+
+  const preferences = consent.preferences;
+  const entries = Object.entries(preferences);
+
+  const enabledCategories = entries
+    .filter(([, value]) => value)
+    // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+    .map(([key]) => mapping[key as keyof ConsentState]);
```
chanceaclark: Might want to add a `.filter` and validate the key as `ConsentNamesSchema` so we don't have to do type assertions.

core/app/[locale]/layout.tsx
```diff
@@ -53,8 +54,12 @@ const RootLayoutMetadataQuery = graphql(
...
 const fetchRootLayoutMetadata = cache(async () => {
+  const consentCookie = (await consentCookieServer()).get();
+  const consentCategories = getConsentCategoriesFromCookie(consentCookie);
+
   return await client.fetch({
     document: RootLayoutMetadataQuery,
+    variables: { consentCategories },
     fetchOptions: { next: { revalidate } },
   });
 });
```
Comment (jorgemoya) [inline core/app/[locale]/layout.tsx]: We need to rethink how we pass `consentCategories` to this query (or maybe create a new query just for scripts). 

The problem with this implementation is that you're reading from cookies inside the cached function. If we want to enable composable caching at some other point, `consentCategories` needs to be a param passed to `fetchRootLayoutMetadata` (which is simple enough).

However, to read from cookies in `layout.tsx` without wrapping it in a `Streamable`, will make this entire layout dynamic, and we want to keep it as static as possible.

I believe we need to figure out a way to stream in what scripts are loaded per selected consent, and render those scripts after the request has streamed in.

Comment (matthewvolk) [reply core/app/[locale]/layout.tsx]: Great catch! Thanks for that, it slipped my mind. Just pushed up a fix, though I think it'd be worthwhile to talk through it a bit more. There's something smelly about my two separate `streamableHeaderScripts` and `streamableFooterScripts` fn's.

Comment (jorgemoya) [reply core/app/[locale]/layout.tsx]: I see what you say!

What if we just have one `streamableScripts` and then `<Stream />` the script components.

```ts
<Stream fallback={null} value={streamableScripts}>
  {(scripts) => <ScriptManagerScripts scripts={scripts. headerScripts} strategy="afterInteractive" />}
</Stream>

...

<Stream fallback={null} value={streamableScripts}>
  {(scripts) => <ScriptManagerScripts scripts={scripts.footerScripts} strategy="lazyOnload" />}
</Stream>
```

I have no clue how this conflicts with the `strategy` tho.

Comment (jorgemoya) [reply core/app/[locale]/layout.tsx]: I guess it is not recommended to Stream `afterInteractive` scripts

Comment (jorgemoya) [reply core/app/[locale]/layout.tsx]: Follow up question, what scripts are loaded in the header vs footer? Are header scripts essential and should always load? Or is it user choice? My thinking is that maybe header scripts are essential and we always load them in the initial layout request, while scripts requiring consent are loaded in the footer and Streamed in.

PR #2659, opened 2025-10-30
feat(scripts): use c15t to load scripts based on shopper consent preferences

matthewvolk:
> [!NOTE]
> This PR is a replacement for #2651 

Closes #2651

## What/Why?

This PR adds script loading functionality to Catalyst's consent manager, achieving parity with Stencil's consent-aware script loading behavior. 

The implementation replaces the custom `<ScriptManagerScripts />` component with the [native Script Loader feature](https://c15t.com/docs/frameworks/next/script-loader) of the c15t.com library.

Additionally, the consent categories exposed by c15t are different than the consent categories available in the BigCommerce API. The consent categories in the c15t library are hardcoded, so we added a `scriptsTransformer` to map BigCommerce's consent categories to C15T's standardized categories:

```typescript
const BC_TO_C15T_CONSENT_CATEGORY_MAP = {
  ESSENTIAL: 'necessary',
  UNKNOWN: 'necessary', // Matches Stencil behavior of always loading "UNKNOWN" scripts
  FUNCTIONAL: 'functionality',
  ANALYTICS: 'measurement',
  TARGETING: 'marketing',
}
```

It's also worth noting that we maintain feature parity with Stencil in determining which scripts load based on consent preferences:
1. If the user has not provided any consent preference (cookie banner still visible): Only `ESSENTIAL` and `UNKNOWN` scripts are loaded.
2. If the user has "Accepted All" consent categories, all scripts load
3. If the user has "Rejected All" consent categories, only `ESSENTIAL` and `UNKNOWN` scripts load.

## Testing

Demo video:
https://github.com/user-attachments/assets/7f859c23-8419-46e7-b86e-a1e24127879a

## Migration

This PR modifies the following files that you may encounter merge conflicts with:

1. **`core/app/[locale]/layout.tsx`**:
   - Added `ScriptsFragment` import and inclusion in `RootLayoutMetadataQuery`
   - Added `scriptsTransformer` import
   - Transformed scripts are now passed to `<ConsentManager scripts={scripts}>`
   - **Resolution tip**: The `ConsentManager` component now requires a `scripts` prop. You'll need to query scripts using the `ScriptsFragment` and transform them via `scriptsTransformer()` before passing them to the component.

4. **`core/components/consent-manager/index.tsx`** and **`consent-providers.tsx`**:
   - Both now accept and forward a `scripts` prop
   - `ConsentManagerProvider` passes scripts to `ClientSideOptionsProvider`
   - **Resolution tip**: Ensure the `scripts` prop flows through: `ConsentManager` → `ConsentManagerProvider` → `ClientSideOptionsProvider`

5. **New files added**:
   - `core/components/consent-manager/scripts-fragment.ts` - GraphQL fragment for querying scripts
   - `core/data-transformers/scripts-transformer.ts` - Transforms BC scripts to C15T format
   - **Resolution tip**: These files should not conflict, but if you've added similar functionality, consider consolidating the logic.

If you've customized your consent manager implementation, ensure that:
- Scripts are queried in your layout/page where the ConsentManager is rendered
- The `scriptsTransformer` is applied to the raw script data before passing to ConsentManager
- The scripts prop is threaded through all consent manager components


core/app/[locale]/layout.tsx
```diff
@@ -124,36 +123,26 @@ export default async function RootLayout({ params, children }: Props) {
...
-          scripts={rootData.data.site.content.headerScripts}
-          strategy="afterInteractive"
-        />
-      </head>
       <body className="flex min-h-screen flex-col">
         <NextIntlClientProvider>
-          <ConsentManagerProvider>
+          <ConsentManager scripts={scripts}>
             <NuqsAdapter>
               <StreamableAnalyticsProvider data={streamableAnalyticsData} />
```
Comment (chanceaclark) [inline core/app/[locale]/layout.tsx]: 🍹 Might need to consider where this analytics provider needs to live in this component tree. I would just test around to see.

core/data-transformers/scripts-transformer.ts
```diff
@@ -0,0 +1,74 @@
...
+): 'necessary' | 'functionality' | 'marketing' | 'measurement' | 'experience' {
+  if (isValidConsentCategory(bigCommerceCategory)) {
+    return BC_TO_C15T_CONSENT_CATEGORY_MAP[bigCommerceCategory];
+  }
+
+  // eslint-disable-next-line no-console
+  console.warn(
+    `[scripts-transformer] Unknown BigCommerce Script Consent Category ` +
+      `"${bigCommerceCategory}", defaulting to UNKNOWN`,
+  );
```
chanceaclark: This will probably log in production. We probably want to remove this.

core/app/[locale]/layout.tsx
Comment (jorgemoya) [reply core/app/[locale]/layout.tsx]: Definitely nested within `ConsentManager` since that will eventually load the gtag script required for the GA4 provider.

core/data-transformers/scripts-transformer.ts
jorgemoya: Remove or just run when not in prod?

core/components/consent-manager/scripts-fragment.ts
```diff
@@ -0,0 +1,26 @@
...
+    scripts(first: 50, filters: { visibilities: [ALL_PAGES, STOREFRONT] }) {
+      edges {
+        node {
+          __typename
+          integrityHashes {
+            hash
+          }
+          entityId
+          consentCategory
+          location
```
jorgemoya: Should we remove `location` since we don't use it?

core/components/consent-manager/scripts-fragment.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export const ScriptsFragment = graphql(`
+  fragment ScriptsFragment on Content {
+    scripts(first: 50, filters: { visibilities: [ALL_PAGES, STOREFRONT] }) {
+      edges {
+        node {
+          __typename
+          integrityHashes {
+            hash
+          }
```
jorgemoya, line 11: Are we still using `integrityHashes` for anything?

core/data-transformers/scripts-transformer.ts
chanceaclark: I'd just remove here, otherwise we'd need to add some sort of invariant logger here.

core/components/consent-manager/scripts-fragment.ts
chanceaclark: We should be.

core/app/[locale]/layout.tsx
Comment (matthewvolk) [reply core/app/[locale]/layout.tsx]: I agree, `<ConsentManager />` needs to be inside of `<NextIntlClientProvider />` for translations, and `<StreamableAnalyticsProvider />` should be inside `<ConsentManager />` like Jorge mentioned.

core/components/consent-manager/scripts-fragment.ts
matthewvolk: Yeah good catch, we can always add it back later when the library adds support (which they are considering)

core/components/consent-manager/scripts-fragment.ts
matthewvolk: Good catch! I added a fix that
1. `.map()`'s over `site.content.scripts.edges.node.integrityHashes`
2. Filters out falsy (null)
3. If resulting array contains hashes, join them with a single space (following HTML spec) 
4. Add them to the [arbitrary `attributes` object](https://c15t.com/docs/frameworks/next/script-loader#types) to be added to the `<script>` element when rendered

PR #2660, opened 2025-10-31
Sync integrations makeswift

chanceaclark:
## What/Why?
Syncs the `canary` branch with the `integrations/makeswift` branch.

> [!NOTE]
> Keep in mind that there is a change to the CLI in this merge, it shouldn't affect anything regarding the `alpha` branch because it was before we released it. 

## Testing
See each PR

## Migration
Pull in the latest tagged code when released.


matthewvolk (approved):
This looks good to me—spot checked for a couple files that usually get written to incorrectly during these sync's, pulled down the branch locally and did some manual testing with Makeswift 👍

PR #2668, opened 2025-11-04
feat: add backorder messages to the PDP

Tharaae:
## What/Why?
On product details page, add the following backorder messages based on the store inventory settings and the product backorders data:
- Backorder availability prompt (notifying the shopper that there are more available of the product that the current stock to order)
- Quantity on backorder (notifying the shopper of how many of the quantity selected will be on backorder if any)
- Backorder message (more information to shopper about the backordered product like expected processing time for example)

## Testing
Screen records
- **LATEST: Final UI updates**
https://github.com/user-attachments/assets/c9bd306b-3075-4c87-a923-8d6b777eca6d

- **When the product has unlimited backorders limit**

https://github.com/user-attachments/assets/547509ed-0a8c-4c91-8020-f74be1bce959

- **When the product has a backorder limit**

https://github.com/user-attachments/assets/c6fe6062-2f47-4f02-a9cd-c892457e7770

- **When the product backordered quantity is equal to the backorder limit**

https://github.com/user-attachments/assets/c0942fa2-0b94-40a5-bf15-bfd3a270541a

- **When the product has no backorder limit**

https://github.com/user-attachments/assets/fffdb6b1-e252-4d28-b829-c602f341bb2c

- **When store backorder message display is not enabled**

https://github.com/user-attachments/assets/57616efb-e1ca-4609-8106-7e333f9432d9

- **When the store backorder quantity display is not enabled**

https://github.com/user-attachments/assets/b1e65fd8-689f-4419-bd2f-412d68ce4383

- **When the store availability prompt display is not enabled**

https://github.com/user-attachments/assets/ea7d1483-8d2c-417a-8b19-d890b5207b2c

## Migration
While rebasing, there might be some conflicts if there is some custom styling of current stock message and/or selected quantity input on PDP. Those conflicts may occur in the following files:
- core/vibes/soul/sections/product-detail/index.tsx
- core/vibes/soul/sections/product-detail/product-detail-form.tsx

### DIFF
--- a/core/app/[locale]/(default)/product/[slug]/page.tsx
+++ b/core/app/[locale]/(default)/product/[slug]/page.tsx
@@ -283,16 +283,22 @@ export default async function Product({ params, searchParams }: Props) {
       }
     }
 
+    const availabilityMessage = getBackorderAvailabilityPrompt({
+      showBackorderAvailabilityPrompt,
+      backorderAvailabilityPrompt,
+      availableForBackorder,
+      unlimitedBackorder,
+    });
+
+    if (!availabilityMessage && stockQuantity === undefined) {
+      return null;
+    }
+
     return {
       stockLevelMessage: t('ProductDetails.currentStock', {
         quantity: stockQuantity ?? 0,
       }),
-      backorderAvailabilityPrompt: getBackorderAvailabilityPrompt({
-        showBackorderAvailabilityPrompt,
-        backorderAvailabilityPrompt,
-        availableForBackorder,
-        unlimitedBackorder,
-      }),
+      backorderAvailabilityPrompt: availabilityMessage,
     };
   });
 


Tharaae:
> Thanks @Tharaae for contributing again back to Catalyst!
> Before we review, mind fixing the linting issue reported by CI?

Hi @chanceaclark ... the issue is due to a broken GQL type check because the new GQL fields used in this PR are made behind an experiment. We already have [a PR to remove the API experiment check](https://github.com/bigcommerce/storefront/pull/4160) that is in code review. Once it's merged and hit production, the type checks should pass then.

core/app/[locale]/(default)/product/[slug]/page.tsx
```diff
@@ -216,28 +224,137 @@ export default async function Product({ params, searchParams }: Props) {
...
+    // if no backorder info is to be displayed, then availableToSell is the stock quantity to be used
+    if (!showBackorderAvailabilityPrompt && !showBackorderMessage && !showQuantityOnBackorder) {
+      // availableToSell can be 0 or undefined while the product is in stock if backorder is UNLIMITED
+      if (!availableToSell) {
         return null;
       }
 
-      if (availableToSell && availableToSell > warningLevel) {
-        return null;
+      if (stockLevelDisplay === 'SHOW_WHEN_LOW') {
```
Comment (animesh1987) [inline core/app/[locale]/(default)/product/[slug]/page.tsx]: Should we use an enum for this?

core/vibes/soul/sections/product-detail/product-detail-form.tsx
```diff
@@ -169,7 +240,17 @@ export function ProductDetailForm<F extends Field>({
               {error}
             </FormStatus>
           ))}
-          <div className="flex gap-x-3 pt-3">
+
+          <div className="flex flex-wrap justify-start gap-x-6 gap-y-2 text-sm text-[var(--product-detail-secondary-text,hsl(var(--contrast-500)))]">
```
animesh1987: Should we wrap this div in backorder information, or we may have an empty div rendered?

core/app/[locale]/(default)/product/[slug]/_actions/get-backorder-messages.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  const t = await getTranslations('Product.ProductDetails');
+
+  return {
+    backorderQuantityMessage: t('backorderQuantity', {
+      quantity: unlimitedBackorder
+        ? orderQuantity - availableOnHand
+        : Math.min(orderQuantity - availableOnHand, availableForBackorder),
+    }),
+    backorderInfoMessage: backorderMessage,
```
Comment (animesh1987) [inline core/app/[locale]/(default)/product/[slug]/_actions/get-backorder-messages.tsx]: Do we need to pass `backorderMessage` in this method? Given we just remap it to a different property and return it?

core/app/[locale]/(default)/product/[slug]/page.tsx
```diff
@@ -216,28 +224,137 @@ export default async function Product({ params, searchParams }: Props) {
...
+      return null;
+    }
+
+    if (!product.variants.edges[0]?.node?.inventory?.byLocation?.edges) {
+      return null;
+    }
+
+    return (
+      product.variants.edges[0].node.inventory.byLocation.edges[0]?.node.backorderMessage || null
+    );
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/product/[slug]/page.tsx]: We have a utility function called `removeEdgesAndNodes` you should try to use instead of attempting to do all of these conditionals. It would be best to also do this within `page-data.ts` so the page component doesn't care about removing edges and nodes.

core/app/[locale]/(default)/product/[slug]/page.tsx
```diff
@@ -370,8 +488,14 @@ export default async function Product({ params, searchParams }: Props) {
             minQuantity: streamableMinQuantity,
             maxQuantity: streamableMaxQuantity,
             stockLevelMessage: streamableStockLevelMessage,
+            availableOnHand: streamableAvailableOnHand,
+            availableForBackorder: streamableAvailableForBackorder,
+            unlimitedBackorder: streamableUnlimitedBackorder,
+            backorderAvailabilityPrompt: streamableAvailabilityPrompt,
+            backorderMessage: streamableBackorderMessage,
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/product/[slug]/page.tsx]: I'm just throwing this idea out there as something to explore to see if it feels like a better approach, but what about passing in an object for stock related display values? I'm thinking that this is pretty similar to the `streamablePrices` return above, where we handle displaying the price internally shifting responsibility lower into the component.

core/app/[locale]/(default)/product/[slug]/_actions/get-backorder-messages.tsx
Comment (chanceaclark) [inline core/app/[locale]/(default)/product/[slug]/_actions/get-backorder-messages.tsx:1]: A server action is not a good tool for returning a simple payload. Can we have it as a client-side hook instead? There should be a `useTranslations` hook. This might be able to remove the `useState` usage in the component too.

core/app/[locale]/(default)/product/[slug]/page.tsx
Comment (Tharaae) [reply core/app/[locale]/(default)/product/[slug]/page.tsx]: The enum is already defined via the GQL data types and the type check is already done using those types. We cannot refer directly to the QGL enum, but if we use a value that does not exist in the enum (like `if (stockLevelDisplay === 'NOT_IN_ENUM')`), the type check will fail and the IDE will complain.

JessicaHu51:
Hi @Tharaae , I’ve caught up with @andrewreifman and @chanceaclark  to clarify the final change:

- reduce the gap between the “Current stock” row and “Backorder quantity” row to half of what it is in the latest version.
- Keep the spacing for the “Backorder quantity” row even when it’s hidden, so the quantity input row stays in a fixed position.

Once this change is made, could you please attach the updated video? After that, we should be good to move forward. Thanks

core/app/[locale]/(default)/product/[slug]/page.tsx
Comment (Tharaae) [reply core/app/[locale]/(default)/product/[slug]/page.tsx]: Thanks for the head-up @chanceaclark 🙏  ... I didn't know about such method.

core/vibes/soul/sections/product-detail/product-detail-form.tsx
Tharaae: Yes, with the changes requested by Chancellor, this will make more sense.

core/app/[locale]/(default)/product/[slug]/_actions/get-backorder-messages.tsx
Comment (Tharaae) [reply core/app/[locale]/(default)/product/[slug]/_actions/get-backorder-messages.tsx]: Applying Chancellor's comments, we will get rid of the whole method and move the assignments logic to the client.

core/app/[locale]/(default)/product/[slug]/_actions/get-backorder-messages.tsx
Comment (Tharaae) [reply core/app/[locale]/(default)/product/[slug]/_actions/get-backorder-messages.tsx:1]: Your approach makes more sense to me, but it will be the first usage of `useTranslations` in _vibes_ folder. Probably this is why I was under the impression that `useTranslations` should be used on server only.

core/app/[locale]/(default)/product/[slug]/page.tsx
Comment (Tharaae) [reply core/app/[locale]/(default)/product/[slug]/page.tsx]: If you mean the GQL response for those fields to be contained in an object of the own the same way `prices` field encapsulates all pricing data, we have some of them  (all the quantities) already encapsulated `inventory.aggregated` object. The other settings fields are either on variant/location level or on store level.

If you mean encapsulating related components props, I've done a related refactor and push it on a separate commit.

Tharaae:
Hi @chanceaclark ... Thanks for the review. Regarding [your suggested refactor here](https://github.com/bigcommerce/catalyst/commit/0eadd37e109f11e287eb86ba7589e1e614c6acb4), I applied some of it in a separate commit, and added some comments on your branch for the other parts that are not applied.

core/app/[locale]/(default)/product/[slug]/page.tsx
```diff
@@ -228,16 +245,112 @@ export default async function Product({ params, searchParams }: Props) {
...
       }
     }
 
     return t('ProductDetails.currentStock', {
-      quantity: availableToSell,
+      quantity: stockQuantity ?? 0,
     });
   });
 
+  const streamableStockDisplayData = Streamable.from(async () => {
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/product/[slug]/page.tsx]: Let's combine `streamableStockLevelMessage` into this function so we reduce how many streamable functions we have.

core/app/[locale]/(default)/product/[slug]/page.tsx
```diff
@@ -228,16 +245,112 @@ export default async function Product({ params, searchParams }: Props) {
...
+    const { showBackorderAvailabilityPrompt, backorderAvailabilityPrompt } = inventorySetting ?? {};
+    const { availableForBackorder, unlimitedBackorder } = product.inventory.aggregated ?? {};
+    const hasBackorderAvailablity = !!availableForBackorder || unlimitedBackorder;
+
+    if (
+      !showBackorderAvailabilityPrompt ||
+      !backorderAvailabilityPrompt ||
+      !product.inventory.isInStock ||
+      !hasBackorderAvailablity
+    ) {
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/product/[slug]/page.tsx]: Couldn't find a good place to put this comment, but if we can, we should be able to abstract some conditional code into utility functions to reduce climatic complexity.

core/vibes/soul/sections/product-detail/product-detail-form.tsx
```diff
@@ -169,7 +227,39 @@ export function ProductDetailForm<F extends Field>({
               {error}
             </FormStatus>
           ))}
-          <div className="flex gap-x-3 pt-3">
+
+          <div className="h-[3.2rem] sm:h-[2.6rem]">
+            {!!stockDisplayData?.stockLevelMessage && (
+              <div
+                className={`flex ${backorderMessages?.backorderQuantityMessage || backorderMessages?.backorderInfoMessage ? 'translate-y-0' : 'translate-y-[calc(100%+4px)]'} flex-wrap justify-start gap-x-2.5 gap-y-2 text-sm text-[var(--product-detail-secondary-text,hsl(var(--contrast-500)))] transition-transform duration-200 ease-in-out`}
```
chanceaclark: Can we explore `clsx` for conditional classes here?

animesh1987 (approved):

PR #2669, opened 2025-11-04
chore: CATALYST-676 add sentry to demo site

chanceaclark:
## What/Why?

This pull request adds comprehensive error tracking and performance monitoring to Catalyst using Sentry. This integration provides production-ready observability for both server-side and client-side errors, helping teams quickly identify and debug issues in their storefronts.

The implementation is fully opt-in via environment variables, meaning it has zero impact on stores that don't configure Sentry. When enabled, it provides:

- **Error Tracking**: Automatic capture of unhandled errors on both client and server
- **Performance Monitoring**: Distributed tracing across server, edge, and client runtimes
- **Source Maps**: Upload of source maps for debugging minified production code
- **Global Error Boundary**: User-friendly error UI with automatic error reporting

### Technical Details

**New Dependencies:**

- `@sentry/nextjs` v10.22.0 added to `core/package.json`

**Configuration Files Created:**

1. `instrumentation.ts` - Main instrumentation with `register()` and `onRequestError()` hooks
2. `instrumentation-client.ts` - Client-side initialization with router transition tracking
3. `sentry.server.config.ts` - Server runtime configuration (Node.js)
4. `sentry.edge.config.ts` - Edge runtime configuration (middleware, edge API routes)
5. `app/global-error.tsx` - Global error boundary component

**Build Configuration:**

- Updated `next.config.ts` to conditionally apply Sentry webpack plugin when `SENTRY_AUTH_TOKEN` is present
- Added `.env.sentry-build-plugin` to `.gitignore`

**Environment Variables Required (all optional):**

- `NEXT_PUBLIC_SENTRY_DSN` - Project DSN from Sentry (enables error tracking)
- `SENTRY_AUTH_TOKEN` - Auth token for source map uploads (enables debugging)
- `SENTRY_ORG` - Organization slug (for source maps)
- `SENTRY_PROJECT` - Project slug (for source maps)

**Key Implementation Features:**

- Conditional initialization - only runs when `NEXT_PUBLIC_SENTRY_DSN` is set
- Integration with existing OpenTelemetry instrumentation
- Separate configurations for server and edge runtimes
- Client-side router instrumentation for navigation tracking
- 100% trace sampling (configurable via `tracesSampleRate`)
- PII enabled (configurable via `sendDefaultPii`)

## Testing

Screenshot 2025-11-04 at 16 52 05

## Migration

No migration required. This is a new opt-in feature that requires no changes to existing code or configuration. Stores that don't set up Sentry environment variables will see no change in behavior.

**For stores wanting to enable Sentry:**

1. Create a Sentry account and project
2. Add the required environment variables to your deployment environment
3. Redeploy the application
4. Monitor errors in the Sentry dashboard

**Production Configuration Recommendations:**

- Adjust `tracesSampleRate` based on traffic volume (default 1.0 = 100%)
- Review `sendDefaultPii` setting based on privacy requirements
- Configure release tracking for better error grouping
- Set up alerts for critical errors

---

**This pull request description was generated with the assistance of AI. Portions of the code and/or implementation ideas in this PR may also have been created or influenced by AI tools.**

### DIFF
--- a/core/sentry.edge.config.ts
+++ b/core/sentry.edge.config.ts
@@ -1,0 +1,14 @@
+// This file configures the initialization of Sentry for edge features (middleware, edge routes, and so on).
+// The config you add here will be used whenever one of the edge features is loaded.
+// Note that this config is unrelated to the Vercel Edge Runtime and is also required when running locally.
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from '@sentry/nextjs';
+
+if (process.env.NEXT_PUBLIC_SENTRY_DSN) {
+  Sentry.init({
+    dsn: process.env.NEXT_PUBLIC_SENTRY_DSN,
+    tracesSampleRate: 1,
+    sendDefaultPii: true,
+  });
+}
--- a/core/.gitignore
+++ b/core/.gitignore
@@ -54,3 +54,6 @@ build-config.json
 # OpenNext
 .open-next
 .wrangler
+
+# Sentry Config File
+.env.sentry-build-plugin
--- a/core/app/global-error.tsx
+++ b/core/app/global-error.tsx
@@ -1,0 +1,26 @@
+'use client';
+
+import * as Sentry from '@sentry/nextjs';
+import NextError from 'next/error';
+import { useEffect } from 'react';
+
+export default function GlobalError({ error }: { error: Error & { digest?: string } }) {
+  useEffect(() => {
+    if (process.env.NEXT_PUBLIC_SENTRY_DSN) {
+      Sentry.captureException(error);
+    }
+  }, [error]);
+
+  return (
+    // eslint-disable-next-line jsx-a11y/html-has-lang
+    <html>
+      <body>
+        {/* `NextError` is the default Next.js error page component. Its type
+        definition requires a `statusCode` prop. However, since the App Router
+        does not expose status codes for errors, we simply pass 0 to render a
+        generic error message. */}
+        <NextError statusCode={0} />
+      </body>
+    </html>
+  );
+}
--- a/core/instrumentation-client.ts
+++ b/core/instrumentation-client.ts
@@ -1,0 +1,22 @@
+// This file configures the initialization of Sentry on the client.
+// The added config here will be used whenever a users loads a page in their browser.
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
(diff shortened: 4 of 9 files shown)


core/sentry.edge.config.ts
```diff
@@ -0,0 +1,14 @@
+// This file configures the initialization of Sentry for edge features (middleware, edge routes, and so on).
+// The config you add here will be used whenever one of the edge features is loaded.
+// Note that this config is unrelated to the Vercel Edge Runtime and is also required when running locally.
+// https://docs.sentry.io/platforms/javascript/guides/nextjs/
+
+import * as Sentry from '@sentry/nextjs';
+
+if (process.env.NEXT_PUBLIC_SENTRY_DSN) {
```
migueloller, line 8: Is it necessary to conditionally initialize Sentry here given that we only use the Sentry plugin (i.e., `withSentryConfg`) if the env var is present? The same applies to `sentry.server.config.ts`

chanceaclark: I'm trying here to reduce the footguns on setting up Sentry here. I am trying to safeguard registering Sentry if a few things are not set. Just using the plugin in `next.config.ts` is not safe enough as they utilize different vars.

migueloller: What I'm trying to get at is that these files aren't imported anywhere in the application, they're inserted by the Next.js plugin and thus the env var is redundant. These files simply won't be included in the bundles at all because of the conditional in the Next.js config.

migueloller: Or put a different way, if somebody has the wrong env var such that the Next.js plugin is including these files but somehow they have the other env var set to something else, I'd rather they get a runtime error because they're missing the Sentry DNS or something than completely miss that they're loading in the Sentry Next.js plugin.

chanceaclark:
>I wonder if we can fix this.

I think what I was seeing was a Sentry-specific issue, Jaeger and Zipkin showed a lot more information here. I also think Sentry had some issues trying to do custom spans as I never saw them populate in the dashboard. I think this is a non-trivial fix which may not be worth our time right now.

---

I forgot to mention that any sort of solution here Next.js has automatic instrumentation for pretty close to most things, including server actions:
>Next.js automatically instruments several spans for you to provide useful insights into your application's performance.
Screenshot 2025-11-05 at 14 42 07

I'll spin up another PR for the generic approach I mentioned above and we can make a decision from there what path we want to do.

PR #2670, opened 2025-11-05
fix(core): Update with-routes to account for trailing slash or capitalizations when redirecting

jordanarldt:
## What/Why?
Update the with-routes middleware to account for the `TRAILING_SLASH` env var, as well as capitalized letters in 301 Redirects.

There was a reported issue where when using `TRAILING_SLASH=false` in the Catalyst environment variables, redirect loops could occur if you had a redirect going from `/example-path -> /example-path/`, or capitalizations like `/some-url -> /some-URL`

## Testing
Tested locally, confirmed fix works.

### Regression test results before changes:
image

### Regression test results after changes:
image

### Before fix
**Dynamic 301 Redirect from `/infinite-redirect-test` to a product with the URL `/infinite-redirect-test/` while `TRAILING_SLASH=false`**

https://github.com/user-attachments/assets/de33a638-88dd-470b-8dc5-7278f4395c7d

**Manual 301 Redirect from `/redirect-test` to `/redirect-TEST`**

https://github.com/user-attachments/assets/d5ae1973-c2e9-49f6-90ab-a42747388889

### After fix

**Dynamic 301 Redirect from `/infinite-redirect-test` to a product with the URL `/infinite-redirect-test/` while `TRAILING_SLASH=false`**

https://github.com/user-attachments/assets/1398df93-92a9-4a7f-8370-2fdff2c871d0

**Manual 301 Redirect from `/redirect-test` to `/redirect-TEST`**

https://github.com/user-attachments/assets/963d71b6-f0f7-43a7-94cc-d6ce72c47edb

## Migration
Copy changes from `with-routes.ts` using this PR as a reference

### DIFF
--- a/core/middlewares/with-routes.ts
+++ b/core/middlewares/with-routes.ts
@@ -12,6 +12,8 @@ import { kv } from '../lib/kv';
 
 import { type MiddlewareFactory } from './compose-middlewares';
 
+const trailingSlashDisabled = process.env.TRAILING_SLASH === 'false';
+
 const GetRouteQuery = graphql(`
   query GetRouteQuery($path: String!) {
     site {
@@ -223,6 +225,21 @@ const clearLocaleFromPath = (path: string, locale: string) => {
   return path;
 };
 
+function normalizeForCompare(url: URL): string {
+  if (trailingSlashDisabled && url.pathname !== '/' && url.pathname.endsWith('/')) {
+    return `${url.pathname.replace(/\/+$/, '')}${url.search}`;
+  }
+
+  if (!trailingSlashDisabled && !url.pathname.endsWith('/')) {
+    return `${url.pathname}/${url.search}`;
+  }
+
+  return `${url.pathname}${url.search}`;
+}
+
+const sameInternalUrl = (a: URL, b: URL) =>
+  a.origin === b.origin && normalizeForCompare(a) === normalizeForCompare(b);
+
 const getRouteInfo = async (request: NextRequest, event: NextFetchEvent) => {
   const locale = request.headers.get('x-bc-locale') ?? '';
   const channelId = request.headers.get('x-bc-channel-id') ?? '';
@@ -269,6 +286,7 @@ const getRouteInfo = async (request: NextRequest, event: NextFetchEvent) => {
 };
 
 export const withRoutes: MiddlewareFactory = () => {
+  // eslint-disable-next-line complexity
   return async (request, event) => {
     const locale = request.headers.get('x-bc-locale') ?? '';
 
@@ -303,6 +321,10 @@ export const withRoutes: MiddlewareFactory = () => {
           // For dynamic redirects, assume an internal redirect and construct the URL from the path
           const redirectUrl = new URL(route.redirect.to.path + searchParams, request.url);
 
+          if (sameInternalUrl(request.nextUrl, redirectUrl)) {
+            break;
+          }
+
           return NextResponse.redirect(redirectUrl, redirectConfig);
         }
 
@@ -314,6 +336,10 @@ export const withRoutes: MiddlewareFactory = () => {
 
           if (redirectUrl.origin === request.nextUrl.origin) {
             redirectUrl.search = searchParams;
+
+            if (sameInternalUrl(request.nextUrl, redirectUrl)) {
+              break;
+            }
           }
 
           return NextResponse.redirect(redirectUrl, redirectConfig);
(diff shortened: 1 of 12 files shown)


core/middlewares/with-routes.ts
```diff
@@ -223,6 +223,23 @@ const clearLocaleFromPath = (path: string, locale: string) => {
...
+  if (trailingSlashDisabled && url.pathname !== '/' && url.pathname.endsWith('/')) {
+    return `${url.pathname.replace(/\/+$/, '')}${url.search}`;
+  }
+
+  if (!trailingSlashDisabled && !url.pathname.endsWith('/')) {
+    return `${url.pathname}/${url.search}`;
+  }
+
+  return `${url.pathname}${url.search}`;
+}
```
migueloller, line 238: 🍹 Is this normalization necessary? I would've thought that comparing the request path to the destination path would've been enough to detect an infinite redirect. Is it because we handle the trailing slash ourselves in a middleware that's later in the chain (or perhaps before)?

core/middlewares/with-routes.ts
```diff
@@ -223,6 +223,23 @@ const clearLocaleFromPath = (path: string, locale: string) => {
   return path;
 };
 
+function normalizeForCompare(url: URL): string {
+  const trailingSlashDisabled = process.env.TRAILING_SLASH === 'false';
```
chanceaclark: We can probably elevate this to the root of this file so we don't redefine this variable each time.

core/middlewares/with-routes.ts
chanceaclark, line 238: I'm just posting my AI response here. I prefer the flow of code + handling the lowercasing of mine, but yours deals with the search params and utilizes a `URL` object. Might want to pick and choose some things between the two.

```ts
const trailingSlashEnabled = process.env.TRAILING_SLASH !== 'false';

const normalizeUrlForComparison = (path: string): string => {
  let normalized = path.toLowerCase();

  if (trailingSlashEnabled) {
    if (normalized !== '/' && !normalized.endsWith('/')) {
      normalized += '/';
    }

    return normalized;
  }

  if (normalized !== '/' && normalized.endsWith('/')) {
    normalized = normalized.slice(0, -1);
  }

  return normalized;
};
```

migueloller: I'm not sure we should be normalizing the casing here since URLs are case sensitive. As long as the request path doesn't match the redirect `Location` header, then we don't get into an infinite loop. I think it should be enough for us to compare the two right before calling `NextResponse.redirect`.

jordanarldt: @chanceaclark 
if a product URL is uppercase specifically like `/my-PRODUCT`, we should reflect that instead of redirecting to `/my-product`. Case sensitivity was part of the original issue so I figured it would make sense to take casing into account.

chanceaclark: That makes sense!

jordanarldt: @migueloller Yes, it's because of the `TRAILING_SLASH` environment variable. In the getRoutes query, the destination path might be `/my-path/` (with trailing slash), but your location might be `/my-path`.

Without this, `/my-path` would redirect to `/my-path/` and then NextJS (TRAILING_SLASH=false) will redirect to `/my-path`, then get redirected to `/my-path/` and so-on. That's why this issue was occurring

core/tests/fixtures/utils/api/redirects/index.ts
```diff
@@ -0,0 +1,30 @@
...
+  readonly id: number;
+  readonly siteId: number;
+  readonly fromPath: string;
+  readonly to: RedirectToEntity | RedirectToUrl;
+}
+
+export interface UpsertRedirectData {
+  fromPath: string;
+  to: RedirectToEntity | RedirectToUrl;
+}
```
chanceaclark, line 23: Do we need some of these types if we already have the zod schema?

migueloller (approved):
The tests are looking great! Thanks @jordanarldt 🙏🏻

core/tests/fixtures/utils/api/redirects/index.ts
jordanarldt: @chanceaclark IIRC, when I was originally making these functional tests we wanted to keep some common/abstract interface so that in the future we can have different clients (http / graphQL), so these interfaces declared here will be those common types.

The zod schemas that i built are specific to the HTTP client 👍

chanceaclark (approved):

PR #2676, opened 2025-11-10
feat(reviews): add reviews form enabling shoppers to submit reviews

matthewvolk:
## What/Why?

Adds the ability for customers to submit product reviews directly from the product detail page. This implementation includes a modal form with validation for rating, title, review text, name, and email fields. The form integrates with BigCommerce's GraphQL API to create reviews and provides real-time validation feedback using Conform and Zod.

Includes:
* The new `ProductReviewButton` component appears in both the empty state and populated reviews sections, allowing customers to easily contribute feedback regardless of whether existing reviews are present.
* Clicking the review button triggers a Modal containing the review form
* Does not include reCAPTCHA
* Email is required until we can read "only allow reviews from previous customers" from GQL
* Error responses from GQL rendered after form submission
* If successfully submitted, show a toast and close the modal
* If user is logged in, read email and name from auth JWT and pre-fill the review form with those values

## Testing
Happy Path:

https://github.com/user-attachments/assets/56ee3b47-53ae-4fb5-944a-b9f33a65c6a0

Error Response:

https://github.com/user-attachments/assets/ef6e78c5-3024-4b13-9336-eb9d3f1038db

## Migration

- **core/messages/en.json** - Added `Product.Reviews.Form` object with 9 new translation keys (button, title, submit, ratingLabel, titleLabel, reviewLabel, nameLabel, emailLabel, successMessage, somethingWentWrong)
- **core/app/[locale]/(default)/product/[slug]/\_components/reviews.tsx** - Added imports: `auth`, `submitReview`, `getStreamableProduct`. Props interface expanded with `streamableImages` and `streamableProduct`. Component creates `streamableProductName` (extracts product name) and `streamableUser` (fetches/obfuscates user info as "FirstName L."). Passes form-related props to `<ReviewsSection>` and `<ReviewsEmptyState>`.
- **core/vibes/soul/sections/reviews/index.tsx** - Props interface expanded with `productId`, `action`, and 8 form label props plus 3 streamable props. Added `<ProductReviewButton>` component in two locations: inside `StickySidebarLayout` sidebar and inside `ReviewsEmptyState`.
- **core/app/[locale]/(default)/product/[slug]/page.tsx** - `<Reviews>` component now receives `streamableImages={streamableImages}` and `streamableProduct={streamableProduct}` props.
- **core/vibes/soul/primitives/modal/index.tsx** - Modal header className changed: removed `mb-5` and `border-b border-b-contrast-200`.
- **core/vibes/soul/primitives/rating/index.tsx** - Changed `const Star` to `export const Star` to allow external usage.

---

Jira: [CATALYST-1563](https://bigcommercecloud.atlassian.net/browse/CATALYST-1563)

Note: This pull request description was generated with the assistance of AI.

[CATALYST-1563]: https://bigcommercecloud.atlassian.net/browse/CATALYST-1563?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx
+++ b/core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx
@@ -148,34 +148,17 @@ export const Reviews = async ({
 
   const streamableUser = Streamable.from(async () => {
     const session = await auth();
-    const fullName = session?.user?.name ?? '';
-    let obfuscatedName = '';
+    const firstName = session?.user?.firstName ?? '';
+    const lastName = session?.user?.lastName ?? '';
 
-    if (fullName === '') {
-      obfuscatedName = '';
-    } else {
-      const parts = fullName.trim().split(/\s+/);
-
-      if (parts.length < 2) {
-        obfuscatedName = fullName;
-      } else {
-        const firstName = parts[0];
-        const lastName = parts[parts.length - 1];
-
-        if (!lastName) {
-          obfuscatedName = fullName;
-        } else {
-          const lastInitial = lastName.charAt(0).toUpperCase();
-
-          obfuscatedName = `${firstName} ${lastInitial}.`;
-        }
-      }
+    if (!firstName || !lastName) {
+      return { email: session?.user?.email ?? '', name: '' };
     }
 
-    return {
-      email: session?.user?.email ?? '',
-      name: obfuscatedName,
-    };
+    const lastInitial = lastName.charAt(0).toUpperCase();
+    const obfuscatedName = `${firstName} ${lastInitial}.`;
+
+    return { email: session?.user?.email ?? '', name: obfuscatedName };
   });
 
   return (


core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx
```diff
@@ -129,16 +140,67 @@ export const Reviews = async ({ productId, searchParams }: Props) => {
...
+        formModalTitle={t('Form.title')}
+        formNameLabel={t('Form.nameLabel')}
+        formRatingLabel={t('Form.ratingLabel')}
+        formReviewLabel={t('Form.reviewLabel')}
+        formSubmitLabel={t('Form.submit')}
+        formTitleLabel={t('Form.titleLabel')}
         nextLabel={t('next')}
         paginationInfo={streamablePaginationInfo}
         previousLabel={t('previous')}
+        productId={productId}
```
Comment (matthewvolk) [inline core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx:181]: Instead of drilling `productId` down from the page level just to include it as a hidden input on the review form, I'm open to binding it to the server action instead. Not sure what pattern we prefer

core/vibes/soul/form/rating-radio-group/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+        {label !== undefined && label !== '' && (
+          <Label colorScheme={colorScheme} id={groupId}>
+            {label}
+          </Label>
+        )}
+        <RadioGroupPrimitive.Root
+          {...rest}
+          aria-labelledby={groupId}
+          className="flex items-center gap-1"
+          onMouseDown={handleMouseDown}
```
matthewvolk, line 85: The reason these mouse handlers + the mouse down ref is needed is because I needed to prevent the flash of "preview stars". I am very open to better ways to do this, I don't think this is the cleanest solution

https://github.com/user-attachments/assets/e760ee24-afc0-4591-8a52-887823b65304

core/vibes/soul/sections/reviews/review-form.tsx
```diff
@@ -0,0 +1,213 @@
...
+              label={formRatingLabel}
+              max={5}
+              name={fields.rating.name}
+              onBlur={ratingControl.blur}
+              onFocus={ratingControl.focus}
+              // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+              onValueChange={ratingControl.change}
+              required={fields.rating.required}
+              value={typeof ratingControl.value === 'string' ? ratingControl.value : ''}
+            />
```
matthewvolk, line 167: Should I just make all the form input fields controlled as well? Not sure what pattern we prefer

matthewvolk: Also... should I rename this file + component `review-form.tsx`/`<ReviewForm />`?

core/app/[locale]/(default)/product/[slug]/_actions/submit-review.ts
```diff
@@ -0,0 +1,92 @@
...
+    });
+
+    const result = response.data.catalog.addProductReview;
+
+    if (result.errors.length > 0) {
+      return {
+        ...prevState,
+        lastResult: submission.reply({ formErrors: result.errors.map(({ message }) => message) }),
+      };
+    }
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/product/[slug]/_actions/submit-review.ts:65]: 🍹 I know this is an existing pattern, but what about building an `AggregateError` instance and doing the error handling in the `catch` block? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError#browser_compatibility
```suggestion
    if (result.errors.length > 0) {
      throw AggregateError(result.errors.map(({ message }) => Error(message)))
    }
```

core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx
```diff
@@ -129,16 +140,67 @@ export const Reviews = async ({ productId, searchParams }: Props) => {
...
+          obfuscatedName = `${firstName} ${lastInitial}.`;
+        }
+      }
+    }
+
+    return {
+      email: session?.user?.email ?? '',
+      name: obfuscatedName,
+    };
+  });
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx:162]: Instead of handling all of this, can we just add the `firstName` and `lastName` fields onto `session.user`? We might also want to remove the `name` field from the session as I think we shouldn't make assumptions of building the whole `name` field (bad initial assumption on my part). If we go down this path, it should be in another PR so we can separate out the changes.

Obligatory blog post about assumptions made about names: https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/

core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx
Comment (chanceaclark) [reply core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx:181]: Either or, but I kinda prefer this as it's more simple.

core/vibes/soul/form/rating-radio-group/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+      }
+    };
+
+    return (
+      <div className={clsx('rating-radio-group space-y-2', className)}>
+        {label !== undefined && label !== '' && (
+          <Label colorScheme={colorScheme} id={groupId}>
+            {label}
+          </Label>
+        )}
```
chanceaclark: Can we make `label` required so we can enforce the accessibility here?

core/vibes/soul/form/rating-radio-group/index.tsx
chanceaclark: I think it's because the `onBlur` is being triggered

core/vibes/soul/sections/reviews/review-form.tsx
chanceaclark: I think controlled would be best within this file.

core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx
Comment (matthewvolk) [reply core/app/[locale]/(default)/product/[slug]/_components/reviews.tsx:162]: I like this! PR: https://github.com/bigcommerce/catalyst/pull/2684

PR #2679, opened 2025-11-10
chore: remove next lint command

chanceaclark:
## What/Why?

This PR updates the `lint` command in the `core` package to use the `eslint` CLI directly instead of `next lint`. With Next.js 16 deprecating `next lint`, this change proactively migrates away from the deprecated command to minimize future migration impact.

**Key changes:**
- Updated `lint` script in `core/package.json` to run `eslint . --ext .js,.jsx,.ts,.tsx` instead of `next lint`
- Added `prelint` script that runs `next typegen` to ensure TypeScript definitions are generated before linting
- Created new `.eslintignore` file to explicitly define ignored paths (dependencies, build outputs, generated files, test outputs, cache)
- Removed `ignorePatterns` from `.eslintrc.cjs` since ignore patterns are now managed in `.eslintignore`

This change maintains the same linting behavior while using the more direct ESLint CLI approach, which is the recommended pattern going forward.

## Testing

1. Run `npm run lint` from the `core` directory to verify linting works correctly
2. Run `pnpm lint` from the root directory to ensure the monorepo-level lint command still functions
3. Verify that the same files are linted as before (check that ignore patterns work correctly)
4. Confirm that TypeScript definitions are generated before linting via the `prelint` hook
5. Intentionally introduce a linting error to verify ESLint catches and reports it properly

## Migration

**For developers working on feature branches:**

No action required. The `lint` command works the same way - you can continue using `npm run lint` or `pnpm lint` as before. The implementation has changed but the interface remains identical.

**If you've customized ignore patterns:**

If you have custom ignore patterns in your local `.eslintrc.cjs`, you may need to move them to the new `.eslintignore` file instead, as we've consolidated ignore patterns there for better clarity and maintainability.

---

This pull request description was generated with the assistance of AI. Portions of the code and/or implementation ideas in this PR may also have been created or influenced by AI tools.

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -9,7 +9,8 @@
     "build": "npm run generate && next build",
     "build:analyze": "ANALYZE=true npm run build",
     "start": "next start",
-    "lint": "next lint",
+    "prelint": "next typegen",
+    "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
     "typecheck": "tsc --noEmit"
   },
   "dependencies": {
--- a/.changeset/neat-coins-stick.md
+++ b/.changeset/neat-coins-stick.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Update the `lint` command to utilize `eslint` CLI directly. `next lint` is deprecated in Next.js version 16 and this provides a lower migration impact when the time comes.
--- a/core/.eslintignore
+++ b/core/.eslintignore
@@ -1,0 +1,24 @@
+# Dependencies
+node_modules/
+
+# Build outputs
+.next/
+.wrangler/
+.open-next/
+out/
+dist/
+build/
+
+# Generated files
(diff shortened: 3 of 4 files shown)


core/package.json
```diff
@@ -9,7 +9,8 @@
     "build": "npm run generate && next build",
     "build:analyze": "ANALYZE=true npm run build",
     "start": "next start",
-    "lint": "next lint",
+    "prelint": "next typegen",
```
chanceaclark, line 12: The `next lint` command essentially ran this before linting, so I opted to have this be a `pre*` hook to the lint task.

jorgemoya: This runs automatically because it has the `pre` prefix?

chanceaclark: Yeah, it should be documented here: https://docs.npmjs.com/cli/v8/using-npm/scripts#pre--post-scripts

PR #2681, opened 2025-11-11
fix(core): CATALYST-1579 Ensure "Proceed to checkout" has loading state, respect trailing slash on checkout action path

jordanarldt:
## What/Why?
Follow-up PR for https://github.com/bigcommerce/catalyst/pull/2680 to ensure that the "Proceed to checkout" button displays a loading spinner while navigating away from the page.

## Testing

https://github.com/user-attachments/assets/0050a01b-cd4a-47d2-9760-84766da231d4

## Migration
Copy all changes from this PR

### DIFF
--- a/core/app/[locale]/(default)/cart/page.tsx
+++ b/core/app/[locale]/(default)/cart/page.tsx
@@ -20,6 +20,8 @@ interface Props {
   params: Promise<{ locale: string }>;
 }
 
+const CHECKOUT_URL = process.env.TRAILING_SLASH !== 'false' ? '/checkout/' : '/checkout';
+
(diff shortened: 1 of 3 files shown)


core/app/[locale]/(default)/cart/page.tsx
```diff
@@ -20,6 +20,8 @@ interface Props {
   params: Promise<{ locale: string }>;
 }
 
+const isUsingTrailingSlashes = process.env.TRAILING_SLASH !== 'false';
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/cart/page.tsx]: 🍹 I would just define the URL globally here
```suggestion
const CHECKOUT_URL = process.env.TRAILING_SLASH !== 'false' ? '/checkout/' : '/checkout';
```

PR #2685, opened 2025-11-12
Bump version to 1.3.1

jordanarldt:
## What/Why?
Add changeset for 1.3.0 -> 1.3.1

## Testing
n/a

## Migration
See changset

### DIFF
--- a/.changeset/blue-snakes-press.md
+++ b/.changeset/blue-snakes-press.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-makeswift": patch
+---
+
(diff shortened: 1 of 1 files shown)


.changeset/blue-snakes-press.md
matthewvolk, line 1: Might be simpler to just replace this with:
```
Includes all changes included in the [1.3.1 release](https://github.com/bigcommerce/catalyst/blob/61183ca1cd2957a2ebc114aa1b15a9d5184c8966/core/CHANGELOG.md#131) of `@bigcommerce/catalyst-core`
```

This is because Changesets takes care of the `### Patch Changes` headings, so if we merge this file as-is, it'll be confusing to read

chanceaclark (approved):

PR #2691, opened 2025-11-14
feat(reviews): implement cache revalidation for reviews after submission

matthewvolk:


core/app/[locale]/(default)/product/[slug]/_actions/submit-review.ts
```diff
@@ -64,6 +66,8 @@ export async function submitReview(
       };
     }
 
+    revalidateTag(TAGS.reviews);
```
Comment (matthewvolk) [inline core/app/[locale]/(default)/product/[slug]/_actions/submit-review.ts:69]: I'm trying to poke holes in this approach... since we don't have reCAPTCHA on the form... does this approach leave the merchant open to bots spamming their reviews and reviews data effectively never being cached?

PR #2694, opened 2025-11-14
Update anonymous-session.ts

BC-AdamWard:
Malformed maxAge was resulting in 7 hour instead of 7 day cookies

### DIFF
--- a/.changeset/fix-anonymous-session-maxage.md
+++ b/.changeset/fix-anonymous-session-maxage.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Fix anonymous session cookie maxAge calculation to correctly set 7 days instead of 7 hours.


chanceaclark (commented):
Thanks @BC-AdamWard for this! Mind if you add a changeset for this PR? Here's a canned response around generating one (should be a `patch` version):

***

To generate one, from the repo root, run:  

```
pnpm changeset
```

This will walk you through a few prompts – select an option with **space**, then press **enter** to continue (with or without a selected option). After you select an option, it'll ask you to write a short message describing the change. After finishing the prompts, it generates a markdown file where you can customize and add more detail; we like to include migration notes to guide users applying this change on top of their customizations of their cloned repository.

More info here: [Changesets documentation](https://github.com/changesets/changesets?tab=readme-ov-file#documentation).

BC-AdamWard:
@chanceaclark I think I added it correctly, but let me know if any other action is needed here.

PR #2723, opened 2025-11-17
Sync integrations makeswift

chanceaclark:
## What/Why?
Syncs `integrations/makeswift` and releases `1.3.4` to fix typecheck issue.

## Testing
N/A

## Migration
N/A


core/components/footer/index.tsx
chanceaclark, line 1: Needed to manually fix these issues as it wasn't fixed in the 1.3.3 version.

PR #2727, opened 2025-11-19
Add Meta Pixel analytics provider support

BC-AdamWard:
## What/Why?

Add Meta Pixel (Facebook Pixel) analytics provider to Catalyst analytics framework. Implements standard e-commerce event tracking (ViewContent, AddToCart, RemoveFromCart, ViewCart, ViewCategory) with consent management support. Enables merchants to track events using Meta Pixel alongside or instead of Google Analytics.

- Add MetaPixelProvider class implementing AnalyticsProvider interface
- Update WebAnalyticsFragment to include metaPixel configuration
- Update AnalyticsProvider component to support multiple providers
- Add consent management integration for GDPR/CCPA compliance

## Testing

1. Add Meta Pixel to store settings or inline in the Catalyst component. 
2. Load storefront and validate pixel fires against meta pixel 

## Migration

Minor change

### DIFF
--- a/core/lib/analytics/providers/meta-pixel/index.ts
+++ b/core/lib/analytics/providers/meta-pixel/index.ts
@@ -1,0 +1,221 @@
+import { AnalyticsProvider } from '~/lib/analytics/types';
+
+export interface MetaPixelConfig {
+  pixelId: string;
+  consentModeEnabled?: boolean;
+  nonce?: string;
+  getConsent?: () => Analytics.Consent.ConsentValues | null;
+}
+
+/**
+ * Meta Pixel (Facebook Pixel) Analytics Provider
+ *
+ * Implements the AnalyticsProvider interface to track e-commerce events using
+ * Meta Pixel. Mirrors the GoogleAnalyticsProvider architecture: the base code
+ * is injected once (defining the global `fbq`), consent is applied via Meta's
+ * Consent Mode, and events are forwarded through `fbq`.
+ *
+ * Standard Meta events are sent with `fbq('track', …)`; events without a Meta
+ * standard equivalent (ViewCart, RemoveFromCart, ViewCategory) are sent with
+ * `fbq('trackCustom', …)`. Each event carries an `eventID` for browser ↔
+ * Conversions API deduplication.
+ *
+ * @example
+ * ```typescript
+ * const metaPixel = new MetaPixelProvider({
+ *   pixelId: '1234567890',
+ *   consentModeEnabled: true,
+ *   getConsent: () => getConsentCookie(),
+ * });
+ * metaPixel.initialize();
+ * ```
+ */
+export class MetaPixelProvider implements AnalyticsProvider {
+  static #instance: MetaPixelProvider | null = null;
+
+  readonly cart = this.getCartEvents();
+  readonly navigation = this.getNavigationEvents();
+  readonly consent = this.getConsentEvents();
+
+  private readonly pixelScriptId = 'meta-pixel-script';
+
+  constructor(private readonly config: MetaPixelConfig) {
+    this.validateConfig();
+
+    if (MetaPixelProvider.#instance) {
+      return MetaPixelProvider.#instance;
+    }
+
+    MetaPixelProvider.#instance = this;
+  }
+
+  initialize() {
+    if (typeof window === 'undefined') {
+      throw new Error('Meta Pixel is only available in the browser environment');
+    }
+
+    // Order matters: define `fbq`, apply consent, then init. Meta's Consent
+    // Mode requires `fbq('consent', 'revoke')` to run *before* `fbq('init')`
+    // so no events fire until marketing consent is granted.
+    this.injectBaseCode();
+    this.initializeConsent();
+    this.trackPageView();
+  }
+
+  private validateConfig() {
+    if (!this.config.pixelId) {
+      throw new Error('Meta Pixel requires a Pixel ID');
+    }
+
+    // Pixel IDs are numeric strings.
+    if (!/^\d+$/.test(this.config.pixelId)) {
+      throw new Error('Meta Pixel ID must be a numeric string');
+    }
+  }
+
+  // Inject the Meta Pixel base code. This defines the global `fbq` function
+  // (which queues calls until fbevents.js loads) and loads the script. It does
+  // NOT call `fbq('init')` — that happens in `trackPageView`, after consent.
+  private injectBaseCode() {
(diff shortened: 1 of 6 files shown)


core/lib/analytics/providers/meta-pixel/index.ts
```diff
@@ -0,0 +1,346 @@
+import { AnalyticsProvider } from '~/lib/analytics/types';
+
+export interface MetaPixelConfig {
+  pixelId: string;
+  consentModeEnabled?: boolean;
+  nonce?: string;
+  getConsent?: () => Analytics.Consent.ConsentValues | null;
+  debugMode?: boolean;
```
chanceaclark: These two properties are based on how the analytics providers manages consent and if it has a debug option.

For the consent piece, do you know if MP handles managing consent on it's own, like GA does? Or does it always assume you have consent when placed on the page?

core/lib/analytics/providers/meta-pixel/index.ts
```diff
@@ -0,0 +1,346 @@
...
+      !function(f,b,e,v,n,t,s)
+      {if(f.fbq)return;n=f.fbq=function(){n.callMethod?
+      n.callMethod.apply(n,arguments):n.queue.push(arguments)};
+      if(!f._fbq)f._fbq=n;n.push=n;n.loaded=!0;n.version='2.0';
+      n.queue=[];t=b.createElement(e);t.async=!0;
+      t.src=v;s=b.getElementsByTagName(e)[0];
+      s.parentNode.insertBefore(t,s)}(window, document,'script',
+      'https://connect.facebook.net/en_US/fbevents.js');
+      fbq('init', '${this.config.pixelId}');
+      ${this.config.debugMode ? "fbq('track', 'PageView');" : ''}
```
chanceaclark: Is this manually added or do they have an example with `debugMode`?

core/lib/analytics/providers/meta-pixel/index.ts
```diff
@@ -0,0 +1,346 @@
...
+    if (!this.config.consentModeEnabled || !this.config.getConsent) {
+      return;
+    }
+
+    const consent = this.config.getConsent();
+
+    if (!consent) {
+      // Set default consent to denied if no consent is available
+      if (typeof window !== 'undefined' && (window as any).fbq) {
+        (window as any).fbq('consent', 'revoke');
```
chanceaclark: `any` should be avoided here and anywhere else in the file. Would also love to see if they have a type library we can use so we don't have to type check this, see `@types/gtag.js`.
```suggestion
      if (typeof window !== 'undefined' && 'fbq' in window) {
        window.fbq('consent', 'revoke');
```

core/lib/analytics/providers/meta-pixel/index.ts
```diff
@@ -0,0 +1,346 @@
...
+    if (this.config.consentModeEnabled && this.config.getConsent) {
+      const consent = this.config.getConsent();
+      if (consent && !consent.marketing) {
+        // Consent denied, don't initialize pixel
+        if (this.config.debugMode) {
+          console.log('[Meta Pixel] Consent denied, pixel not initialized');
+        }
+        return;
+      }
+    }
```
chanceaclark: We can probably move this logic into the `initializeConsent` function.

chanceaclark (commented):
Love how simple this add was! Wonder who set this up to be easy... 😏

Have a few comments specific to the provider + some type stuff otherwise looking good.

core/lib/analytics/providers/meta-pixel/index.ts
BC-AdamWard: Done — added `@types/facebook-pixel`, which declares a typed global `fbq` (same mechanism as `@types/gtag.js` for the GA provider), so all the `(window as any).fbq` casts are gone and the calls are fully typed.

BC-AdamWard:
Thanks for the review, @chanceaclark — sorry for the long pause on this one! I've refreshed the PR against the latest `canary` and addressed all the feedback:

- **No more `any`**: switched to `@types/facebook-pixel`, which declares a global `fbq` the same way `@types/gtag.js` backs the GA provider — so calls are fully typed and there are zero `any` casts left.
- **Consent**: reworked to mirror `GoogleAnalyticsProvider`. Meta *does* have a Consent Mode, so rather than skipping init when consent is denied, the pixel always initializes and applies `fbq('consent', 'grant'|'revoke')` — before `fbq('init')` per Meta's docs — so nothing fires until marketing consent is granted.
- **`debugMode`**: removed. Unlike GA4's real `debug_mode` param, Meta has no equivalent (debugging is via Pixel Helper / Test Events), so it was hand-rolled — dropped it and the console logging, and fixed the stray debug-gated `PageView`.
- **Standard vs custom events**: `AddToCart`/`ViewContent` use `fbq('track', …)`; `ViewCart`/`RemoveFromCart`/`ViewCategory` (no Meta standard equivalent) use `fbq('trackCustom', …)`. Each event now carries an `eventID` for browser ↔ Conversions API dedup.

Also — the schema dependency this originally flagged is resolved: `webAnalytics.metaPixel { pixelId }` now ships in the Storefront API, so the fragment validates and Lint/Typecheck is green. Verified events firing end-to-end with Meta Pixel Helper. Ready for another look 🙏

core/lib/analytics/providers/meta-pixel/index.ts
BC-AdamWard: Meta Pixel does support its own consent management via `fbq('consent', 'grant'|'revoke')` (Meta Consent Mode), so I've kept the consent path and reworked it to mirror the GA provider — see the updated `initializeConsent`. `debugMode` is removed (see the thread below).

core/lib/analytics/providers/meta-pixel/index.ts
BC-AdamWard: Refactored — the init-time consent handling now lives in `initializeConsent`, and the provider no longer skips initialization when consent is denied. It always initializes and applies `revoke`/`grant`, matching the GA provider and Meta's recommended Consent Mode flow.

core/lib/analytics/providers/meta-pixel/index.ts
BC-AdamWard: Removed `debugMode` entirely — unlike GA4's real `debug_mode` config param, Meta Pixel has no official equivalent (debugging is via the Pixel Helper extension / Events Manager Test Events), so this was hand-rolled. `PageView` now fires unconditionally after init as the standard base event.

PR #2739, opened 2025-12-02
feat: Add out-of-stock/backorder message to product lists

Tharaae:
## What/Why?
Display the backorder message or out-of-stock messages on the product card in different product lists according to the product stock status and the store settings.

## Testing
Screenshots
**Home page featured products lists**
Screenshot 2025-12-02 at 3 41 24 pm

**Category PLP**
Screenshot 2025-12-02 at 3 40 42 pm

**Brand PLP**
Screenshot 2025-12-02 at 3 40 00 pm

**Searc
Screenshot 2025-12-02 at 3 39 10 pm
h PLP**

## Migration
While rebasing, there might be some conflicts if there is some custom styling of the product card component. Those conflicts may occur in the following files:
- core/vibes/soul/primitives/product-card/index.tsx

### DIFF
--- a/.changeset/tricky-plants-teach.md
+++ b/.changeset/tricky-plants-teach.md
@@ -1,0 +1,105 @@
+---
+"@bigcommerce/catalyst-core": minor
+---
+
+Add out-of-stock / backorder message to product cards on PLPs based on store settings:
+- Add out of stock message if the product is out of stock and stock is set to display it.
+- Add the backorder message if the product has no on-hand stock and is available for backorder and the store/product is set to display the backorder message
+
+## Migration
+
+### Option 1: Automatic Migration (Recommended)
+For existing Catalyst stores, the simplest way to get the newly added feature is to rebase the existing code with the new release code. The files that will be updated are listed below.
+
+### Option 2: Manual Migration
+If you prefer not to rebase or have made customizations that prevent rebasing, follow these manual steps:
+
+#### Step 1: Update GraphQL Fragment
+Add the inventory fields to your product card fragment in `core/components/product-card/fragment.ts` under `Product`:
+```graphql
+inventory {
+  hasVariantInventory
+  isInStock
+  aggregated {
+    availableForBackorder
+    unlimitedBackorder
+    availableOnHand
+  }
+}
+variants(first: 1) {
+  edges {
+    node {
+      entityId
+      sku
+      inventory {
+        byLocation {
+          edges {
+            node {
+              locationEntityId
+              backorderMessage
+            }
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+#### Step 2: Update Product interface in Product Card component
+Update the `Product` interface in `core/vibes/soul/primitives/product-card/index.tsx` adding the following field to it:
+
+`inventoryMessage?: string;`
+
+#### Step 3: Update Data Transformer
+Modify `core/data-transformers/product-card-transformer.ts` to include inventory message in the transformed data. You can simply copy the whole file from this release as it does not have UI breaking changes.
+
+#### Step 4: Update Product Card Layout
+Update `core/vibes/soul/primitives/product-card/index.tsx` layout to display the new `inventoryMessage` product field.
+
+#### Step 5: Update Page Data GraphQL queries
+Add inventory settings queries to the pages data. Add the following query to the main GQL query under `site.settings`:
+```
+inventory {
+  defaultOutOfStockMessage
+  showOutOfStockMessage
+  showBackorderMessage
+}
+```
+to the following page data files:
+- `core/app/[locale]/(default)/(faceted)/brand/[slug]/page-data.ts`
+- `core/app/[locale]/(default)/(faceted)/category/[slug]/page-data.ts`
+- `core/app/[locale]/(default)/(faceted)/search/page-data.ts`
+- `core/app/[locale]/(default)/page-data.ts`
+
+#### Step 6: Update Page Components
+Update the corresponding page components to use the `productCardTransformer` method (if not already using it) to get the product card, and pass inventory data to those product cards based on the store inventory settings. Use the following code while retrieving the product lists:
+```
+    const { defaultOutOfStockMessage, showOutOfStockMessage, showBackorderMessage } =
+      data.site.settings?.inventory ?? {};
+
+    return productCardTransformer(
+      featuredProducts,
+      format,
+      showOutOfStockMessage ? defaultOutOfStockMessage : undefined,
+      showBackorderMessage,
+    );
+```
+in the following files:
+- `core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx`
+- `core/app/[locale]/(default)/(faceted)/category/[slug]/page.tsx`
+- `core/app/[locale]/(default)/(faceted)/search/page.tsx`
+- `core/app/[locale]/(default)/page.tsx`
+
+### Files Modified in This Change
+- `core/app/[locale]/(default)/(faceted)/brand/[slug]/page-data.ts`
+- `core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx`
+- `core/app/[locale]/(default)/(faceted)/category/[slug]/page-data.ts`
+- `core/app/[locale]/(default)/(faceted)/category/[slug]/page.tsx`
+- `core/app/[locale]/(default)/(faceted)/search/page-data.ts`
+- `core/app/[locale]/(default)/(faceted)/search/page.tsx`
+- `core/app/[locale]/(default)/page-data.ts`
+- `core/app/[locale]/(default)/page.tsx`
+- `core/components/product-card/fragment.ts`
+- `core/data-transformers/product-card-transformer.ts`
+- `core/vibes/soul/primitives/product-card/index.tsx`
--- a/core/app/[locale]/(default)/(faceted)/brand/[slug]/page-data.ts
+++ b/core/app/[locale]/(default)/(faceted)/brand/[slug]/page-data.ts
@@ -16,6 +16,11 @@ const BrandPageQuery = graphql(`
         }
       }
       settings {
+        inventory {
+          defaultOutOfStockMessage
+          showOutOfStockMessage
+          showBackorderMessage
+        }
         storefront {
           catalog {
             productComparisonsEnabled
--- a/core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx
+++ b/core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx
@@ -11,7 +11,7 @@ import { getFilterParsers } from '@/vibes/soul/sections/products-list-section/fi
 import { getSessionCustomerAccessToken } from '~/auth';
 import { facetsTransformer } from '~/data-transformers/facets-transformer';
 import { pageInfoTransformer } from '~/data-transformers/page-info-transformer';
-import { pricesTransformer } from '~/data-transformers/prices-transformer';
+import { productCardTransformer } from '~/data-transformers/product-card-transformer';
 import { getPreferredCurrencyCode } from '~/lib/currency';
 
 import { MAX_COMPARE_LIMIT } from '../../../compare/page-data';
@@ -132,16 +132,15 @@ export default async function Brand(props: Props) {
     const search = await streamableFacetedSearch;
     const products = search.products.items;
 
-    return products.map((product) => ({
-      id: product.entityId.toString(),
-      title: product.name,
-      href: product.path,
-      image: product.defaultImage
-        ? { src: product.defaultImage.url, alt: product.defaultImage.altText }
-        : undefined,
-      price: pricesTransformer(product.prices, format),
-      subtitle: product.brand?.name ?? undefined,
-    }));
+    const { defaultOutOfStockMessage, showOutOfStockMessage, showBackorderMessage } =
+      settings?.inventory ?? {};
+
+    return productCardTransformer(
+      products,
+      format,
+      showOutOfStockMessage ? defaultOutOfStockMessage : undefined,
+      showBackorderMessage,
+    );
   });
 
   const streamableTotalCount = Streamable.from(async () => {
--- a/core/components/product-card/fragment.ts
+++ b/core/components/product-card/fragment.ts
@@ -15,9 +15,36 @@ export const ProductCardFragment = graphql(
         name
         path
       }
+      inventory {
+        hasVariantInventory
+        isInStock
+        aggregated {
+          availableForBackorder
+          unlimitedBackorder
+          availableOnHand
+        }
+      }
       reviewSummary {
         numberOfReviews
         averageRating
+      }
+      variants(first: 1) {
+        edges {
+          node {
+            entityId
+            sku
+            inventory {
+              byLocation {
+                edges {
+                  node {
+                    locationEntityId
+                    backorderMessage
+                  }
+                }
+              }
+            }
+          }
+        }
       }
       ...PricingFragment
     }
--- a/core/data-transformers/product-card-transformer.ts
+++ b/core/data-transformers/product-card-transformer.ts
@@ -1,3 +1,4 @@
+import { removeEdgesAndNodes } from '@bigcommerce/catalyst-client';
 import { ResultOf } from 'gql.tada';
 import { getFormatter } from 'next-intl/server';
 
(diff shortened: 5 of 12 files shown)


core/app/[locale]/(default)/(faceted)/brand/[slug]/page-data.ts
```diff
@@ -16,6 +16,11 @@ const BrandPageQuery = graphql(`
         }
       }
       settings {
+        inventory {
+          defaultOutOfStockMessage
+          showOutOfStockMessage
+          showBackorderMessage
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/(faceted)/brand/[slug]/page-data.ts:22]: Is there a `defaultBackroderMessage`? Or plans to add it?

core/app/[locale]/(default)/(faceted)/brand/[slug]/page-data.ts
```diff
@@ -16,6 +16,11 @@ const BrandPageQuery = graphql(`
         }
       }
       settings {
+        inventory {
+          defaultOutOfStockMessage
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/(faceted)/brand/[slug]/page-data.ts:20]: Is this translatable in the CP? Mostly curious since I don't know much about this!

core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx
```diff
@@ -132,16 +132,15 @@ export default async function Brand(props: Props) {
...
-      title: product.name,
-      href: product.path,
-      image: product.defaultImage
-        ? { src: product.defaultImage.url, alt: product.defaultImage.altText }
-        : undefined,
-      price: pricesTransformer(product.prices, format),
-      subtitle: product.brand?.name ?? undefined,
-    }));
+    const { defaultOutOfStockMessage, showOutOfStockMessage, showBackorderMessage } =
+      settings?.inventory || {};
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx]: Would swap for nullish coalescing for all these instances

```suggestion
      settings?.inventory ?? {};
```

core/data-transformers/product-card-transformer.ts
```diff
@@ -7,9 +8,43 @@ import { ProductCardFragment } from '~/components/product-card/fragment';
 
 import { pricesTransformer } from './prices-transformer';
 
+const getInventoryMessage = (
+  product: ResultOf<typeof ProductCardFragment>,
+  outOfStockMessage?: string,
```
jorgemoya, line 13: Should we default this value in case no string is provided? Add to the Catalyst dictionary.

core/data-transformers/product-card-transformer.ts
```diff
@@ -7,9 +8,43 @@ import { ProductCardFragment } from '~/components/product-card/fragment';
...
+  showBackorderMessage?: boolean,
+) => {
+  if (!product.inventory.isInStock) {
+    return outOfStockMessage;
+  }
+
+  if (!showBackorderMessage || product.inventory.hasVariantInventory) {
+    return undefined;
+  }
```
jorgemoya, line 22: So if `showBackorderMessage` is false, we don't show a message. If product `hasVariantInventory` is true, we don't show a message. Is there a case where `hasVariantInventory` is true, but `showBackorderMessage` is also true? If this is possible, shouldn't we show a message?

core/data-transformers/product-card-transformer.ts
```diff
@@ -7,9 +8,43 @@ import { ProductCardFragment } from '~/components/product-card/fragment';
...
+
+  const { availableForBackorder, unlimitedBackorder } = product.inventory.aggregated || {};
```
jorgemoya:
```suggestion
  const { availableForBackorder, unlimitedBackorder } = product.inventory.aggregated ?? {};
```

core/data-transformers/product-card-transformer.ts
```diff
@@ -7,9 +8,43 @@ import { ProductCardFragment } from '~/components/product-card/fragment';
...
+
+  const hasBackorderAvailablity = !!availableForBackorder || unlimitedBackorder;
+
+  if (!hasBackorderAvailablity) {
+    return undefined;
+  }
```
jorgemoya, line 35: Why isn't `showBackorderMessage` coming with a value of `true` or `false` from the BE without having to do these checks in the FE? Any reasoning behind it that you know of?

core/data-transformers/product-card-transformer.ts
```diff
@@ -7,9 +8,43 @@ import { ProductCardFragment } from '~/components/product-card/fragment';
...
+
+  const baseVariant = removeEdgesAndNodes(product.variants).at(0);
+
+  if (!baseVariant?.inventory?.byLocation) {
+    return undefined;
+  }
+
+  const inventoryByLocation = removeEdgesAndNodes(baseVariant.inventory.byLocation).at(0);
+
+  return inventoryByLocation?.backorderMessage || undefined;
```
jorgemoya: Can you add a comment of what this is doing? I'm not familiar why this `!baseVariant?.inventory?.byLocation` warrants no message.

Should we also default a `backorderMessage` that we define in the Catalyst dictionary instead of returning `undefined`?

core/vibes/soul/primitives/product-card/index.tsx
```diff
@@ -133,20 +136,29 @@ export function ProductCard({
               {title}
             </span>
 
-            {subtitle != null && subtitle !== '' && (
```
jorgemoya, line 136: Is this conditional no longer needed?

core/vibes/soul/primitives/product-card/index.tsx
```diff
@@ -167,7 +179,7 @@ export function ProductCard({
         )}
       </div>
       {showCompare && (
-        <div className="mt-0.5 shrink-0">
+        <div className="mt-auto shrink-0">
```
jorgemoya, line 183: Is this spacing change required?

core/app/[locale]/(default)/(faceted)/brand/[slug]/page-data.ts
Comment (jorgemoya) [reply core/app/[locale]/(default)/(faceted)/brand/[slug]/page-data.ts:22]: I see now that this comes from `byLocation`

core/app/[locale]/(default)/(faceted)/brand/[slug]/page-data.ts
Comment (Tharaae) [reply core/app/[locale]/(default)/(faceted)/brand/[slug]/page-data.ts:20]: It's set by the merchant, and I'm not sure if it is or planned to be multi-lang.

core/data-transformers/product-card-transformer.ts
Tharaae: We don't want that as this prop has a value only if the inventory settings `showOutOfStockMessage` is true and `defaultOutOfStockMessage` is set by the merchant. Otherwise, it should be undefined.

core/data-transformers/product-card-transformer.ts
Tharaae: > Is there a case where hasVariantInventory is true, but showBackorderMessage is also true? 

Yes there is. But in this case, we only show the backorder message on the PDP when a variant is selected if the variant stock status requires to display it.

> If this is possible, shouldn't we show a message?

No, we shouldn't because if the product has variants, then each variant has its own stock status according to which we either display the backorder message or not on PDP when a variant is selected. On PLP, no variant is selected. Moreover, the backorder message can be set on the variant level, so we don't know which one to display unless a variant is selected.

core/data-transformers/product-card-transformer.ts
Tharaae: `showBackorderMessage` is an inventory settings set on the store level, while backorder availability is driven from a product's stock variables. Whether to display the backorder message or not depends on both factors.

core/data-transformers/product-card-transformer.ts
Tharaae: It's up to the merchant to decide whether to display a backorder message or not by allocating a backorder message to a product / product variant. So, it can be undefined.

core/vibes/soul/primitives/product-card/index.tsx
Tharaae: That was removed because I wanted to fix the line height even when empty. I will return the condition as we are not fixing the line height anymore.

core/vibes/soul/primitives/product-card/index.tsx
Tharaae: Yes, because was want this item to be aligned at the bottom of the card container.

core/app/[locale]/(default)/(faceted)/brand/[slug]/page-data.ts
Comment (jorgemoya) [reply core/app/[locale]/(default)/(faceted)/brand/[slug]/page-data.ts:20]: I don't think we can use this field if it is not multi-lang.

core/app/[locale]/(default)/(faceted)/brand/[slug]/page-data.ts
```diff
@@ -16,6 +16,11 @@ const BrandPageQuery = graphql(`
...
+          showOutOfStockMessage
```
Comment (jorgemoya) [inline core/app/[locale]/(default)/(faceted)/brand/[slug]/page-data.ts]: This setting only applies in PLP?

core/components/product-variants-inventory/fragment.ts
```diff
@@ -0,0 +1,24 @@
+import { graphql } from '~/client/graphql';
+
+export const ProductVariantsInventoryFragment = graphql(`
```
jorgemoya: I would define this in the existing [core/components/product-card/fragment.ts](https://github.com/bigcommerce/catalyst/pull/2739/files#diff-c34832222d1298f35975dc4965d7c30def4b702c2bf34d43c2ad4a0039876039)

core/components/product-variants-inventory/fragment.ts
```diff
@@ -0,0 +1,24 @@
...
+                  backorderMessage
+                }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+`);
```
jorgemoya: Seems to me you only use the base variant. Could you modify this fragment to only include the first (unless I'm missing something)?

core/data-transformers/product-card-transformer.ts
```diff
@@ -7,9 +8,43 @@ import { ProductCardFragment } from '~/components/product-card/fragment';
...
+
+  const baseVariant = removeEdgesAndNodes(product.variants).at(0);
+
+  if (!baseVariant?.inventory?.byLocation) {
+    return undefined;
+  }
+
+  const inventoryByLocation = removeEdgesAndNodes(baseVariant.inventory.byLocation).at(0);
+
+  return inventoryByLocation?.backorderMessage ?? undefined;
```
jorgemoya, line 45: Same issue with this. If this is not multi-lang we will need to rely on the Catalyst dictionary afaik.

core/vibes/soul/primitives/product-card/index.tsx
```diff
@@ -123,7 +126,7 @@ export function ProductCard({
           <div className="flex-1 text-sm @[16rem]:text-base">
             <span
               className={clsx(
-                'block font-semibold',
+                'line-clamp-2 font-semibold',
```
jorgemoya, line 129: Can we include changes like these in a separate PR for review?

core/vibes/soul/primitives/product-card/index.tsx
```diff
@@ -65,7 +68,7 @@ export function ProductCard({
   return (
     <article
       className={clsx(
-        'group flex min-w-0 max-w-md flex-col gap-2 font-[family-name:var(--card-font-family,var(--font-family-body))] @container',
+        'group flex min-w-0 max-w-md flex-col gap-3 font-[family-name:var(--card-font-family,var(--font-family-body))] @container',
```
jorgemoya, line 71: Just want to make sure this change was approved by a member of the design team.

jorgemoya (commented):
The problem I'm seeing with this feature is using strings defined by the merchant in the CP that are not multi-lang or translatable (as far as we know). We've skipped using these fields in Catalyst so that the translated experience remains consistent. To me, a good workaround is to simply use the Catalyst dictionary instead of fetching from GQL. 

Tagging our Catalyst PM to provide some input on this. @bc-erich

core/vibes/soul/primitives/product-card/index.tsx
Tharaae: I believe that @andrewreifman already discussed and finalised this UI with Jessica, our designer.

core/components/product-variants-inventory/fragment.ts
Tharaae: I created it to avoid duplicate code as the same fragment will be used in PDP and maybe cart page.

jorgemoya: Got it!

core/components/product-variants-inventory/fragment.ts
Tharaae: The same fragment will be used in product details page as well. So, I wanted to avoid duplicate code.

core/data-transformers/product-card-transformer.ts
Tharaae: @animesh1987 @bc-luke ... Do you have any info about the multi-lang support of backorder availability prompt, backorder message, out-of-stock message, etc?

core/vibes/soul/primitives/product-card/index.tsx
Tharaae: Why can't we get it reviewed in this PR? 🤔

core/data-transformers/product-card-transformer.ts
animesh1987: > @animesh1987 @bc-luke ... Do you have any info about the multi-lang support of backorder availability prompt, backorder message, out-of-stock message, etc?

Hey @jorgemoya we have multi lang in scope for the final product. There is work on inventory side that needs to be done, to make backorder message translated. It will be done when this project is GA.

core/components/product-variants-inventory/fragment.ts
jorgemoya: This is a very taxing query that is multiplied by number of products in a given PLP page. For PDP it's not a big deal, but if you are showing 20-30 products in a given PLP, I believe this is simply too much. 

There's two options:

1. Duplicate the code
2. Fragment can include a `first` prop (example `PricingFragment`)

core/data-transformers/product-card-transformer.ts
jorgemoya: Awesome, thanks @animesh1987 for clarifying!

core/app/[locale]/(default)/(faceted)/brand/[slug]/page-data.ts
Comment (jorgemoya) [reply core/app/[locale]/(default)/(faceted)/brand/[slug]/page-data.ts:20]: Resolving since @animesh1987 clarified this fields will be multi lang.

core/vibes/soul/primitives/product-card/index.tsx
jorgemoya: This is a complex PR that will be difficult for consumers of Catalyst to merge into their custom codebases. If this change is not related to the intent of this PR, it is much easier for consumers to see this change as a separate PR that is tracked with it's own migration and changeset release notes.

core/components/product-variants-inventory/fragment.ts
Tharaae: I'll go with option one

core/vibes/soul/primitives/product-card/index.tsx
Tharaae: This change to limit the product name to product name on the card to two line followed by "..." if longer, is requested by the Jess as per her agreement with @andrewreifman. It's related to this PR to control the card size/layout while adding more info to it which is the backorder message.

core/components/product-card/fragment.ts
```diff
@@ -15,10 +15,36 @@ export const ProductCardFragment = graphql(
...
+        aggregated {
+          availableForBackorder
+          unlimitedBackorder
+        }
+      }
       reviewSummary {
         numberOfReviews
         averageRating
       }
+      variants {
```
jorgemoya: Still need to query only the first one, since it seems that is all you need.

```suggestion
      variants(first: 1) {
```

.changeset/tricky-plants-teach.md
```diff
@@ -0,0 +1,21 @@
+---
+"@bigcommerce/catalyst-core": minor
+---
+
+Add out-of-stock / backorder message to product cards on PLPs based on store settings:
+- Add out of stock message if the product is out of stock and stock is set to display it.
+- Add the backorder message if the product has no on-hand stock and is available for backorder and the store/product is set to display the backorder message
+
+## Migration
+For existing Catalyst stores, to get the newly added feature, simply rebase the existing code with the new release code. The files to be rebased for this change to be applied are:
```
jorgemoya: Can you leverage AI and see if it can give you a better migration path for merchants that don't rebase?

Tharaae: AI did not do a good job extracting the detailed steps for migration. So, I explained it myself in more details.

PR #2740, opened 2025-12-03
Next 16

jorgemoya:

### DIFF
--- a/.changeset/strong-mammals-take.md
+++ b/.changeset/strong-mammals-take.md
@@ -88,7 +88,11 @@ Remove or comment out eslint config
 
 ### Step 3: Replace instances of `unstable_expireTag` and `unstable_expirePath` 
 
-Use `revalidatePath` and `revalidateTag` instead.
+Use `revalidatePath` and `revalidateTag` instead. To use the new API, utilize the `max` cache life argument:
+```typescript
+// Include the 'max' revalidation behavior
+revalidateTag(TAGS.checkout, 'max');
+```
 
 ### Step 4: 
 
@@ -103,11 +107,6 @@ Rename `priority` prop in `<Image>` for `preload`:
   sizes="100vw"
   src={image.src}
 />
-```
-
-```typescript
-// Include the 'max' revalidation behavior
-revalidateTag(TAGS.checkout, 'max');
 ```
 
 ## What's Coming Next


core/.eslintrc.cjs
```diff
@@ -82,6 +82,22 @@ const config = {
...
+    '.next/**',
+    'out/**',
+    'coverage/**',
+    'node_modules/**',
+    '.turbo/**',
+    '.wrangler/**',
+    '.bigcommerce/**',
+    'next-env.d.ts',
+    'messages/*.d.json.ts',
+  ],
```
chanceaclark: We can omit this as it's in `.eslintignore`: https://github.com/bigcommerce/catalyst/blob/canary/core/.eslintignore

.changeset/strong-mammals-take.md
```diff
@@ -0,0 +1,80 @@
...
+```typescript
+// Remove or disable:
+// experimental: {
+//   ppr: 'incremental',
+// }
+```
+
+### Step 3: Replace instances of `unstable_expireTag` and `unstable_expirePath` 
+
+Use `revalidatePath` and `revalidateTag` instead.
```
chanceaclark: Let's include the pattern of using the `max` expiration.

.changeset/strong-mammals-take.md
```diff
@@ -0,0 +1,80 @@
...
+
+# Next.js 16 Upgrade
+
+Catalyst has been upgraded to Next.js 16, bringing improved performance, new features, and better developer experience. This is a **minor version upgrade** that requires migration steps for existing stores.
+
+## Key Changes
+
+- ⚡ **Next.js 16.0.7**: Upgraded from Next.js 15.x to 16.0.7 (no more canary)
+- ⚛️ **React 19**: Upgraded to React 19.2.1 and React DOM 19.2.1
+- 🔧 **ESLint Migration**: Migrated from `next lint` to direct ESLint CLI (Next.js deprecated `next lint` in v16)
```
migueloller: Looking at the diff it seems that we had already migrated to `eslint` in a previous change set. Although it's not a problem to mention it here again.

.changeset/strong-mammals-take.md
```diff
@@ -0,0 +1,83 @@
...
+  }
+}
+```
+
+Then run:
+```bash
+pnpm install
+```
+
+### Step 2: Update next.config.ts
```
chanceaclark, line 57: Maybe we add something like this?
```suggestion

Note, `next` will automatically update your `tsconfig.json` file.

### Step 2: Update next.config.ts
```

core/package.json
```diff
@@ -10,7 +10,7 @@
     "build:analyze": "ANALYZE=true npm run build",
     "start": "next start",
     "prelint": "next typegen",
-    "lint": "eslint . --ext .js,.jsx,.ts,.tsx",
```
chanceaclark, line 13: Probably want to go back to this

PR #2744, opened 2025-12-03
chore: update next to latest v15

chanceaclark:
## What/Why?

This PR upgrades Catalyst to Next.js 15.5.7 to address a **critical security vulnerability** ([CVE-2025-55182](https://react.dev/blog/2025/12/03/critical-security-vulnerability-in-react-server-components)) in React Server Components that allowed unauthenticated remote code execution. This is a security patch upgrade that all users should apply immediately.

### Key Changes

**Security Updates:**

- Upgraded Next.js from 15.5.1-canary.4 to 15.5.7 (stable release, no longer canary)
- Upgraded React from 19.1.0 to 19.2.1
- Upgraded React DOM from 19.1.0 to 19.2.1

**Configuration Changes:**

- Removed Partial Prerendering (PPR) experimental flag from `next.config.ts`
  - PPR is unsupported in non-canary versions of Next.js 15
  - Full support for Next.js 16 cache component patterns will be added in a future release
- Removed ESLint configuration block from `next.config.ts`
- Removed `experimental_ppr` exports from layout and page files

**Dependency Updates:**

- Updated @next/bundle-analyzer from 15.2.3 to 15.5.7
- Updated eslint-config-next from 15.2.3 to 15.5.7
- Updated related development dependencies

**Performance Impact:**
The removal of PPR may result in different performance characteristics compared to the previous setup. This is a necessary trade-off to move to a stable release and address the security vulnerability.

## Testing
N/A

## Migration

If you're maintaining a custom Catalyst store, follow these steps:

### 1. Update Dependencies

Update your `core/package.json` dependencies:

```json
{
  "dependencies": {
    "next": "15.5.7",
    "react": "^19.2.1",
    "react-dom": "^19.2.1"
  },
  "devDependencies": {
    "@next/bundle-analyzer": "15.5.7",
    "eslint-config-next": "15.5.7"
  }
}
```

Then run:

```bash
pnpm install
```

### 2. Update next.config.ts

Remove or comment out the PPR configuration:

```typescript
// Remove this block:
// experimental: {
//   ppr: 'incremental',
// }
```

Remove or comment out the ESLint configuration block:

```typescript
// Remove this block:
// eslint: {
//   ignoreDuringBuilds: !!process.env.CI,
//   dirs: [
//     'app',
//     'auth',
//     'build-config',
//     'client',
//     'components',
//     'data-transformers',
//     'i18n',
//     'lib',
//     'middlewares',
//     'scripts',
//     'tests',
//     'vibes',
//   ],
[...]

PR #2755, opened 2025-12-08
chore: update next to latest v15

jorgemoya:
## What/Why?
Syncs next versions and fixes security vulnerability merged in #2744

## Testing
N/A

## Migration
Detailed in #2744


.changeset/green-rockets-chew.md
```diff
@@ -0,0 +1,91 @@
+---
+"@bigcommerce/catalyst-core": patch
```
jorgemoya: Does this needs to be different?

PR #2768, opened 2025-12-12
chore(core): upgrade c15t to 1.8.2

jorgemoya:
## What/Why?
Upgrade c15t to 1.8.2, migrate from custom mode to offline mode, refactor consent cookie handling to use c15t's compact format, add script location support for HEAD/BODY rendering, and add privacy policy link support to CookieBanner.

### What Changed

- Upgraded `@c15t/nextjs` to version `1.8.2`
- Changed consent manager mode from `custom` (with endpoint handlers) to `offline` mode
  - Removed custom `handlers.ts` implementation
- Added `enabled` prop to `C15TConsentManagerProvider` to control consent manager functionality
- Removed custom consent cookie encoder/decoder implementations (`decoder.ts`, `encoder.ts`)
- Added `parse-compact-format.ts` to handle c15t's compact cookie format
  - Compact format: `i.t:timestamp,c.necessary:1,c.functionality:1,etc...`
- Updated cookie parsing logic in both client and server to use the new compact format parser
- Scripts now support `location` field from BigCommerce API and can be rendered in `<head>` or `<body>` based on the `target` property
- `CookieBanner` now supports the `privacyPolicyUrl` field from BigCommerce API and will be rendered in the banner description if available.

## Testing
https://github.com/user-attachments/assets/edc06a3f-e65d-4339-a626-c7cea367e6a7

### Privacy Policy
Screenshot 2025-12-12 at 1 00 55 PM

### Consent Manager Provider Changes

The `ConsentManagerProvider` now uses `offline` mode instead of `custom` mode with endpoint handlers. The provider configuration has been simplified:

**Before:**
```typescript
<C15TConsentManagerProvider
  options={{
    mode: 'custom',
    consentCategories: ['necessary', 'functionality', 'marketing', 'measurement'],
    endpointHandlers: {
      showConsentBanner: () => showConsentBanner(isCookieConsentEnabled),
      setConsent,
      verifyConsent,
    },
  }}
>
  <ClientSideOptionsProvider scripts={scripts}>
    {children}
  </ClientSideOptionsProvider>
</C15TConsentManagerProvider>
```

**After:**
```typescript
<C15TConsentManagerProvider
  options={{
    mode: 'offline',
    storageConfig: {
      storageKey: CONSENT_COOKIE_NAME,
      crossSubdomain: true,
    },
    consentCategories: ['necessary', 'functionality', 'marketing', 'measurement'],
    enabled: isCookieConsentEnabled,
  }}
>
  <ClientSideOptionsProvider scripts={scripts}>
    {children}
  </ClientSideOptionsProvider>
</C15TConsentManagerProvider>
```

**Key changes:**
- `mode` changed from `'custom'` to `'offline'`
- Removed `endpointHandlers` - no longer needed in offline mode
- Added `enabled` prop to control consent manager functionality
- Added `storageConfig` for cookie storage configuration

### Cookie Handling

If you have custom code that directly reads or writes consent cookies, you'll need to update it:

**Before:**
The previous implementation used custom encoding/decoding. If you were directly accessing consent cookie values, you would have needed to use the custom decoder.

**After:**
The consent cookie now uses c15t's compact format. The public API for reading cookies remains the same:

```typescript
import { getConsentCookie } from '~/lib/consent-manager/cookies/client'; // client-side
// or
import { getConsentCookie } from '~/lib/consent-manager/cookies/server'; // server-side

const consent = getConsentCookie();
```

The `getConsentCookie()` function now internally uses `parseCompactFormat()` to parse the compact format cookie string. If you were directly parsing cookie values, you should now use the `getConsentCookie()` helper instead.

`getConsentCookie` now returns a compact version of the consent values:

```typescript
{
  i.t: 123456789,
  c.necessary: true,
  c.functionality: true,
  c.marketing: false,
  c.measurment: false
}
```

Updated instances where `getConsentCookie` is used to reflect this new schema.

Removed `setConsentCookie` from server and client since this is now handled by the c15t library.

### Script Location Support

Scripts now support rendering in either `<head>` or `<body>` based on the `location` field from the BigCommerce API:

```typescript
// Scripts transformer now includes target based on location
target: script.location === 'HEAD' ? 'head' : 'body'
```

The `ScriptsFragment` GraphQL query now includes the `location` field, allowing scripts to be placed in the appropriate DOM location. `FOOTER` location is still not supported.

### Privacy Policy

The `RootLayoutMetadataQuery` GraphQL query now includes the `privacyPolicyUrl` field, which renders a provicy policy link in the `CookieBanner` description.

```typescript
<CookieBanner 
  privacyPolicyUrl="https://example.com/privacy-policy"
  // ... other props
/>
```

The privacy policy link:
- Opens in a new tab (`target="_blank"`)
- Only renders if `privacyPolicyUrl` is provided as a non-empty string

Add translatable `privacyPolicy` field to `Components.ConsentManager.CookieBanner` translation namespace for the privacy policy link text.


core/app/[locale]/(default)/checkout/route.ts
```diff
@@ -76,9 +76,9 @@ export async function GET(req: NextRequest, { params }: { params: Promise<{ loca
         cartId,
         visitId: visitId ?? '',
         visitorId: visitorId ?? '',
-        analyticsConsent: consent?.preferences.measurement ?? false,
-        functionalConsent: consent?.preferences.functionality ?? false,
-        targetingConsent: consent?.preferences.marketing ?? false,
+        analyticsConsent: consent?.['c.measurement'] ?? false,
+        functionalConsent: consent?.['c.functionality'] ?? false,
+        targetingConsent: consent?.['c.marketing'] ?? false,
```
Comment (matthewvolk) [inline core/app/[locale]/(default)/checkout/route.ts:81]: 🍹 Is there a good reason to mix dot and bracket property accessors like this?

core/components/consent-manager/consent-providers.tsx
```diff
@@ -21,15 +20,13 @@ export function ConsentManagerProvider({
...
-
-        // @ts-expect-error endpointHandlers type is not yet exposed by the package
-        endpointHandlers: {
-          showConsentBanner: () => showConsentBanner(isCookieConsentEnabled),
-          setConsent,
-          verifyConsent,
+        mode: 'offline',
+        storageConfig: {
+          storageKey: CONSENT_COOKIE_NAME,
+          crossSubdomain: true,
```
matthewvolk, line 26: Enabling cross-subdomain doesn't interfere with consent cookie names that may be set on the hosted checkout, does it?

core/data-transformers/scripts-transformer.ts
```diff
@@ -66,6 +66,7 @@ export function scriptsTransformer(scripts: BigCommerceScripts): C15tScripts {
     const baseConfig: C15tScript = {
       category: mapConsentCategory(script.consentCategory),
       id: script.entityId,
+      target: script.location === 'HEAD' ? 'head' : 'body',
```
matthewvolk, line 69: 👏 Awesome that we support this now. Maybe you already mentioned it to Chris on the call we had, but we should include in the release notes since [the last release notes mentioned this feature was coming soon](https://developer.bigcommerce.com/docs/storefront/catalyst/release-notes/1-3-0#consent-manager:~:text=We%E2%80%99re%20collaborating%20closely%20with%20the%20c15t%20team%20to%20extend%20support%20for%20additional%20features%20such%20as%20footer%20script%20placement.%20If%20your%20storefront%20relies%20on%20specific%20script%20placement%20or%20privacy%20handling%20use%20cases%2C%20we%E2%80%99d%20love%20to%20hear%20your%20feedback%20as%20we%20refine%20this%20integration%20further.)

core/lib/consent-manager/cookies/parse-compact-format.ts
```diff
@@ -0,0 +1,25 @@
+/**
+ * Parses the compact format: i.t:1765485149496,c.necessary:1,c.functionality:1,etc.
```
matthewvolk, line 2: Ah... maybe this is the answer to my previous comment about property accessors

matthewvolk (approved):
Looks good, just want to make sure that if someone changes the default cookie name to match stencil, the `crossSubdomain` property doesn't cause spooky behavior

core/app/[locale]/(default)/checkout/route.ts
Comment (jorgemoya) [reply core/app/[locale]/(default)/checkout/route.ts:81]: Just not possible with the way the cookie schema is defined

core/components/consent-manager/consent-providers.tsx
jorgemoya: Per an offline conversation, it could conflict but cookie names are different so it should be fine.

core/data-transformers/scripts-transformer.ts
jorgemoya: Technically it's not rendered in the `footer` but I believe the behavior works as expected (priority loading).

core/lib/consent-manager/cookies/parse-compact-format.ts
jorgemoya: Yes, `c.necessary` is the key

PR #2786, opened 2025-12-18
ci: deploy `latest` tags to catalyst-demo.site

migueloller:
## What/Why?

This PR sets up a GitHub Action to deploy `latest` tags to Vercel.

## Testing

Tested locally.

## Migration

This workflow file is not intended to be used by Catalyst users and can safely be ignored or deleted.

```bash
rm .github/workflows/deploy.yml
```

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,0 +1,44 @@
+name: Production Tag Deployment
+env:
+    VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
+    VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+on:
+    push:
+        tags:
+            - "@bigcommerce/catalyst-core@latest"
+            - "@bigcommerce/catalyst-makeswift@latest"
+            - "@bigcommerce/catalyst-b2b-makeswift@latest"
+jobs:
+    deploy-tag:
+        name: Deploy `{{ github.ref_name }}` tag
+        runs-on: ubuntu-latest
+        concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,41 @@
...
+    VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+on:
+    push:
+        tags:
+            - "@bigcommerce/catalyst-core@latest"
+            - "@bigcommerce/catalyst-makeswift@latest"
+            - "@bigcommerce/catalyst-b2b-makeswift@latest"
+jobs:
+    Deploy-Tag:
+        runs-on: ubuntu-latest
```
chanceaclark: 🍹 
```suggestion
    deploy-tag:
        name: Deploy tag {{ github.ref_name }}
        runs-on: ubuntu-latest
```

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,41 @@
...
+        steps:
```
chanceaclark, line 17: We probably should add `concurrency` since it can be triggered by multiple tags at once.
```suggestion
        concurrency: 
          group: ${{ github.workflow }}-${{ github.ref_name }}
        steps:
```

.github/workflows/deploy.yml
migueloller: Updated!

.github/workflows/deploy.yml
migueloller: Updated!

migueloller:
Thanks for the suggestions @chanceaclark, PR has been updated.

matthewvolk (approved):

PR #2791, opened 2025-12-19
fix: order of `additionalProducts` in carousel

### LINKED ISSUE #2790: Product Carousel reordering is not working when I drag and drop products in the editor
**Describe the bug**
When using the Product Carousel component with Makeswift, manually reordering products in the Makeswift editor does not affect the render order on the storefront. The products always render in the default order returned by the BigCommerce API (usually sorted by ID) rather than the specific order defined in the productIds prop.

**To Reproduce**

1. Open the Makeswift builder/editor.
2. Add a Product Carousel component to a page.
3. Select specific products manually (e.g., Product A (ID: 10), Product B (ID: 5)).
4. Drag and drop to rearrange them so that Product B (ID: 5) comes before Product A (ID: 10).
5. Save/Publish and view the component.

Result: The carousel displays Product A first, then Product B (ignoring the manual order).

**Previously working?**
I don't this so it was ever working as this looks like a miss at start.

**Expected behavior**
The carousel should display products in the exact order arranged by the user in the Makeswift editor.

**Evidence**
[Vimeo Link](https://vimeo.com/1147911850)

**Additional context**
It appears the component renders the list based directly on the API response order, which defaults to sorting by ID, rather than preserving the order of the product IDs passed from the editor configuration.

migueloller:
## What/Why?

Fixes #2790.

### Before

https://github.com/user-attachments/assets/27a21c13-2e20-44db-a445-892f533dac88

### After

https://github.com/user-attachments/assets/db845aa9-a775-4ba1-96af-91ad4cd42d14

## Migration

Apply this PR's changes to the `core/lib/makeswift/utils/use-bc-product-to-vibes-product/use-bc-product-to-vibes-product.ts` and `core/lib/makeswift/utils/use-products.ts` files.

### DIFF
--- a/.changeset/heavy-hotels-lie.md
+++ b/.changeset/heavy-hotels-lie.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-makeswift": patch
+---
+
+Fix sort order of `additionalProducts` prop in `ProductsCarousel` Makeswift component.
--- a/core/lib/makeswift/utils/use-bc-product-to-vibes-product/use-bc-product-to-vibes-product.ts
+++ b/core/lib/makeswift/utils/use-bc-product-to-vibes-product/use-bc-product-to-vibes-product.ts
@@ -32,6 +32,8 @@ export const BcProductSchema = z.object({
 
 export type BcProductSchema = z.infer<typeof BcProductSchema>;
 
+export type { Product };
+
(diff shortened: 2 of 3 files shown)


matthewvolk (approved):
Looks good! Makes me wonder if people would ever want to change the order of products returned in the collection... but that's outside the scope of this PR. Thanks for taking care of this!

PR #2795, opened 2025-12-22
fix(core): show success message when a subscribed user resubscribes

jorgemoya:
## What/Why?
Show same success message when a user subscribes for the first time, or is already subscribed.

## Testing
Locally

## Migration
Update `core/components/subscribe/_actions/subscribe.ts` action to include new logic for resubscribed use case.


migueloller (approved):
LGTM! 👍🏻

PR #2800, opened 2025-12-30
feat(catalyst): CATALYST-1665 Hide "Exclusive Offers" field from account registration [temporary]

jordanarldt:
## What/Why?
Hide the "Exclusive Offers" field from the Account Registration form.

Currently, the field is not fully implemented in GraphQL, so it may be misleading to display it on the storefront if it's not actually doing anything when registering a customer.

Once the Register Customer operation takes this field into account, we can display it again.

## Testing
Tested locally

## Migration

Update `core/app/[locale]/(default)/(auth)/register/page.tsx` and add the function:
```ts
// There is currently a GraphQL gap where the "Exclusive Offers" field isn't accounted for
// during customer registration, so the field should not be shown on the Catalyst storefront until it is hooked up.
function removeExlusiveOffersField(field: Field | Field[]): boolean {
  if (Array.isArray(field)) {
    // Exclusive offers field will always have ID '25', since it is made upon store creation and is also read-only.
    return !field.some((f) => f.id === '25');
  }

  return field.id !== '25';
}
```

Then, add the following code at the end of the `const fields` declaration:
```ts
    })
    .filter(exists)
    .filter(removeExlusiveOffersField); // <---
```

### DIFF
--- a/core/app/[locale]/(default)/(auth)/register/page.tsx
+++ b/core/app/[locale]/(default)/(auth)/register/page.tsx
@@ -2,6 +2,7 @@ import { Metadata } from 'next';
 import { notFound } from 'next/navigation';
 import { getTranslations, setRequestLocale } from 'next-intl/server';
 
+import { Field } from '@/vibes/soul/form/dynamic-form/schema';
 import { DynamicFormSection } from '@/vibes/soul/sections/dynamic-form-section';
 import {
   formFieldTransformer,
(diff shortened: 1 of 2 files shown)


core/app/[locale]/(default)/(auth)/register/page.tsx
```diff
@@ -90,7 +102,8 @@ export default async function Register({ params }: Props) {
 
       return injectCountryCodeOptions(field, countries ?? []);
     })
-    .filter(exists);
+    .filter(exists)
+    .filter(removeExlusiveOffersField);
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/(auth)/register/page.tsx:106]: Do we want to filter this out completely or just set it as a hidden field. A merchant can make this field required which removing it from the form will cause the form submission to error IIRC.

chanceaclark (dismissed):

core/app/[locale]/(default)/(auth)/register/page.tsx
Comment (jordanarldt) [reply core/app/[locale]/(default)/(auth)/register/page.tsx:106]: @chanceaclark You can't make the field required:
image

And if the "Require Consent" checkbox is true in the Miscellaneous store settings, not including the field in graphQL is treated as `false` (not giving consent), so it should be all good 👍

Comment (chanceaclark) [reply core/app/[locale]/(default)/(auth)/register/page.tsx:106]: Sweet, thanks for that context!

chanceaclark (approved):

PR #2801, opened 2025-12-31
feat: Remove caching on fetching product inventory data

Tharaae:
## What/Why?
Changes in a product inventory does not reflect right away on product details page due GQL caching in Catalyst. To fix this issue, we will fetch the product/variant inventory data separately with no caching.

## Testing
Screen record
https://github.com/user-attachments/assets/e060094d-c076-49c9-86e6-cc0797c1aac1

## Migration
Rebase

### DIFF
--- a/.changeset/eager-nails-bake.md
+++ b/.changeset/eager-nails-bake.md
@@ -1,0 +1,10 @@
+---
+"@bigcommerce/catalyst-core": minor
+---
+
+Fetch product inventory data with a separate GQL query with no caching
+
+## Migration
+The files to be rebased for this change to be applied are:
+- core/app/[locale]/(default)/product/[slug]/page-data.ts
+- core/app/[locale]/(default)/product/[slug]/page.tsx
--- a/core/app/[locale]/(default)/product/[slug]/page-data.ts
+++ b/core/app/[locale]/(default)/product/[slug]/page-data.ts
@@ -210,7 +210,7 @@ export const getProduct = cache(async (entityId: number, customerAccessToken?: s
   return data.site;
 });
 
-const StreamableProductVariantBySkuQuery = graphql(`
+const StreamableProductVariantInventoryBySkuQuery = graphql(`
   query ProductVariantBySkuQuery($productId: Int!, $sku: String!) {
     site {
       product(entityId: $productId) {
@@ -246,15 +246,15 @@ const StreamableProductVariantBySkuQuery = graphql(`
   }
 `);
 
-type VariantVariables = VariablesOf<typeof StreamableProductVariantBySkuQuery>;
+type VariantInventoryVariables = VariablesOf<typeof StreamableProductVariantInventoryBySkuQuery>;
 
-export const getStreamableProductVariant = cache(
-  async (variables: VariantVariables, customerAccessToken?: string) => {
+export const getStreamableProductVariantInventory = cache(
+  async (variables: VariantInventoryVariables, customerAccessToken?: string) => {
     const { data } = await client.fetch({
-      document: StreamableProductVariantBySkuQuery,
+      document: StreamableProductVariantInventoryBySkuQuery,
       variables,
       customerAccessToken,
-      fetchOptions: { cache: 'no-store' },
+      fetchOptions: customerAccessToken ? { cache: 'no-store' } : { next: { revalidate: 60 } },
     });
 
     return data.site.product?.variants;
@@ -370,7 +370,7 @@ export const getStreamableProductInventory = cache(
       document: StreamableProductInventoryQuery,
       variables,
       customerAccessToken,
-      fetchOptions: { cache: 'no-store' },
+      fetchOptions: customerAccessToken ? { cache: 'no-store' } : { next: { revalidate: 60 } },
     });
 
     return data.site.product;
--- a/core/app/[locale]/(default)/product/[slug]/page-data.ts
+++ b/core/app/[locale]/(default)/product/[slug]/page-data.ts
@@ -254,7 +254,7 @@ export const getStreamableProductVariant = cache(
       document: StreamableProductVariantBySkuQuery,
       variables,
       customerAccessToken,
-      fetchOptions: customerAccessToken ? { cache: 'no-store' } : { next: { revalidate } },
+      fetchOptions: { cache: 'no-store' },
     });
 
     return data.site.product?.variants;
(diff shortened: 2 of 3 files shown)


.changeset/eager-nails-bake.md
```diff
@@ -0,0 +1,10 @@
+---
+"@bigcommerce/catalyst-core": minor
+---
+
+Fetch product inventory data with a separate GQL query with no caching
```
migueloller, line 5: Is the rationale for marking this as a `minor` the impact it could have on performance and/or costs for the PDP?

This is the primary question that I have when introducing queries that aren't cached. Are we good with every single PDP load invoking the API? Are we certain merchants won't run into rate limits?

core/app/[locale]/(default)/product/[slug]/page-data.ts
```diff
@@ -255,7 +255,7 @@ export const getStreamableProductVariant = cache(
       document: StreamableProductVariantBySkuQuery,
       variables,
       customerAccessToken,
-      fetchOptions: customerAccessToken ? { cache: 'no-store' } : { next: { revalidate } },
+      fetchOptions: { cache: 'no-store' },
```
Comment (migueloller) [inline core/app/[locale]/(default)/product/[slug]/page-data.ts]: This seems to be removing caching from the existing query for product variants. Was this intentional? If the query is specifically for variant inventory quantities, can we update the query name to be explicit about that.

core/app/[locale]/(default)/product/[slug]/page-data.ts
```diff
@@ -320,25 +350,23 @@ const StreamableProductQuery = graphql(
...
-  async (variables: Variables, customerAccessToken?: string) => {
+export const getStreamableProductInventory = cache(
+  async (variables: ProductInventoryVariables, customerAccessToken?: string) => {
     const { data } = await client.fetch({
-      document: StreamableProductQuery,
+      document: StreamableProductInventoryQuery,
       variables,
       customerAccessToken,
-      fetchOptions: customerAccessToken ? { cache: 'no-store' } : { next: { revalidate } },
+      fetchOptions: { cache: 'no-store' },
```
Comment (migueloller) [inline core/app/[locale]/(default)/product/[slug]/page-data.ts]: Is `no-store` the appropriate value here or would a very small revalidation period be desirable? There's is always a chance for there to be stale inventory data, even if it's always fetched without caching since there could be race conditions when loading the page. This could be addressed with polling, but I question whether that's a good approach. Instead, a warning when stock is very low might be good enough to let the user know that the item product might be out of stock soon.

core/app/[locale]/(default)/product/[slug]/page.tsx
```diff
@@ -117,8 +118,22 @@ export default async function Product({ params, searchParams }: Props) {
...
+    if (!product) {
+      return notFound();
+    }
+
+    return product;
+  });
+
   const streamableProductVariant = Streamable.from(async () => {
-    const product = await streamableProduct;
+    const product = await streamableProductInventory;
```
Comment (migueloller) [inline core/app/[locale]/(default)/product/[slug]/page.tsx]: Shouldn't the variable name change if this is specifically tracking product inventory?

core/app/[locale]/(default)/product/[slug]/page.tsx
Comment (migueloller) [inline core/app/[locale]/(default)/product/[slug]/page.tsx]: If the result here is coupled to both queries and we don't use `getStreamableProductVariant` elsewhere then we should combine this all into a single query so that we have all the data we need in a single query.

core/app/[locale]/(default)/product/[slug]/page-data.ts
Comment (jorgemoya) [reply core/app/[locale]/(default)/product/[slug]/page-data.ts]: I feel revalidation period + a GQL error message when trying to add a product to cart (which could revalidate the request) should be the way to go.

Comment (migueloller) [reply core/app/[locale]/(default)/product/[slug]/page-data.ts]: Agreed!

Comment (jorgemoya) [reply core/app/[locale]/(default)/product/[slug]/page-data.ts]: Is this something that we can explore @Tharaae?

core/app/[locale]/(default)/product/[slug]/page-data.ts
Comment (animesh1987) [reply core/app/[locale]/(default)/product/[slug]/page-data.ts]: Sorry bit behind on this due to various PTOs. I believe we only want to remove caching for inventory related information, and not all of the product details. 

@Tharaae is back from leave this week and we will work on addressing and cleaning up this PR.

.changeset/eager-nails-bake.md
Tharaae: I think it's minor because it has not major changes that impacts the UX much.

What happens with the current change is that only the product variant inventory data will be fetched with the PDP reload. So, product details and images as well as other settings data are not refetched.

core/app/[locale]/(default)/product/[slug]/page-data.ts
Comment (Tharaae) [reply core/app/[locale]/(default)/product/[slug]/page-data.ts]: As Animesh said, it's only for the product inventory data which is placed on the variant level. I will update the query name to state that.

core/app/[locale]/(default)/product/[slug]/page-data.ts
Comment (Tharaae) [reply core/app/[locale]/(default)/product/[slug]/page-data.ts]: Well, a GQL error message when trying to add a non-available product to cart is already implemented. I can add a revalidation period as well. The question is: what is the suitable period in this case? 🤔 Currently, the revalidation period is set as an env variable `DEFAULT_REVALIDATE_TARGET` with a default value 3600 (1 hour) which I think is too long for this query. I'll set it to 1 minute for now and let me know if you have a concern. @migueloller @jorgemoya @animesh1987

core/app/[locale]/(default)/product/[slug]/page.tsx
Comment (Tharaae) [reply core/app/[locale]/(default)/product/[slug]/page.tsx]: This inventory data is already in `inventory` child object, so we use it as `product.inventory.xxx` which makes more sense than `productInventory.inventory.xxx`, doesn't it?

I will change the main `streamableProductVariant` variable name to `streamableProductVariantInventory` though.

core/app/[locale]/(default)/product/[slug]/page.tsx
Comment (Tharaae) [reply core/app/[locale]/(default)/product/[slug]/page.tsx]: We use `getStreamableProductVariant` only if the product has different variants, otherwise not used.

Tharaae:
Hey @animesh1987 
@jorgemoya and @migueloller agreed on the following:

>I feel revalidation period + a GQL error message when trying to add a product to cart (which could revalidate the request) should be the way to go.

The error message when attempting to add an unavailable item to the cart is already implemented. For inventory data GQL, instead of real-time fetching, I added a revalidation period as advised, but I did not use the env variable one defaulted to one hour as I think this is too long. I set it to 1 minute to start the discussion about how long the revalidation period should be. What do you think? TBH, I still think that inventory data should be real-time with no caching.

jorgemoya (approved):

core/app/[locale]/(default)/product/[slug]/page-data.ts
```diff
@@ -324,25 +354,23 @@ const StreamableProductQuery = graphql(
...
         }
       }
     }
   `,
-  [ProductViewedFragment, ProductSchemaFragment, ProductVariantsInventoryFragment],
+  [ProductVariantsInventoryFragment],
 );
 
-type Variables = VariablesOf<typeof StreamableProductQuery>;
+type ProductInventoryVariables = VariablesOf<typeof StreamableProductQuery>;
```
Comment (agchan12) [inline core/app/[locale]/(default)/product/[slug]/page-data.ts:365]: Would this be better?

`type ProductInventoryVariables = VariablesOf<typeof StreamableProductInventoryQuery>;`

agchan12 (approved):

core/app/[locale]/(default)/product/[slug]/page-data.ts
```diff
@@ -246,15 +246,15 @@ const StreamableProductVariantBySkuQuery = graphql(`
...
-  async (variables: VariantVariables, customerAccessToken?: string) => {
+export const getStreamableProductVariantInventory = cache(
+  async (variables: VariantInventoryVariables, customerAccessToken?: string) => {
     const { data } = await client.fetch({
-      document: StreamableProductVariantBySkuQuery,
+      document: StreamableProductVariantInventoryBySkuQuery,
       variables,
       customerAccessToken,
-      fetchOptions: customerAccessToken ? { cache: 'no-store' } : { next: { revalidate } },
+      fetchOptions: customerAccessToken ? { cache: 'no-store' } : { next: { revalidate: 60 } },
```
Comment (animesh1987) [inline core/app/[locale]/(default)/product/[slug]/page-data.ts:257]: This seems fine for now, but I have a feeling we might get push back from product on aligning on fresh data for stock count.

animesh1987 (approved):

PR #2807, opened 2026-01-05
chore(ci): remove conditionally based on deployments

chanceaclark:
## What/Why?
We recently changed out deployments process to just one deployment target, thus we need to fix this CI action to account for it.

## Testing
See CI.

## Migration
N/A

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,7 +1,8 @@
 name: Production Tag Deployment
 env:
-    VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
-    VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+    # secrets is for dependabot compatibility; prefer vars when available
+    VERCEL_ORG_ID: ${{ vars.VERCEL_ORG_ID != '' && vars.VERCEL_ORG_ID || secrets.VERCEL_ORG_ID }}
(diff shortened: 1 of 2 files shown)


matthewvolk (approved):
Thank you sir! I'm sure you already did but I double checked and couldn't find any more references to the project name in other workflow definition files

jorgemoya (approved):

.github/workflows/deploy.yml
```diff
@@ -1,7 +1,8 @@
 name: Production Tag Deployment
 env:
-    VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
-    VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
+    # secrets is for dependabot compatibility; prefer vars when available
+    VERCEL_ORG_ID: ${{ vars.VERCEL_ORG_ID != '' && vars.VERCEL_ORG_ID || secrets.VERCEL_ORG_ID }}
+    VERCEL_PROJECT_ID: ${{ vars.VERCEL_PROJECT_ID != '' && vars.VERCEL_PROJECT_ID || secrets.VERCEL_PROJECT_ID }}
```
chanceaclark, line 5: Noticed the Github action wasn't deploying so changed this bit.

https://github.com/bigcommerce/catalyst/actions/runs/20732376914/job/59600526182

chanceaclark (commented):
@bigcommerce/team-catalyst need another review as I added some things.

jorgemoya (approved):

PR #2813, opened 2026-01-07
fix(core): delete duplicate select component

jorgemoya:
## What/Why?
`SelectField` was wrapping the `Select` component and causing double label/errors to be displayed. Deletes `SelectField` in favor of `Select`. 

Added the workaround for the [Radix issue](https://github.com/radix-ui/primitives/issues/3198) to the `Select` component. Confirmed issue is not resolved.

## Testing
Before:
image

After:
Screenshot 2026-01-07 at 12 12 10 PM

## Migration
Remove `SelectField` and update components to use `Select` instead.


chanceaclark (commented):
I fully forget why we did it, but I remember having two "Selects" was intentional based on this PR https://github.com/bigcommerce/catalyst/pull/2326 Do we want to keep the two still or does it make sense to deduplicate based on that PR?

jorgemoya:
> I fully forget why we did it, but I remember having two "Selects" was intentional based on this PR #2326 Do we want to keep the two still or does it make sense to deduplicate based on that PR?

Good callout @chanceaclark. I can't see a reason to separate these, maybe @hunterbecton can provide some context.

jorgemoya:
@chanceaclark I just noticed that at some point, the `Select` primitive component was modified to include a label, thus making the `SelectField` component obsolete.

We can keep just one (`Select`), or if we want to keep them separate, we need to make the `Select` component act as a true primitive with no `label` or error message logic.

chanceaclark:
I believe the `SelectField` contains the input, but I think there are some UI's like the PDP that utilize `Select` so moving over to utilizing an component with an input breaks when trying to add to the cart. I would double check though. @hunterbecton might know more though.

jorgemoya:
> I believe the `SelectField` contains the input, but I think there are some UI's like the PDP that utilize `Select` so moving over to utilizing an component with an input breaks when trying to add to the cart. I would double check though. @hunterbecton might know more though.

`SelectField` contains `Select`, which is the original problem, since both included a `label` and `errors`.

With the removal of `SelectField`, I have tested all UIs that use the `Select` component and see no issues.

hunterbecton:
> > I believe the `SelectField` contains the input, but I think there are some UI's like the PDP that utilize `Select` so moving over to utilizing an component with an input breaks when trying to add to the cart. I would double check though. @hunterbecton might know more though.
> 
> `SelectField` contains `Select`, which is the original problem, since both included a `label` and `errors`.
> 
> With the removal of `SelectField`, I have tested all UIs that use the `Select` component and see no issues.

I don't recall why we have a separate `SelectField`, but I'm introducing a `Field` component in `storefront-kit` so we don't need one-off components that might need labels, input descriptions, or field errors.

hunterbecton (approved):
I don't see an issue with dropping the `SelectField`

chanceaclark (approved):

PR #2815, opened 2026-01-08
Feat: add required indicators to form fields

jamesqquick:
## What/Why?
Added `(optional)` text to optional fields in forms so that it's clear which ones are required vs not

## Testing
Open the `/register` route and see a combination of required and optional fields.

image

## Migration
The new `required` props are optional, so they are backwards compatible. However, this does mean that the `(optional)` text will now show up on fields that aren't explicitly marked as required by passing the `required` prop to the `Label` component.

[CATALYST-1672](https://bigcommercecloud.atlassian.net/browse/CATALYST-1672)

[CATALYST-1672]: https://bigcommercecloud.atlassian.net/browse/CATALYST-1672?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/vibes/soul/form/label/index.tsx
```diff
@@ -30,6 +35,9 @@ export function Label({
         }[colorScheme],
         className,
       )}
-    />
+    >
+      {children}
+      {!required && <span className="ml-1 normal-case">(optional)</span>}
```
jorgemoya: We don't simply use strings as is, since we want translatable strings.

The old way of doing is by having a `optionalLabel` prop with a default value, but then make sure you prop drill the translated string all the way to this component.

The easy way is to simply use `getTranslations`/`useTranslations` in this component. Before, we avoided using these utilities directly in vibes components, but I don't think it matters anymore. cc @chanceaclark @hunterbecton

chanceaclark: Yeah, go ahead and use it.

matthewvolk: I agree, since we want to start leaning into "the Next.js way" of doing things in future UI library iterations, I vote using translations hooks inside components is OK.

matthewvolk (commented):
Could you generate a changeset for this? You can run `pnpm changeset` from the project root and walk through the prompts. Just include a 1-2 sentence description of the feature, and copy/paste the migration guide from your PR description and the changeset should be good to commit and add to this PR 👍

jamesqquick:
Added translations and a changeset cc @jorgemoya @matthewvolk

core/messages/da.json
```diff
@@ -607,5 +607,8 @@
                 }
             }
         }
+    },
+    "Form": {
+        "optional": "valgfri"
```
jorgemoya: These translated labels get auto generated every Monday, so for future references as long as we have the `en` label that is fine. `next-intl` will fallback to the `en` label if there is no available translation.

core/vibes/soul/form/button-radio-group/index.tsx
```diff
@@ -48,15 +48,24 @@ export const ButtonRadioGroup = React.forwardRef<
...
   (
-    { label, options, errors, className, onOptionMouseEnter, colorScheme = 'light', ...rest },
+    {
+      label,
+      options,
+      errors,
+      className,
+      onOptionMouseEnter,
+      colorScheme = 'light',
+      required,
```
jorgemoya, line 58: 🚨 Since you are destructuring the `required` prop and it is no longer contained within the `...rest` property, this will break the component since [RadioGroupPrimitive.Root](https://www.radix-ui.com/primitives/docs/components/radio-group#root) also needs `required` for form validation. 

So to fix this, simply pass `required` to `RadioGroupPrimitive.Root`.

This same issue applies to all other form components.

jorgemoya (changes requested):
Marking as request changes only because there's an oversight that will break form components until fixed.

core/vibes/soul/form/button-radio-group/index.tsx
jamesqquick: Updated each of the components that referenced `...rest` to additionally pass the `required` prop

jamesqquick:
Removed all translations except for english and made sure to pass `required` prop down for any component that referenced `...rest`

core/messages/da.json
jamesqquick: Updated to only keep the english version

core/vibes/soul/form/checkbox/index.tsx
```diff
@@ -120,6 +120,7 @@ export function Checkbox({
             id={id !== undefined ? `${id}-label` : `${generatedId}-label`}
           >
             {label}
+            {!props.required && <span className="ml-1 normal-case">(optional)</span>}
```
jorgemoya: Still not using the translated labels

jamesqquick: Removed this text from checkbox since inherently it doesn't make sense. Checkboxes are inherently optinoal

jorgemoya: Checkbox fields can still be required by the Bigcommerce GQL api.

Thinking of general cases, a ToS checkbox would be required.

jamesqquick: There's a difference though between requiring a value and requiring that the box be checked right? Technically, an unchecked box is still a valid user input? Does that make sense?

jamesqquick: Isn't an unckecked checkbox looked at as a boolean `false`?

jamesqquick: @jorgemoya updated to pass `field.required` to `CheckboxGroup` which handles displaying the label using the `Label` component for a group of individual `Checkbox` components. This matches the Stencil functionality. The difference is that we display optional text instead of required text.

CleanShot 2026-01-12 at 14 49 01@2x

CleanShot 2026-01-12 at 14 49 53@2x

core/messages/en.json
```diff
@@ -607,5 +607,8 @@
         }
       }
     }
+  },
+  "Form": {
+    "optional": "optional"
```
matthewvolk, line 612: 🍹Should punctuation e.g., `(`, `)` be included in `next-intl` strings? Wondering if other languages use a different convention.

core/vibes/soul/form/label/index.tsx
```diff
@@ -30,6 +40,9 @@ export function Label({
         }[colorScheme],
         className,
       )}
-    />
+    >
+      {children}
+      {!required && <span className="ml-1 normal-case">({t('optional')})</span>}
```
matthewvolk, line 45: I just realized that `<ToggleGroup />` is untouched in this PR, which may be intentional... but since `<ToggleGroup />` is used to power the filters panel on faceted pages (PLP), will this cause filter panels to display "(optional)" on PDP's if a filter renders as a `"toggle-group"`?

matthewvolk: We can disregard this comment - it looks like we don't even render a label in the filter panel for ToggleGroup

jorgemoya (approved):

core/messages/en.json
jamesqquick: I wondered about that. Not that I know of. I also wouldn't think the automation we have set up would have an automatic/appropriate translation if that was the case. I wouldn't think our automatic translations would understand the use case just based on the text enough to make a decision?

core/vibes/soul/form/label/index.tsx
jamesqquick: looks like that's handlded already

PR #2817, opened 2026-01-08
fix(core): persist checkbox modifier in PDP

jorgemoya:
## What/Why?
Persist the checkbox product modifier since it can modify pricing and other product data. By persisting this and tracking in the url, this will trigger a product refetch when added or removed. Incidentally, now we manually control what fields are persisted, since `option.isVariantOption` doesn't apply to `checkbox`, additionally multi options modifiers that are not variant options can also modify price and other product data.

## Testing
Without insurance:
Screenshot 2026-01-08 at 3 03 16 PM

With insurance (insurance is $100 in this example):
Screenshot 2026-01-08 at 3 03 28 PM

### Step 1

Update `product-options-transformer.ts` to manually track persisted fields:

```ts
case 'DropdownList': {
    return {
        // before
        // persist: option.isVariantOption,
        // after (manually persist)
        persist: true,
        type: 'select',
        label: option.displayName,
        required: option.isRequired,
        name: option.entityId.toString(),
        defaultValue: values.find((value) => value.isDefault)?.entityId.toString(),
        options: values.map((value) => ({
        label: value.label,
        value: value.entityId.toString(),
        })),
    };
}
```

Fields that persist and can affect product pricing when selected:
- Swatch
- RectangleBoxes
- RadioButtons
- ProductPickList
- ProductPickListWithImages
- CheckboxOption

### Step 2

Remove `isVariantOption` from GQL query since we no longer use it:

```ts
export const ProductOptionsFragment = graphql(
  `
    fragment ProductOptionsFragment on Product {
      entityId
      productOptions(first: 50) {
        edges {
          node {
            __typename
            entityId
            displayName
            isRequired
            isVariantOption // remove this
            ...MultipleChoiceFieldFragment
            ...CheckboxFieldFragment
            ...NumberFieldFragment
            ...TextFieldFragment
            ...MultiLineTextFieldFragment
            ...DateFieldFragment
          }
        }
      }
    }
  `,
  [
    MultipleChoiceFieldFragment,
    CheckboxFieldFragment,
    NumberFieldFragment,
    TextFieldFragment,
    MultiLineTextFieldFragment,
    DateFieldFragment,
  ],
);
```

### Step 3
Update `product-detail-form.tsx` to include separate handing of the checkbox field:

```ts
const defaultValue = fields.reduce<{
  [Key in keyof SchemaRawShape]?: z.infer<SchemaRawShape[Key]>;
}>(
  (acc, field) => {
    // Checkbox field has to be handled separately because we want to convert checked or unchecked value to true or undefined respectively.
    // This is because the form expects a boolean value, but we want to store the checked or unchecked value in the query params.
    if (field.type === 'checkbox') {
      if (params[field.name] === field.checkedValue) {
        return {
          ...acc,
          [field.name]: 'true',
        };
      }

      if (params[field.name] === field.uncheckedValue) {
        return {
          ...acc,
          [field.name]: undefined,
        };
      }

      return {
        ...acc,
        [field.name]: field.defaultValue, // Default value is either 'true' or undefined
      };
    }

    return {
      ...acc,
      [field.name]: params[field.name] ?? field.defaultValue,
    };
  },
  { quantity: minQuantity ?? 1 },
);

...

const handleChange = useCallback(
  (value: string) => {
    // Checkbox field has to be handled separately because we want to convert 'true' or '' to the checked or unchecked value respectively.
    if (field.type === 'checkbox') {
      void setParams({ [field.name]: value ? field.checkedValue : field.uncheckedValue });
    } else {
      void setParams({ [field.name]: value || null }); // Passing `null` to remove the value from the query params if fieldValue is falsey
    }

    controls.change(value || ''); // If fieldValue is falsey, we set it to an empty string
  },
  [setParams, field, controls],
);
```

### Step 4

Update schema in `core/vibes/soul/sections/product-detail/schema.ts`:

```ts
type CheckboxField = {
  type: 'checkbox';
  defaultValue?: string;
  checkedValue: string; // add
  uncheckedValue: string; // add
} & FormField;
```

### DIFF
--- a/core/vibes/soul/sections/product-detail/product-detail-form.tsx
+++ b/core/vibes/soul/sections/product-detail/product-detail-form.tsx
@@ -118,10 +118,21 @@ export function ProductDetailForm<F extends Field>({
   const defaultValue = fields.reduce<{
     [Key in keyof SchemaRawShape]?: z.infer<SchemaRawShape[Key]>;
   }>(
-    (acc, field) => ({
-      ...acc,
-      [field.name]: params[field.name] ?? field.defaultValue,
-    }),
+    (acc, field) => {
+      // Checkbox fields need to be handled differently since we keep track of the checkedValue and not the boolean value of the default value.
+      if (field.type === 'checkbox') {
+        return {
+          ...acc,
+          [field.name]:
+            (params[field.name] ?? field.defaultValue === 'true') ? field.checkedValue : '',
+        };
+      }
+
+      return {
+        ...acc,
+        [field.name]: params[field.name] ?? field.defaultValue,
+      };
+    },
     { quantity: minQuantity ?? 1 },
   );
 
--- a/core/vibes/soul/sections/product-detail/product-detail-form.tsx
+++ b/core/vibes/soul/sections/product-detail/product-detail-form.tsx
@@ -327,20 +327,16 @@ function FormField({
 }) {
   const controls = useInputControl(formField);
 
-  const [params, setParams] = useQueryStates(
+  const [, setParams] = useQueryStates(
     field.persist === true ? { [field.name]: parseAsString.withOptions({ shallow: false }) } : {},
   );
 
   const handleChange = useCallback(
     (value: string) => {
-      // Ensure that if page is reached without a full reload, we are still setting the selection properly based on query params.
-      const fieldValue = value || params[field.name];
-
-      void setParams({ [field.name]: fieldValue || null }); // Passing `null` to remove the value from the query params if fieldValue is falsey
-
-      controls.change(fieldValue ?? ''); // If fieldValue is falsey, we set it to an empty string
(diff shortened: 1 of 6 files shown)


core/vibes/soul/sections/product-detail/product-detail-form.tsx
```diff
@@ -327,20 +327,16 @@ function FormField({
...
 
-  const [params, setParams] = useQueryStates(
+  const [, setParams] = useQueryStates(
     field.persist === true ? { [field.name]: parseAsString.withOptions({ shallow: false }) } : {},
   );
 
   const handleChange = useCallback(
     (value: string) => {
-      // Ensure that if page is reached without a full reload, we are still setting the selection properly based on query params.
-      const fieldValue = value || params[field.name];
```
jorgemoya, line 337: Verified this is no longer needed.

chanceaclark (approved):
Love it when the changesets is more code than the actual changes ❤️

chanceaclark (approved):

PR #2819, opened 2026-01-09
[Catalyst 1677] Dont reset login form data

jamesqquick:
## What/Why?
Don't reset login form data on a failed login attempt. This way the user doesn't have to enter their email address again from scratch,

## Testing
Navigate to `/login` and attempt to login with invalid credentials. You should see an error message but the form data should be maintained.

https://github.com/user-attachments/assets/ef56e270-3e68-42bb-bb4c-9b5792874e2c

## Migration
Added a call to `event.preventDefault()` in the `onSubmission` callback to ensure default resetting of form data doesn't not happen in the sign up form.

```ts
const [form, fields] = useForm({
  //...
  onSubmit(event, { formData }) {
    event.preventDefault();
    startTransition(() => {
      formAction(formData);
    });
  }
});
```

### DIFF
--- a/core/vibes/soul/sections/sign-in-section/sign-in-form.tsx
+++ b/core/vibes/soul/sections/sign-in-section/sign-in-form.tsx
@@ -2,7 +2,7 @@
 
 import { getFormProps, getInputProps, SubmissionResult, useForm } from '@conform-to/react';
 import { getZodConstraint, parseWithZod } from '@conform-to/zod';
-import { useActionState, useEffect } from 'react';
+import { startTransition, useActionState, useEffect } from 'react';
 import { useFormStatus } from 'react-dom';
 
 import { FormStatus } from '@/vibes/soul/form/form-status';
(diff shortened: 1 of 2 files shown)


core/vibes/soul/sections/sign-in-section/sign-in-form.tsx
```diff
@@ -37,6 +37,12 @@ export function SignInForm({
     constraint: getZodConstraint(schema),
     shouldValidate: 'onBlur',
     shouldRevalidate: 'onInput',
+    onSubmit(event, { formData }) {
+      event.preventDefault();
```
matthewvolk, line 41: 🍹 I could be wrong, but I think `preventDefault()` allows us to remove the `action={formAction}` attribute on the form element below. Since we're now calling it here in `onSubmit`, I _think_ the action prop is redundant. Double check me on that.

matthewvolk (approved):
Looks good, if you don't mind double checking my comment below that would be awesome!

core/vibes/soul/sections/sign-in-section/sign-in-form.tsx
jamesqquick: Updated to remove passing the action!

matthewvolk (approved):

PR #2822, opened 2026-01-13
chore(catalyst): CATALYST-1297 Run E2E tests on PRs

jordanarldt:
## What/Why?
Ensure Catalyst E2E tests are executed on pull requests. This will help prevent further breaking changes from contributions. There are 3 separate jobs:
- `E2E Functional Tests` - Executes all e2e tests
- `E2E Functional Tests (alternate locale)` - Executes all tests that require an alternate locale to be used
- `E2E Functional Tests (TRAILING_SLASH=false)` - Executes the functional tests that require the `TRAILING_SLASH` env var to be disabled. This will prevent 301 redirect loop regressions

## Testing
Test suites executed on this PR and they are passing.

## Migration

Add `required` prop to the Country selector and Country input fields in `core/vibes/soul/sections/cart/shipping-form/index.tsx` on lines 280 and 28

For all other changes, use this PR as a reference.

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -87,9 +87,31 @@ jobs:
         run: pnpm run test
 
   e2e-tests:
-    name: E2E Functional Tests
+    name: E2E Functional Tests (${{ matrix.name }})
 
     runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        include:
+          - name: default
+            browsers: chromium webkit
+            test-filter: tests/ui/e2e
+            trailing-slash: true
+            locale-var: TESTS_LOCALE
+            artifact-name: playwright-report
+          - name: TRAILING_SLASH=false
+            browsers: chromium
+            test-filter: tests/ui/e2e --grep @no-trailing-slash
+            trailing-slash: false
+            locale-var: TESTS_LOCALE
+            artifact-name: playwright-report-no-trailing
+          - name: alternate locale
+            browsers: chromium
+            test-filter: tests/ui/e2e --grep @alternate-locale
+            trailing-slash: true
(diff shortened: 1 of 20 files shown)


.github/workflows/basic.yml
```diff
@@ -75,3 +85,179 @@ jobs:
...
+
+      - name: Install Playwright browsers
+        run: pnpm exec playwright install --with-deps chromium webkit
+        working-directory: ./core
+
+      - name: Build catalyst-client
+        run: pnpm --filter @bigcommerce/catalyst-client build
+
+      - name: Build catalyst-core
+        run: pnpm --filter @bigcommerce/catalyst-core build
```
chanceaclark: 🍹 Could we just run `pnpm build` and let turbo cache the other projects? We'll need to provide the turbo env var here I think to enable the build cache, but allows us to not worry about each thing that needs to the built.

.github/workflows/basic.yml
```diff
@@ -75,3 +85,179 @@ jobs:
 
       - name: Run Tests
         run: pnpm run test
+
+  e2e-tests:
```
chanceaclark, line 89: 🍹 What about instead of defining each job manually, we use a matrix to run the tests. I threw AI to showcase what I mean on this branch (not sure it actually works though): https://github.com/bigcommerce/catalyst/compare/CATALYST-1297...refactor/e2e-tests

chanceaclark (approved):
Two 🍹's but looks good on my end.

jorgemoya (approved):
What an :mvp:!

.github/workflows/basic.yml
jordanarldt: Looks promising! I'll try it.

PR #2825, opened 2026-01-16
feat: TRAC-268 add graphql proxy for client-side storefront API access

chanceaclark:
## What/Why?

Add a GraphQL proxy endpoint to the Catalyst proxy layer that allows client-side JavaScript (e.g., checkout-sdk-js) to make GraphQL Storefront API requests without exposing storefront tokens to the browser.

This is a foundational piece that enables:
- **PROJECT-6580** — eWallet buttons on PDP (Apple Pay, Google Pay, etc.) require checkout-sdk-js to make GraphQL calls from the browser to initialize wallet payment methods.
- **PROJECT-6074** — Saved payment methods management in My Account may require client-side GraphQL calls for payment instrument CRUD operations.

### How it works

1. **Request matching** — Only `POST /graphql` requests with a valid `x-catalyst-graphql-proxy-requester` header are intercepted. All others pass through.
2. **Requester allowlist** — The header value must be in `ALLOWED_REQUESTERS` (currently: `checkout-sdk-js`).
3. **Unauthenticated token** — Authenticates to the BigCommerce GraphQL API using `BIGCOMMERCE_STOREFRONT_UNAUTHENTICATED_TOKEN`, a dedicated scoped-down token separate from the primary `BIGCOMMERCE_STOREFRONT_TOKEN`.
4. **Customer access token** — The `auth()` wrapper extracts the customer access token from the session and passes it via `X-Bc-Customer-Access-Token` for customer-specific queries (e.g., cart, wishlists).
5. **No cookie forwarding** — Browser cookies are not forwarded to the GraphQL API.
6. **Body validation** — Request bodies are validated with Zod before forwarding.

```
Browser (checkout-sdk-js)
  │  POST /graphql
  │  x-catalyst-graphql-proxy-requester: checkout-sdk-js
  │  { query, variables }
  ▼
Next.js Proxy Layer (withGraphqlProxy)
  │  Authorization: Bearer <BIGCOMMERCE_STOREFRONT_UNAUTHENTICATED_TOKEN>
  │  X-Bc-Customer-Access-Token: <from session>
  │  { query, variables }
  ▼
BigCommerce GraphQL Storefront API
```

### Valid request to the proxy:

Screenshot 2026-01-16 at 15 48 56

### Request without valid header:
Screenshot 2026-01-16 at 15 49 03

### Request with invalid query (since it's a browser request):
Screenshot 2026-01-22 at 13 48 30

### Request with valid query:
Screenshot 2026-01-22 at 13 48 38

## Migration

See changeset for full migration steps. Key changes:
1. Add `BIGCOMMERCE_STOREFRONT_UNAUTHENTICATED_TOKEN` env var
2. Create `core/proxies/with-graphql-proxy.ts`
3. Register `withGraphqlProxy` in `core/proxy.ts` (after `withChannelId`, before `withRoutes`)

### DIFF
--- a/.changeset/foo-bar-baz.md
+++ b/.changeset/foo-bar-baz.md
@@ -1,0 +1,116 @@
+---
+"@bigcommerce/catalyst-core": minor
+---
+
+Add GraphQL proxy to enable client-side GraphQL requests through the storefront. This proxy forwards requests from allowed clients (such as checkout-sdk-js) to the BigCommerce Storefront API using a dedicated unauthenticated storefront token for API authorization, while still passing through the customer access token for customer-specific data. No browser cookies are forwarded.
+
+## Migration
+
+### Step 1: Add environment variable
+
+Add a `BIGCOMMERCE_STOREFRONT_UNAUTHENTICATED_TOKEN` to your `.env.local`. This should be a storefront API token scoped for client-side proxy use with minimal permissions.
+
+### Step 2: Create proxy
+
+Create a new file `core/proxies/with-graphql-proxy.ts`:
+
+```ts
+import { NextResponse, URLPattern } from 'next/server';
+import { z } from 'zod';
+
+import { auth } from '~/auth';
+import { client } from '~/client';
+
+import { type ProxyFactory } from './compose-proxies';
+
+const ALLOWED_REQUESTERS = ['checkout-sdk-js'];
+const graphqlPathPattern = new URLPattern({ pathname: '/graphql' });
+
+const bodySchema = z.object({
+  query: z.unknown(),
+  variables: z.record(z.unknown()).default({}),
+});
+
+export const withGraphqlProxy: ProxyFactory = (next) => {
+  return async (request, event) => {
+    // Only handle /graphql path
+    if (!graphqlPathPattern.test(request.nextUrl.toString())) {
+      return next(request, event);
+    }
+
+    const requester = request.headers.get('x-catalyst-graphql-proxy-requester');
+
+    // Validate required header
+    if (!requester || !ALLOWED_REQUESTERS.includes(requester)) {
+      return next(request, event);
+    }
+
+    // Only handle POST requests
+    if (request.method !== 'POST') {
+      return new NextResponse('Method not allowed', { status: 405 });
+    }
+
+    // Wrap in auth to get customer access token for customer-specific data
+    return auth(async (req) => {
+      try {
+        // Parse incoming GraphQL request body
+        const body: unknown = await req.json();
+        const { query, variables } = bodySchema.parse(body);
+
+        if (!query) {
+          return NextResponse.json({ error: 'Missing query' }, { status: 400 });
+        }
+
+        // Get customer access token if authenticated
+        const customerAccessToken = req.auth?.user?.customerAccessToken;
+
+        // Proxy the request using the existing client with an unauthenticated storefront token
+        const response = await client.fetch({
+          document: query,
+          variables,
+          customerAccessToken,
+          fetchOptions: {
+            headers: {
+              Authorization: `Bearer ${process.env.BIGCOMMERCE_STOREFRONT_UNAUTHENTICATED_TOKEN}`,
+            },
+            next: { revalidate: 0 },
+          },
+        });
+
+        return NextResponse.json(response);
+      } catch (error) {
+        // eslint-disable-next-line no-console
+        console.error(error);
+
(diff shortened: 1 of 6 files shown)


jorgemoya (approved):

chanceaclark:
Marking as `do not merge` since this is a minor and we probably want to add it as apart of the next minor release.

.changeset/foo-bar-baz.md
```diff
@@ -0,0 +1,114 @@
...
+        const customerAccessToken = req.auth?.user?.customerAccessToken;
+
+        // Proxy the request using the existing client
+        const response = await client.fetch({
+          document: query,
+          variables,
+          customerAccessToken,
+          fetchOptions: {
+            headers: {
+              // DO NOT REMOVE: We pass the request cookies to emulate that this is coming from the browser to the GraphQL API
```
davidchin: Do we actually need this? Since we’ll be using scoped API tokens without the customer scope for the proxy endpoint, it won’t be able to access sensitive fields such as the customer access token.

The issue with forwarding all browser cookies to the underlying BC SF API is that those cookies are intended for Catalyst, not the BC SF API. They may contain sensitive data (e.g.: third-party analytics cookies), cookie names that BC might unintentionally read if they overlap (e.g.: CSRF tokens), or cause BC to assume a storefront session exists and start treating proxied requests as stateful.

chanceaclark: >Do we actually need this? Since we’ll be using scoped API tokens without the customer scope for the proxy endpoint, it won’t be able to access sensitive fields such as the customer access token.

We are thinking about releasing it in the next minor version whether or not the scoped API tokens are ready or not.

>The issue with forwarding all browser cookies to the underlying BC SF API is that those cookies are intended for Catalyst, not the BC SF API. They may contain sensitive data (e.g.: third-party analytics cookies), cookie names that BC might unintentionally read if they overlap (e.g.: CSRF tokens), or cause BC to assume a storefront session exists and start treating proxied requests as stateful.

We don't necessarily need to forward all the cookies then, we can pass a dummy value instead. I looked at the storefront service code, and essentially passing this header makes the request behave like it's a request coming directly from the client which applies the same rules as we want from a functionality standpoint as the scoped API token approach. 

Also, we don't pass back the response `Set-Cookie` value back to the client so we won't be passing back stateful information back. I get the point you are trying to make here, but I don't think it's a cause for concern as we are not passing the stateful session cookies back.

davidchin: > We don't necessarily need to forward all the cookies then, we can pass a dummy value instead.

Thanks for the explanation. Sending a specific cookie set by Catalyst could definitely alleviate some of my concerns. That said, I’m not 100% sure about the potential unintended side effects of this change. It’s worth checking the following to make sure it’s safe to proceed or if it's better to wait for scoped tokens.

1. We should verify whether sending a cookie would make the request stateful, meaning starting a new BC storefront session. If so, it could have several implications, e.g.: cart access would require the corresponding session token to be sent back, spam protection might be enforced etc...
2. This change could also impact analytics data and potentially skew metrics, since shopper activity is tracked for requests coming from browsers.
3. Stateful requests may have to different rate limits than stateless requests. Also, I'm not sure whether including a cookie header could interfere with the trusted proxy setup.

core/proxies/with-graphql-proxy.ts
```diff
@@ -0,0 +1,88 @@
...
+        console.log('[TRAC-521 DEMO] proxy → client.fetch');
+        console.log('[TRAC-521 DEMO]   override key casing : "Authorization" (capital A)');
+        console.log(
+          '[TRAC-521 DEMO]   override token      : Bearer',
+          `${unauthToken.slice(0, 8)}…${unauthToken.slice(-4)}`,
+          `(UNAUTH token, len=${unauthToken.length})`,
+        );
+        console.log('[TRAC-521 DEMO]   expected behavior   : client.ts default Authorization');
+        console.log('[TRAC-521 DEMO]                         (storefront token) is overridden');
+        /* eslint-enable no-console */
```
chanceaclark: Make sure to remove all this debugging code.

.changeset/foo-bar-baz.md
parthshahp: For added context here: this doesn't apply anymore as we aren't proxying cookies anymore.

core/proxies/with-graphql-proxy.ts
parthshahp: 🤦

rtalvarez (approved):

PR #2830, opened 2026-01-20
Fix cart success message translations in French and Danish

funivan:
## Overview

This PR fixes translation issues in French and Danish JSON files where `CartItems` (capital C) was used instead of `cartItems` (lowercase), and where `<cartLink>` incorrectly wrapped the entire message.

## Root Cause
The code in `core/app/[locale]/(default)/product/[slug]/_actions/add-to-cart.tsx` (passes `cartItems: quantity` (lowercase) to the translation function. When translation files used `CartItems` (capital C), the pluralization failed because the variable name didn't match.

Additionally, the `<cartLink>` tag should only wrap the clickable text (like "your cart"), not the entire message including the plural variable.

## Testing

image

image

- The cart link is clickable and navigates to `/cart`

### DIFF
--- a/core/messages/da.json
+++ b/core/messages/da.json
@@ -373,7 +373,7 @@
         "otherDetails": "Andre oplysninger",
         "noOtherDetails": "Der er ingen andre oplysninger.",
         "viewOptions": "Se muligheder",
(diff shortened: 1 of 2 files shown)


chanceaclark:
@funivan should this be fixed in the next round of translations? How is manually updating going to prevent it from being overwritten in the future?

funivan:
@chanceaclark We need to improve tooling in our translations system. We'll try to fix this issue, I have a ticket for that.

PR #2832, opened 2026-01-22
fix(core): add root level not-found page

jorgemoya:
## What/Why?
Adds a root level `core/app/not-found.tsx` that redirects to our catch all route, which in turns renders the `core/app/[locale]/not-found.tsx`.

This is to circumvent the issue that `/404` is handled by Vercel in the edge and looks for a `not-found` page at the root level.

## Testing
Screenshot 2026-01-27 at 4 32 59 PM

## Migration
Add new `core/app/not-found.tsx` page.

### DIFF
--- a/.changeset/large-trams-rule.md
+++ b/.changeset/large-trams-rule.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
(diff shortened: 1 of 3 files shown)


chanceaclark (approved):

jorgemoya:
Closing until we get a response from Vercel on how to manage this in a better way.

PR #2835, opened 2026-01-22
chore: add deployment for native hosting

chanceaclark:
## What/Why?

Adds CI/CD for deploying the Catalyst storefront to BigCommerce Native Hosting using the `@bigcommerce/catalyst` alpha CLI and `@opennextjs/cloudflare`.

This PR includes:

- **Native Hosting workflow** (`.github/workflows/native-hosting.yml`) — builds and deploys the storefront on pushes to `canary` and on PRs
- **Proxy-to-middleware conversion step** — automatically converts Next.js 16's `proxy.ts` to `middleware.ts` at build time (renames the file, swaps the export, adds `runtime: 'experimental-edge'`) so the OpenNext/Cloudflare build works without maintaining a separate `middleware.ts` in the repo
- **Alpha CLI + changeset pre-release setup** — versions `@bigcommerce/catalyst` as `1.0.0-alpha.x` with changesets in pre mode
- **CLI improvements included in alpha** — auth commands, logs command, `--prebuilt` deploy, credential resolution, deployment error mapping, telemetry updates, and removal of the dev command

## Testing

- Workflow runs on PR open/sync and push to `canary`
- Verify the "Convert proxy.ts to middleware.ts" step correctly transforms the file before the Next.js build
- Confirm `catalyst build` and `catalyst deploy --prebuilt` succeed with the configured secrets/vars

## Migration

N/A

### DIFF
--- a/.github/workflows/native-hosting.yml
+++ b/.github/workflows/native-hosting.yml
@@ -1,0 +1,69 @@
+name: Native Hosting
+
+on:
+  push:
+    branches: [canary]
+  pull_request:
+    types: [opened, synchronize]
+
+jobs:
+  build-and-deploy:
+    name: Build and Deploy
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - uses: pnpm/action-setup@v3
+
+      - name: Use Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: '.nvmrc'
+          cache: 'pnpm'
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Install Catalyst CLI
+        run: pnpm add @bigcommerce/catalyst@alpha @opennextjs/cloudflare
+        working-directory: core
+
+      - name: Update package scripts
+        run: |
+          jq '.scripts.build = "npm run generate && catalyst build --framework catalyst" |
+              .scripts.dev = "npm run generate && catalyst dev" |
+              .scripts.start = "catalyst start" |
+              .scripts["deploy"] = "catalyst deploy"' \
+          package.json > package.tmp.json
+          mv package.tmp.json package.json
+        working-directory: core
+
+      - name: Build Dependencies
+        run: pnpm --filter "./packages/*" build
+
+      - name: Catalyst Build
+        run: pnpm run build
+        working-directory: core
+        # These environment variables are required for the build step to succeed
+        env:
+          AUTH_SECRET: ${{ secrets.NATIVE_HOSTING_AUTH_SECRET }}
+          BIGCOMMERCE_CHANNEL_ID: ${{ vars.NATIVE_HOSTING_BIGCOMMERCE_CHANNEL_ID }}
+          BIGCOMMERCE_STORE_HASH: ${{ vars.NATIVE_HOSTING_BIGCOMMERCE_STORE_HASH }}
+          BIGCOMMERCE_STOREFRONT_TOKEN: ${{ secrets.NATIVE_HOSTING_BIGCOMMERCE_STOREFRONT_TOKEN }}
+          BIGCOMMERCE_PROJECT_UUID: ${{ secrets.NATIVE_HOSTING_BIGCOMMERCE_PROJECT_UUID }}
+
+      - name: Catalyst Deploy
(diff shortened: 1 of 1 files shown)


.github/workflows/native-hosting.yml
```diff
@@ -0,0 +1,71 @@
...
+        with:
+          node-version-file: '.nvmrc'
+          cache: 'pnpm'
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Install Catalyst CLI
+        run: pnpm add @bigcommerce/catalyst@alpha @opennextjs/cloudflare
+        working-directory: core
```
chanceaclark, line 30: If this is our example gh-action, should we conditionally run this if it's already installed?

.github/workflows/native-hosting.yml
```diff
@@ -0,0 +1,71 @@
...
+
+      - name: Update package scripts
```
chanceaclark: I could see this script being a little finicky if merchants are trying to use this. Might need to think of a better way here, e.g. just running the commands in the Github action.

.github/workflows/native-hosting.yml
```diff
@@ -0,0 +1,71 @@
...
+        run: |
+          jq '.scripts.build = "npm run generate && catalyst build --framework catalyst" |
```
chanceaclark: @bigcommerce/team-catalyst this is an interesting workaround I had to do. Basically, if the `project.json` `framework` field is populated we use that value. However in CI, `project.json` doesn't exist so we manually need to add `--framework` here.

.github/workflows/native-hosting.yml
```diff
@@ -0,0 +1,71 @@
...
+            --secret BIGCOMMERCE_CHANNEL_ID=${{ vars.NATIVE_HOSTING_BIGCOMMERCE_CHANNEL_ID }} \
+            --secret AUTH_SECRET=${{ secrets.NATIVE_HOSTING_AUTH_SECRET }} \
+            --secret OPEN_NEXT_ERROR_LOG_LEVEL=debug
+        working-directory: core
+        # These environment variables are for the deploy command only
+        # Added here to avoid polluting the run script with additional flags
+        env:
+          BIGCOMMERCE_ACCESS_TOKEN: ${{ secrets.NATIVE_HOSTING_BIGCOMMERCE_ACCESS_TOKEN }}
+          BIGCOMMERCE_PROJECT_UUID: ${{ secrets.NATIVE_HOSTING_BIGCOMMERCE_PROJECT_UUID }}
+          BIGCOMMERCE_STORE_HASH: ${{ vars.NATIVE_HOSTING_BIGCOMMERCE_STORE_HASH }}
```
chanceaclark: I think the conversation we are having in Slack is crucial to this mess.

1. I think building during the deploy command will resolve having two commands.
2. The separation of CLI environment variables vs. Build-time environment variables vs. runtime environment variables is a topic of conversation.

.github/workflows/native-hosting.yml
matthewvolk: I agree, I don't think there's much benefit to changing scripts every CI run, since the main benefit behind changing scripts is convenience for developers typing in the commands locally

.github/workflows/native-hosting.yml
matthewvolk: Does this change if we move forward with not having the CLI fall back to `next` CLI?

.github/workflows/native-hosting.yml
matthewvolk: I was thinking about this earlier this morning, I say we keep it until the `create-catalyst` CLI installs Native Hosting deps for you; technically if someone follows documentation correctly, they won't need this step but right now there's not a huge risk to leaving it in (unless we ship a breaking change to the CLI on alpha, or something similar, which installing `@whatever` could actually break things)

.github/workflows/native-hosting.yml
matthewvolk: Resolved by using `pnpm exec catalyst` directly

.github/workflows/native-hosting.yml
matthewvolk: CLI was simplified to no longer support `--framework`, resolved

jorgemoya (approved):

PR #2848, opened 2026-01-28
chore(ci): extract e2e tests to separate workflow

matthewvolk:
## What/Why?

Extracts E2E tests into a separate workflow file to optimize CI resource usage.

**Before:** E2E tests ran on all triggers including merge commits to `canary`/integration branches
**After:** E2E tests only run on PRs and merge queue (before merging)

This ensures E2E tests still validate changes before they land, while avoiding redundant runs after merge.

## Testing

- Open a PR targeting canary → E2E tests should run
- Merge queue entry → E2E tests should run

## Migration

None required.


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,112 @@
+name: E2E Tests
+
+on:
+  pull_request:
+    types: [opened, synchronize]
+    branches: [canary, integrations/makeswift, integrations/b2b-makeswift]
+  merge_group:
```
chanceaclark: Let's remove the merge group too for now.

matthewvolk: I was thinking it might be useful to run in the merge_queue still, but maybe my understanding is off... I thought `merge_group` would catch cases where two PR's individually pass, but break when merged together. 

That said, I think the merge queue is less frequently "full" these days, so I guess I see both sides

chanceaclark: Yeah but because this workflow isn't required it's going to just merge in when all the required workflows pass thus making this just run in limbo.

tl'dr there is no real benefit in including it in the merge group unless the workflow is required.

matthewvolk: bvEMf2S

True true - I'll remove it!

jordanarldt (approved):

jorgemoya (approved):

chanceaclark (approved):

PR #2855, opened 2026-01-30
fix(core): use state abbreviation instead of entityId on shipping form

jorgemoya:
## What/Why?
Use state abbreviation instead of entityId for cart shipping form state values. The shipping API expects state abbreviations, not entityIds. Since some states share the same abbreviation across countries (causing Radix UI duplicate key issues), we use a composite value format (`entityId-abbreviation`) and extract the abbreviation when submitting.

## Testing
Alaska has a specific state shipping rate compared to other states.
Screenshot 2026-01-30 at 2 21 14 PM

### Step 1

Update `core/app/[locale]/(default)/cart/_actions/update-shipping-info.ts` to extract the abbreviation from the composite value (in both the add and update consignment branches):

```diff
            address: {
              countryCode: submission.value.country,
              city: submission.value.city,
-             stateOrProvince: submission.value.state,
+             stateOrProvince: submission.value.state?.split('-')[1],
              postalCode: submission.value.postalCode,
            },
```

### Step 2

Update the state select options in `core/app/[locale]/(default)/cart/page.tsx` to use composite values:

```diff
  const statesOrProvinces = shippingCountries.map((country) => ({
    country: country.code,
    states: country.statesOrProvinces.map((state) => ({
-     value: state.entityId.toString(),
+     value: `${state.entityId}-${state.abbreviation}`,
      label: state.name,
    })),
  }));
```

### Step 3

Add the selected state mapping logic in `core/app/[locale]/(default)/cart/page.tsx` before the return statement to properly initialize the form with the saved shipping address:

```diff
  const checkoutUrl = data.site.settings?.url.checkoutUrl;

+ const selectedStateOrProvince = statesOrProvinces
+   .find((country) => country.country === shippingConsignment?.address.countryCode)
+   ?.states.find(
+     (state) => state.value.split('-')[1] === shippingConsignment?.address.stateOrProvince,
+   )?.value;
+
+ const selectedAddress = shippingConsignment?.address
+   ? {
+       country: shippingConsignment.address.countryCode,
+       city:
+         shippingConsignment.address.city !== ''
+           ? (shippingConsignment.address.city ?? undefined)
+           : undefined,
+       state:
+         shippingConsignment.address.stateOrProvince !== '' ? selectedStateOrProvince : undefined,
+       postalCode:
+         shippingConsignment.address.postalCode !== ''
+           ? (shippingConsignment.address.postalCode ?? undefined)
+           : undefined,
+     }
+   : undefined;

  return (
```

### Step 4

Update the `ShippingEstimator` address prop in `core/app/[locale]/(default)/cart/page.tsx`:

```diff
            action: updateShippingInfo,
            countries,
            states: statesOrProvinces,
-           address: shippingConsignment?.address
-             ? {
-                 country: shippingConsignment.address.countryCode,
-                 city:
-                   shippingConsignment.address.city !== ''
-                     ? (shippingConsignment.address.city ?? undefined)
-                     : undefined,
-                 state:
-                   shippingConsignment.address.stateOrProvince !== ''
-                     ? (shippingConsignment.address.stateOrProvince ?? undefined)
-                     : undefined,
-                 postalCode:
-                   shippingConsignment.address.postalCode !== ''
-                     ? (shippingConsignment.address.postalCode ?? undefined)
-                     : undefined,
-               }
-             : undefined,
+           address: selectedAddress,
```

### DIFF
--- a/.changeset/hip-women-design.md
+++ b/.changeset/hip-women-design.md
@@ -2,40 +2,98 @@
 "@bigcommerce/catalyst-core": patch
 ---
 
-Use state abbreviation instead of entityId for cart shipping form state values. The shipping API expects state abbreviations, not entityIds. Since some states share the same abbreviation across countries (causing Radix UI duplicate key issues), we use a composite value format (`entityId-abbreviation`) and extract the abbreviation when submitting.
+Use state abbreviation instead of entityId for cart shipping form state values. The shipping API expects state abbreviations, not entityIds. Since some states share the same abbreviation across countries (causing Radix UI duplicate key issues), we use a composite value format (`entityId|abbreviation`) and extract the abbreviation when submitting.
 
 ## Migration steps
 
 ### Step 1
 
-Update `core/app/[locale]/(default)/cart/_actions/update-shipping-info.ts` to extract the abbreviation from the composite value (in both the add and update consignment branches):
+Create a new utility file `core/lib/state-utils.ts` with helper functions for handling composite state values:
+
+```typescript
+/**
+ * Separator used in composite state values.
+ * We use a pipe character to avoid conflicts with hyphens that may appear in abbreviations.
+ */
+const STATE_VALUE_SEPARATOR = '|';
+
+/**
+ * Creates a composite state value from entityId and abbreviation.
+ * This format ensures unique values for Radix UI Select components.
+ *
+ * @param {number} entityId - The state entity ID
+ * @param {string} abbreviation - The state abbreviation
+ * @returns {string} A composite value in the format "entityId|abbreviation"
+ */
+export function createStateValue(entityId: number, abbreviation: string): string {
+  return `${entityId}${STATE_VALUE_SEPARATOR}${abbreviation}`;
+}
+
+/**
+ * Extracts the state abbreviation from a composite state value.
+ *
+ * @param {string | undefined} compositeValue - The composite value in format "entityId|abbreviation"
+ * @returns {string | undefined} The state abbreviation, or undefined if the value is invalid
+ */
+export function parseStateAbbreviation(compositeValue: string | undefined): string | undefined {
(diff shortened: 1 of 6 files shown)


.changeset/hip-women-design.md
```diff
@@ -0,0 +1,96 @@
...
+### Step 1
+
+Update `core/app/[locale]/(default)/cart/_actions/update-shipping-info.ts` to extract the abbreviation from the composite value (in both the add and update consignment branches):
+
+```diff
+            address: {
+              countryCode: submission.value.country,
+              city: submission.value.city,
+-             stateOrProvince: submission.value.state,
++             stateOrProvince: submission.value.state?.split('-')[1],
```
migueloller: What do you think about providing helper functions instead of having the guts of the encoding/decoding out like this?

core/app/[locale]/(default)/cart/page.tsx
```diff
@@ -195,7 +195,9 @@ export default async function Cart({ params }: Props) {
   const statesOrProvinces = shippingCountries.map((country) => ({
     country: country.code,
     states: country.statesOrProvinces.map((state) => ({
-      value: state.entityId.toString(),
+      // Use abbreviation and entityId as value to avoid duplicate values (Radix uses value as key)
+      // Workaround until all statesOrProvince are unique values
+      value: `${state.entityId}-${state.abbreviation}`,
```
Comment (migueloller) [inline core/app/[locale]/(default)/cart/page.tsx]: What if we replaced this with `getStateValue(state)` or `encodeState(state)` and then below used `decodeStateValue(value).entityId` or `getStateValueEntityId(state)`?

Comment (jorgemoya) [reply core/app/[locale]/(default)/cart/page.tsx]: Great call out. Updated!

jorgemoya:
Closing in favor of an easier solution

PR #2860, opened 2026-02-03
feat: add explicit Makeswift SEO  metadata support

jorgemoya:
## What/Why?

Add explicit Makeswift SEO metadata support to public-facing pages. When configured in Makeswift, the SEO title and description will take priority over the default values from BigCommerce or static translations.

The following pages now support Makeswift SEO metadata:

- Home page (`/`)
- Catch-all page (`/[...rest]`)
- Product page (`/product/[slug]`)
- Brand page (`/brand/[slug]`)
- Category page (`/category/[slug]`)
- Blog list page (`/blog`)
- Blog post page (`/blog/[blogId]`)
- Search page (`/search`)
- Cart page (`/cart`)
- Compare page (`/compare`)
- Gift certificates page (`/gift-certificates`)
- Gift certificates balance page (`/gift-certificates/balance`)
- Contact webpage (`/webpages/[id]/contact`)
- Normal webpage (`/webpages/[id]/normal`)

## Testing

Without Makeswift metadata:
Screenshot 2026-02-03 at 11 58 00 AM

With Makeswift metadata:
Screenshot 2026-02-03 at 11 56 08 AM
metadata:

### Step 1: Add `getMakeswiftPageMetadata` function

Add the `getMakeswiftPageMetadata` function to `core/lib/makeswift/client.ts`:

```diff
+ export async function getMakeswiftPageMetadata({ path, locale }: { path: string; locale: string }) {
+   const snapshot = await getPageSnapshot({ path, locale });
+
+   if (snapshot == null) {
+     return null;
+   }
+
+   const { meta } = snapshot.document;
+
+   return {
+     ...(meta.title && { title: meta.title }),
+     ...(meta.description && { description: meta.description }),
+   };
+ }
```

Export the function from `core/lib/makeswift/index.ts`:

```diff
  export { Page } from './page';
- export { client } from './client';
+ export { client, getMakeswiftPageMetadata } from './client';
```

### Step 2: Update page metadata

Each page's `generateMetadata` function has been updated to fetch Makeswift metadata and use it as the primary source, falling back to existing values. Here's an example using the cart page:

Update `core/app/[locale]/(default)/cart/page.tsx`:

```diff
  import { getPreferredCurrencyCode } from '~/lib/currency';
+ import { getMakeswiftPageMetadata } from '~/lib/makeswift';
  import { Slot } from '~/lib/makeswift/slot';
```

```diff
  export async function generateMetadata({ params }: Props): Promise<Metadata> {
    const { locale } = await params;

    const t = await getTranslations({ locale, namespace: 'Cart' });
+   const makeswiftMetadata = await getMakeswiftPageMetadata({ path: '/cart', locale });

    return {
-     title: t('title'),
+     title: makeswiftMetadata?.title || t('title'),
+     description: makeswiftMetadata?.description || undefined,
    };
  }
```

Apply the same pattern to the other pages listed above, using the appropriate path for each page (e.g., `/blog`, `/search`, `/compare`, etc.).


core/lib/makeswift/client.ts
```diff
@@ -32,3 +32,18 @@ export const getComponentSnapshot = async (snapshotId: string) => {
 function normalizeLocale(locale: string): string | undefined {
   return locale === defaultLocale ? undefined : locale;
 }
+
+export async function getPageMetadata({ path, locale }: { path: string; locale: string }) {
+  const snapshot = await getPageSnapshot({ path, locale });
```
migueloller: We shouldn't be using `getPageSnapshot` for this since the snapshot is meant to be opaque. We should have a `getPage` method on the client that provides the relevant metadata.

core/app/[locale]/(default)/page.tsx
```diff
@@ -1,18 +1,27 @@
...
-  return locales.map((locale) => ({ locale }));
-}
-
 interface Props {
   params: Promise<Params>;
 }
 
+export async function generateMetadata({ params }: Props): Promise<Metadata> {
+  const { locale } = await params;
+  const metadata = await getPageMetadata({ path: '/', locale });
```
Comment (migueloller) [inline core/app/[locale]/(default)/page.tsx]: This name is a bit generic. Is there a reason we need the helper as opposed to calling the client directly? Or perhaps we can rename it to `getMakeswiftPageMetadata`?

migueloller (commented):
We're adding Makeswift metadata to every single page, regardless of whether it's editable in Makeswift or not. We might just want to add it to the pages that are visually editable. @jamesqquick, what do you think?

core/app/[locale]/(default)/page.tsx
Comment (jamesqquick) [reply core/app/[locale]/(default)/page.tsx]: Yeah I think specifying Makeswift in the function name is nice!

core/lib/makeswift/client.ts
jorgemoya: `client.getPage(path, { locale })` returns `null`, but `getPageSnapshot` does work, do you know why?

core/app/[locale]/(default)/page.tsx
Comment (jorgemoya) [reply core/app/[locale]/(default)/page.tsx]: Done!

core/lib/makeswift/client.ts
jorgemoya: `getPageSnapshot` calls `client.getPageSnapshot`, I prefer to keep `getMakeswiftPageMetadata` because it handles how to create the metadata object properly so that nextjs templates work.

jorgemoya: `client.getPage` doesn't seem to return metadata: `{ pathname: 'page', locale: 'en-US', alternate: [] }`

jorgemoya: I can use `client.getPages` to get the fields:

```
  await client.getPages({
    siteVersion: await getSiteVersion(),
    locale: normalizeLocale(locale),
    pathPrefix: path,
  });
```  

Is this better than using `getPageSnapshot`?

migueloller: Hmm, that doesn't seem like a great option. Maybe we need to do an API call directly to the endpoint to get a page. Or perhaps contribute a fix so that `getPage` provides the relevant metadata. I know that this might make solving this problem more involved, but IMO it's worth doing right.

jorgemoya: I fail to see the point of calling the in beta REST API directly over using the client, which already returns the information we need in two separate functions. I will open a discussion for this but I'm not sure if having another function available in the client that will return the same information is worth it?

chanceaclark (approved):

PR #2866, opened 2026-02-06
Catalyst 1744 refactor configuration variables

jamesqquick:
## What/Why?
Refactor config variables:

### Rename environment variables from `BIGCOMMERCE_` prefix to `CATALYST_` prefix

`BIGCOMMERCE_STORE_HASH` -> `CATALYST_STORE_HASH`

`BIGCOMMERCE_PROJECT_UUID` -> `CATALYST_PROJECT_UUID`

`BIGCOMMERCE_FRAMEWORK` -> `CATALYST_FRAMEWORK`

`BIGCOMMERCE_ACCESS_TOKEN` - > `CATALYST_ACCESS_TOKEN`

### Implement new priority for config variables

1. Individual parameter flags (e.g. --store-hash)
2. --env-file flag
3. process.env
4. .bigcommerce/project.json

### Add --env-file flag
This allows the user to specify a environment variable file to load variables from

### Remove default environment variable file

Remove the reading of default environment variable files. This requires the user to pass `--file-env` if they want variables to be read from a file.

### Expand project.json

Write additional config variables to project.json file.

`projectUuid`
`framework`
`storeHash`
`accessToken`

### Improved logging
Update logs to details which variables are required and how/where to set them

## Testing
Tests have been added to test basic functionality. One thing the tests do not cover is when the user is passing the `--env-file` flag to specify an environment variable file to read from. For testing, 

- rename environment variables to match `CATALYST_` prefix
- run CLI command (ex. `project list`) without passing any flags. It should fail since no env file is being picked up automatically
- run `catalyst project create` command to generate a new project and verify the correct properties are set in `project.json`
- `catalyst project list` - should now work by reading variables from `project.json`
- run `catalyst project list --access-token dummy-token` and pass dummy variable for access token via `--access-token` flag. This should fail since the explicitly passed access token isn't valid
- create `.env.test` with invalid access token. Then, run `catalyst project list --env-file .env.test` and show that the command fails since it picked up the invalid access token
- test other combinations of variable configurations to confirm priority order is respected


packages/catalyst/src/cli/index.ts
```diff
@@ -1,4 +1,23 @@
...
+import { config } from 'dotenv';
+import { resolve } from 'node:path';
+
 import { program } from './program';
 
+// Load --env-file before parse so option resolution uses this order:
+// 1. Explicit flags  2. --env-file (if passed)  3. process.env (shell)  4. project.json
+const envFileIndex = process.argv.findIndex(
+  (arg) => arg === '--env-file' || arg.startsWith('--env-file='),
+);
```
matthewvolk, line 11: This works but I think it's fragile... I'm not sure it'll handle edge cases like `--env-file` being passed after `--` (which Commander uses to stop option parsing).

If we're ok with that, that's fine. We can always merge and re-evaluate if it starts behaving undesirably.

packages/catalyst/src/cli/commands/link.spec.ts
```diff
@@ -0,0 +1,349 @@
...
+
+  expect(exitMock).toHaveBeenCalledWith(0);
+
+  expect(config.get('projectUuid')).toBe(projectUuid3);
+  expect(config.get('framework')).toBe('catalyst');
+
+  consolaPromptMock.mockRestore();
+});
+
+test('prompts to create a new project', async () => {
```
matthewvolk, line 221: I think this test name is a duplicate of the test on line 164 in this same file. I see the following:

- line 164: `test('prompts to create a new project', ...)` - the happy path where the user selects "create" and it succeeds
- line ~221: `test('prompts to create a new project', ...)` - the error path where the API returns a 502

Maybe we rename the second one to "errors when project creation fails" or something?

packages/catalyst/src/cli/commands/deploy.spec.ts
```diff
@@ -306,19 +306,19 @@ test('--dry-run skips upload and deployment', async () => {
 
 test('reads from env options', () => {
   const envVariables = parseEnvironmentVariables([
-    'BIGCOMMERCE_STORE_HASH=123',
-    'BIGCOMMERCE_STOREFRONT_TOKEN=456',
+    'CATALYST_STORE_HASH=123',
+    'CATALYST_ACCESS_TOKEN=456',
```
matthewvolk, line 310: Was this intentional, to rename `BIGCOMMERCE_STOREFRONT_TOKEN` to `CATALYST_ACCESS_TOKEN`?

packages/catalyst/src/cli/commands/deploy.spec.ts
```diff
@@ -306,19 +306,19 @@ test('--dry-run skips upload and deployment', async () => {
...
     {
       type: 'secret',
-      key: 'BIGCOMMERCE_STORE_HASH',
+      key: 'CATALYST_STORE_HASH',
       value: '123',
     },
     {
       type: 'secret',
-      key: 'BIGCOMMERCE_STOREFRONT_TOKEN',
+      key: 'CATALYST_ACCESS_TOKEN',
```
matthewvolk, line 321: Was this intentional, to rename `BIGCOMMERCE_STOREFRONT_TOKEN` to `CATALYST_ACCESS_TOKEN`?

matthewvolk (changes requested):
⚠️ I just realized now that we're storing access token in `.bigcommerce/`, the root `.gitignore` contains a line to ignore `.bigcommerce` but the `core/.gitignore` does not. Since eventually, `core/.gitignore` will be the ignore file shipped to customers, and they won't have the root ignore file, can we please add `.bigcommerce` to `core/.gitignore`?

chanceaclark (changes requested):
Can we split this up into separate changes for each thing you are doing here?

jamesqquick:
> Can we split this up into separate changes for each thing you are doing here?

Into separate PRs you mean? I wondered about it being too much in one. I can do that!

packages/catalyst/src/cli/commands/deploy.spec.ts
jamesqquick: I actually just realized I made that jump. I think I just had `access_token` in my mind and put that in the different docs, then copied in. I don't think we made a collective decision on that. I think the correct thing to do would be `CATALYST_STOREFRONT_TOKEN`. Does you agree with that?

jamesqquick: Wait ignore the above lol  I get it now. This is supposed to be testing parsing secret runtime variables. Great catch!

jamesqquick:
Splitting this into separate PRs

chanceaclark:
I believe we've done this already in the `alpha` branch. Closing this one out.

PR #2872, opened 2026-02-10
fix(cli): change --secret from variadic to repeatable

matthewvolk:
## What/Why?

The `--secret` flag on the `deploy` command used Commander.js variadic syntax (`--secret <secrets...>`), which accepts space-separated values. This is problematic because secret values themselves could contain spaces.

This changes `--secret` to a repeatable option (`--secret KEY=VALUE --secret KEY2=VALUE2`) using Commander's custom argument parser, making each secret explicit and unambiguous. The parsed result is still a `string[]`, so no changes were needed to `parseEnvironmentVariables` or `createDeployment`.

## Testing

Existing tests pass — `parseEnvironmentVariables` already expects a `string[]` input. Updated the command options test to match the new flag format.

## Migration

**Before:** `catalyst deploy --secret SECRET_1=FOO SECRET_2=BAR`
**After:** `catalyst deploy --secret SECRET_1=FOO --secret SECRET_2=BAR`


jorgemoya (approved):

migueloller (approved):
LGTM 👍🏻

chanceaclark (approved):

PR #2876, opened 2026-02-11
[Catalyst-1744] Write additional variables to project config

jamesqquick:
## What/Why?
Write additional config variables to project.json file so that we have a centralized place for all required variables.

`projectUuid`
`framework`
`storeHash`
`accessToken`

## Testing
`pnpm build --filter @bigcommerce/catalyst` compiles without errors
`pnpm --filter @bigcommerce/catalyst test` — all 44 tests pass

Manually, you can run a `project create` command and verify that the 4 properties above are written to the `project.json` file.

### DIFF
--- a/packages/catalyst/src/cli/commands/deploy.spec.ts
+++ b/packages/catalyst/src/cli/commands/deploy.spec.ts
@@ -336,13 +336,9 @@ test('errors when store hash is missing and not in .bigcommerce/project.json', a
   config.set('accessToken', accessToken);
   config.delete('storeHash');
 
-  const savedStoreHash = process.env.CATALYST_STORE_HASH;
-
-  delete process.env.CATALYST_STORE_HASH;
+  vi.stubEnv('CATALYST_STORE_HASH', undefined);
 
   await program.parseAsync(['node', 'catalyst', 'deploy']);
-
-  if (savedStoreHash !== undefined) process.env.CATALYST_STORE_HASH = savedStoreHash;
 
   expect(consola.error).toHaveBeenCalledWith(
     expect.objectContaining({
@@ -351,6 +347,8 @@ test('errors when store hash is missing and not in .bigcommerce/project.json', a
     }),
   );
   expect(exitMock).toHaveBeenCalledWith(1);
+
+  vi.unstubAllEnvs();
 });
 
 test('errors when access token is missing and not in .bigcommerce/project.json', async () => {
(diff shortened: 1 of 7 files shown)


packages/catalyst/src/cli/commands/deploy.spec.ts
```diff
@@ -312,6 +313,68 @@ test('--dry-run skips upload and deployment', async () => {
...
+  config.set('accessToken', accessToken);
+  config.delete('storeHash');
+
+  const savedStoreHash = process.env.CATALYST_STORE_HASH;
+
+  delete process.env.CATALYST_STORE_HASH;
+
+  await program.parseAsync(['node', 'catalyst', 'deploy']);
+
+  if (savedStoreHash !== undefined) process.env.CATALYST_STORE_HASH = savedStoreHash;
```
matthewvolk: This pattern (here and in the test below) seems a little clumsy to me... can you look into https://vitest.dev/api/vi.html#vi-stubenv and https://vitest.dev/api/vi.html#vi-unstuballenvs ? IMO that's the correct way we should be setting/unsetting env vars in these tests

packages/catalyst/src/cli/commands/deploy.ts
```diff
@@ -373,8 +369,10 @@ export const deploy = new Command('deploy')
   .action(async (options) => {
     try {
       const config = getProjectConfig();
+      const storeHash = options.storeHash ?? config.get('storeHash');
+      const accessToken = options.accessToken ?? config.get('accessToken');
 
-      await telemetry.identify(options.storeHash);
+      await telemetry.identify(storeHash);
```
matthewvolk: Since you removed `makeOptionMandatory` from the `options.storeHash`, you might be sending an undefined identifier when you get to this line

matthewvolk (commented):
A couple changes please!

packages/catalyst/src/cli/commands/deploy.spec.ts
jamesqquick: Oooh I like that! Good call. Updated!

packages/catalyst/src/cli/commands/deploy.ts
jamesqquick: Good point! Updated to check for valid options (storeHash, accessToken, and projectUuid) first before continuignng

packages/catalyst/src/cli/commands/project.ts
```diff
@@ -165,10 +171,19 @@ export const link = new Command('link')
     try {
       const config = getProjectConfig(options.rootDir);
 
-      const writeProjectConfig = (uuid: string) => {
+      const writeProjectConfig = (
```
matthewvolk, line 174: If a user runs `catalyst project link --project-uuid X --store-hash Y --access-token Z`, the store hash and access token won't be persisted, because the `link --project-uuid` path calls `writeProjectConfig(options.projectUuid) — so a subsequent catalyst deploy (without explicit flags) will fail. 

This is probably an edge case, so I'm cool merging as-is and if this becomes a problem we can come back and think it through. Might get easier once we have `catalyst auth` merged too.

matthewvolk (approved):
Looks good 👍

PR #2877, opened 2026-02-11
CATALYST-1744 - Remove default loading of environment variable files

jamesqquick:
## What/Why?
Remove the reading of default environment variable files. 

## Testing
`pnpm build --filter @bigcommerce/catalyst` compiles without errors
`pnpm --filter @bigcommerce/catalyst test` — all tests pass

TO manually test. First, delete an existing `.bigcommerce/project.json` file if it exists. Then, show that running `catalyst project create` will not work if you don't pass variables in explicit flags.

## Migration
Pass environment variable files explicitly using flags in the command or include them in your project.json

### DIFF
--- a/packages/catalyst/src/cli/program.ts
+++ b/packages/catalyst/src/cli/program.ts
@@ -1,7 +1,5 @@
 import { Command } from 'commander';
 import { colorize } from 'consola/utils';
-import { config } from 'dotenv';
-import { resolve } from 'node:path';
 
 import PACKAGE_INFO from '../../package.json';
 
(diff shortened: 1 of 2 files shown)


packages/catalyst/src/cli/program.ts
```diff
@@ -1,7 +1,5 @@
 import { Command } from 'commander';
 import { colorize } from 'consola/utils';
-import { config } from 'dotenv';
```
matthewvolk, line 3: I think this might be the last place we're using `dotenv` in the `@bigcommerce/catalyst` package - if so, can you be sure to run `pnpm --filter "./packages/catalyst" uninstall dotenv` as part of this PR?

matthewvolk (commented):
Just one comment

packages/catalyst/src/cli/program.ts
chanceaclark: I'm thinking that if we include the `--env-file` flag, we can use `dotenv` to load in the env vars onto `process.env` if provided.

jamesqquick: Yeah I left it because it will be used in the other PR for the new flag as @chanceaclark mentioned!

matthewvolk (approved):

PR #2878, opened 2026-02-11
[CATALYST-1744] - add --env-path flag

jamesqquick:
## What/Why?
This allows the user to specify a environment variable file to load variables from

## Testing
Create a `.env.test` file with credentials. Start by running `catalyst project list` with no variables and show that the command fails. Then, run `catalyst project list --env-path .env.test` with valid variables and show that it works.

To test the priority of variables loaded from the file taking precedence over existing `process.env` variables, set an access token in environment variable with an invalid take. `CATALYST_ACCESS_TOKEN=dummy_token`. Run the `catalyst project list` command again, and prove that it still works.

## Migration
If you want to load environment variables from a file, update your commands to include `--env-path` flag.

### DIFF
--- a/packages/catalyst/src/cli/program.ts
+++ b/packages/catalyst/src/cli/program.ts
@@ -16,19 +16,14 @@ import { version } from './commands/version';
 import { telemetryPostHook, telemetryPreHook } from './hooks/telemetry';
 import { consola } from './lib/logger';
 
-/**
- * Load env file from --env-file in argv before Commander parses.
- * Must run before program.parse() so process.env is populated when
- * Commander reads .env('VAR') for option defaults.
- */
 export function loadEnvFileFromArgv(argv: string[]): void {
   const envFileIdx = argv.findIndex((arg) => arg === '--env-file' || arg.startsWith('--env-file='));
 
   if (envFileIdx === -1) return;
 
-  const arg = argv[envFileIdx];
-  const value = arg.startsWith('--env-file=')
-    ? arg.slice('--env-file='.length)
+  const envFileArg = argv[envFileIdx];
+  const value = envFileArg.startsWith('--env-file=')
+    ? envFileArg.slice('--env-file='.length)
     : argv[envFileIdx + 1];
 
   if (value && !value.startsWith('-')) {
@@ -37,8 +32,9 @@ export function loadEnvFileFromArgv(argv: string[]): void {
 
     if (result.error) {
       const err = result.error as NodeJS.ErrnoException;
-      console.log(result.error?.message);
-      console.log(result.error?.name);
+
+      console.log(result.error.message);
+      console.log(result.error.name);
       consola.warn(
         err.code === 'ENOENT'
           ? `Env file not found: ${resolvedPath}`
--- a/packages/catalyst/src/cli/program.ts
+++ b/packages/catalyst/src/cli/program.ts
@@ -1,5 +1,7 @@
 import { Command } from 'commander';
 import { colorize } from 'consola/utils';
+import { config } from 'dotenv';
+import { resolve } from 'node:path';
 
 import PACKAGE_INFO from '../../package.json';
 
@@ -16,12 +18,33 @@ import { consola } from './lib/logger';
 
 export const program = new Command();
 
+function loadEnvFilePreHook(thisCommand: Command) {
+  let root: typeof thisCommand | undefined = thisCommand;
+
+  while (root.parent) {
+    root = root.parent as typeof thisCommand;
+  }
+
+  const opts = root.opts() as { envFile?: string };
+  const envFile = opts.envFile;
+
+  if (envFile) {
+    const resolvedPath = resolve(process.cwd(), envFile);
+
+    config({ path: resolvedPath, override: true });
+  }
+}
+
 consola.log(colorize('cyanBright', `◢ ${PACKAGE_INFO.name} v${PACKAGE_INFO.version}\n`));
 
 program
   .name(PACKAGE_INFO.name)
   .version(PACKAGE_INFO.version)
   .description('CLI tool for Catalyst development')
+  .option(
+    '--env-file <path>',
+    'Path to environment file to load (relative to current working directory)',
+  )
   .addCommand(version)
   .addCommand(dev)
   .addCommand(start)
@@ -30,5 +53,6 @@ program
   .addCommand(project)
   .addCommand(auth)
   .addCommand(telemetry)
+  .hook('preAction', loadEnvFilePreHook)
   .hook('preAction', telemetryPreHook)
   .hook('postAction', telemetryPostHook);
--- a/packages/catalyst/src/cli/program.ts
+++ b/packages/catalyst/src/cli/program.ts
@@ -31,12 +31,15 @@ export function loadEnvFileFromArgv(argv: string[]): void {
     const result = config({ path: resolvedPath, override: true });
 
     if (result.error) {
-      const err = result.error as NodeJS.ErrnoException;
+      const errCode =
+        'code' in result.error && typeof result.error.code === 'string'
+          ? result.error.code
+          : undefined;
 
       console.log(result.error.message);
       console.log(result.error.name);
       consola.warn(
-        err.code === 'ENOENT'
+        errCode === 'ENOENT'
           ? `Env file not found: ${resolvedPath}`
           : `Failed to load --env-file ${value}: ${result.error.message}`,
       );
--- a/packages/catalyst/src/cli/program.ts
+++ b/packages/catalyst/src/cli/program.ts
@@ -1,8 +1,8 @@
+import { Option } from '@commander-js/extra-typings';
 import { Command } from 'commander';
 import { colorize } from 'consola/utils';
 import { config } from 'dotenv';
 import { resolve } from 'node:path';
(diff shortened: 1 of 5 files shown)


packages/catalyst/src/cli/program.ts
```diff
@@ -13,6 +15,37 @@ import { version } from './commands/version';
...
+    const result = config({ path: resolvedPath, override: true });
+
+    if (result.error) {
+      const errCode =
+        'code' in result.error && typeof result.error.code === 'string'
+          ? result.error.code
+          : undefined;
+
+      console.log(result.error.message);
+      console.log(result.error.name);
```
matthewvolk: Are these leftover debug statements from testing? Looks like you already have a `console.warn` below for user-facing error messages.

packages/catalyst/src/cli/program.ts
```diff
@@ -13,6 +15,37 @@ import { version } from './commands/version';
...
+      consola.warn(
```
matthewvolk: Is `console.warn` without crashing the program the behavior we want? If the user supplied an invalid `--env-file` I'd be willing to bet they don't have/don't want to use fallback values in `process.env` or `.bigcommerce/project.json`. So should we just crash the program instead?

packages/catalyst/src/cli/program.ts
```diff
@@ -13,6 +15,37 @@ import { version } from './commands/version';
 import { telemetryPostHook, telemetryPreHook } from './hooks/telemetry';
 import { consola } from './lib/logger';
 
+export function loadEnvFileFromArgv(argv: string[]): void {
```
matthewvolk: This whole function is a  bit odd to me... but if you tested locally and it behaves as you'd like, you can ignore this comment. Since we're manually parsing args in this function, does it conflict with commander.js's default `parse()` behavior? Why do we need this manual parsing function?

matthewvolk (commented):
Just a big question about the `loadEnvFileFromArgv` function, and whether `console.warn` is the behavior we want.

packages/catalyst/src/cli/program.ts
jamesqquick: As far as I can tell, this is necessary to customize the order of precedence that we want. The specific tricky part is between variables loaded from env file and ones loaded from `process.env`. If we didn't do this custom implementation, it would take the `process.env` variables without factoring in the ones loaded from `--env-file`. 

If you just take default `parse` behavior, it would load `process.env` variables, then just pass us the path to the file from `--env-file`. Since the variables had already been loaded by commander and included as options, us then loading a manual env file wouldn't be tapped into the options that commander has already loaded.

packages/catalyst/src/cli/program.ts
jamesqquick: Yep, they are! I'll delete

packages/catalyst/src/cli/program.ts
jamesqquick: I think that's fair. Will update!

packages/catalyst/src/cli/program.ts
matthewvolk: Gotcha, this makes sense when I think about it. `options.storeHash` could be defined either from the `--store-hash` flag or the `process.env` var, and since we can't detect which one it set, we have to jailbreak essentially.

packages/catalyst/src/cli/program.ts
```diff
@@ -13,6 +15,36 @@ import { version } from './commands/version';
...
+          ? result.error.code
+          : undefined;
+
+      const message =
+        errCode === 'ENOENT'
+          ? `Env file not found: ${resolvedPath}`
+          : `Failed to load --env-file ${value}: ${result.error.message}`;
+
+      throw new Error(message);
+    }
```
matthewvolk: 🍹 Not a common edge case, but know that if a user does something like `--env-file --other-option` (basically forgets to add a value for --env-file), this function silently does nothing. Might be an opportunity to add something like:

```js
if (!value || value.startsWith('-')) {
  throw new Error('--env-file requires a path argument');
}
```

packages/catalyst/src/cli/program.ts
```diff
@@ -13,6 +15,36 @@ import { version } from './commands/version';
...
+        'code' in result.error && typeof result.error.code === 'string'
+          ? result.error.code
+          : undefined;
+
+      const message =
+        errCode === 'ENOENT'
+          ? `Env file not found: ${resolvedPath}`
+          : `Failed to load --env-file ${value}: ${result.error.message}`;
+
+      throw new Error(message);
```
matthewvolk: 🍹 The throw new Error() from loadEnvFileFromArgv will surface as an unhandled exception with a full stack trace — not great UX for a CLI. Wrap it in a try/catch for a clean exit

```js
try {
  loadEnvFileFromArgv(process.argv);
} catch (e) {
  consola.error(e instanceof Error ? e.message : e);
  process.exit(1);
}
```

packages/catalyst/src/cli/program.ts
matthewvolk: 🍹 Might be worth adding tests in the future for this if it becomes confusing

matthewvolk (approved):

packages/catalyst/src/cli/program.ts
chanceaclark: What about something like this instead of this function?
```ts
  .addOption(
    new Option(
      '--env-file <path>',
      'Path to environment file to load (relative to current working directory)',
    ),
  )
  .hook('preAction', (thisCommand) => {
    const envFile = thisCommand.opts().envFile;

    if (envFile) {
      const envFilePath = resolve(process.cwd(), envFile);

      config({
        path: [envFilePath],
      });

      consola.log(colorize('cyanBright', `Loaded environment variables from ${envFilePath}\n`));
    }
  })
```

chanceaclark: I tested above ^ already and it works great:
```bash
 ~/repositories/catalyst/packages/catalyst  canary  node dist/cli.js version                                                                                                                                                                                                                                                                                                                                               ✔ 
◢ @bigcommerce/catalyst v0.1.0                                                                                                                                                                                                                                                                                                                                                                                       2:28:09 PM

Version Information:                                                                                                                                                                                                                                                                                                                                                                                                 2:28:09 PM
CLI Version: 0.1.0                                                                                                                                                                                                                                                                                                                                                                                                   2:28:09 PM
Node Version: v24.13.1                                                                                                                                                                                                                                                                                                                                                                                               2:28:09 PM
Platform: darwin (arm64)                                                                                                                                                                                                                                                                                                                                                                                             2:28:09 PM

 ~/repositories/catalyst/packages/catalyst  canary  node dist/cli.js version --env-file=../../.env.local                                                                                                                                                                                                                                                                                                                   ✔ 
◢ @bigcommerce/catalyst v0.1.0                                                                                                                                                                                                                                                                                                                                                                                       2:28:38 PM

Loaded environment variables from /Users/chancellor.clark/repositories/catalyst/.env.local                                                                                                                                                                                                                                                                                                                           2:28:38 PM

Version Information:                                                                                                                                                                                                                                                                                                                                                                                                 2:28:38 PM
CLI Version: 0.1.0                                                                                                                                                                                                                                                                                                                                                                                                   2:28:38 PM
Node Version: v24.13.1                                                                                                                                                                                                                                                                                                                                                                                               2:28:38 PM
Platform: darwin (arm64)
```

jamesqquick: @chanceaclark did you test that? I just tried that locally, and I'm seeing that `process.env` is still taking precedency over the environment variable file.

chanceaclark: I removed everything in this PR then added those lines. I then rebuilt the CLI `NODE_ENV="production" pnpm run build` then ran the script `node ./dist/cli.js version`.

_Edit:_ I also added the following to the `tsup.config.ts` file, as I was getting some issues:
```
external: ['commander', '@commander-js/extra-typings'],
```

packages/catalyst/src/cli/program.ts
jamesqquick: It looks like commander handles that by default interestingly. Do you think it's still worth adding a check?
CleanShot 2026-02-16 at 15 56 04@2x

matthewvolk (changes requested):
Ok, figured out the issue with `--env-file`:

Node.js itself added support for the same exact flag starting in Node 20.6.0:
https://nodejs.org/api/environment_variables.html#env-files

Node is intercepting --env-file .en and trying to load it _before your script even executes._

When you pass a valid file, Node loads it successfully, then your command's own `--env-file` handling in `loadEnvFileFromArgv` processes it a second time.

TL;DR, I think the best option is to rename the flag, e.g., `--env-path` (is that confusing?). I don't think there's a clean way to disable Node.js's built-in flag parsing, and honestly I don't think we'd want to.

packages/catalyst/src/cli/program.ts
jamesqquick: updated!

chanceaclark (approved):
LGTM with the approach.

jorgemoya (changes requested):
Missing some tests just to make sure this is working as expected.

jorgemoya (approved):

matthewvolk (approved):

PR #2879, opened 2026-02-11
feat(cli): map ignition error codes to actionable messages

matthewvolk:
## What/Why?

When a deployment fails via the SSE event stream, users currently see generic messages like `Deployment failed with error code: 30` with no explanation of what went wrong or how to fix it.

This maps the 6 known Ignition error codes (10–60) to human-readable, actionable messages. The Ignition gRPC API only surfaces numeric codes — no error message strings — so the CLI owns the entire mapping.

**Jira:** [CATALYST-1724](https://bigcommercecloud.atlassian.net/browse/CATALYST-1724)

### Changes

- **New `ignition-errors.ts`** — error code map + `getIgnitionErrorMessage()` function with fallback for unknown codes
- **Updated `deploy.ts`** — import + use `getIgnitionErrorMessage` in the deployment error handler
- **New `ignition-errors.spec.ts`** — unit tests for all 6 known codes + unknown code fallback
- **Updated `deploy.spec.ts`** — updated "handles deployment errors" assertion to match new message format

## Testing

```bash
cd packages/catalyst && pnpm exec vitest run src/cli/lib/ignition-errors.spec.ts src/cli/commands/deploy.spec.ts
```

All 21 tests pass (7 new + 14 existing).

## Migration

N/A — no breaking changes.

[CATALYST-1724]: https://bigcommercecloud.atlassian.net/browse/CATALYST-1724?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/catalyst/src/cli/lib/deployment-errors.ts
+++ b/packages/catalyst/src/cli/lib/deployment-errors.ts
@@ -10,7 +10,7 @@ const ERROR_MESSAGES: Record<number, string> = {
   60: "Your code was deployed, but we couldn't determine your deployment URL. Please try deploying again. If the issue persists, contact support.",
 };
 
-export function getIgnitionErrorMessage(code: number): string {
+export function getDeploymentErrorMessage(code: number): string {
   return (
     ERROR_MESSAGES[code] ??
     `Deployment failed with an unexpected error (code: ${code}). Please try again. If the issue persists, contact support.`
--- a/packages/catalyst/src/cli/lib/ignition-errors.ts
+++ b/packages/catalyst/src/cli/lib/ignition-errors.ts
@@ -1,0 +1,18 @@
+const ERROR_MESSAGES: Record<number, string> = {
+  10: 'Something went wrong on our end. Please try again. If the issue persists, contact support.',
+  20: "We couldn't retrieve your bundle. This is usually a temporary issue — please try deploying again. If the problem continues, contact support.",
+  30: 'Your bundle could not be extracted. This may mean your build output is too large (max 64 MB compressed / 512 MB uncompressed) or the archive is corrupted. Try reducing your build size or rebuilding your project and deploying again.',
+  40: `There's a problem with your build output. This could be caused by:
+- A worker.js file larger than 40 MB
+- An individual asset file larger than 25 MB
+- More than 1,000 total files in the bundle`,
+  50: 'Deployment failed. This is usually a temporary issue — please try again. If the problem persists, contact support.',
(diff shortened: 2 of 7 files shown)


packages/catalyst/src/cli/lib/deployment-errors.ts
chanceaclark, line 1: Do you think this might be better to have an error class instead? For example,
```ts
export class DeploymentError extends Error {
  constructor(code: number) {
    super(`Deployment failed (error code ${code}): ${this.getErrorMessage(code)}`);

    this.name = "DeploymentError";
  }

  private getErrorMessage(code: number) {...}
}

// ...

throw new DeploymentError(data.error.code)

```

packages/catalyst/src/cli/lib/ignition-errors.ts
```diff
@@ -0,0 +1,18 @@
...
+  30: 'Your bundle could not be extracted. This may mean your build output is too large (max 64 MB compressed / 512 MB uncompressed) or the archive is corrupted. Try reducing your build size or rebuilding your project and deploying again.',
+  40: `There's a problem with your build output. This could be caused by:
+- A worker.js file larger than 40 MB
+- An individual asset file larger than 25 MB
+- More than 1,000 total files in the bundle`,
+  50: 'Deployment failed. This is usually a temporary issue — please try again. If the problem persists, contact support.',
+  60: "Your code was deployed, but we couldn't determine your deployment URL. Please try deploying again. If the issue persists, contact support.",
+};
+
+export function getIgnitionErrorMessage(code: number): string {
```
chanceaclark: "Ignition" is kinda an internal name, we should probably remove that here.

jordanarldt (approved):

packages/catalyst/src/cli/lib/ignition-errors.ts
jordanarldt: Good call here

packages/catalyst/src/cli/lib/deployment-errors.ts
matthewvolk: I'd agree with you if, (1) we needed something like the error code to be programmatically accessible (e.g., `error.code`) — but right now nothing reads it, it's just a function used to interpolate the human-readable message string into the error thrown by the command. Or (2) if callers would benefit from `instanceof` but since we're just printing a message I think we can skip and add later if any of that changes

packages/catalyst/src/cli/lib/ignition-errors.ts
matthewvolk: Updated all occurrences 👍

PR #2880, opened 2026-02-13
feat(cli): add crash reporting with trace ids

matthewvolk:
## What/Why?

When the CLI crashes, customers send screenshots to support which aren't actionable. This PR instruments the CLI so that on any error, a **trace ID** (UUID) is displayed. The customer shares it with support, who can look up the session across telemetry systems.

This works because every outgoing BigCommerce API request now includes an `X-Correlation-Id` header set to the trace ID.

### Changes

**Centralized error handling**
- New `withErrorHandler()` HOF that wraps Commander `.action()` callbacks
- On error: tracks via telemetry, displays trace ID, exits with code 1
- Removes duplicated try/catch + `consola.error` + `process.exit(1)` from all 6 command actions (deploy, build, dev, start, project list/create/link)

**Telemetry singleton**
- `getTelemetry()` / `resetTelemetry()` singleton pattern replaces per-module `new Telemetry()` instances
- `sessionId` changed from `randomBytes(32).toString('hex')` to `randomUUID()` for a human-readable, support-friendly format
- New `traceId()`, `durationMs()`, and `trackError()` methods
- All `track()` calls enriched with `traceId`, `nodeVersion`, `platform`, `arch`, `cliVersion`

**API correlation**
- `X-Correlation-Id` header added to all 6 BigCommerce API fetch calls (4 in deploy.ts, 2 in project.ts)

**Global safety net**
- `uncaughtException` and `unhandledRejection` handlers in `index.ts`

**User-facing output on crash:**
```
Error: Failed to fetch upload signature: 401 Unauthorized

Trace ID: a1b2c3d4-e5f6-7890-abcd-ef1234567890
Share this Trace ID with BigCommerce support.
```

## Testing

- 67 tests passing across 13 test files
- New test suites for `error-handler.spec.ts` and `telemetry.spec.ts`
- Updated assertions in `deploy.spec.ts` and `project.spec.ts`
- Lint passes with 0 errors/warnings

## Migration

N/A — no breaking changes to public API or file structure.


.changeset/brave-tigers-trace.md
chanceaclark, line 1: You can omit the changeset here

packages/catalyst/src/cli/lib/telemetry.ts
```diff
@@ -41,6 +51,11 @@ export class Telemetry {
       properties: {
         ...payload,
         sessionId: this.sessionId,
+        traceId: this.traceId(),
```
chanceaclark: Trace id and session id above are the same thing. Can we just normalize this all to use session id internally but show it as trace id externally?

packages/catalyst/src/cli/hooks/telemetry.ts
```diff
@@ -46,5 +45,8 @@ export const telemetryPreHook = async (command: Command) => {
 };
 
 export const telemetryPostHook = async () => {
```
chanceaclark, line 61: Are there arguments on this that we can determine if the command failed with an error? It would be cleaner than having to add a closure around each command and is more scalable.

packages/catalyst/src/cli/lib/telemetry.ts
```diff
@@ -51,6 +66,18 @@ export class Telemetry {
     });
   }
 
+  async trackError(commandName: string, error: unknown) {
```
chanceaclark: Do we need to add a whole new function if we are only calling it once? Can we just use the existing track function?

packages/catalyst/src/cli/lib/telemetry.ts
```diff
@@ -99,3 +126,15 @@ export class Telemetry {
     return generated;
   }
 }
+
+let telemetryInstance: Telemetry | undefined;
```
chanceaclark, line 114: Can you explain why we need a singleton here?

chanceaclark (commented):
I feel like a lot of this PR can be simplified. Let's work together tomorrow to trim it down.

packages/catalyst/src/cli/lib/telemetry.ts
matthewvolk: Yup!

packages/catalyst/src/cli/lib/telemetry.ts
matthewvolk: I'll combine 👍

packages/catalyst/src/cli/lib/telemetry.ts
matthewvolk: Right now it's because the pre-hook, post-hook, error handler, and command action callback (for X-Correlation-Id) all call getTelemetry() independently — without the singleton they'd each get a fresh sessionId... though after I incorporate your feedback to leverage the postHook more, maybe this changes

packages/catalyst/src/cli/hooks/telemetry.ts
matthewvolk: Ok I was actually just playing around with this in a sandbox and here's what I found:

```js
const { Command } = require('commander');

const p = new Command();

p.command('test').action(async () => { throw new Error('boom'); });
p.hook('postAction', () => { console.log('POST_ACTION_FIRED'); });
p.parseAsync(['node', 'test', 'test']).catch(e => console.log('CAUGHT:', e.message));
```

If you run that, you don't see `POST_ACTION_FIRED`. I guess hooks in Commander work like `.then()`?

TL;DR... to your original question, I don't think so... but open to your thoughts

packages/catalyst/src/cli/lib/telemetry.ts
matthewvolk: Actually, see https://github.com/bigcommerce/catalyst/pull/2880#discussion_r2805046859

matthewvolk:
@chanceaclark just updated, let me know what you think 👍

packages/catalyst/src/cli/lib/telemetry.ts
chanceaclark: I think a fresh session per command call is expected? As in you call the CLI once, you get a new session, then call it again it's a separate session.

matthewvolk: No, I don't want the trace ID to change throughout the command execution

packages/catalyst/src/cli/index.ts
```diff
@@ -1,4 +1,48 @@
...
+  void handleFatalError(error);
+});
+
+process.on('unhandledRejection', (reason) => {
+  void handleFatalError(reason);
+});
+
+try {
+  await program.parseAsync(process.argv);
+  await getTelemetry().analytics.closeAndFlush();
```
chanceaclark: Won't this get handled by the post hook?
```suggestion
```

packages/catalyst/src/cli/index.ts
```diff
@@ -1,4 +1,48 @@
...
+    // Don't mask the original error
+  }
+
+  consola.error(errorMessage);
+
+  const traceMessage = telemetry.isEnabled()
+    ? 'Share this Trace ID with BigCommerce support.'
+    : 'Enable telemetry (`catalyst telemetry enable`) so this Trace ID can be looked up by support.';
+
+  consola.info(`\nTrace ID: ${telemetry.sessionId}\n${traceMessage}`);
```
chanceaclark: Should we only show the trace id whenever telemetry is enabled? Users might be weirded out that a trace id is showing when telemetry is disabled.

matthewvolk: Good call, just updated

packages/catalyst/src/cli/index.ts
matthewvolk: Good call, just updated

chanceaclark (approved):

PR #2881, opened 2026-02-13
feat(cli): add whoami command

matthewvolk:
## What/Why?

CATALYST-1759: Adds a `catalyst whoami` command that verifies stored credentials and displays store/project info.

Credential resolution follows the same priority as other CLI commands: CLI flags → env vars (`CATALYST_STORE_HASH`, `CATALYST_ACCESS_TOKEN`) → `.bigcommerce/project.json` config. Calls the V2 Store Information API to verify the token and fetch the store name, and optionally looks up the linked project name if `projectUuid` is set in config.

**Dependency on #2876 (CATALYST-1744):** This branch adds `storeHash`/`accessToken` to `ProjectConfigSchema` so it works standalone. The rebase onto `alpha` after #2876 merges will be a trivial conflict.

### Changes
- **`packages/catalyst/src/cli/commands/whoami.ts`** — new `whoami` command
- **`packages/catalyst/src/cli/lib/project-config.ts`** — add `storeHash`/`accessToken` to schema (matches #2876)
- **`packages/catalyst/src/cli/program.ts`** — register `whoami` command
- **`packages/catalyst/tests/mocks/handlers.ts`** — add MSW handler for `GET /v2/store`

## Testing

```bash
# no credentials — helpful message
catalyst whoami

# with flags — displays store info
catalyst whoami --store-hash <hash> --access-token <token>

# invalid token — reports auth error
catalyst whoami --store-hash <hash> --access-token bad-token

# via env vars
CATALYST_STORE_HASH=xxx CATALYST_ACCESS_TOKEN=yyy catalyst whoami
```

Build, lint, and all 47 existing tests pass.

## Migration

None.

### DIFF
--- a/.changeset/brave-maps-arrive.md
+++ b/.changeset/brave-maps-arrive.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst": minor
+---
+
+Add `catalyst whoami` command to verify stored credentials and display store/project info.
--- a/packages/catalyst/src/cli/commands/auth.ts
+++ b/packages/catalyst/src/cli/commands/auth.ts
@@ -1,0 +1,106 @@
+import { Command, Option } from 'commander';
+import { z } from 'zod';
+
+import { consola } from '../lib/logger';
+import { fetchProjects } from '../lib/project';
+import { getProjectConfig } from '../lib/project-config';
+
+const StoreProfileSchema = z.object({
+  data: z.object({
+    store_name: z.string(),
+  }),
+});
+
+async function fetchStoreProfile(storeHash: string, accessToken: string, apiHost: string) {
+  const response = await fetch(`https://${apiHost}/stores/${storeHash}/v3/settings/store/profile`, {
+    method: 'GET',
+    headers: {
+      'X-Auth-Token': accessToken,
+      Accept: 'application/json',
+    },
+  });
+
+  if (!response.ok) {
+    throw new Error(`${response.status} ${response.statusText}`);
+  }
+
+  const res: unknown = await response.json();
+  const { data } = StoreProfileSchema.parse(res);
+
+  return data;
+}
+
+const whoami = new Command('whoami')
+  .description('Verify stored credentials and display store/project info.')
+  .addOption(
+    new Option(
+      '--store-hash <hash>',
+      'BigCommerce store hash. Can be found in the URL of your store Control Panel.',
+    ).env('CATALYST_STORE_HASH'),
+  )
+  .addOption(
+    new Option(
+      '--access-token <token>',
+      'BigCommerce access token. Can be found after creating a store-level API account.',
+    ).env('CATALYST_ACCESS_TOKEN'),
+  )
+  .addOption(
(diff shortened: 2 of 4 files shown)


packages/catalyst/src/cli/commands/auth.ts
```diff
@@ -0,0 +1,101 @@
...
+import { getProjectConfig } from '../lib/project-config';
+
+const StoreInfoSchema = z.object({
+  id: z.string(),
+  name: z.string(),
+  domain: z.string(),
+});
+
+async function fetchStoreInformation(storeHash: string, accessToken: string, apiHost: string) {
+  const response = await fetch(`https://${apiHost}/stores/${storeHash}/v2/store`, {
```
chanceaclark: Hmm is there another api we could use instead of a V2 endpoint?

.changeset/brave-maps-arrive.md
chanceaclark, line 1: Don't need a changeset yet for this stuff

packages/catalyst/src/cli/commands/auth.ts
matthewvolk: I'm not aware of one for getting store name

.changeset/brave-maps-arrive.md
matthewvolk: Even if this PR has `alpha` as the base branch? I thought if base branch === canary, no changeset, but if base branch === alpha, we can add it

packages/catalyst/src/cli/commands/auth.ts
matthewvolk: Might be able to use Admin GraphQL API, if it accepts the same token in `Authorization` header as I'm passing here in `X-Auth-Token`

matthewvolk: Actually, TIL https://developer.bigcommerce.com/docs/rest-management/settings/store-profile 

I'll update

matthewvolk: Fixed

.changeset/brave-maps-arrive.md
chanceaclark: Comment made before our convo this morning. Yeah it's fine to leave if merging in with the `alpha` branch.

packages/catalyst/src/cli/commands/auth.ts
```diff
@@ -0,0 +1,106 @@
...
+  return data;
+}
+
+const whoami = new Command('whoami')
+  .description('Verify stored credentials and display store/project info.')
+  .addOption(
+    new Option(
+      '--store-hash <hash>',
+      'BigCommerce store hash. Can be found in the URL of your store Control Panel.',
+    ).env('CATALYST_STORE_HASH'),
```
chanceaclark, line 39: 🍹 At this point, I think it might be time to do a refactor and abstract this to a global variable that we can inject in when needed. For example:
```ts
.addOption(storeHashOption)
```

chanceaclark (approved):

PR #2882, opened 2026-02-13
feat(ci): enable unlighthouse performance audits for vercel

jorgemoya:
## What/Why?
Re-enable the performance category in Unlighthouse with mitigations for hardware throttling: maxConcurrency=1 and samples=3 for stable medians. Limit testing to known paths to remove redundant testing (PDP and PLP). Rename artifacts with a "vercel" label for future cross-platform comparison.

## Testing
Unlighthouse mobile and desktop reports below

## Migration
N/A

### DIFF
--- a/unlighthouse.config.ts
+++ b/unlighthouse.config.ts
@@ -15,6 +15,28 @@ export default {
     // Run each page multiple times and use the median to absorb cold start
     // outliers across all discovered pages.
     samples: 3,
+    // Only audit one representative URL per unique page template. PLP/PDP and
+    // other dynamic routes share the same layout, so scanning more instances of
+    // the same template yields redundant data.
+    include: [
+      "/",
+      "/fog-linen-chambray-towel-beige-stripe", // PDP
+      "/bath", // PLP categories
+      "/brands/ofs", // PLP brands
+      "/search",
+      "/cart",
+      "/login",
+      "/login/forgot-password",
+      "/register",
+      "/blog",
+      "/your-first-blog-post", // Blog post
+      "/compare",
+      "/gift-certificates", // Gift certificates page
+      "/gift-certificates/balance", // Gift certificates balance page
+      "/gift-certificates/purchase", // Gift certificates purchase page
+      "/contact-us",
+      "/shipping-returns",
+    ],
   },
   lighthouseOptions: {
     onlyCategories: ["best-practices", "accessibility", "seo", "performance"],
--- a/unlighthouse.config.ts
+++ b/unlighthouse.config.ts
@@ -31,9 +31,8 @@ export default {
       "/smith-journal-13/|/dustpan-brush/|/utility-caddy/|/canvas-laundry-cart/|/laundry-detergent/|/tiered-wire-basket/|/oak-cheese-grater/|/1-l-le-parfait-jar/|/chemex-coffeemaker-3-cup/|/sample-able-brewing-system/|/orbit-terrarium-small/|/orbit-terrarium-large/|/fog-linen-chambray-towel-beige-stripe/|/zz-plant/":
         { name: "PDP" },
       "/shop-all/|/bath/|/garden/|/kitchen/|/publications/|/early-access/": {
-        name: "Categories PLP",
+        name: "PLP",
       },
-      "/brands.*": { name: "Brands PLP" },
     },
     // Disable throttling to avoid issues with cold start and cold cache.
     throttle: false,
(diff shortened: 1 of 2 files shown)


unlighthouse.config.ts
```diff
@@ -1,27 +1,57 @@
...
+    maxConcurrency: 1,
+  },
+  scanner: {
+    // Run each page multiple times and use the median to absorb cold start
+    // outliers across all discovered pages (no explicit warm-up step needed).
+    samples: 3,
+    // Only audit one representative URL per unique page template. PLP/PDP and
+    // other dynamic routes share the same layout, so scanning more instances of
+    // the same template yields redundant data.
+    include: [
```
chanceaclark: 🍹 I am generally fine with this. Playing devils advocate, I think running it on all pages helps with the sample size and averages the overall Lighthouse scores to be better reflective of the overall health. Up to you, but just my 🪙 🪙

chanceaclark (approved):

unlighthouse.config.ts
jorgemoya: Understood. I will test this a little more. Problem with having multiple runs and no concurrency is that it adds up if we start to add all PDPs/PLPs. But now that I think about it, adding multiple passes with concurrency enabled should help diminish the issue of having throttled tests, since it will average the runs.

jorgemoya: I tried the sample size and concurrency, and it took a long time to complete. I feel this only works if the sample size is 1.

jorgemoya:
@chanceaclark updated the config to disable throttiling, add custom sampling and dynamic sampling.

unlighthouse.config.ts
```diff
@@ -1,27 +1,52 @@
...
+      "/invoices/",
+      "/bath/*/*",
+      "/garden/*/*",
+      "/kitchen/*/*",
+      "/publications/*/*",
+      "/early-access/*/*",
+      "/digital-test-product/",
+      "/blog/\\?tag=*",
+    ],
+    customSampling: {
```
chanceaclark, line 30: Should we also add the home page, content pages, blog pages, shopper registration/login pages, etc...?

unlighthouse.config.ts
```diff
@@ -1,27 +1,52 @@
...
+    samples: 3,
+    dynamicSampling: 5,
+    exclude: [
+      "/bundleb2b/",
+      "/invoices/",
+      "/bath/*/*",
+      "/garden/*/*",
+      "/kitchen/*/*",
+      "/publications/*/*",
+      "/early-access/*/*",
```
chanceaclark, line 26: How does this work? We are excluding them here, but have them in custom sampling below?

unlighthouse.config.ts
```diff
@@ -1,27 +1,52 @@
-
-import type { UserConfig } from 'unlighthouse';
+import type { UserConfig } from "unlighthouse";
 
 export default {
   ci: {
     buildStatic: true,
-    // Disabling the budget so we can audit and fix the issues first
+    reporter: "jsonExpanded",
```
chanceaclark, line 6: Curious about this change?

jorgemoya: `jsonExpanded` includes CWV metrics.

unlighthouse.config.ts
jorgemoya: We exclude child categories, mostly because these are empty.

unlighthouse.config.ts
jorgemoya: This is only to tell Unlighthouse that these pages are of the same group, since they have custom paths.

I guess for `brands` this is not needed since Unlighthouse should detect the `/brand/` path and group accordingly.

unlighthouse.config.ts
chanceaclark: Ah, perfect

unlighthouse.config.ts
chanceaclark: What about blogs?

jorgemoya: For blogs it should be picked up automatically too: https://unlighthouse.dev/guide/guides/dynamic-sampling#how-it-works 

I will remove brands since `/brands/` is part of the path tree so it should pick it up easily.

jorgemoya: Basically, if we used the internal urls `/products/1`, `/products/2` etc we wouldn't need to do this.

chanceaclark: Ah okay

chanceaclark (approved):

PR #2887, opened 2026-02-16
feat(cli): add auth login and auth logout commands

matthewvolk:
## What/Why?

Adds `catalyst auth login` and `catalyst auth logout` subcommands to the CLI. Currently, developers must manually provide `--store-hash` and `--access-token` flags (or environment variables) for every authenticated command. These new commands use the OAuth device code flow to authenticate via browser and store credentials locally in `.bigcommerce/project.json`.

- **`auth login`** — Requests a device code, displays a one-time code, opens the browser to the BigCommerce login page, polls until the user authorizes, then stores `storeHash` and `accessToken` in the project config.
- **`auth logout`** — Deletes `storeHash` and `accessToken` from the project config.

Both are registered as sibling subcommands alongside the existing `auth whoami`.

### Changes

- **`src/cli/lib/auth.ts`** (new) — Pure functions for the OAuth device flow: `requestDeviceCode`, `pollDeviceToken`, `waitForDeviceToken`, plus Zod schemas and constants (client ID, scopes, default login URL).
- **`src/cli/commands/auth.ts`** (modified) — Added `login` and `logout` subcommands with spinner, browser opening (`open` package), and error handling matching existing patterns.
- **`tests/mocks/handlers.ts`** (modified) — Added MSW handler for `POST /device/token` that distinguishes initial vs poll requests.
- **`src/cli/lib/auth.spec.ts`** (new) — 9 unit tests for the auth lib functions.
- **`src/cli/commands/auth.spec.ts`** (new) — 11 tests covering whoami, login (happy path, already logged in, API error, browser failure, spinner), and logout.
- **`src/cli/index.spec.ts`** (modified) — Added `auth` and its subcommands to expected command assertions.
- **`package.json`** (modified) — Added `open` dependency (`^10.1.0`).

## Testing

- `pnpm --filter @bigcommerce/catalyst test` — 82/82 tests pass
- `pnpm --filter @bigcommerce/catalyst typecheck` — clean
- `pnpm --filter @bigcommerce/catalyst lint` — clean (0 errors, 0 warnings)
- Manual: `catalyst auth login` displays device code and opens browser
- Manual: `catalyst auth logout` clears credentials
- Manual: `catalyst auth whoami` still works as before

## Demo

https://github.com/user-attachments/assets/3d37eaf3-055e-4388-af1a-274421d4611a

## Migration

No breaking changes. The `auth` parent command retains its existing `whoami` subcommand; `login` and `logout` are additive.

PR #2888, opened 2026-02-17
feat(cli): integrate stored auth into existing commands

matthewvolk:
> [!WARNING]
> This PR depends on #2887

## What/Why?

The CLI now has `auth login`/`logout`/`whoami` commands that store credentials in `.bigcommerce/project.json`, but existing authenticated commands (`project create`, `project list`, `project link`, `deploy`) handle credential resolution inconsistently — some fall back to stored config, some don't, and error messages don't mention `catalyst auth login`.

This PR creates a centralized `resolveCredentials()` utility and applies it to all authenticated commands, making `--store-hash`/`--access-token` flags optional everywhere when credentials are stored via `catalyst auth login`.

**Changes:**

- **New `resolveCredentials()` utility** (`packages/catalyst/src/cli/lib/resolve-credentials.ts`) — resolves credentials from options (flags/env vars) with config fallback, exits with unified error mentioning `catalyst auth login`
- **Updated `project create`** — now falls back to stored config (previously required flags)
- **Updated `project list`** — replaced inline fallback + error with `resolveCredentials`
- **Updated `project link`** — replaced guard + error block with `resolveCredentials`
- **Updated `deploy`** — replaced three separate validation blocks with single `resolveCredentials` call
- **Updated `auth whoami`** — error message now mentions `catalyst auth login` first
- **New tests** for `resolveCredentials` (6 test cases)
- **Updated existing tests** for new error messages and behavior

## Testing

- All 88 tests pass across 15 test files (`pnpm --filter @bigcommerce/catalyst test`)
- Lint and typecheck pass
- CI compatibility preserved: flags/env vars still take precedence, config is only consulted as fallback

### Before

https://github.com/user-attachments/assets/8ebe22fe-211f-4116-9d0a-860b792950d7

### After

https://github.com/user-attachments/assets/cc797cc3-6fa1-4e1f-8e83-56e1f15e8ca3

## Migration

No breaking changes. Commands that previously required `--store-hash`/`--access-token` flags now optionally fall back to credentials stored via `catalyst auth login`. Error messages have been updated to mention `catalyst auth login` as the primary resolution path.

### DIFF
--- a/packages/catalyst/src/cli/commands/auth.ts
+++ b/packages/catalyst/src/cli/commands/auth.ts
@@ -67,7 +67,7 @@ const whoami = new Command('whoami')
       if (!storeHash || !accessToken) {
         consola.info('Not logged in: no credentials found.');
         consola.info(
-          'Provide --store-hash and --access-token flags, set CATALYST_STORE_HASH and CATALYST_ACCESS_TOKEN environment variables, or run `catalyst project create` / `catalyst project link` with credentials.',
+          'Run `catalyst auth login`, or provide --store-hash and --access-token flags (or set CATALYST_STORE_HASH and CATALYST_ACCESS_TOKEN environment variables).',
         );
         process.exit(1);
 
--- a/packages/catalyst/src/cli/commands/auth.spec.ts
+++ b/packages/catalyst/src/cli/commands/auth.spec.ts
@@ -91,6 +91,9 @@ describe('whoami', () => {
     await program.parseAsync(['node', 'catalyst', 'auth', 'whoami']);
 
     expect(consola.info).toHaveBeenCalledWith('Not logged in: no credentials found.');
+    expect(consola.info).toHaveBeenCalledWith(
+      'Run `catalyst auth login`, or provide --store-hash and --access-token flags (or set CATALYST_STORE_HASH and CATALYST_ACCESS_TOKEN environment variables).',
+    );
     expect(exitMock).toHaveBeenCalledWith(1);
   });
 
--- a/packages/catalyst/src/cli/commands/deploy.spec.ts
+++ b/packages/catalyst/src/cli/commands/deploy.spec.ts
@@ -339,8 +339,11 @@ test('errors when store hash is missing and not in .bigcommerce/project.json', a
   vi.stubEnv('CATALYST_STORE_HASH', undefined);
 
   await expect(program.parseAsync(['node', 'catalyst', 'deploy'])).rejects.toThrow(
-    'Store hash is required',
+    'Missing credentials',
   );
+
+  expect(consola.error).toHaveBeenCalledWith('Missing credentials.');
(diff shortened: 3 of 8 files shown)


packages/catalyst/src/cli/commands/auth.ts
```diff
@@ -101,6 +105,96 @@ const whoami = new Command('whoami')
...
+  .addOption(
+    new Option('--login-url <url>', 'BigCommerce login URL.')
+      .env('BIGCOMMERCE_LOGIN_URL')
+      .default(DEFAULT_LOGIN_URL),
+  )
+  .action(async (options) => {
+    try {
+      const config = getProjectConfig();
+
+      const storeHash = config.get('storeHash');
```
jamesqquick: should we factor in environment variables here in aditio to config?

packages/catalyst/src/cli/commands/auth.ts
```diff
@@ -101,6 +105,96 @@ const whoami = new Command('whoami')
...
+    }
+  });
+
+const logout = new Command('logout')
+  .description('Remove stored credentials for the current project.')
+  .action(() => {
+    try {
+      const config = getProjectConfig();
+
+      const storeHash = config.get('storeHash');
```
jamesqquick, line 190: why do we only check storehash and not accesstoken? I would think accessToken is the more relevant one to check since that's the actual credential?

packages/catalyst/src/cli/commands/auth.ts
jamesqquick: is there any potential error thrown in the parse that we need to handle?

packages/catalyst/src/cli/commands/auth.ts
jamesqquick: It looks like we factor it in for whoami

jamesqquick: Could we also not use `resolveCredentials` function here?

matthewvolk: Good call on environment vars - however, I didn't use resolveCredentials directly here because it calls `process.exit(1)` when credentials are missing, which is the opposite of what we want in `login`

packages/catalyst/src/cli/commands/auth.ts
matthewvolk: Good catch — now checking both `storeHash` and `accessToken`. If either is present, we clean up. Originally I figured just checking for one should imply the presence of the other, but I'm fine with making sure explicitly

packages/catalyst/src/cli/commands/auth.ts
matthewvolk: Good catch - switched to `.safeParse()` and throwing a friendlier `"Unexpected response from BigCommerce API"` error instead.

jorgemoya (approved):
Looks good to me. I would only suggest we prompt the user to see if they want to start the auth flow, but we can do that in a different PR. 🍹

PR #2889, opened 2026-02-17
fix(core): allow disabling canonical SEO via env variable

jorgemoya:
## What/Why?
Add `NEXT_DISABLE_METADATA_BASE` environment variable to disable canonical URL and hreflang alternate tags. When set to `'true'`, `metadataBase`, canonical URLs, and hreflang alternates are omitted from all pages. This is useful for preview deployments (Vercel, Cloudflare, etc.) where the vanity URL from BigCommerce points to production and causes SEO mismatch warnings. Default behavior (canonical tags enabled) is unchanged.

## Testing
Preview url, canonical and alternates not set.

## Migration
N/A

### DIFF
--- a/.changeset/quick-lies-lie.md
+++ b/.changeset/quick-lies-lie.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Add `NEXT_DISABLE_METADATA_BASE` environment variable to disable canonical URL and hreflang alternate tags. When set to `'true'`, `metadataBase`, canonical URLs, and hreflang alternates are omitted from all pages. This is useful for preview deployments (Vercel, Cloudflare, etc.) where the vanity URL from BigCommerce points to production and causes SEO mismatch warnings. Default behavior (canonical tags enabled) is unchanged.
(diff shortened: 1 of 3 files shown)


.changeset/quick-lies-lie.md
```diff
@@ -0,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Add `NEXT_DISABLE_METADATA_BASE` environment variable to disable canonical URL and hreflang alternate tags. When set to `'true'`, `metadataBase`, canonical URLs, and hreflang alternates are omitted from all pages. This is useful for preview deployments (Vercel, Cloudflare, etc.) where the vanity URL from BigCommerce points to production and causes SEO mismatch warnings. Default behavior (canonical tags enabled) is unchanged.
```
chanceaclark, line 5: 🍹 My only callout here is that now we have yet another configuration option that we need to document on this page: https://developer.bigcommerce.com/docs/storefront/catalyst/development/environment-variables

When it comes to merchants they will need to set this up in their preview environments otherwise they might be confused if they run lighthouse tests and see a failing rule on their preview deployments. I am curious if we've also explored using `VERCEL_ENV` or something. **This ultimately seems like a fix to make our unlighthouse tests pass.**

Also, how are we going to account for this in Native Hosting? I know Native Hosting is only production deployments, but when we eventually figure out how to tackle preview deployments, how is this going to look?

jorgemoya:
Closing in favor of another approach

PR #2890, opened 2026-02-17
fix(core): preview deployments now use correct metadataBase

jorgemoya:
## What/Why?
On Vercel preview deployments (`VERCEL_ENV=preview`), `metadataBase` and canonical/hreflang URLs use `VERCEL_URL` instead of the production vanity URL to prevent preview environments from generating SEO metadata pointing to production.

## Testing
Preview deployment/unlighthouse score

## Migration
N/A

### DIFF
--- a/core/app/[locale]/layout.tsx
+++ b/core/app/[locale]/layout.tsx
@@ -73,8 +73,16 @@ export async function generateMetadata(): Promise<Metadata> {
 
   const vanityUrl = data.site.settings?.url.vanityUrl;
 
+  const isPreview = process.env.VERCEL_ENV === 'preview';
+  const baseUrl =
+    isPreview && process.env.VERCEL_URL
+      ? new URL(`https://${process.env.VERCEL_URL}`)
+      : vanityUrl
+        ? new URL(vanityUrl)
+        : undefined;
+
   return {
-    metadataBase: vanityUrl ? new URL(vanityUrl) : undefined,
+    metadataBase: baseUrl,
     title: {
       template: `%s - ${storeName}`,
       default: pageTitle || storeName,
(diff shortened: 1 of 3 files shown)


core/app/[locale]/layout.tsx
```diff
@@ -73,8 +73,17 @@ export async function generateMetadata(): Promise<Metadata> {
 
   const vanityUrl = data.site.settings?.url.vanityUrl;
 
+  // Use preview deployment URL so metadataBase (canonical, og:url) points at the preview, not production.
+  let baseUrl: URL | undefined;
+
+  if (process.env.VERCEL_ENV === 'preview' && process.env.VERCEL_URL) {
```
Comment (migueloller) [inline core/app/[locale]/layout.tsx]: 🍹 What if instead of checking for a truthy value for `process.env.VERCEL_URL` we used `URL.canParse` or at least verified that the URL constructor won't throw.

Same for parsing `vanityUrl`—do we want to add some logic to check for potential parsing failures?

core/lib/seo/canonical.ts
```diff
@@ -46,6 +46,12 @@ const VanityUrlQuery = graphql(`
 `);
 
 const getVanityUrl = cache(async () => {
+  const isPreview = process.env.VERCEL_ENV === 'preview';
+
+  if (isPreview && process.env.VERCEL_URL) {
+    return `https://${process.env.VERCEL_URL}`;
+  }
```
migueloller: 🍹 I don't love how this logic is coupled to the `getVanityUrl` function. It feels like the fallback logic should live at a higher level than here

core/app/[locale]/layout.tsx
Comment (chanceaclark) [reply core/app/[locale]/layout.tsx]: TIL about `URL.canParse`!

chanceaclark (approved):

PR #2891, opened 2026-02-19
feat: report bundle size delta

jorgemoya:
## What/Why?
This feature adds automated bundle size tracking to the CI pipeline. Every time a PR is opened or updated, the build is compiled and its output is compared against a stored baseline snapshot, with the size differences posted as a PR comment. To keep that baseline from going stale, a separate workflow automatically regenerates and commits it whenever code is merged into canary or integrations/makeswift.

## Testing
Bundle reporting tested below. Testing the workflow to keep baseline is sync is TBD.

No changes:
Screenshot 2026-02-24 at 3 23 45 PM

Global metrics changed, routes identical:
Screenshot 2026-02-24 at 3 24 38 PM

Route changes (with warning):
Screenshot 2026-02-24 at 3 25 00 PM

## Migration
N/A


jordanarldt (approved):

.github/scripts/bundle-size.mjs
```diff
@@ -0,0 +1,377 @@
...
+const args = process.argv.slice(2);
+const command = args[0];
+
+function getFlag(name, defaultValue) {
+  const idx = args.indexOf(`--${name}`);
+
+  if (idx === -1 || idx + 1 >= args.length) return defaultValue;
+
+  return args[idx + 1];
+}
```
matthewvolk: 🍹 Less fragile option: https://nodejs.org/api/util.html#utilparseargsconfig

.github/scripts/bundle-size.mjs
```diff
@@ -0,0 +1,377 @@
...
+function getCommitSha() {
+  try {
+    return execSync('git rev-parse --short HEAD', {
+      encoding: 'utf-8',
+      stdio: ['pipe', 'pipe', 'pipe'],
+    }).trim();
+  } catch {
+    return 'unknown';
+  }
+}
```
matthewvolk: 🍹 Since `HEAD` is hardcoded (assuming you won't be using this to get the sha of any other commit), you can probably just use `github.sha` ([docs](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context)) in the Action and pass it to your script like

```
run: node .github/scripts/bundle-size.mjs generate --output .github/bundle-baseline.json --sha ${{ github.sha }} 
```

.github/scripts/bundle-size.mjs
```diff
@@ -0,0 +1,377 @@
...
+  console.error('Commands:');
+  console.error('  generate  Analyze .next/ build output and produce bundle size JSON');
+  console.error('    --output <path>  Write JSON to file instead of stdout');
+  console.error('');
+  console.error('  compare   Compare current bundle against a baseline');
+  console.error('    --baseline <path>  Path to baseline JSON (default: ./bundle-baseline.json)');
+  console.error('    --current <path>   Path to current bundle JSON (required)');
+  console.error('    --threshold <n>    Warning threshold percentage (default: 5)');
+  process.exit(1);
+}
```
matthewvolk: 🍹Same as above https://nodejs.org/api/util.html#utilparseargsconfig

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,85 @@
...
+                body,
+              });
+            } else {
+              await github.rest.issues.createComment({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: context.issue.number,
+                body,
+              });
+            }
```
migueloller: Why not have this in a JS file that we can reference by path here? Does the `@actions/github-script` action support providing a path to the script?

.github/scripts/bundle-size.mjs
jorgemoya: Updated to us this!

.github/scripts/bundle-size.mjs
jorgemoya: Good idea

.github/workflows/bundle-size.yml
jorgemoya: Good callout. Split to a JS file.

.github/scripts/bundle-size.mjs
jorgemoya: `command` is now parsed via parseArgs. Or where you thinking something else?

jorgemoya:
Instead of comparing against a committed baseline snapshot, the workflow now builds both the PR branch and the appropriate base branch live in the same job and diffs them directly. When a PR is opened or updated, it checks out the PR at ./pr and inspects core/package.json to determine whether the base should be canary or integrations/makeswift. It then checks out that branch at ./baseline, installs and builds both independently, and generates a bundle size report for each. The two reports are compared and the result is written to the job summary and posted (or updated) as a PR comment. This eliminates the need for a committed bundle-baseline.json and a separate workflow to keep it fresh — the baseline is always the exact current state of the target branch.

.github/workflows/bundle-size.yml
```diff
@@ -0,0 +1,104 @@
...
+
+      - name: Generate baseline bundle size
+        run: |
+          SHA=$(git rev-parse --short HEAD)
+          node .github/scripts/bundle-size.mjs generate --output /tmp/bundle-baseline.json --sha $SHA
+        working-directory: baseline
+
+      - name: Compare against baseline
+        run: node .github/scripts/bundle-size.mjs compare --baseline /tmp/bundle-baseline.json --current /tmp/bundle-current.json > /tmp/bundle-report.md
+        working-directory: pr
```
chanceaclark: This all happens synchronously which might take a long time. Is it possible we can do this async within the script itself? I know that's how the Next.js bundle size differ does it.

jorgemoya: We can split these to run as separate jobs, or where you thinking something else?

jorgemoya: Updated

.github/scripts/bundle-size.mjs
```diff
@@ -0,0 +1,369 @@
...
+
+import { existsSync, readFileSync, writeFileSync } from 'node:fs';
+import { dirname, join, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { parseArgs } from 'node:util';
+import { gzipSync } from 'node:zlib';
+
+// eslint-disable-next-line no-underscore-dangle
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const CORE_DIR = resolve(__dirname, '../../core');
```
migueloller: 🍹 I think that the point of `resolve` is to abstract OS-specific separators so we'd want to do `resolve(__dirname, '..', '..', 'core')` instead?

.github/scripts/bundle-size.mjs
```diff
@@ -0,0 +1,369 @@
...
+    baseline:  { type: 'string' },
+    current:   { type: 'string' },
+    threshold: { type: 'string' },
+    sha:       { type: 'string' },
+    dir:       { type: 'string' },
+  },
+});
+
+const NEXT_DIR = values.dir ? resolve(values.dir) : join(CORE_DIR, '.next');
+const command = positionals[0];
```
migueloller: 🍹 Should we use `.at(0)` here since `positionals` might be empty?

.github/scripts/bundle-size.mjs
```diff
@@ -0,0 +1,369 @@
...
+    output:    { type: 'string' },
+    baseline:  { type: 'string' },
+    current:   { type: 'string' },
+    threshold: { type: 'string' },
+    sha:       { type: 'string' },
+    dir:       { type: 'string' },
+  },
+});
+
+const NEXT_DIR = values.dir ? resolve(values.dir) : join(CORE_DIR, '.next');
```
migueloller: 🍹 Will `resolve` just throw here is `values.dir` is an invalid directory? Is that intentional?

.github/scripts/bundle-size.mjs
migueloller, line 1: 🍹 I started reading this file but I'm realizing this is mostly AI-generated and given its determinism, probably the better thing to focus on is tests/that it does what it's supposed to. Do we have a test suite or perhaps some test runs that show that?

.github/scripts/bundle-size.mjs
```diff
@@ -0,0 +1,369 @@
...
+      firstLoadJs: round1(firstLoadJs + sizes.js + sizes.css),
+    };
+  }
+
+  return routes;
+}
+
+function generate() {
+  const appManifestPath = join(NEXT_DIR, 'app-build-manifest.json');
+  const buildManifestPath = join(NEXT_DIR, 'build-manifest.json');
```
migueloller: 🍹 Seeing us look into Next.js internals like these makes me wonder if there's an official tool or interface that supports what we need 🤔

migueloller (approved):
Changes look good, but with something like this I'd like to see some sort of video recording or test or something else that shows the script working.

The PR description points to the PR comment but that hasn't been updated in ~4 days.

jorgemoya:
@migueloller Just to be clear, the PR comment only gets updated on push. Since no changes have been made since 4 days ago, it hasn't changed. I will add some screenshots of how the comment looks if there are changes or above threshold.

.github/scripts/bundle-size.mjs
jorgemoya: Nope, but I can look into it.

.github/scripts/bundle-size.mjs
jorgemoya: `resolve` returns an empty string on an invalid path. The code that validates this is path is:

```ts
  if (!existsSync(appManifestPath)) {
    console.error('Error: .next/app-build-manifest.json not found. Run `next build` first.');
    process.exit(1);
  }
```

.github/scripts/bundle-size.mjs
jorgemoya: I looked at how some analyzers did this and it seemed like they were doing the same thing, however Next.js does have some tools to analyze bundles: https://nextjs.org/docs/app/guides/package-bundling

However, it seems overkill for what we're doing, but I can look into it again.

jorgemoya: It seems these tools' `--output` is static HTML and don't export a simple `json` that we can use to compare bundle sizes.

jorgemoya: [next-bundle-analysis](https://github.com/hashicorp/nextjs-bundle-analysis) the gold standard is, but I believe the current approach works well for our needs.

.github/scripts/bundle-size.mjs
jorgemoya: Added tests, more info below

jorgemoya:
Generated some tests for the scripts added and made sure the tests pass succesfully:

```bash
❯ pnpm run test:scripts

> @bigcommerce/catalyst@0.1.0 test:scripts /Users/jorge.moya/dev/catalyst
> node --test .github/scripts/__tests__/*.test.mjs

▶ round1
  ✔ rounds up at .05 (0.35425ms)
  ✔ rounds down below .05 (0.061209ms)
  ✔ returns 0 unchanged (0.050125ms)
  ✔ handles negative values (0.056209ms)
✔ round1 (0.993375ms)
▶ parseManifestEntries
  ✔ routes /layout entries to layouts (0.438ms)
  ✔ routes /page entries to pages (0.058417ms)
  ✔ ignores entries ending in neither /layout nor /page (0.06125ms)
  ✔ returns empty objects for empty input (0.075959ms)
  ✔ handles multiple layouts and pages together (0.103542ms)
✔ parseManifestEntries (0.878709ms)
▶ computeRootLayout
  ✔ selects shortest path as root when multiple layouts exist (0.250458ms)
  ✔ returns null rootLayoutPath when layoutPaths is empty (0.087125ms)
  ✔ excludes sharedChunks from rootLayoutChunks (1.913459ms)
  ✔ rootLayoutChunks contains all non-shared layout chunks (0.313875ms)
  ✔ computes non-zero sizes when real files exist (0.158291ms)
✔ computeRootLayout (2.827458ms)
▶ computeRouteMetrics
  ✔ firstLoadJs equals firstLoadJs arg when all chunks are non-existent (0.156416ms)
  ✔ firstLoadJs is greater than firstLoadJs arg when real chunk files exist (0.28675ms)
  ✔ excludes sharedChunks from route chunk set (0.277583ms)
  ✔ excludes rootLayoutChunks from route chunk set (0.288125ms)
  ✔ includes non-root ancestor layout chunks in route size (0.240625ms)
  ✔ does not include root ancestor layout chunks in route size (0.090042ms)
  ✔ applies round1 to all output values (0.0615ms)
✔ computeRouteMetrics (1.49325ms)
▶ compareReport
  ✔ shows "No route changes detected." when routes are identical (0.275917ms)
  ✔ does not show global metrics table when global metrics are unchanged (0.050083ms)
  ✔ shows global metrics table only when metrics changed (0.070542ms)
  ✔ shows only the changed global metrics (0.049917ms)
  ✔ shows NEW row for added route (0.066833ms)
  ✔ shows REMOVED row for deleted route (0.042709ms)
  ✔ does not show warning for increase under threshold (0.048542ms)
  ✔ shows warning for increase over threshold (over 1kB AND over threshold percent) (0.051167ms)
  ✔ does not warn when delta is over threshold percent but 1kB or less (0.045959ms)
  ✔ does not warn when delta is over 1kB but at or under threshold percent (0.03925ms)
  ✔ respects custom threshold: no warning when under (0.038125ms)
  ✔ respects custom threshold: warning when over (0.053167ms)
  ✔ uses default threshold of 5 percent when not specified (0.042833ms)
  ✔ formats positive delta with + sign and percent (0.0405ms)
  ✔ formats negative delta with minus sign and percent (0.062125ms)
  ✔ sorts routes alphabetically (0.077708ms)
  ✔ strips the /[locale] prefix from display names (0.038167ms)
  ✔ omits near-zero deltas that round to 0.0 (0.031375ms)
  ✔ always shows Per-Route First Load JS section (0.027666ms)
  ✔ includes threshold value in footer (0.028167ms)
  ✔ shows header with baseline commitSha and updatedAt (0.044083ms)
  ✔ handles empty routes in both baseline and current (0.031125ms)
  ✔ shows table header when routes have changes (0.039292ms)
✔ compareReport (1.545958ms)
▶ getGzipSize
  ✔ returns 0 when file does not exist (0.089791ms)
  ✔ returns a positive number for an existing file (0.197666ms)
  ✔ caches results and returns same value on second call (0.474167ms)
  ✔ clearSizeCache resets the cache (0.806459ms)
✔ getGzipSize (1.6495ms)
▶ post-bundle-comment
  ✔ creates a new comment when no existing comment contains the marker (1.394291ms)
  ✔ updates existing comment when marker found (0.41025ms)
  ✔ body always starts with marker and newline (0.330083ms)
  ✔ updated comment body also starts with marker and newline (0.260583ms)
  ✔ includes report file content in the comment body (0.384292ms)
  ✔ reads report from a custom reportPath (0.896875ms)
  ✔ passes correct owner, repo, issue_number from context to listComments (0.540875ms)
  ✔ passes correct owner, repo, issue_number from context to createComment (0.469708ms)
  ✔ passes correct owner and repo to updateComment (0.892167ms)
  ✔ uses the first comment that contains the marker (not just exact match) (0.507083ms)
  ✔ creates comment when existing comments do not contain the marker (0.353417ms)
✔ post-bundle-comment (7.124083ms)
ℹ tests 59
ℹ suites 7
ℹ pass 59
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 560.026416
```

migueloller (approved):
Thanks for the tests and screenshots @jorgemoya 

Two minor notes 🍹 
- Why not use TypeScript and get the benefit of type checking? Node.js can strip types natively. 
- The report when no changes happen is a bit confusing since it shows the callout with the threshold. Maybe we can simplify the comment when there are no changes?

jorgemoya:
@migueloller Updated the reporting to be less confusing (see PR description) + migrated to TS (`post-bundle-comment.js` had to remain as javascript since `actions/github-script` uses `eval()/vm.runInThisContext()` internally and it won't invoke node with type stripping).

PR #2892, opened 2026-02-19
feat(cli): remove --root-dir flag from project create and link commands

matthewvolk:
## What/Why?

The `--root-dir` flag on `project create` and `project link` is a footgun — if a user passes it once but forgets it on subsequent commands (e.g. `catalyst deploy`), they get confusing errors.

This removes the `--root-dir` option from both commands and hardcodes `process.cwd()` in `getProjectConfig()`, simplifying the CLI surface and eliminating a class of user mistakes.

**Changes:**
- Remove `rootDir` parameter from `getProjectConfig()` — always use `process.cwd()`
- Remove `--root-dir` option from `project create` and `project link` commands
- Update tests to mock `process.cwd()` instead of passing `--root-dir`

## Testing

- `vitest run src/cli/commands/project.spec.ts` — 13 tests pass
- `vitest run src/cli/lib/project-config.spec.ts` — 3 tests pass

## Migration

The `--root-dir` flag has been removed. `process.cwd()` is now used as the hardcoded default. Run CLI commands from your project directory instead of passing `--root-dir`.

### DIFF
--- a/.changeset/remove-root-dir-opt.md
+++ b/.changeset/remove-root-dir-opt.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst": minor
+---
+
+Remove `--root-dir` flag from `project create` and `project link` commands. Use `process.cwd()` as a hardcoded default for usage simplicity.
--- a/packages/catalyst/src/cli/commands/project.spec.ts
+++ b/packages/catalyst/src/cli/commands/project.spec.ts
@@ -55,7 +55,9 @@ beforeAll(async () => {
 
   [tmpDir, cleanup] = await mkTempDir();
 
-  config = getProjectConfig(tmpDir);
+  vi.spyOn(process, 'cwd').mockReturnValue(tmpDir);
+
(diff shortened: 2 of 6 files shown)


.changeset/remove-root-dir-opt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@bigcommerce/catalyst": minor
```
jamesqquick, line 2: I think we aren't supposed to include changesets for CLI right now

jamesqquick (changes requested):

.changeset/remove-root-dir-opt.md
matthewvolk: Based on this, I believe we are: https://bigcommerce.slack.com/archives/C06U627FQKW/p1771270128933839?thread_ts=1770755736.597359&cid=C06U627FQKW

jordanarldt (approved):

jamesqquick (approved):

PR #2894, opened 2026-02-19
feat(cli): remove next start proxy from start command

matthewvolk:
## What/Why?

The `catalyst start` command currently supports two modes via a `--framework` option: proxying `next start` (for vanilla Next.js users) or running `opennextjs-cloudflare preview` (for native hosting users). The CLI is moving away from proxying Next.js commands — users who don't use native hosting should simply run `next start` directly.

This simplifies `start` to only run `opennextjs-cloudflare preview`, removing:
- The `--framework` option and conditional code path
- The `.allowUnknownOption(true)` passthrough hack and `[start-options...]` argument
- The `existsSync` / `getProjectConfig` dependencies
- The Next.js binary resolution logic

Net result: **-108 lines** across implementation and tests.

## Testing

- `pnpm --filter @bigcommerce/catalyst test` — all 87 tests pass (15 files)
- Verified no other files depend on the removed `--framework` option or `next start` path

## Migration

Users who previously ran `catalyst start --framework nextjs` should use `next start` (or `npx next start`) directly instead.

### DIFF
--- a/packages/catalyst/src/cli/commands/start.ts
+++ b/packages/catalyst/src/cli/commands/start.ts
@@ -1,43 +1,12 @@
-import { Command, Option } from 'commander';
+import { Command } from 'commander';
 import { execa } from 'execa';
-import { existsSync } from 'node:fs';
 import { join } from 'node:path';
 
-import { getProjectConfig } from '../lib/project-config';
-
 export const start = new Command('start')
-  .description('Start your Catalyst storefront in optimized production mode.')
-  .allowUnknownOption(true)
-  // The unknown options end up in program.args, not in program.opts(). Commander does not take a guess at how to interpret the unknown options.
-  .argument(
-    '[start-options...]',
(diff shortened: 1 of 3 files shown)


packages/catalyst/src/cli/commands/start.ts
```diff
@@ -1,43 +1,12 @@
...
-    ]),
+  .description(
+    'Start a local preview of your Catalyst storefront using the OpenNext Cloudflare adapter.',
   )
-  .action(async (startOptions, options) => {
-    const config = getProjectConfig();
-    const framework = options.framework ?? config.get('framework');
-
-    if (framework === 'nextjs') {
-      const nextBin = join('node_modules', '.bin', 'next');
```
jamesqquick, line 27: Is nextjs not still a dependency for running even with open next installed?

matthewvolk: It is, `opennextjs-cloudflare` calls it under the hood for the commands that require it (basically just `build` IIRC)

jordanarldt (approved):

PR #2895, opened 2026-02-20
feat(cli): remove framework config and simplify build command

matthewvolk:
## What/Why?

The `framework` configuration concept (`catalyst` | `nextjs`) is no longer needed. The CLI will only be installed when using native hosting, so the `nextjs` framework path is a dead code path. This PR removes it entirely and simplifies the build command to always call `buildCatalystProject()` directly.

### Changes

- **`project-config.ts`** — Remove `framework` from `ProjectConfigSchema` and its Conf schema definition
- **`build.ts`** — Remove `.allowUnknownOption()`, `[next-build-options...]` argument, `--framework` option, and the nextjs branch in the action handler; remove unused `existsSync` import
- **`project.ts`** — Remove `config.set('framework', 'catalyst')` from `create` and `link` actions
- **Tests** — Remove all framework-related assertions, mocks, and test cases across `build.spec.ts`, `project.spec.ts`, and `project-config.spec.ts`

## Testing

- `vitest run` — all tests pass (2 pre-existing failures in `index.spec.ts` and `version.spec.ts` are unrelated)
- `eslint .` — clean
- `tsc --noEmit` — clean

## Migration

The `framework` key in `.bigcommerce/project.json` is no longer read or written. Existing config files with this key will be silently ignored by `conf`.

### DIFF
--- a/packages/catalyst/src/cli/lib/project-config.ts
+++ b/packages/catalyst/src/cli/lib/project-config.ts
@@ -3,7 +3,8 @@ import { join } from 'path';
 
 export interface ProjectConfigSchema {
   projectUuid: string;
-  framework: 'catalyst' | 'nextjs';
+  // Reserved for future use
+  framework: 'catalyst';
   storeHash?: string;
   accessToken?: string;
   telemetry: {
@@ -21,8 +22,8 @@ export function getProjectConfig() {
       projectUuid: { type: 'string', format: 'uuid' },
       framework: {
         type: 'string',
-        enum: ['catalyst', 'nextjs'],
-        default: 'nextjs',
+        enum: ['catalyst'],
+        default: 'catalyst',
       },
       storeHash: { type: 'string' },
       accessToken: { type: 'string' },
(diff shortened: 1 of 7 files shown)


packages/catalyst/src/cli/lib/project-config.ts
```diff
@@ -3,7 +3,6 @@ import { join } from 'path';
 
 export interface ProjectConfigSchema {
   projectUuid: string;
-  framework: 'catalyst' | 'nextjs';
```
chanceaclark, line 6: Could we actually keep this field for future consumption and just make a note about it? The type union can just be `catalyst`.

jorgemoya (approved):

jamesqquick (approved):

packages/catalyst/src/cli/lib/project-config.ts
matthewvolk: I'm cool with that 👍

PR #2896, opened 2026-02-20
feat(core): enable recaptcha in forms

jamesqquick:
Jira [CATALYST-1806](https://bigcommercecloud.atlassian.net/browse/CATALYST-1806)

## What/Why?
Match functionality of Stencil with regards to reCAPTCHA. Catalyst should respect the reCAPTCHA settings from the Control Panel and apply reCAPTCHA to forms appropriately based on those settings.

CleanShot 2026-03-02 at 12 20 01@2x

Forms that should support reCAPTCHA if enabled include:

- product reviews
- user registration
- contact

### with reCAPTCHA enabled in the Control Panel

**Register & contact**
- Submit without completing the reCAPTCHA checkbox → submission is blocked and “Please complete the reCAPTCHA verification.” is shown.
- Complete reCAPTCHA and submit with valid data → form submits and succeeds as expected.

**Product review**
- Open “Write a review” and submit without completing reCAPTCHA → same error, no submit.
- Complete reCAPTCHA and submit → review submits and success toast appears.

### without reCAPTCHA enabled in the Control Panel

When reCAPTCHA is disabled in the control panel, all three forms submit without a reCAPTCHA widget or “complete reCAPTCHA” errors.

## Migration

[CATALYST-1806]: https://bigcommercecloud.atlassian.net/browse/CATALYST-1806?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


core/vibes/soul/form/dynamic-form/index.tsx
```diff
@@ -145,11 +156,28 @@ export function DynamicForm<F extends Field>({
...
     shouldRevalidate: 'onInput',
-    onSubmit(event, { formData }) {
+    async onSubmit(event, { formData }) {
       event.preventDefault();
 
+      let payload: FormData = formData;
+
+      if (recaptchaSiteKey && recaptchaRef.current) {
+        try {
+          const token = await recaptchaRef.current.executeAsync();
```
matthewvolk: I think this `executeAsync()` might cause some weird bugs. I could totally be wrong, but did you have a chance to manually test that this works?

In my head, I see this sequence of events:
1. User clicks to submit the form
2. Browser executes the submit event and puts the form in `event.currentTarget`
3. Then this line says "wait let me ask the recaptcha lib for a token first"
4. That happens asynchronously, so while waiting for recaptcha, the browser considers the submit event "done" and nulls out `event.currentTarget`
5. Down below, when you try to set `payload = new FormData(event.currentTarget);`, what is the value of `event.currentTarget`?

Again, I could be wrong, but that might be what the `unsafe argument of type error` is complaining about

core/components/recaptcha-provider/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+  return useContext(ReCaptchaContext);
+}
+
+export function useReCaptchaSiteKey(): string | undefined {
+  const settings = useReCaptchaSettings();
+
+  return settings?.isEnabledOnStorefront === true && settings.siteKey
+    ? settings.siteKey
+    : undefined;
+}
```
matthewvolk: Is this actually being used anywhere?

To me, it looks like you have two different ways you're getting the recaptcha key and adding it to forms:

1. This `ReCaptchaProvider` - in the default layout.tsx, it [looks like you fetch the recaptcha settings](https://github.com/bigcommerce/catalyst/pull/2896/changes#diff-c80e80d54ec7a7e1a0d056ce661f16c6ec21b7a90103b4eb4de9b6dfe39f9091R18) and wrap the app in this `<ReCaptchaProvider>`. This makes the site key available to child components with `useReCaptchaSiteKey()`... but I don't actually see this being used anywhere
2. Separately, in non-client pages like register + contact, it [looks like we're also fetching the site key server side](https://github.com/bigcommerce/catalyst/pull/2896/changes#diff-9d80c267487bc1f18bb4ae50fa840e2979cd8aae354d0d27305a960df7220ed2R67) and [passing it as a prop to DynamicFormSection](https://github.com/bigcommerce/catalyst/pull/2896/changes#diff-9d80c267487bc1f18bb4ae50fa840e2979cd8aae354d0d27305a960df7220ed2R115)

I think right now the forms use the prop (point 2), and nothing actually consumes the provider (point 1).

matthewvolk (commented):
High level, I think this is pretty good - the main "architectural" feedback I have is that it looks like you introduce two ways to get the recaptcha key but are only using one of them... wondering if that's redundant. I could be misunderstanding intent here, so I definitely encourage a second reviewer to weigh in!

Outside of this, my only other concern is the library is a bit old... not sure if it's being maintained, especially looking at some of the recent issues: https://github.com/dozoisch/react-google-recaptcha/issues/301

core/vibes/soul/form/dynamic-form/index.tsx
matthewvolk: Admittedly I didn't have time to set up a reCAPTCHA key on my demo store, I can do that once this is converted from a draft... but as a little experiment, if you save this HTML and open in your browser, it does show that `e.currentTarget` is `null` after awaiting the promise in the listener callback when you click "Submit".

```html
<!DOCTYPE html>
<html>
<head><title>event.currentTarget after await</title></head>
<body>
  <form>
    <button type="submit">Submit</button>
  </form>
  <pre id="output"></pre>
  <script>
    const output = document.getElementById('output');
    const log = (msg) => { output.textContent += msg + '\n'; };

    document.querySelector('form').addEventListener('submit', async (e) => {
      e.preventDefault();
      log('before await: ' + e.currentTarget);
      await new Promise((r) => setTimeout(r, 100));
      log('after await:  ' + e.currentTarget);
    });
  </script>
</body>
</html>
```

core/components/recaptcha-provider/index.tsx
jamesqquick: Yeah you're right. Looks like this was leftover from a previous iteration. Deleted!

core/vibes/soul/form/dynamic-form/index.tsx
jamesqquick: You were right. That was the wrong function to use. async was only for invisible recaptcha. I updated now to use `getValue()`.

core/vibes/soul/sections/reviews/review-form.tsx
```diff
@@ -87,8 +100,26 @@ export const ReviewForm = ({
...
 
+      setRecaptchaError(null);
+
+      let payload: FormData = formData;
+
+      if (recaptchaSiteKey && recaptchaRef.current) {
+        const token = recaptchaRef.current.getValue();
+
+        if (!token || typeof token !== 'string') {
+          setRecaptchaError(t('recaptchaRequired'));
```
jorgemoya: Is there no way to validate for recaptcha within the form action? And return the errors accordingly? 🤔

jamesqquick: Good call. Refactored to move that logic to the server and create utility functions to parse and validate the token.

core/lib/recaptcha.ts
```diff
@@ -0,0 +1,83 @@
...
+export { RECAPTCHA_TOKEN_FORM_KEY } from './recaptcha/constants';
+export type { ReCaptchaSettings } from './recaptcha/constants';
+
+export const ReCaptchaSettingsQuery = graphql(`
+  query ReCaptchaSettingsQuery {
+    site {
+      settings {
+        reCaptcha {
+          failedLoginLockoutDurationSeconds
+          isEnabledOnCheckout
```
jorgemoya: I don't think these two are being used. Can we use it for something here? I would remove if not needed.

core/lib/recaptcha.ts
```diff
@@ -0,0 +1,83 @@
...
+  formData: FormData,
+): Promise<{ siteKey: string | undefined; token: string }> {
+  const siteKey = await getRecaptchaSiteKey();
+  const raw = formData.get(RECAPTCHA_TOKEN_FORM_KEY);
+  const token = typeof raw === 'string' ? raw : '';
+
+  return { siteKey, token };
+}
+
+export function validateRecaptchaToken(
```
jorgemoya: Maybe change to `assertRecaptchaTokenPresent` since it's not really validating anything (this happens server side)?

core/vibes/soul/form/dynamic-form/index.tsx
```diff
@@ -148,8 +162,26 @@ export function DynamicForm<F extends Field>({
...
+        if (!token || typeof token !== 'string') {
+          setRecaptchaError(t('recaptchaRequired'));
+
+          return;
+        }
+
+        payload = new FormData(event.currentTarget);
+        payload.set(RECAPTCHA_TOKEN_FORM_KEY, token);
+        recaptchaRef.current.reset();
+      }
```
jorgemoya: The cool thing about validating on the server is that we no longer need this.

core/vibes/soul/form/dynamic-form/index.tsx
```diff
@@ -206,6 +241,7 @@ export function DynamicForm<F extends Field>({
               {submitLabel}
             </SubmitButton>
           </div>
+          {recaptchaError ? <FormStatus type="error">{recaptchaError}</FormStatus> : null}
```
jorgemoya: No longer needed too.

core/vibes/soul/form/dynamic-form/index.tsx
```diff
@@ -77,9 +82,14 @@ export interface DynamicFormProps<F extends Field> {
   onSuccess?: (lastResult: SubmissionResult, successMessage: ReactNode) => void;
   passwordComplexity?: PasswordComplexitySettings | null;
   errorTranslations?: FormErrorTranslationMap;
+  recaptchaSiteKey?: string;
 }
 
-export function DynamicForm<F extends Field>({
+export function DynamicForm<F extends Field>(props: DynamicFormProps<F>) {
+  return <DynamicFormInner {...props} />;
+}
```
jorgemoya: Why did we add this `Inner` component? 🤔

core/vibes/soul/sections/reviews/review-form.tsx
jorgemoya: Even with the refactor this is still present. I think we can remove all client facing logic for showing the errors and validating.

jorgemoya (commented):
Thanks for looking into validating and erroring from the actions, but I still see some leftover logic from the previous implementation. Should be easy to clean up, I believe.

Another question, should we include reCaptcha in other pages? Login, reset password? I forget if once we enable reCaptcha on a storefront, forms automatically require a token to succeed?

Another detail missing is how to make sure the forms still succeed in our e2e tests when reCaptcha is enabled. Our [old approach](https://github.com/bigcommerce/catalyst/blob/be80d14c1c1051189e629d182dfd46fc60c363b1/core/lib/bypass-recaptcha.ts) was to use the `X-Vercel-Set-Bypass-Cookie: true`, however another approach is to make sure the E2E test store has Google's test key 🤔? This would work now since our E2E test store is separate from our demo sites.

core/vibes/soul/sections/reviews/review-form.tsx
jamesqquick: Got it. Removed all client logic in lieu of leveraging logic on the server

core/lib/recaptcha.ts
jamesqquick: Great catch. Removed.

core/lib/recaptcha.ts
jamesqquick: Updated!

core/vibes/soul/form/dynamic-form/index.tsx
jamesqquick: Removed. Great call.

core/vibes/soul/form/dynamic-form/index.tsx
jamesqquick: Removed

core/vibes/soul/form/dynamic-form/index.tsx
jamesqquick: Removed

jamesqquick:
> Thanks for looking into validating and erroring from the actions, but I still see some leftover logic from the previous implementation. Should be easy to clean up, I believe.
> 
> Another question, should we include reCaptcha in other pages? Login, reset password? I forget if once we enable reCaptcha on a storefront, forms automatically require a token to succeed?
> 
> Another detail missing is how to make sure the forms still succeed in our e2e tests when reCaptcha is enabled. Our [old approach](https://github.com/bigcommerce/catalyst/blob/be80d14c1c1051189e629d182dfd46fc60c363b1/core/lib/bypass-recaptcha.ts) was to use the `X-Vercel-Set-Bypass-Cookie: true`, however another approach is to make sure the E2E test store has Google's test key 🤔? This would work now since our E2E test store is separate from our demo sites.

I'm not sure what we want to do for forms on other pages. I started with mirroring the Stencil functionality which is specific to these forms and not all of them.

jorgemoya (approved):

matthewvolk (approved):

PR #2903, opened 2026-03-02
fix(core): render branded page for /404 instead of Vercel default

malewis5:
Visiting /404 directly shows the default Vercel black screen because
Next.js special-cases this path with a pre-rendered static 404.html,
bypassing the [locale] segment entirely.

Fix by splitting the layout into a true root layout (app/layout.tsx)
owning <html>, <body>, fonts, and globals.css, and converting the
locale layout into a nested layout. A new root not-found.tsx renders
a minimal branded 404 page for this edge case.

### DIFF
--- a/.changeset/fix-root-not-found.md
+++ b/.changeset/fix-root-not-found.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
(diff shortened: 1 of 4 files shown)


jorgemoya:
This PR is superseded by #2947. Thanks for the example @malewis5!

PR #2913, opened 2026-03-09
feat: enables client-side navigation in Makeswift builder

agurtovoy:
## What/Why?

Enables client-side navigation in Makeswift builder by upgrading Catalyst to the latest Makeswift runtime.

## Testing

https://github.com/user-attachments/assets/9cdc1b1b-b5be-42d6-aaad-137f6955dde8

## Migration

N/A

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -22,7 +22,7 @@
     "@conform-to/react": "^1.6.1",
     "@conform-to/zod": "^1.6.1",
     "@icons-pack/react-simple-icons": "^11.2.0",
-    "@makeswift/runtime": "0.26.4",
+    "@makeswift/runtime": "0.28.5",
     "@opentelemetry/api": "^1.9.0",
     "@opentelemetry/api-logs": "^0.208.0",
     "@opentelemetry/instrumentation": "^0.208.0",
--- a/.changeset/shiny-cougars-drop.md
+++ b/.changeset/shiny-cougars-drop.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-makeswift": minor
+---
+
+Enable interaction-mode site navigation in Makeswift builder
--- a/core/app/api/makeswift/[...makeswift]/route.ts
+++ b/core/app/api/makeswift/[...makeswift]/route.ts
@@ -24,8 +24,6 @@ const defaultVariants: Font['variants'] = [
 
 const handler = MakeswiftApiHandler(process.env.MAKESWIFT_SITE_API_KEY, {
   runtime,
-  apiOrigin: process.env.NEXT_PUBLIC_MAKESWIFT_API_ORIGIN ?? process.env.MAKESWIFT_API_ORIGIN,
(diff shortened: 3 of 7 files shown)


core/package.json
```diff
@@ -19,7 +19,7 @@
     "@conform-to/react": "^1.6.1",
     "@conform-to/zod": "^1.6.1",
     "@icons-pack/react-simple-icons": "^11.2.0",
-    "@makeswift/runtime": "^0.26.0",
+    "@makeswift/runtime": "https://pkg.pr.new/makeswift/makeswift/@makeswift/runtime@1259",
```
agurtovoy, line 22: This needs to be changed to a released `runtime` version before we can merge this PR.

chanceaclark: I also think we need Next 16 + React 19 in order to merge a new runtime version as well. @matthewvolk is tackling this right now.

fikrikarim: The new runtime version should still work with Next 15 + React 18. The runtime uses Suspense instead of Activity when it's on React 18.

agurtovoy: @chanceaclark Just to clarify, we don't *have* to upgrade to Next 16 + React 19 at the same time; Makeswift runtime still supports React 18 and older Next.js versions all the way down to 13.4, so we can tackle the Next 16/React 19 upgrade to get the [`<Activity>` performance improvements](https://github.com/makeswift/makeswift/releases/tag/%40makeswift%2Fruntime%400.26.3) independently of upgrading Makeswift runtime to pull in other features/bug fixes.

chanceaclark: Interesting 🤔 I've been going with the assumption that it's hand-in-hand, specifically to fix that performance issue with `<Activity />`. TIL that we should be good then.

agurtovoy: Yep, we need runtime@0.26.3+, Next 16, and React 19 to get the `<Activity />` performance improvements, but otherwise, these are not coupled. If we want to, we can upgrade to runtime@0.26.3 right now without forcing the  Next 16 / React 19 upgrade.

jorgemoya (approved):
Should we include a changeset?

jorgemoya:
Tagging with `do not merge` until we have that released `runtime` version.

Edit: nevermind, it's a draft PR 🤦

core/package.json
```diff
@@ -22,7 +22,7 @@
     "@conform-to/react": "^1.6.1",
     "@conform-to/zod": "^1.6.1",
     "@icons-pack/react-simple-icons": "^11.2.0",
-    "@makeswift/runtime": "0.26.4",
+    "@makeswift/runtime": "0.28.5",
```
agurtovoy, line 25: Makeswift upgrade guides for 0.26 -> 0.28 transition:
* https://docs.makeswift.com/developer/docs/guides/upgrading/0.27.0
* https://docs.makeswift.com/developer/docs/guides/upgrading/0.28.0

parthshahp (approved):
Looks good!

PR #2938, opened 2026-03-19
fix: show readable error when project creation errors with 422

jorgemoya:
## What/Why?
Returns a readable error message when creating a project with an invalid name.

## Testing
```
~/dev/catalyst/core on  catalyst-1823-add-message-422-error [✘!?]
❯ node ../packages/catalyst/dist/cli.js project create
◢ @bigcommerce/catalyst v1.0.0-alpha.0                                                           3:29:10 PM


✔ Enter a name for the new project:
This is a bad project name

[3:29:18 PM]  ERROR  The project name you entered doesn't meet the requirements. It must be 3–32 characters long and use only letters, numbers, hyphens (-), underscores (_), and periods (.)

ℹ                                                                                               3:29:18 PM
Trace ID: 0aed477d-42ff-445f-b6e0-6dab56056813
Share this Trace ID with BigCommerce support.
```

## Migration
N/A

### DIFF
--- a/packages/catalyst/src/cli/lib/project.ts
+++ b/packages/catalyst/src/cli/lib/project.ts
@@ -95,6 +95,12 @@ export async function createProject(
     );
   }
 
+  if (response.status === 422) {
+    throw new Error(
+      "The project name you entered doesn't meet the requirements. It must be 3–32 characters long and use only letters, numbers, hyphens (-), underscores (_), and periods (.)",
+    );
+  }
+
   if (!response.ok) {
     throw new Error(`Failed to create project: ${response.statusText}`);
   }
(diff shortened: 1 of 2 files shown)


rtalvarez (approved):

packages/catalyst/src/cli/lib/project.ts
```diff
@@ -95,6 +95,12 @@ export async function createProject(
     );
   }
 
+  if (response.status === 422) {
+    throw new Error(
+      "The project name you entered doesn't meet the requirements. It must be 3–32 characters long and use only letters, numbers, hyphens (-), underscores (_), and periods (.)",
+    );
+  }
```
chanceaclark, line 102: 🍹 I would love for this to be returned from the API if we can. Can we just backlog a task to do this?

PR #2948, opened 2026-03-23
fix: improve regression-tests workflow structure and concurrency

jorgemoya:
## Summary

- Remove top-level workflow concurrency group — each job now manages its own concurrency independently
- Simplify `detect-provider` job: remove `is-canary-branch` output, consolidate duplicated logic, and hoist `is-preview` / `PKG_NAME` to avoid repeated conditionals
- Fix preview concurrency groups: use `deployment.ref` instead of `deployment.sha` so same-branch runs actually cancel stale audits (`cancel-in-progress` was a no-op before since each SHA produced a unique group)
- Fix production concurrency groups: use ref-based groups for preview environments (stale baseline audits get cancelled) and SHA-based groups for production (each canary deploy stays isolated)
- Simplify `unlighthouse-audit-production` condition from `(is-preview == 'true' && production-url != '') || is-canary-branch == 'true'` to just `production-url != ''`
- Fix `unlighthouse-report` condition to use `is-preview == 'false' && production-url != ''` instead of the removed `is-canary-branch` output

## Test plan

- [ ] Push two commits quickly to an open PR → first run's preview + production baseline audits get cancelled by the second
- [ ] Merge to canary → production audit runs to completion, never cancelled
- [ ] Two different PRs deploying simultaneously → separate groups (different refs), no interference
- [ ] Cloudflare and unknown provider paths still set all required outputs


.github/workflows/regression-tests-preview.yml
```diff
@@ -1,103 +1,63 @@
...
 on:
   deployment_status:
     states: ["success"]
 
-env:
-  VERCEL_PROTECTION_BYPASS: ${{ secrets.VERCEL_AUTOMATION_BYPASS_SECRET }}
-
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.event.deployment_status.target_url }}
+  group: regression-preview-${{ github.event.deployment.sha }}
```
chanceaclark: I believe you can move this down to the job level so you don't need to split this into two separate workflow files https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency (and key off the outputs from `detect-provider`)

jorgemoya: Updated 👍

chanceaclark (approved):

PR #2951, opened 2026-03-24
chore(core): revert Node to v22

jorgemoya:
## What/Why?

Fixes a compatibility issue between Next.js 16, Node 24+, and Makeswift runtime 0.25+ that causes the following error at runtime:

```
TypeError: Cannot read private member #state from an object whose class did not declare it
    at new l (.next/server/chunks/9ce95_next_294d13e4._.js:12:501)
    at customRoutes (.next/server/chunks/_8840a2c7._.js:70:534)
    at nz.runtime (.next/server/chunks/_8840a2c7._.js:97:32223)
```

This error only triggers when all three version ranges intersect — Next.js 16 + Makeswift runtime 0.25+ works fine on Node < v24. Since Next.js 16 only requires Node 18.18+, downgrading to Node 22 fixes the issue.

The only Node 24-specific API used in core is `URLPattern` (a global in Node 24 but not in Node 22). Next.js re-exports `globalThis.URLPattern` from `next/server` without polyfilling it, so on Node 22 it would be `undefined`. To address this, `urlpattern-polyfill` is added as a dependency and `URLPattern` is imported directly from the polyfill in `with-auth.ts`.

### Changes

- **`.nvmrc`** — `24` → `22`
- **`core/package.json`** — engine `>=24.0.0` → `>=22.0.0`, added `urlpattern-polyfill` dependency
- **`core/proxies/with-auth.ts`** — import `URLPattern` from `urlpattern-polyfill` instead of `next/server`
- **`packages/create-catalyst/package.json`** — engine `^24.0.0` → `^22.0.0`
- **`packages/create-catalyst/bin/supported-node-versions.cjs`** — `['^24']` → `['^22']`
- **`core/README.md`** and **`README.md`** — updated Node version references from 24 to 22

## Testing

1. Switch to Node 22 (`nvm use 22`) and run `pnpm install`
2. `pnpm run build` in core — confirm it builds without the `#state` TypeError
3. Visit `/account` while unauthenticated and confirm redirect to `/login` (validates `URLPattern` polyfill works in auth middleware)

## Migration

Update your local Node.js version to 22 (via `nvm use` or equivalent). No code changes required for downstream consumers.


chanceaclark (commented):
We either need node@24 or edge runtime for the middleware in order to support `URLPattern` FYI

PR #2976, opened 2026-04-13
feat(core): add X-Correlation-ID header to all GraphQL requests

jorgemoya:
Linear: [LTRAC-226](https://linear.app/commerce/issue/LTRAC-226/)

## What/Why?

Adds a stable `X-Correlation-ID` UUID header to every GraphQL request made during a page render. The UUID is generated once per render via `React.cache` and reused across all fetches in that render, making it straightforward to correlate related requests in server logs.

This is the first in a series of PRs splitting the work from #2910 into smaller, reviewable pieces.

## Rollout/Rollback

No feature flag needed — the header is additive and has no effect on existing behavior. Rollback is a simple revert.

## Testing

- `pnpm build` passes
- `pnpm lint` passes
- Verify in dev tools Network tab that all GraphQL requests within a single page load share the same `X-Correlation-ID` value, and that a new page navigation produces a different UUID

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2910, opened 2026-03-06
feat(core): add X-Correlation-ID header, cache guest queries separately

jorgemoya:
## What/Why?
All GraphQL requests now include an `X-Correlation-ID` header containing a UUID that is stable for the duration of a single page render (via `React.cache`), making it easier to trace and correlate all requests made during a single render in server logs.

Guest (unauthenticated) queries are now cached using `unstable_cache` with the configured revalidation interval, while authenticated requests continue to use `cache: 'no-store'`. This separates cacheable public data from session-specific data, improving performance for unauthenticated visitors. The `X-Forwarded-For` and `True-Client-IP` headers are only forwarded on uncached (`no-store`) requests since they are unavailable inside `unstable_cache`.

## Testing
Locally + E2E

Screenshot 2026-03-09 at 12 31 31 PM

### Step 1: Add the correlation ID helper

Create `core/client/correlation-id.ts`:

```ts
import { cache } from 'react';

/**
 * Returns a stable correlation ID for the current request.
 * React.cache ensures the same UUID is returned for all fetches within a
 * single page render, while being unique across renders/requests.
 */
export const getCorrelationId = cache((): string => crypto.randomUUID());
```

### Step 2: Update `core/client/index.ts`

Update the `beforeRequest` hook to add the `X-Correlation-ID` header to all requests and to only forward `X-Forwarded-For` / `True-Client-IP` on uncached requests:

```diff
+ import { getCorrelationId } from './correlation-id';

  export const client = createClient({
    ...
    beforeRequest: async (fetchOptions) => {
      const requestHeaders: Record<string, string> = {};

-     try {
-       const ipAddress = (await headers()).get('X-Forwarded-For');
-       if (ipAddress) {
-         requestHeaders['X-Forwarded-For'] = ipAddress;
-         requestHeaders['True-Client-IP'] = ipAddress;
-       }
-     } catch {
-       // Not in a request context
-     }
+     if (fetchOptions?.cache && ['no-store', 'no-cache'].includes(fetchOptions.cache)) {
+       try {
+         // headers() is a dynamic API unavailable inside unstable_cache; skip IP forwarding in that context
+         const ipAddress = (await headers()).get('X-Forwarded-For');
+         if (ipAddress) {
+           requestHeaders['X-Forwarded-For'] = ipAddress;
+           requestHeaders['True-Client-IP'] = ipAddress;
+         }
+       } catch {
+         // Not in a request context (e.g. inside unstable_cache); IP forwarding not available
+       }
+     }
+
[...]

PR #2977, opened 2026-04-13
feat(core): add locale parameter to client.fetch()

jorgemoya:
Linear: [LTRAC-226](https://linear.app/commerce/issue/LTRAC-226/)

## What/Why?

Prerequisite for guest query caching with `unstable_cache`.

- Adds an optional `locale` parameter to `client.fetch()` for use in cached contexts where `getLocale()` is unavailable
- Wraps the `headers()` call in `beforeRequest` with try/catch so it gracefully degrades inside `unstable_cache`

No changes to `getChannelId` or `beforeRequest` callback signatures.

Second in a series of PRs splitting #2910 into smaller pieces. Depends on #2976.

## Rollout/Rollback

Backwards compatible — `locale` is optional and all existing call sites continue to work unchanged. Rollback is a simple revert.

## Testing

- `pnpm build` passes
- `pnpm lint` passes
- Verify multi-locale storefronts still resolve the correct channel ID
- Verify IP forwarding headers still appear on uncached requests

### DIFF
--- a/core/client/index.ts
+++ b/core/client/index.ts
@@ -41,11 +41,11 @@ export const client = createClient({
   logger:
     (process.env.NODE_ENV !== 'production' && process.env.CLIENT_LOGGER !== 'false') ||
     process.env.CLIENT_LOGGER === 'true',
-  getChannelId: async (defaultChannelId: string) => {
-    const locale = await getLocale();
+  getChannelId: async (defaultChannelId: string, locale?: string) => {
+    const resolvedLocale = locale ?? (await getLocale());
 
     // We use the default channelId as a fallback, but it is not ideal in some scenarios.
-    return getChannelIdFromLocale(locale) ?? defaultChannelId;
+    return getChannelIdFromLocale(resolvedLocale) ?? defaultChannelId;
   },
   beforeRequest: async (fetchOptions) => {
     // We can't serialize a `Headers` object within this method so we have to opt into using a plain object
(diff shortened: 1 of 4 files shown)


core/client/index.ts
```diff
@@ -41,29 +41,38 @@ export const client = createClient({
...
+      } catch {
+        // headers() is unavailable inside unstable_cache / 'use cache' contexts
       }
     }
 
-    if (locale) {
-      requestHeaders['Accept-Language'] = locale;
+    // Use explicit locale when available (required inside unstable_cache where
+    // getLocale() would fail), fall back to next-intl for normal requests.
+    const resolvedLocale = locale ?? (await getLocale());
```
chanceaclark: Instead of passing in the resolved locale, could we do the same as above where we try/catch instead? Mostly asking here cause I am not sure we want to modify the client and I am trying to think of a way where we don't have to.

jorgemoya: Curious, why not modify the client?

chanceaclark: I just don't want to make a change that we don't have to. I would also say changing the client deserves it's own PR.

jorgemoya: So I played with this for a bit. Honestly, if we can't modify `getChannelId` or `beforeRequest`, what is the point of this PR? Keep in mind the interface adds an optional prop (a `patch` change imo).

Passing locale is the only way to have this available in `unstable_cache` and (eventually) `use cache` . It will be **required** for `use cache`. Since we will be passing locale to all queries within these cache contexts, I think it makes sense that we phase out `getLocale` entirely.

In fact, I would make `locale` mandatory, not optional. I have a PR for it already 😆 .

PR #2989, opened 2026-04-23
fix(core): restore locale-aware lang attribute on html

jorgemoya:
Linear: [LTRAC-578](https://linear.app/commerce/issue/LTRAC-578/pages-always-serve-langen-regardless-of-locale)

## What/Why?

PR #2947 introduced a root `app/layout.tsx` that owns `<html>`/`<body>` so `/404` renders a branded page instead of the default Vercel error screen. Because that root layout sits above `[locale]`, the `lang` attribute was hardcoded to `lang="en"` for every locale — regressing accessibility and SEO for non-English storefronts.

Full `rootParams` (Next.js 16.2) would be the clean fix but isn't usable on Native Hosting yet. This PR adopts [next-intl's documented workaround](https://next-intl.dev/docs/environments/error-files#catching-non-localized-requests) instead:

- `app/layout.tsx` → passthrough (`return children`). Next.js still requires the file to exist, but it owns no markup.
- `app/[locale]/layout.tsx` → owns `<html lang={locale}>` + `<body>` + fonts + `globals.css` again, so each locale gets its correct language attribute.
- `app/not-found.tsx` → self-sufficient, renders its own `<html lang="en">`/`<body>` + fonts + `globals.css`. Keeps the branded 404 from #2947 working for non-localized requests.

Why a passthrough and not deleting the root layout: deleting it forces Next.js 16 to synthesize a fallback shell, which produces nested `<html>`/`<body>` tags in the prerendered `_not-found.html`. A passthrough tells Next.js "trust the children" and avoids the synthesized shell.

## Testing

- `pnpm build` succeeds. Prerendered `.next/server/app/_not-found.html` contains a single `<html lang="en" class="...fonts...">` and a single `<body class="flex min-h-screen flex-col">` wrapping the branded "Not found" content.
- `pnpm start`:
  - `GET /` → 200, `<html lang="en" class="...fonts...">`
  - Any `/[locale]/...` route will render with `lang={locale}` (ticket fix)
  - `GET /admin/nonexistent/` → 404, renders branded not-found. Runtime serves Next.js's `<html id="__next_error__">` streaming shell; verified this is **identical on canary** — baseline Next.js 16 behavior for dynamic 404s, not a regression from this PR. On Vercel edge, the prerendered static file is served directly.

## Migration

- Customers that customized the root `app/layout.tsx` (e.g., added `<head>` elements, body classes, providers) need to move those customizations up to `app/[locale]/layout.tsx` (for locale-aware pages) and/or `app/not-found.tsx` (for the non-localized fallback).
- If a custom root layout was added since #2947 to set `<html lang>` via a cookie or similar workaround, it can be deleted — the locale layout handles this now.

Fixes LTRAC-578

### DIFF
--- a/.changeset/fix-html-lang-locale.md
+++ b/.changeset/fix-html-lang-locale.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": minor
+---
+
+Restore locale-aware `lang` attribute on the root `<html>` tag. The previous root layout hardcoded `lang="en"` for all locales; ownership of `<html>`/`<body>` now lives in `app/[locale]/layout.tsx` so `lang={locale}` reflects the active locale. The root `app/layout.tsx` is now a passthrough, and `app/not-found.tsx` is self-sufficient (renders its own `<html>`/`<body>`) to preserve the branded 404 for non-localized requests.
--- a/core/app/[locale]/layout.tsx
+++ b/core/app/[locale]/layout.tsx
@@ -1,5 +1,6 @@
 import { Analytics } from '@vercel/analytics/react';
 import { SpeedInsights } from '@vercel/speed-insights/next';
+import { clsx } from 'clsx';
 import type { Metadata } from 'next';
 import { notFound } from 'next/navigation';
 import { NextIntlClientProvider } from 'next-intl';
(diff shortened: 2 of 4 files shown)


chanceaclark (approved):
Mind just adding a TODO in code for use to move this back using `rootParams` when we have it available?

agurtovoy (approved):

migueloller:
@jorgemoya, minor nit but the Jira link in the PR description actually points to a Linear issue.

PR #2995, opened 2026-04-27
TRAC-137: Add native hosting option to create-catalyst

jordanarldt:
Jira: [TRAC-137](https://bigcommercecloud.atlassian.net/browse/TRAC-137)

## What/Why?

Adds a `--hosting` flag and prompt to `create-catalyst` so merchants can scaffold a project pre-wired for **Commerce Hosting** in one step instead of running `catalyst project create` / `link` themselves.

The hosting prompt is gated on `GET /v3/infrastructure/projects` (200 → prompt, 403 → silent self-hosted default, other failures → warn and fall back). When Commerce Hosting is selected, the user can pick an existing Commerce Hosting project or create a new one; the resulting UUID is wired into the scaffold automatically. New-project names are checked against existing projects on the store (case-insensitive, matching API behavior) before submission so name collisions surface as a friendly validator error instead of a server roundtrip.

`--hosting` accepts:
- `self-hosted` — Next.js (default if omitted)
- `commerce` — Commerce Hosting (Native)

`--project-name <name>`, when passed alongside `--hosting commerce`, skips the Commerce Hosting project selection prompts and auto-creates a project with that name. If a project with that name already exists on the store, the user is asked Yes/No whether to reuse it (in non-interactive environments / with `--use-existing`, the answer is automatic).

`--use-existing` — only used with `--hosting commerce` and `--project-name`. When the named project already exists on the store, reuse it instead of prompting. Has no effect without `--hosting commerce` (a warning is logged if it's set in another mode).

Native-mode post-clone mutations (`setupCommerceHosting`):

- `core/package.json`: `next build` → `catalyst build`, add `deploy`, drop `start`, add `@bigcommerce/catalyst@1.0.0-alpha.3` + `@opennextjs/cloudflare@1.17.3` (per the [Native Hosting docs](https://docs.bigcommerce.com/developer/docs/storefront/catalyst/deployment/native-hosting/getting-started))
- `core/.bigcommerce/project.json`: written with `projectUuid`, `framework`, plus `storeHash` / `accessToken` when available
- `core/.env.local`: symlinked to `../.env.local` (warns and continues if symlinks aren't permitted; tagged with `TODO(LTRAC-409): remove this symlink`)
- `core/proxy.ts` → `core/middleware.ts` with the edge runtime injection ([Next.js 16 migration note](https://docs.bigcommerce.com/developer/docs/storefront/catalyst/deployment/native-hosting/getting-started#nextjs-16-migration-note))

Versions are pinned exactly — these need a deliberate bump alongside `@bigcommerce/catalyst` releases.

## User-facing terminology

To keep the two "project" concepts distinct, the CLI distinguishes:

- **Project directory** — the local folder being scaffolded. Controlled by `--project-name` / `--project-dir`; the interactive prompt is `"What is the name of your project directory?"`.
- **Commerce Hosting project** — the remote record on Commerce that hosts the storefront. Created or selected during the hosting flow.

Files and functions touching the user-facing flow were renamed to match:
- `setup-native-hosting.ts` → `setup-commerce-hosting.ts` (`setupNativeHosting` → `setupCommerceHosting`)
- `prompt-infrastructure-project.ts` → `prompt-commerce-hosting-project.ts` (`promptForInfrastructureProject` → `promptForCommerceHostingProject`)
- API-client identifiers (`listInfrastructureProjects`, `createInfrastructureProject`, `InfrastructureProjectValidationError`) keep the "infrastructure" name since they map directly to `/v3/infrastructure/projects`.

## Error handling

API methods now wrap fetch / parse / non-OK-status failures in a single friendly Error (`Could not load Commerce Hosting projects: ...`, `Could not create Commerce Hosting project: ...`) with the original error preserved as `cause`. `InfrastructureProjectValidationError` is intentionally re-thrown unwrapped so consumer `instanceof` checks still work.

## Rollout/Rollback

Behind the existing Infrastructure Projects API feature flag — merchants without access keep the current self-hosted scaffold unchanged. Rollback is a code revert; no migrations or runtime changes.

## Testing

71 unit tests across the API client, hosting-mode resolution, the prompt + retry / auto-create / collision flow, the `--use-existing` flag (TTY and non-TTY paths), and the file mutations.

Manual end-to-end:
```bash
cd packages/create-catalyst && pnpm build
node bin/index.cjs --project-name test --hosting commerce
# verify core/package.json scripts, core/.bigcommerce/project.json, core/.env.local symlink
```

Scaffolding works end-to-end against a store with infrastructure access. `pnpm build` / `pnpm deploy` on the resulting project are blocked on a separate downstream issue, out of scope for this PR.

## Proof of life

https://github.com/user-attachments/assets/1386398b-6d96-4d56-86d9-0de7924f2fb8

[TRAC-137]: https://bigcommercecloud.atlassian.net/browse/TRAC-137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/create-catalyst/src/commands/create.ts
```diff
@@ -13,9 +13,13 @@ import { Https } from '../utils/https';
 import { installDependencies } from '../utils/install-dependencies';
 import { getAvailableLocales } from '../utils/localization';
 import { login, storeCredentials } from '../utils/login';
+import { promptForInfrastructureProject } from '../utils/prompt-infrastructure-project';
+import { setupNativeHosting } from '../utils/setup-native-hosting';
 import { Telemetry } from '../utils/telemetry/telemetry';
 import { writeEnv } from '../utils/write-env';
 
+type HostingMode = 'next' | 'catalyst';
```
chanceaclark: Should we change the naming to be the hosting providers instead of the frameworks?
```suggestion
type HostingMode = 'vercel' | 'commerce';
```
or
```suggestion
type HostingMode = 'self-hosted' | 'commerce';
```

packages/create-catalyst/src/commands/create.ts
```diff
@@ -283,6 +287,116 @@ async function setupProject(options: {
...
+  }
+
+  const cliApi = new CliApi({
+    origin: cliApiOrigin,
+    storeHash,
+    accessToken,
+    apiHostname,
+  });
+
+  let hasAccess = false;
```
chanceaclark: 🍹 see if we can write this without `let`'s.

jordanarldt: @chanceaclark Good call. I hate `let`'s

packages/create-catalyst/src/commands/create.ts
jordanarldt: @chanceaclark Ahh yeah I was actually thinking something similar yesterday. I do like `self-hosted` | `commerce`, because it is possible someone might opt for `self-hosted` instead of using vercel

chanceaclark (approved):
Overall LGTM. I'd let a second opinion from @bigcommerce/team-trac have some input before merging though.

packages/create-catalyst/src/commands/create.ts
```diff
@@ -316,6 +438,12 @@ export const create = new Command('create')
   .option('--reset-main', 'Reset the main branch to the gh-ref')
   .option('--repository <repository>', 'GitHub repository to clone from', 'bigcommerce/catalyst')
   .option('--env <vars...>', 'Arbitrary environment variables to set in .env.local')
+  .addOption(
+    new Option(
+      '--hosting <mode>',
+      'Hosting mode: "self-hosted" for Next.js or "commerce" for Commerce-hosted',
```
parthshahp: ⛏️  I'd like to think a bit more about these hosting flags, same with the help message.

Is the logic here: "self-hosted" == Vercel right now, but in theory can be any hosting provider (including self-hosted), and "commerce" == Native Hosting on CF? 

I guess my question here is: why not put "vercel" and "commerce" as the options?

packages/create-catalyst/src/utils/prompt-infrastructure-project.ts
```diff
@@ -0,0 +1,73 @@
...
+  defaultName?: string,
+): Promise<InfrastructureProject> {
+  const existingProjects = await cliApi.listInfrastructureProjects();
+
+  if (existingProjects.length === 0) {
+    return promptAndCreateInfrastructureProject(cliApi, defaultName);
+  }
+
+  const action = await select<'existing' | 'create'>({
+    message: 'Would you like to use an existing infrastructure project or create a new one?',
```
jorgemoya: Do we surface that these projects are "infrastructure" anywhere else? Sounds very techy/formal haha.

jorgemoya (approved):

packages/create-catalyst/src/utils/prompt-infrastructure-project.ts
jordanarldt: @jorgemoya Good point. We do surface this "infrastructure" verbiage in the API endpoint for projects: `/v3/infrastructure/projects`, but maybe the wording here in the CLI is unnecessary. Although that means when you run the `create-catalyst` command, there is a prompt asking you for your "project name" (e.g. the folder to create the next app in), and then later you'll be asked again for your "project name", but this time it's for the native hosting project name 🤔  How do you think we should distinguish these?

packages/create-catalyst/src/commands/create.ts
jordanarldt: @parthshahp I think `self-hosted` makes the most sense, because it doesn't necessarily mean Vercel. When not using Native Hosting, the consumer can host their NextJS app anywhere that supports hosting NextJS (heroku, vercel, google cloud, etc)

parthshahp: That makes sense. I think the message `"self-hosted" for Next.js` is a bit confusing here, because technically it's Next.js either way, right? If I saw this message, I would think:

If "self-hosted" is how I use Next.js, what framework is "commerce"?

jordanarldt: @parthshahp Yeah good point, let me fix that!

chanceaclark: I think that `self-hosted` here means that it's hosted elsewhere. We have merchants hosting on Vercel, but also in AWS or other hosting providers right now. The point of `self-hosted` here is more on the path that we ship default Next.js commands, not our Native Hosting specific setup. `self-hosted` also opens up the door that we can support `aws` as an option and tweak what the command ships to be aligned with AWS.

I'm not married to the name, but we should probably avoid "framework" here. I think of "framework" as `Next.js`, `Vue`, or `React`, but not a "hosting framework". If we wanted to remove `self-hosted` we could say the flag follows this logic:
- Only provide `commerce` (or whatever we want to name it) as the only option.
- If `--hosting commerce` is provided, ship with Native Hosting augments
- If no flag is provided or option selected, ship with default Next.js
- If invalid flag name, error.

I really want Catalyst to lean into this "composability" aspect, and this flag is probably one of those design choices. I want developers to have the option of taking a base output and being able to change what they need to fit their needs. If they want to ship it to AWS, they can; Vercel, easy; etc....

migueloller:
I feel like we're mixing concerns here. I'm not sure that we should have this option on the `create-catalyst` CLI, it feels like this logic should live in the `catalyst` CLI as part of the deploy command. In particular, after running `create-catalyst`, if I run `catalyst deploy` we should be smart and take the user through all these steps.

We're adding a ton of logic to `create-catalyst` to set up hosting configuration. `create-catalyst` should be just about scaffolding a Catalyst codebase, and should delegate to `catalyst init` probably. It feels like a lot of what we're doing here is the concern of the Catalyst CLI and we'd want that to happen at deploy time.

An example of where decoupling `create-catalyst` from `catalyst deploy` helps: a user already has an existing Catalyst site and they want to use Commerce Hosting for Catalyst. They're not going to run `create-catalyst`, they already have an app. Instead, what they're going to do is reach for `catalyst deploy`. But since we've put all this logic here, then now they don't get this nice experience.

PR #3001, opened 2026-05-02
feat(cli): add `channel update` command to update channel site url

jorgemoya:
Jira: [LTRAC-446](https://linear.app/commerce/issue/LTRAC-446/cli-auto-update-bc-channel-site-url-on-deploy)

## What/Why?

Adds a new top-level subcommand, `catalyst channel update`, and a boolean `--update-site-url` flag on `catalyst deploy` that both run the same interactive flow:

1. Resolve the linked project (or fall back to `selectOrCreateInfrastructureProject` if none is linked or the linked UUID no longer exists).
2. Fetch available channels, **filter to `platform === 'catalyst'`**, and prompt the user to pick one.
3. Read the project's `deployment_hostnames` — prompt the user to pick one (when invoked from `deploy --update-site-url`, the just-deployed hostname is placed first as the default).
4. PUT the channel's site URL.

This is a more discoverable replacement for the previous design (which exposed `--update-site-url <channelId>` only on deploy and required the caller to already know the channel id). Standalone, the new command lets merchants point any Catalyst channel at any of the project's deployment hostnames — useful for switching channel→deployment mappings, recovering after a manual control-panel change, or wiring up a channel for the first time without re-deploying.

A few decisions worth flagging for review:

- **Standalone command supports `--channel-id` / `--hostname` overrides** so it can be scripted. The deploy flag stays minimal (boolean only) because the just-deployed hostname is the obvious default.
- **The shared flow lives in `lib/channel-site-flow.ts`** so both surfaces hit identical code paths. The deploy wrapper soft-fails on errors (bundle is already live, non-zero exit would mislead); the standalone command hard-fails so problems are loud.
- **The channel picker filters to `platform === 'catalyst'`** — Stencil and other non-Catalyst storefronts can't meaningfully be pointed at a Catalyst deployment hostname, so they're omitted. Empty list surfaces "No Catalyst channels found on this store. Create one with `catalyst create` and try again."
- **OAuth scope** `store_channel_settings` was added to `DEVICE_OAUTH_SCOPES` in the earlier commit on this branch. Existing logged-in users need to re-run `catalyst auth login` to mint a token that includes it; the 401/403 path surfaces a re-auth hint instead of a generic error.

## Testing

Unit tests:

- `pnpm --filter @bigcommerce/catalyst test` — 265/265 passing (11 in `channel-site-flow.spec.ts`, 6 in `channel.spec.ts`, plus updates to the existing `deploy.spec.ts` `--update-site-url` block).
- `pnpm --filter @bigcommerce/catalyst typecheck` and `pnpm --filter @bigcommerce/catalyst lint` clean.

Manual end-to-end (real store + at least one Catalyst channel + a project with at least one deployment):

1. `catalyst auth login` — confirm the new scope is requested.
2. `catalyst channel update` — confirm the channel picker shows only Catalyst-platform storefronts and the hostname picker shows the linked project's `deployment_hostnames`. PUT succeeds; verify in the BC control panel.
3. `catalyst channel update --channel-id <id> --hostname <host>` — confirm no prompts fire and PUT succeeds.
4. `catalyst channel update` on a store with no Catalyst channels — confirm the "No Catalyst channels found" message with the `catalyst create` hint.
5. `catalyst channel update` with no linked project — confirm it falls back to `selectOrCreateInfrastructureProject`. Declining a create exits cleanly with the `catalyst project create` hint.
6. `catalyst deploy --update-site-url` — confirm the channel prompt fires, the just-deployed hostname is the default (top of the hostname list), PUT succeeds. Verify in control panel.
7. `catalyst deploy` without the flag — confirm no channel API call is made.
8. With a token missing `store_channel_settings`, run the standalone command and confirm the re-auth hint surfaces.

https://github.com/user-attachments/assets/8d065c84-e454-4b8b-b3d9-ee8016c2e723

## Migration

None. The flag is opt-in and the `channel` command is new. Users who want the auto-update should re-run `catalyst auth login` (new OAuth scope) before invoking either entry point.

### DIFF
--- a/packages/catalyst/src/cli/commands/deploy.ts
+++ b/packages/catalyst/src/cli/commands/deploy.ts
@@ -6,7 +6,7 @@ import { dirname, join } from 'node:path';
 import yoctoSpinner from 'yocto-spinner';
 import { z } from 'zod';
 
-import { updateChannelSiteUrl } from '../lib/channels';
+import { runChannelSiteUrlFlow } from '../lib/channel-site-flow';
 import {
   NoLinkedProjectError,
   selectOrCreateInfrastructureProject,
@@ -347,14 +347,12 @@ export const getDeploymentStatus = async (
   spinner.success('Deployment completed successfully.');
 
   if (deploymentHostname) {
-    const url = `https://${deploymentHostname}`;
-
-    consola.success(`View your deployment at: ${colorize('blue', url)}`);
-
-    return url;
+    consola.success(
+      `View your deployment at: ${colorize('blue', `https://${deploymentHostname}`)}`,
+    );
   }
 
-  return undefined;
+  return deploymentHostname;
 };
 
 export const fetchProject = async (
@@ -399,13 +397,9 @@ Example:
   .addOption(accessTokenOption())
   .addOption(apiHostOption())
   .addOption(projectUuidOption())
-  .addOption(
-    new Option(
-      '--update-site-url <channelId>',
-      "BigCommerce channel ID whose site URL should be updated with this deployment's URL. When omitted, no channel is updated.",
-    )
-      .env('CATALYST_UPDATE_SITE_URL')
-      .argParser((value: string) => Number(value)),
+  .option(
+    '--update-site-url',
+    "After a successful deploy, prompt to update a channel's site URL to the new hostname.",
   )
   .addOption(
     new Option(
@@ -558,34 +552,33 @@ Example:
       environmentVariables,
     );
 
-    const deploymentUrl = await getDeploymentStatus(
+    const deploymentHostname = await getDeploymentStatus(
       deploymentUuid,
       storeHash,
       accessToken,
       options.apiHost,
     );
 
-    const channelId: number | undefined = options.updateSiteUrl;
-
-    if (!channelId) {
-      return;
-    }
-
-    if (!deploymentUrl) {
-      consola.warn('Skipping channel site URL update: deployment did not return a URL.');
-
+    if (!options.updateSiteUrl) {
       return;
     }
 
     try {
-      await updateChannelSiteUrl(channelId, deploymentUrl, storeHash, accessToken, options.apiHost);
-      consola.success(`Updated channel ${channelId} site URL to ${deploymentUrl}.`);
+      await runChannelSiteUrlFlow({
+        storeHash,
+        accessToken,
+        apiHost: options.apiHost,
+        projectUuid,
+        preferHostname: deploymentHostname,
+      });
     } catch (error) {
+      // Soft-fail: the deploy already succeeded and the bundle is live. A
+      // non-zero exit here would be misleading.
       consola.warn(
         `Failed to update channel site URL: ${error instanceof Error ? error.message : String(error)}`,
       );
       consola.info(
-        'Update it manually in the control panel, or re-run after `catalyst auth login` if the token is missing the store_channel_settings scope.',
+        'Update it manually in the control panel, or re-run `catalyst auth login` if the token is missing the store_channel_settings scope.',
       );
     }
   });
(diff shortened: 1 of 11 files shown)


packages/catalyst/src/cli/commands/deploy.ts
```diff
@@ -479,5 +497,54 @@ Example:
...
+      return;
+    }
+
+    try {
+      const current = await fetchChannelSite(channelId, storeHash, accessToken, options.apiHost);
+
+      if (current?.url === deploymentUrl) {
+        consola.info(`Channel ${channelId} site URL already up to date (${deploymentUrl}).`);
+      } else {
+        await updateChannelSiteUrl(
```
chanceaclark: This actually might mess up people testing deployments to native hosting and would be very destructive.

For example, I might have channel `123` being deployed via Vercel and I am dual-deploying this to Vercel and Native Hosting, deploying will change the channels site, thus messing up the production deployment.

chanceaclark (changes requested):
I think the intention behind this change is good, but I think it might be very destructive and we should probably avoid this. Maybe we can go in the direction that you can pass a flag to do this?

packages/catalyst/src/cli/commands/deploy.ts
jorgemoya: Good callout, 100% agree, lets make this work opt in via an explicit flag that needs to be passed when deploying.

Having said that, do we want to remove all logic from `project create` and `project link` to add the channel to `project.json`? Or we keep that and just add the explicit flag?

chanceaclark: I don't think hosting should be tied to a channel.

jorgemoya: I have simplified it considerably by simply having a `--update-site-url` flag that requires a `channelId`.  Still kept the `store_channel_settings` oauth scope to allow this for users who are authing via the CLI.

The unknown I have is that once we merge the `deployment_hostnames` changes, the returned array can contain N urls. How will one select the one they want to update to? Obviously we could default to the first one, but maybe it needs to be user selected?

chanceaclark: What if instead of doing this on the deploy command we have a separate CLI command to manage this? We'd need to think through the UX for it, but I think it might be better than just a flag?

jorgemoya: Hmm, I see your point. I would wait to get the `deployment_hostnames` work merged first then.

What I would do is a new command that updates the site channel id. It will get all hostnames under the defined project, and then allow the user to select which hostname they want to be used for site url.

chanceaclark: Yeah something like that.

packages/catalyst/src/cli/lib/channel-site-flow.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  if (project.deployment_hostnames.length === 0) {
+    throw new Error(
+      `Project "${project.name}" has no deployment hostnames yet. Run \`catalyst deploy\` first to create one.`,
+    );
+  }
+
+  // When the caller knows which hostname they want surfaced first (e.g. the
+  // freshly-deployed one from `catalyst deploy --update-site-url`), order it to
+  // the top of the list so it's the default selection.
```
parthshahp, line 106: Nice detail!

packages/catalyst/src/cli/lib/auth.ts
```diff
@@ -6,6 +6,7 @@ export const DEVICE_OAUTH_SCOPES = [
   'store_infrastructure_deployments_manage',
   'store_infrastructure_logs_read_only',
   'store_infrastructure_projects_manage',
+  'store_channel_settings',
```
parthshahp, line 9: Double checking, we need to add this scope for this endpoint? Also, this means that any existing cli users who already have a token will need to re-create a token using the `auth` command?

parthshahp (approved):

packages/catalyst/src/cli/lib/auth.ts
jorgemoya: Yes, unfortunately! But we do show a message if that is the case (if they need to re-auth).

PR #3003, opened 2026-05-04
TRAC-137: Add native hosting option in catalyst CLI

jordanarldt:
Jira: [LTRAC-137](https://bigcommercecloud.atlassian.net/browse/LTRAC-137)

## What/Why?

Following review of #2995, the team agreed that hosting/deploy concerns belong in `catalyst` (the CLI), not `create-catalyst` (the scaffolder). Since the broader plan is to absorb `create-catalyst` into `catalyst` anyway, this PR ports the work directly into `packages/catalyst` as a new `catalyst create` command, plus a few related improvements that fell out of building on that foundation.

### `catalyst create`

New command. Scaffold-only by default — no interactive hosting prompt, since most users will reach Commerce Hosting via `catalyst deploy`. `--hosting commerce` opts into the full Commerce Hosting setup eagerly (proxy → middleware swap, OpenNext dep, `.bigcommerce/project.json`, `core/.env.local` symlink). Reuses the existing `catalyst auth` device flow.

`setupCoreProject` runs unconditionally at create time, wiring `catalyst build` / `catalyst start` / `catalyst deploy` scripts and the `@bigcommerce/catalyst` dep into `core/package.json` regardless of hosting choice — the polymorphic dispatcher (below) makes this safe for self-hosted projects too.

### Polymorphic `catalyst build` / `catalyst start`

Both commands dispatch via `lib/project-state.ts`. Untransformed projects fall through to `pnpm exec next build` / `next start`; transformed projects run the OpenNext pipeline. Lets the create-time scripts work end-to-end whether or not the user opts into Commerce Hosting.

### Shared `lib/commerce-hosting.ts`

Setup logic from #2995 extracted into a reusable module: `setupCommerceHosting`, `promptForCommerceHostingProject`, `selectOrCreateInfrastructureProject`, `runCommerceHostingSetup`, plus a `NoLinkedProjectError` sentinel for graceful decline handling. `selectOrCreateInfrastructureProject` is shared between `catalyst project link` and `catalyst deploy`, decorating the matching project with green `[linked]` and skipping the select prompt entirely on empty stores.

### `catalyst project link` / `list`

`list` and `link`'s select prompt now mark the currently linked project with green `[linked]`. After a successful link, `link` offers to run Commerce Hosting setup if the project isn't already transformed.

### `catalyst deploy`

Verifies the linked project still exists on the store before any build/upload work. If missing or 404, falls through to `selectOrCreateInfrastructureProject`. Includes a transformation guard that prompts for Commerce Hosting setup when the project isn't transformed. Graceful info + exit on user decline.

### Misc / quality

- **No more `.bigcommerce/` cwd pollution.** Telemetry moved to a user-scoped `Conf` (lives in OS config dir under `catalyst-cli/`); project state checks moved to a read-only `getProjectState()` helper that doesn't instantiate `Conf`.
- **Wired core scripts.** `core/package.json` always gets `build: "npm run generate && catalyst build"` / `start: "catalyst start"` / `deploy: "npm run generate && catalyst deploy"`.
- **Canonical `package.json` field order** (`lib/sort-package-json.ts`) applied whenever we mutate a `package.json`.
- **Shared option helpers** in `lib/shared-options.ts` now used across `project`, `create`, `deploy` (each helper accepts an optional description override).
- **Consistent HTTP layer.** Removed `lib/https.ts` (stateful client class) and `lib/user-agent.ts`. All API code in `lib/` (project.ts, channels.ts, localization.ts) now follows the same pattern: positional `(...data, storeHash, accessToken, host)` args, schemas + parsing internal to the lib, throws on bad responses. `lib/channels.ts` now owns all channel-related schemas/types (previously scattered across `commands/create.ts`).
- **`--env` and `--channel-id` validate at parse time.** Custom argParsers throw `InvalidArgumentError` on malformed input; values containing `=` in `--env` are preserved.
- **`buildWorkspacePackages`** runs after `installDependencies` at create time when the cloned repo is the catalyst monorepo (`core/`, `packages/catalyst`, `packages/client` all present), so `core` can resolve workspace deps.

## Testing

228 unit tests, all passing. New: `commands/create.spec.ts` (12 tests covering happy paths, parser validation, ordering invariants, failure handling, `--hosting commerce` precondition), `lib/project-state.spec.ts`, `lib/setup-core-project.spec.ts`, `lib/sort-package-json.spec.ts`. Extended: `commerce-hosting`, `project`, `deploy`, `build`, `start` specs.

`pnpm --filter @bigcommerce/catalyst test`, `typecheck`, `lint`, and `build` all clean.

Manually verified the major paths against a real store: `catalyst create` (interactive, `--hosting commerce`, `--hosting self-hosted`), `catalyst project link` / `list` with `[linked]` markers, `catalyst deploy` first-run transformation prompt, `catalyst deploy` against a deleted project (re-link prompt + graceful decline).

### Proof of life

https://github.com/user-attachments/assets/f13ee750-50f6-472a-a86d-91f44628aba8

https://github.com/user-attachments/assets/6a9a238c-ce28-472e-82be-6db488d38674

## Migration

`packages/create-catalyst` is **not modified** in this PR — the staged #2995 changes there were reverted. A follow-up PR will update `create-catalyst create` to shim into `catalyst create` so `pnpm create @bigcommerce/catalyst` keeps working.

External behavior changes:
- `BIGCOMMERCE_STOREFRONT_API_TOKEN` env var (only ever written by the create flow, never read by the storefront) is now correctly `BIGCOMMERCE_STOREFRONT_TOKEN`.
- The catalyst-CLI no longer sets a custom `User-Agent` header on requests to the cli-api gateway or BC store API. Server-side telemetry that buckets on `catalyst-cli/...` UA will see the Node default UA instead.


jorgemoya:
Should we split up `deploy` and `link` changes from this PR?

packages/catalyst/src/cli/program.ts
```diff
@@ -17,6 +19,22 @@ import { version } from './commands/version';
 import { telemetryPostHook, telemetryPreHook } from './hooks/telemetry';
 import { consola } from './lib/logger';
 
+// Auto-load .env.local from cwd if present, matching Next.js convention.
+// `--env-path` (loaded later via argParser) overrides anything we load here.
+const defaultEnvPath = resolve(process.cwd(), '.env.local');
+
+if (existsSync(defaultEnvPath)) {
+  config({ path: defaultEnvPath });
+}
```
jorgemoya, line 28: We're removing any implicit declaration of env vars soon so this probably needs to go too (unless I'm mistaken about the ask).

packages/catalyst/src/cli/program.ts
```diff
@@ -17,6 +19,22 @@ import { version } from './commands/version';
...
+
+// CATALYST_STORE_HASH falls back to BIGCOMMERCE_STORE_HASH so freshly-scaffolded
+// projects work without duplicating the same value under two names. Aliasing
+// here means every command's `.env('CATALYST_STORE_HASH')` binding picks it up
+// without per-command changes.
+if (!process.env.CATALYST_STORE_HASH && process.env.BIGCOMMERCE_STORE_HASH) {
+  process.env.CATALYST_STORE_HASH = process.env.BIGCOMMERCE_STORE_HASH;
+}
+
```
jorgemoya, line 37: Same here

packages/catalyst/src/cli/commands/build.ts
```diff
@@ -96,6 +97,23 @@ Examples:
     ).env('CATALYST_PROJECT_UUID'),
   )
   .action(async (options) => {
+    // Project must be transformed (middleware swapped in, OpenNext dep installed)
+    // before the OpenNext build pipeline can run. If it isn't, fall through to
+    // `next build` so this command works for self-hosted Catalyst projects too.
```
jorgemoya, line 102: Are we expecting people to build with the cli instead of `next build`?

packages/catalyst/src/cli/commands/deploy.ts
```diff
@@ -378,29 +391,17 @@ Example:
   $ catalyst deploy --secret BIGCOMMERCE_STORE_HASH=<YOUR_STORE_HASH> --secret BIGCOMMERCE_STOREFRONT_TOKEN=<YOUR_STOREFRONT_TOKEN>`,
   )
   .addOption(
-    new Option(
-      '--store-hash <hash>',
-      'BigCommerce store hash. Can be found in the URL of your store Control Panel. Read from .bigcommerce/project.json when not provided.',
-    ).env('CATALYST_STORE_HASH'),
+    storeHashOption(
+      'BigCommerce store hash. Can be found in the URL of your store Control Panel. Read from .bigcommerce/project.json or .env when not provided.',
```
jorgemoya: Shouldn't the message be included in `storeHashOption`, or why would it be different?

packages/catalyst/src/cli/commands/deploy.ts
```diff
@@ -378,29 +391,17 @@ Example:
...
-      .hideHelp(),
-  )
-  .addOption(
-    new Option(
-      '--project-uuid <uuid>',
-      'BigCommerce intrastructure project UUID. Can be found via the BigCommerce API (GET /v3/infrastructure/projects).',
-    ).env('CATALYST_PROJECT_UUID'),
+    accessTokenOption(
+      'BigCommerce access token. Can be found after creating a store-level API account. Read from .bigcommerce/project.json or .env when not provided.',
+    ),
```
jorgemoya: Same here

packages/catalyst/src/cli/program.ts
jordanarldt: @jorgemoya Yeah I'm aware we want to remove the implicit declarations, but I think that is more in regards to supplying the .env.local vars to the deployment secrets implicitly, where this line is just to make sure that the CLI also has access to the `.env.local` file

packages/catalyst/src/cli/commands/build.ts
jordanarldt: @jorgemoya It will still use `next build` under the hood, but we want to bootstrap the new projects using the `catalyst build` command so that we don't have to make any `package.json` changes later (if a user decides to use native hosting)

So although package.json always uses `catalyst build`, the real flow looks like this:
```
catalyst build -> (is user on native hosting) YES -> native hosting build (wrangler)
                  (is user on native hosting) NO -> next build
```

packages/catalyst/src/cli/commands/deploy.ts
jordanarldt: @jorgemoya For some reason the messaging was already different for the deploy command, so I just maintained it. I moved the `--store-hash` argument to a shared definition (`storeHashOption`) instead of redefining it, so I just added a description override in the `storeHashOption` function.

The only difference in the description is this part:
> Read from .bigcommerce/project.json or .env when not provided.

I felt like it might be necessary to let users know where the variables are read for the deploy command

jorgemoya: If the change is minor, I would add it or remove from all and remove the message override tbh. 🍹

jorgemoya (approved):

packages/catalyst/src/cli/commands/deploy.ts
jordanarldt: @jorgemoya Fair enough!

PR #3006, opened 2026-05-08
chore(common): Create Claude skill to setup Catalyst on dev store

Tharaae:
## What/Why?
Adds a Claude Code skill (/setup-catalyst-cdvm) that automates setting up Catalyst against a local CDVM dev store. The skill walks Claude through the full 12-step setup flow: detecting/resuming the CDVM cluster, picking an active store, creating an OAuth token, activating the Catalyst profile, enabling AdditionalStorefronts and Multi-Storefront flags, creating a Catalyst channel + site, generating a storefront API token, populating core/.env.local, and launching pnpm run dev in a new Terminal window.                                                           
               
Two modes are offered:                                                                                                                           
  - Automated (recommended) — auto-detects cluster and store via bcli; only the manual store-config UI steps need user input.
  - Manual — user supplies cluster name, store ID, and store hash upfront. Faster when you already have a running cluster and active store.   

## Testing
[Hackathon presentation](https://docs.google.com/presentation/d/1euMKW1pXxG64_zZYgWT04qgRzVBI7CZy/edit?usp=drive_link&ouid=100578269999864089929&rtpof=true&sd=true)

## Migration
None

### DIFF
--- a/.claude/skills/setup-catalyst-cdvm/SKILL.md
+++ b/.claude/skills/setup-catalyst-cdvm/SKILL.md
@@ -1,0 +1,249 @@
+---
+name: setup-catalyst-cdvm
+description: Use when setting up Catalyst on a local CDVM dev store. Requires an existing CDVM cluster.
+---
+
+# Setup Catalyst on a Local Dev Store
+
+## Prerequisite: Confirm cloud-dev-vm Path
+
+Check if `cloud-dev-vm` exists as a sibling of the current Catalyst directory:
+
+```bash
+ls "$(dirname $(pwd))/cloud-dev-vm" 2>/dev/null && echo "exists" || echo "not found"
+```
+
+- If it exists → use `$(dirname $(pwd))/cloud-dev-vm` as `cdvmPath`
+- If not found → ask the user:
+
+> Could not find `cloud-dev-vm` at the expected path (`../cloud-dev-vm` relative to this repo). Please provide the full path to your `cloud-dev-vm` directory.
+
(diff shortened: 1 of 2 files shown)


.claude/skills/setup-catalyst-cdvm/SKILL.md
chanceaclark, line 1: I love this, but since CDVM is more internal, can we move this into our `cc-plugins-marketplace` repo instead!? It can probably go into a new `bc-cdvm` plugin in that repo.

Tharaae: I was actually thinking the same right after I pushed the PR 😄

Tharaae: However, this is a Catalyst-specific skill, same as `release-catalyst` and `release-catalyst-patch` skills here in the repo as well, although they are also internal, aren't they? Is there a way to keep those skills here where they belong but at the same time, not copied via `npm create @bigcommerce/catalyst@latest`? 🤔

chanceaclark: Those skills are specific to the repo itself. I think the callout here is that we are exposing CDVM, an internal dev environment, rather than skills to help release new versions of Catalyst. The skills we ship are more for public visibility into our release process.

The `npm create @bigcommerce/catalyst@latest` command just clones the whole monorepo, yeah we could filter it out, but some people choose to clone it directly instead of using that command so we can filter it that way.

It doesn't have to be a `bc-cdvm` plugin, it can be something else too, e.g. `bc-catalyst` 🤷

PR #3009, opened 2026-05-11
fix(core): TRAC-277 translate Show more/less labels on product compare page

bc-yevhenii-buliuk:
## What/Why?
This PR fixes the 'Show more' / 'Show less' labels in the Reveal component on the Product compare page.
The labels were hardcoded in English because `showMoreLabel`/`showLessLabel` props were not passed through the component chain (CompareSection → CompareCard → Reveal).

## Testing
locally
**before:**
trac_277_bug

**after:**
trac_277_fixed

## Migration
n/a

### DIFF
--- a/core/vibes/soul/primitives/compare-card/index.tsx
+++ b/core/vibes/soul/primitives/compare-card/index.tsx
@@ -33,6 +33,8 @@ export interface CompareCardProps {
   noOtherDetailsLabel?: string;
   viewOptionsLabel?: string;
   preorderLabel?: string;
+  showMoreLabel?: string;
+  showLessLabel?: string;
   addToCartAction?: CompareAddToCartAction;
   imageSizes?: string;
 }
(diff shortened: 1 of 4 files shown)


jairo-bc (approved):

core/vibes/soul/primitives/compare-card/index.tsx
```diff
@@ -66,6 +68,8 @@ export function CompareCard({
   noOtherDetailsLabel = 'There are no other details.',
   viewOptionsLabel = 'View options',
   preorderLabel = 'Preorder',
+  showMoreLabel = 'Show more',
+  showLessLabel = 'Show less',
```
jorgemoya, line 72: We've moved away from prop drilling, and instead we can simply call the `t` helper within this component. BUT merging as is, is also fine. 🍹

jorgemoya (approved):

PR #3015, opened 2026-05-18
feat(core): TRAC-421 consume merchant-configured locale subfolders from BC backend

mfaris9:
## What / Why

[PROJECT-5002](https://bigcommercecloud.atlassian.net/browse/PROJECT-5002) added merchant-configurable per-locale URL subfolders in the BC Control Panel — e.g., `fr-CA` can be served at `/fr/` instead of `/fr-CA/`. The Storefront GraphQL API exposes this as `Locale.fullPath`.

This PR teaches Catalyst to use that configured subfolder for URL routing, while keeping the locale code for translations, `<html lang>`, `Accept-Language`, and BC content negotiation.

## JIRA
[TRAC-421](https://bigcommercecloud.atlassian.net/browse/TRAC-421)

## Testing

Tested 5 cases in Integration.

Case 1 — No custom subfolders (legacy)

{locale = fr, isDefault=true, path=""}
{locale = es-es, isDefault=false, path=""}
{locale = es-mx, isDefault=false, path=""}
{locale = uk, isDefault=false, path=""}

Expected URLs:

site.com/               <-- fr locale
site.com/es-es/    <-- es-es locale
site.com/es-mx/   <-- es-mx locale
site.com/uk/          <-- uk locale

- [x] Tested

Case 2 — One non-default has a subfolder

{locale = fr, isDefault=true, path="" }
{locale = es-es, isDefault=false, path="es"}
{locale = es-mx, isDefault=false, path=""}
{locale = uk, isDefault=false, path=""}

Expected URLs:

site.com/               <-- fr locale
site.com/es/          <-- es-es locale
site.com/es-mx/   <-- es-mx locale
site.com/uk/          <-- uk locale

- [x] Tested

Case 3 — Default and one non-default have subfolders

{locale = fr, isDefault=true, path="fr" }
{locale = es-es, isDefault=false, path="es"}
{locale = es-mx, isDefault=false, path=""}
{locale = uk,    isDefault=false, path=""}

Expected URLs:

site.com/fr/           <-- fr locale
site.com/es/          <-- es-es locale
site.com/es-mx/   <-- es-mx locale
site.com/uk/          <-- uk locale

Note: nobody lives at bare /. Hitting site.com/ redirects to site.com/fr/.

- [x] Tested

Case 4 — Every non-default has a subfolder, default blank

{locale = fr, isDefault=true, path="" }
{locale = es-es, isDefault=false, path="es"}
{locale = es-mx, isDefault=false, path="mx"}
{locale = uk,    isDefault=false, path="uk"}

Expected URLs:

site.com/                 <-- fr locale
site.com/es/            <-- es-es locale
site.com/mx/           <-- es-mx locale
site.com/uk/            <-- uk locale

- [x] Tested

Case 5 — Default has a path, one non-default left blank

{locale = fr, isDefault=true, path="fr" }
{locale = es-es, isDefault=false, path="es"}
{locale = es-mx, isDefault=false, path="ex-mx"}
{locale = uk,    isDefault=false, path=""}

- [x] Tested

site.com/fr/             <-- fr locale
site.com/es/            <-- es-mx locale
site.com/es-mx/     <-- es-mx locale
site.com/                 <-- uk locale

- [x] Tested

## Migration
None needed.

[PROJECT-5002]: https://bigcommercecloud.atlassian.net/browse/PROJECT-5002?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[TRAC-421]: https://bigcommercecloud.atlassian.net/browse/TRAC-421?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/i18n/locales.ts
+++ b/core/i18n/locales.ts
@@ -1,6 +1,33 @@
 import { buildConfig } from '~/build-config/reader';
 
 const localeNodes = buildConfig.get('locales');
+const defaultLocaleNode = localeNodes.find((locale) => locale.isDefault);
 
 export const locales = localeNodes.map((locale) => locale.code);
-export const defaultLocale = localeNodes.find((locale) => locale.isDefault)?.code ?? 'en';
+export const defaultLocale = defaultLocaleNode?.code ?? 'en';
+
+export const prefixes = localeNodes.reduce<Record<string, `/${string}`>>((acc, locale) => {
+  if (locale.path) acc[locale.code] = `/${locale.path}`;
+
+  return acc;
+}, {});
+
+// The locale that should occupy the bare root URL ("/").
+//
+// Rule:
+//   1. If the default locale has no CP path, default lives at "/".
+//   2. Else, if exactly one non-default locale has no CP path, that one lives at "/".
+//   3. Otherwise, no locale lives at "/" and every locale gets a prefix.
+function computeRootLocale(): string | null {
+  if (!defaultLocaleNode) return defaultLocale;
+  if (!defaultLocaleNode.path) return defaultLocaleNode.code;
+
+  const blankNonDefaults = localeNodes.filter((locale) => !locale.isDefault && !locale.path);
+  const [onlyBlank, ...rest] = blankNonDefaults;
+
+  if (onlyBlank && rest.length === 0) return onlyBlank.code;
+
+  return null;
+}
+
+export const rootLocale = computeRootLocale();
--- a/core/i18n/routing.ts
+++ b/core/i18n/routing.ts
@@ -1,21 +1,28 @@
 import { createNavigation } from 'next-intl/navigation';
 import { defineRouting } from 'next-intl/routing';
 
-import { defaultLocale, locales } from './locales';
+import { defaultLocale, locales, prefixes, rootLocale } from './locales';
 
 enum LocalePrefixes {
   ALWAYS = 'always',
-  // Don't use NEVER as there is a issue that causes cache problems and returns the wrong messages.
+  // Don't use NEVER as there is an issue that causes cache problems and returns the wrong messages.
   // More info: https://github.com/amannn/next-intl/issues/786
   // NEVER = 'never',
   ASNEEDED = 'as-needed', // removes prefix on default locale
 }
 
-const localePrefix = LocalePrefixes.ASNEEDED;
+// `rootLocale` is the locale (if any) that occupies the bare "/" URL. We hand it
+// to next-intl as the routing default so URLs like "/" resolve correctly. When no
+// locale lives at "/", every locale needs a prefix, so we use `always` mode and
+// fall back to the BC default for next-intl's defaultLocale.
+const localePrefix =
+  rootLocale === null
+    ? { mode: LocalePrefixes.ALWAYS, prefixes }
+    : { mode: LocalePrefixes.ASNEEDED, prefixes };
 
 export const routing = defineRouting({
   locales,
-  defaultLocale,
+  defaultLocale: rootLocale ?? defaultLocale,
   localePrefix,
 });
 
--- a/core/next.config.ts
+++ b/core/next.config.ts
@@ -25,6 +25,7 @@ const SettingsQuery = graphql(`
         locales {
           code
           isDefault
+          path
         }
       }
     }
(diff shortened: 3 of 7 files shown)


mfaris9:
Question: since `fullPath` is removed from the schema entirely when the `availableLocaleFullPathEnabled` flag is off, I went with a try/catch fallback to a legacy query. I am open to suggestions if there's a cleaner way to do this.

core/i18n/locales.ts
```diff
@@ -2,5 +2,24 @@ import { buildConfig } from '~/build-config/reader';
 
 const localeNodes = buildConfig.get('locales');
 
+function pathFromFullPath(fullPath: string | null | undefined): string | undefined {
```
funivan: We can output also `path` if needed, this is an easy task. 
```
store {	
  locales{
    code   : 'fr'
    isDefault : true
    path: 'france'
    fullPath : 'site.com/france'
 
  }
}
```

mfaris9: That would be great, it would be cleaner to have the path output as well

core/next.config.ts
```diff
@@ -25,14 +26,64 @@ const SettingsQuery = graphql(`
...
+
+function isFullPathUnavailableError(err: unknown): boolean {
+  const parsed = GraphqlErrorShape.safeParse(err);
+
+  if (!parsed.success) return false;
+
+  return parsed.data.graphqlErrors.some((e) => e.message.includes("Cannot query field 'fullPath'"));
+}
+
+async function fetchSettings() {
```
parthshahp: Is this the standard pattern for new GQL schemas in `storefront`? I would double check if we can just change the schema without the flag, it should just return null anyway. May not be possible.

core/i18n/locales.ts
```diff
@@ -2,5 +2,24 @@ import { buildConfig } from '~/build-config/reader';
...
+    return pathname || undefined;
+  } catch {
+    return undefined;
+  }
+}
+
 export const locales = localeNodes.map((locale) => locale.code);
 export const defaultLocale = localeNodes.find((locale) => locale.isDefault)?.code ?? 'en';
+
+export const prefixes: Record<string, `/${string}`> = Object.fromEntries(
```
parthshahp: Lots of casting here, is it all needed? Can we simplify this a bit?

core/i18n/routing.ts
```diff
@@ -1,22 +1,15 @@
 import { createNavigation } from 'next-intl/navigation';
 import { defineRouting } from 'next-intl/routing';
 
-import { defaultLocale, locales } from './locales';
-
-enum LocalePrefixes {
-  ALWAYS = 'always',
```
parthshahp, line 7: Just curious: why did we remove this?

mfaris9: I didn't think we needed the enum anymore as we are only using as-needed. I can add a comment like this if you prefer:

localePrefix: {
  // 'as-needed' removes the prefix for the default locale.
  // 'always' would prefix every locale including the default.
  // 'never' has cache-related issues — avoid:
  //   https://github.com/amannn/next-intl/issues/786
  mode: 'as-needed',
  prefixes,
}

core/i18n/locales.ts
mfaris9: Thanks, simplified

core/next.config.ts
mfaris9: I didn't love this change either because we'll have to go back and clean it up once the flag graduates... BC's storefront removes from the schema entirely when the flag is off. I agree it feels weird and I feel as though this issue probably comes up often i'll send a message in trac-team to see if anyone else handled this differently.

mfaris9: Hey Parth, looks like this is standard pattern on the storefront side (see SearchSchema.scala:28). Thanks for the review.

core/i18n/routing.ts
mfaris9: Enum change reverted to provide context for merchant configuration

parthshahp (approved):

PR #3018, opened 2026-05-18
TRAC-525: fix - client.ts not using `Header` object for default values

parthshahp:
## What/Why?

Spreading plain object literals into the `fetch` `headers` field left case-variant keys (e.g. `Authorization` and `authorization`) as distinct properties, which `fetch` then appended together into a single comma-joined value. Build the header set with a `Headers` object so overrides from `beforeRequest` and per-call `headers` replace the defaults instead of duplicating them.

## Testing

Tested manually. Previously, passing in a header with `Authorization` would append the key to the `authorization` header rather than replacing it.

## Migration

No migration needed.

### DIFF
--- a/.changeset/fresh-steaks-remain.md
+++ b/.changeset/fresh-steaks-remain.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-client": patch
+---
+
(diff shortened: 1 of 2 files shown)


jordanarldt (approved):

.changeset/fresh-steaks-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+"@bigcommerce/catalyst-client": minor
```
chanceaclark: 🍹 I'd say it's a patch, since we are technically fixing an intended functionality.

chanceaclark (approved):

.changeset/fresh-steaks-remain.md
parthshahp: You're right, I thought I changed that but I didn't. Updating.

PR #3021, opened 2026-05-20
fix(catalyst): TRAC-279 default brand PLP sort to featured

mfaris9:
## JIRA
[TRAC-279](https://bigcommercecloud.atlassian.net/browse/TRAC-279)

## What / Why
Brand pages weren't showing products in the order the merchant set in the
Control Panel.

Catalyst wasn't telling the backend which sort to use when a shopper landed
on a brand page without picking one. The backend then fell back to its own
default, which doesn't honor the merchant's chosen order. Asking for
"featured" explicitly fixes this and matches the sort dropdown's default
selection.

## Rollout / Rollback
No flag. Backwards compatible: the new default only applies when no `?sort=` param is set, so any existing URL with a sort param is unaffected. Rollback is reverting the single line.

## Testing
Tested in catalyst store. The first screenshot is featured items all with sort = 0. The second screenshot is featured items where Spray Bottle was assigned 1, Snake Plant was assigned -1, and ZZ Plant was left at 0.

Screenshot 2026-05-20 at 1 40 35 PM

Screenshot 2026-05-20 at 2 15 41 PM

:robot: Generated with [Claude Code](https://claude.com/claude-code)

[TRAC-279]: https://bigcommercecloud.atlassian.net/browse/TRAC-279?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx
+++ b/core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx
@@ -116,12 +116,14 @@ export default async function Brand(props: Props) {
 
     const loadSearchParams = await createBrandSearchParamsLoader(slug, customerAccessToken);
     const parsedSearchParams = loadSearchParams?.(searchParams) ?? {};
+    const sort = typeof searchParams.sort === 'string' ? searchParams.sort : 'featured';
 
     const search = await fetchFacetedSearch(
       {
         ...searchParams,
         ...parsedSearchParams,
         brand: [slug],
+        sort,
       },
       currencyCode,
       customerAccessToken,
--- a/.changeset/fix-brand-plp-sort.md
+++ b/.changeset/fix-brand-plp-sort.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Default brand PLP sort to `featured` to honor the merchant's product sort order.


core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx
```diff
@@ -122,6 +122,7 @@ export default async function Brand(props: Props) {
         ...searchParams,
         ...parsedSearchParams,
         brand: [slug],
+        sort: typeof searchParams.sort === 'string' ? searchParams.sort : 'featured',
```
Comment (parthshahp) [inline core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx]: 🍹 It would probably be better to put this into it's own variable.

Also, looks like searchParams type is `string | string[] | undefined`. Is this the correct behavior for the `string[]` case?

parthshahp:
Also, can we clean up the commits a bit here. Looks like most of them are just changing wording? You can either amend or squash as needed.

core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx
Comment (mfaris9) [reply core/app/[locale]/(default)/(faceted)/brand/[slug]/page.tsx]: Updated sort into its own variable.

From the UI only string is able to be created (from the dropdown).

parthshahp (approved):
Looks good, approved! Also just to double check: "featured" is meant to be the default, correct?

bc-yevhenii-buliuk (approved):

PR #3028, opened 2026-05-27
fix(cli): strip @vercel/otel hook in Commerce Hosting setup

jorgemoya:
Linear: [LTRAC-807](https://linear.app/commerce/issue/LTRAC-807) (parent: [LTRAC-615](https://linear.app/commerce/issue/LTRAC-615))

## What/Why?

`catalyst logs tail` against a Cloudflare-hosted Catalyst storefront prints `[ERROR] Failed to prepare server Error: An error occurred while loading the instrumentation hook` on every cold start. Root cause: `core/instrumentation.ts` registers `@vercel/otel`, and because OpenNext builds the worker as a Node-style runtime (`nodejs_compat`), the bundler resolves `@vercel/otel` through its `node` export condition — pulling `@opentelemetry/sdk-node` and its Node-only side-effects into the worker chunk. At cold start, `workerd` evaluates those side-effects, hits an API `nodejs_compat` doesn't cover, throws, and Next.js's instrumentation loader catches it as the generic "Failed to prepare server" line.

The fix lives in a new exported `cleanupCloudflareIncompatibilities(projectDir)` helper called from every Commerce Hosting entry point so both fresh setups and already-transformed projects pick it up:

- `setupCommerceHosting` — fresh users via the regular setup flow
- `offerCommerceHostingSetup` on `catalyst project link` — already-transformed users (the path that gates on `!isTransformed` and otherwise short-circuits)
- `catalyst deploy` — already-transformed users on every deploy, so the fix lands without a re-link

### Behavior — prompt-driven, not silent

A Vercel-deploying Catalyst user may have **customized** `core/instrumentation.ts` (Sentry, custom OTel providers, span enrichment, etc.). Silent deletion would wipe that work the first time they try Cloudflare. So the cleanup asks first:

| State | TTY | What happens |
|---|---|---|
| `core/instrumentation.ts` missing | any | No prompt, no action, no dep change. Silent. |
| File present | TTY | Prompt: "Catalyst found `core/instrumentation.ts` … Remove it and drop `@vercel/otel`?" Default Yes. On accept → delete file + drop dep. On decline → keep both, log a one-line note that the cold-start error will persist until resolved manually. |
| File present | non-TTY | `consola.warn` explaining the incompatibility; skip the cleanup. No file or dep change. Recommend re-running interactively or removing/gating manually. |

The dep removal is tied to the file decision — if the user kept their customized file, `@vercel/otel` stays too (they're probably importing it).

Self-hosted / Vercel deploys never hit any of these Commerce Hosting paths, so they keep the file regardless.

Not in scope: this isn't "OTel can't work on Cloudflare." A real Workers-compatible OTel setup (e.g. `@microlabs/otel-cf-workers`) is a separate workstream if Native Hosting wants storefront instrumentation beyond Cloudflare's built-in observability.

## Testing

- `pnpm vitest run src/cli/lib/commerce-hosting.spec.ts` — 52 passed. New cases cover TTY accept (file deleted + dep dropped), TTY decline (file kept + dep kept), non-TTY skip (no prompt, no change), and the "file missing" no-op.
- `pnpm test` (full `@bigcommerce/catalyst` package) — 279 passed across 23 files.
- `pnpm lint` + `pnpm tsc --noEmit` — clean.

End-to-end on Native Hosting:

1. **Default file path (happy)** — on an already-transformed project that still has the default `core/instrumentation.ts` + `@vercel/otel`: run `catalyst project link` (or `catalyst deploy`) interactively, accept the prompt, see the `[info]` confirmations. Re-deploy → `[ERROR] Failed to prepare server` is gone on cold start.
2. **Customized file path** — edit `core/instrumentation.ts` to add a comment or import, re-run `catalyst project link`, decline the prompt → confirm file and dep are untouched.
3. **CI / non-TTY path** — run `catalyst project link --project-uuid <uuid> < /dev/null` → see the `[warn]` line, confirm nothing was modified.

The `[WARN] env.IMAGES binding is not defined` is intentionally still present — tracked separately in [LTRAC-808](https://linear.app/commerce/issue/LTRAC-808).

```
~/dev/catalyst/core on  jorgemoya/ltrac-807-cli-strip-vercelotel-instrumentation-hook-during-commerce [?]
❯ node ../packages/catalyst/dist/cli.js project link
◢ @bigcommerce/catalyst v1.0.0-alpha.4                                                                                                                10:43:35 AM

◐ Fetching projects...                                                                                                                                10:43:35 AM
✔ Projects fetched.                                                                                                                                  10:43:35 AM

✔ Select a project or create a new project (Press <enter> to select).
catalyst-jm-test [linked]
◐ Writing project UUID to .bigcommerce/project.json...                                                                                                10:43:36 AM
✔ Project UUID written to .bigcommerce/project.json.                                                                                                 10:43:36 AM

✔ Your project has been linked, but is not fully set up for Commerce Hosting deployments yet. Would you like to run the setup now?
Yes

✔ Catalyst found core/instrumentation.ts, which is incompatible with the Cloudflare Workers bundle when it uses @vercel/otel (causes "Failed to prepare server" at
cold start). Remove it and drop @vercel/otel from core/package.json?
Yes
                                                                                                                                                      10:43:40 AM
ℹ Removed core/instrumentation.ts (incompatible with Cloudflare Workers).                                                                            10:43:40 AM
ℹ Dropped @vercel/otel from core/package.json.                                                                                                       10:43:40 AM
✔ Commerce Hosting setup complete.                                                                                                                   10:43:40 AM
✔ Dependencies installed successfully.
```

## Migration

None required. Cleanup is automatic on the next interactive `catalyst project link` or `catalyst deploy` invocation. Customized users see a prompt before any change is made. CI deploys never auto-remove.

Fixes LTRAC-807

### DIFF
--- a/.changeset/cli-strip-instrumentation-hook.md
+++ b/.changeset/cli-strip-instrumentation-hook.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst": patch
+---
+
+Remove `core/instrumentation.ts` and the `@vercel/otel` dependency during Commerce Hosting setup. The hook isn't compatible with the OpenNext + Cloudflare Workers bundling path and caused a "Failed to prepare server" error on every cold start in `catalyst logs tail`. Self-hosted (non-Commerce Hosting) deployments are unaffected.
--- a/packages/catalyst/src/cli/commands/create.ts
+++ b/packages/catalyst/src/cli/commands/create.ts
@@ -478,7 +478,7 @@ Examples:
       setupCoreProject(projectDir);
 
       if (useCommerceHosting && commerceHostingProject && storeHash && accessToken) {
-        setupCommerceHosting({
+        await setupCommerceHosting({
           projectDir,
           projectUuid: commerceHostingProject.uuid,
           storeHash,
--- a/packages/catalyst/src/cli/commands/deploy.ts
+++ b/packages/catalyst/src/cli/commands/deploy.ts
@@ -8,6 +8,7 @@ import { z } from 'zod';
 
 import { runChannelSiteUrlFlow } from '../lib/channel-site-flow';
 import {
+  cleanupCloudflareIncompatibilities,
   NoLinkedProjectError,
   selectOrCreateInfrastructureProject,
   setupCommerceHosting,
@@ -467,6 +468,12 @@ Example:
       consola.success('Commerce Hosting setup complete.');
 
       await installDependencies(projectDir);
+    } else {
+      // Existing Commerce Hosting users may carry artifacts incompatible with
+      // the Cloudflare worker bundle from earlier Catalyst versions
+      // (`core/instrumentation.ts`, `@vercel/otel`). Sweep them on every deploy
+      // so the fix lands without forcing a re-link.
+      cleanupCloudflareIncompatibilities(dirname(process.cwd()));
     }
 
     if (options.prebuilt) {
--- a/packages/catalyst/src/cli/commands/deploy.ts
+++ b/packages/catalyst/src/cli/commands/deploy.ts
@@ -464,7 +464,7 @@ Example:
 
       const projectDir = dirname(process.cwd());
 
-      setupCommerceHosting({ projectDir, projectUuid, storeHash, accessToken });
+      await setupCommerceHosting({ projectDir, projectUuid, storeHash, accessToken });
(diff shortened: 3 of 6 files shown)


jorgemoya:
Tested manually and I no longer see the error.

parthshahp (approved):

PR #3040, opened 2026-06-03
LTRAC-441: fix(cli) - Surface clear re-auth error on expired token

jorgemoya:
Jira: [LTRAC-441](https://linear.app/commerce/issue/LTRAC-441) · [TRAC-431](https://bigcommercecloud.atlassian.net/browse/TRAC-431)

## What/Why?

Auth tokens stored in `.bigcommerce/project.json` stop working between sessions. The CLI still believes it's authorized (a token is present), but API calls return **401 Unauthorized** on `project list`, `deploy`, etc. Because every authenticated `fetch` only checked `response.ok` and threw a generic `Failed to ...: Unauthorized`, users had no idea they needed to re-authenticate.

Root cause: credentials come from the device-code OAuth flow (short-lived, **expiring** token) or a manual store-level API account (non-expiring). The device-code success response is parsed for only `access_token`/`store_hash` — any expiry/refresh metadata is discarded — so **silent refresh isn't possible** today. Auto-refresh (refresh token) and longer token lifetimes both require login-service/backend changes outside this repo; this PR implements the self-contained fix: detect 401 consistently and tell the user to re-authenticate.

Key decisions:
- Only **401** is treated as an auth failure. **403** is intentionally left untouched because it's overloaded in this codebase to mean "Infrastructure Projects API not enabled" (scope/feature flag).
- New shared `UnauthorizedError` + `assertAuthorized(response)` (`lib/auth-errors.ts`), called at every authenticated fetch site (`project.ts`, `channels.ts`, `localization.ts`, `deploy.ts`).
- Messaging covers **any rejected token** — expired/revoked *or* never-valid (e.g. a typo'd `--access-token` flag or stale `CATALYST_ACCESS_TOKEN`): _"Your access token is invalid or has expired. Run `catalyst auth login` to re-authenticate."_ The "no credentials at all" case is **not** a 401 — it's caught earlier by `resolveCredentials` (and `auth whoami`'s own guard) with a "Missing credentials" message before any request is made.
- `logs tail` raises a **fatal** stream error on 401 so the reconnect loop stops with the re-auth message instead of burning retries.
- The top-level handler in `index.ts` prints the re-auth guidance without the "share your Correlation ID with support" bug-report framing (telemetry still tracked).
- `lib/login.ts` `manualLogin` is deliberately unchanged — a 401 during interactive login already shows "Could not validate credentials".

Follow-ups (out of scope, backend-dependent): auto-refresh via a persisted `refresh_token`, and proactive expiry warnings via persisted `expires_in`.

## Testing

- `pnpm typecheck`, `pnpm lint` (max-warnings 0), and `pnpm test` (**314 passing**) all green.
- Added/updated specs: `auth-errors.spec.ts` (401 throws, 200/403/500 no-op), a `project list` 401 regression test, and updated `auth.spec.ts` / `logs.spec.ts` to assert the new re-auth messaging.
- Manual smoke: set a bogus `accessToken` in `.bigcommerce/project.json`, then run `catalyst project list`, `catalyst deploy`, `catalyst auth whoami`, `catalyst logs tail` — each prints "Your access token is invalid or has expired ... Run \`catalyst auth login\`" and exits non-zero (logs does not retry-loop). A valid token still works end-to-end.

## Migration

None. No breaking changes or moved files.

[TRAC-431]: https://bigcommercecloud.atlassian.net/browse/TRAC-431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/catalyst/src/cli/commands/auth.ts
+++ b/packages/catalyst/src/cli/commands/auth.ts
@@ -1,6 +1,7 @@
 import { Command, Option } from 'commander';
 import { z } from 'zod';
 
+import { assertAuthorized, UnauthorizedError } from '../lib/auth-errors';
 import { consola } from '../lib/logger';
 import { LoginAbortedError, login as runInteractiveLogin } from '../lib/login';
 import { fetchProjects } from '../lib/project';
@@ -21,6 +22,8 @@ async function fetchStoreProfile(storeHash: string, accessToken: string, apiHost
       Accept: 'application/json',
     },
   });
+
+  assertAuthorized(response);
 
   if (!response.ok) {
     throw new Error(`${response.status} ${response.statusText}`);
@@ -107,7 +110,11 @@ Example:
     } catch (error) {
       const message = error instanceof Error ? error.message : String(error);
 
-      if (message.includes('401') || message.includes('403')) {
+      if (error instanceof UnauthorizedError) {
+        consola.error(
+          'Not logged in: your access token is invalid or has expired. Run `catalyst auth login`.',
+        );
+      } else if (message.includes('401') || message.includes('403')) {
         consola.error(`Not logged in: invalid credentials (${message})`);
       } else {
         consola.error(`Failed to verify credentials: ${message}`);
--- a/packages/catalyst/src/cli/index.ts
+++ b/packages/catalyst/src/cli/index.ts
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+import { UnauthorizedError } from './lib/auth-errors';
 import { consola } from './lib/logger';
 import { getTelemetry } from './lib/telemetry';
 import { program } from './program';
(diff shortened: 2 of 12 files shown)


packages/catalyst/src/cli/commands/auth.ts
```diff
@@ -107,7 +110,11 @@ Example:
     } catch (error) {
       const message = error instanceof Error ? error.message : String(error);
 
-      if (message.includes('401') || message.includes('403')) {
+      if (error instanceof UnauthorizedError) {
+        consola.error(
```
bc-yevhenii-buliuk, line 114: should it be `console.error()`?

packages/catalyst/src/cli/index.ts
```diff
@@ -21,6 +22,13 @@ const handleFatalError = async (error: unknown) => {
     // Don't mask the original error
   }
 
+  // An invalid/expired token is user-actionable, not a bug to report — print the
+  // re-auth guidance without the "share your Correlation ID with support" noise.
+  if (error instanceof UnauthorizedError) {
+    consola.error(errorMessage);
```
bc-yevhenii-buliuk, line 28: `console`?

bc-yevhenii-buliuk (approved):
looks good, just small question about console)

packages/catalyst/src/cli/commands/auth.ts
jorgemoya: We use [consola](https://github.com/unjs/consola) for the CLI.

PR #3046, opened 2026-06-11
feat(core): TRAC-813 Gate analytics cookies behind shopper consent

chanceaclark:
Jira: [TRAC-813](https://bigcommercecloud.atlassian.net/browse/TRAC-813)

## What/Why?

`catalyst.visitorId`, `catalyst.visitId`, and `currencyCode` were set unconditionally, ignoring the shopper's consent decision. This gates them behind consent.

**Consent decision logic** (new `hasConsentFor(category)` in `lib/consent-manager`): the `c15t-consent` cookie is the source of truth. When no consent cookie exists yet, consent is treated as not given — the client-side consent manager writes the cookie once the shopper decides (or c15t auto-grants when consent is disabled), and `startVisit` handles recording the visit at that point.

**Category mapping** mirrors `scripts-transformer`'s BC-to-c15t map: the visit cookies are ANALYTICS → `measurement`; `currencyCode` is a preference → `functionality`. Without functionality consent, switching currency still updates the cart but the preference isn't persisted.

**Withdrawal**: the proxy deletes both analytics cookies when a consent cookie exists without measurement granted, so identifiers don't linger for the cookie's remaining 400-day lifetime after the shopper opts out.

**Firing the visit event after mid-session consent** (the open question in the ticket): the proxy only starts visits on full-page navigations (prefetch/RSC requests are skipped), so accepting the banner wouldn't record a visit until the next hard reload. A new `startVisit` server action sets the cookies and fires `visitStartedEvent`, triggered by a small client listener when measurement consent flips to granted. Two non-obvious details:

- The listener polls `document.cookie` directly before calling the action — c15t updates its React state before persisting the cookie, so the action (which validates consent from the cookie server-side) would no-op if called immediately after the state change.
- The proxy skips visit creation for server-action POSTs (`Next-Action` header). Cookies set in the proxy aren't visible to the action handler's cookie store; without this, accepting the banner fired two `visitStartedEvent`s with different visit IDs.

Also fixed a latent bug in two e2e expiry tests: they looked up `visitId`/`visitorId` instead of `catalyst.visitId`/`catalyst.visitorId`, so their assertions never executed.

Refs TRAC-813

## Testing

`pnpm lint`, `pnpm typecheck`, and all 6 `analytics-session.spec.ts` e2e tests pass locally (tests now seed accept/decline consent cookies, so they're deterministic regardless of the store's consent setting; new tests cover the declined and withdrawal paths).

Verified against a local dev server connected to a store with cookie consent enabled:

| Request state | Result |
|---|---|
| Consent cookie with `c.measurement:1` | visitorId + visitId set, one `VisitStarted` mutation |
| Consent cookie without measurement | no analytics cookies set |
| Consent withdrawn, stale analytics cookies sent | both cookies expired via Set-Cookie |
| No consent cookie | no analytics cookies set |

Browser flow via Playwright: loading the home page sets no analytics cookies, clicking "Accept All" sets both cookies without a reload and fires exactly one `VisitStarted` mutation, and a subsequent reload keeps the same visitId (sliding-window refresh, no duplicate event).

## Migration

None. No files moved; `withAnalyticsCookies` is back in its original position in the proxy chain.

[TRAC-813]: https://bigcommercecloud.atlassian.net/browse/TRAC-813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/ltrac-846-gate-analytics-cookies-behind-consent.md
+++ b/.changeset/ltrac-846-gate-analytics-cookies-behind-consent.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Gate `catalyst.visitorId`, `catalyst.visitId`, and `currencyCode` cookies behind shopper consent. The visitor and visit cookies now require measurement consent and the currency preference cookie requires functionality consent. When consent is absent, existing analytics cookies are deleted on the next request. When measurement consent is granted mid-session, a new `startVisit` server action sets the cookies and fires the server-side `visitStartedEvent` immediately rather than waiting for the next full-page navigation.
--- a/core/components/consent-manager/_actions/start-visit.ts
+++ b/core/components/consent-manager/_actions/start-visit.ts
@@ -1,0 +1,52 @@
+'use server';
+
+import { headers } from 'next/headers';
+import { validate as isUuid, v4 as uuidv4 } from 'uuid';
+
+import {
+  getVisitIdCookie,
+  getVisitorIdCookie,
+  setVisitIdCookie,
+  setVisitorIdCookie,
+} from '~/lib/analytics/bigcommerce';
+import { sendVisitStartedEvent } from '~/lib/analytics/bigcommerce/data-events';
+import { getConsentCookie } from '~/lib/consent-manager/cookies/server';
+
+// Starts an analytics visit after the shopper grants measurement consent.
+// The proxy only starts visits on full-page navigations, so without this a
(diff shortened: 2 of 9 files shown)


chanceaclark:
https://github.com/user-attachments/assets/c0b8cfda-46bc-4813-90f2-e30fed426256

jorgemoya (approved):

parthshahp (approved):

PR #3047, opened 2026-06-12
fix: TRAC-814 Make session token cookies expire with the session

chanceaclark:
Jira: [TRAC-814](https://bigcommercecloud.atlassian.net/browse/TRAC-814)

## What/Why?

`authjs.session-token` and `authjs.anonymous-session-token` must be session cookies (no `Expires` attribute) to comply with Essential cookie classification requirements.

**Anonymous session token:** removed `maxAge` from `anonymousSignIn` — without it, Next.js omits the `Max-Age`/`Expires` attribute and the cookie becomes a session cookie.

**Auth session token:** Auth.js v5 unconditionally sets `Expires` on the session cookie based on `session.maxAge` — there is no config option to disable this. The fix post-processes `Set-Cookie` headers to strip the `Expires` attribute from matching cookies. This is applied in two places:

- `proxies/with-auth.ts` — covers session refreshes on every page request (the middleware matcher excludes `/api/*`)
- `app/api/auth/[...nextauth]/route.ts` — covers sign-in, where Auth.js sets the initial session cookie (not reachable by middleware)

The regex matches `authjs.session-token`, `__Secure-authjs.session-token` (HTTPS prefix), and chunked variants (`authjs.session-token.0`, etc.). Only `Expires` is stripped — `Max-Age=0` (used by Auth.js for cookie deletion) is left intact.

This workaround is tracked upstream in [nextauthjs/next-auth#13457](https://github.com/nextauthjs/next-auth/discussions/13457), where we've proposed a first-class config option so this kind of post-processing isn't necessary.

## Testing

1. Start the dev server and open DevTools → Application → Cookies
2. Navigate to any page — verify `authjs.anonymous-session-token` has no `Expires` column value (session cookie)
3. Sign in — verify `authjs.session-token` has no `Expires` column value
4. Sign out and back in — verify the cookie is still a session cookie after the refresh cycle
5. Close the browser and reopen — both cookies should be gone

## Migration

No migrations required.

[TRAC-814]: https://bigcommercecloud.atlassian.net/browse/TRAC-814?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/proxies/with-auth.ts
+++ b/core/proxies/with-auth.ts
@@ -6,14 +6,36 @@ import { type ProxyFactory } from './compose-proxies';
 
 // Path matcher for any routes that require authentication
 const protectedPathPattern = new URLPattern({ pathname: `{/:locale}?/(account)/*` });
+const SESSION_TOKEN_COOKIE_RE = /^(__Secure-)?authjs\.session-token(\.\d+)?=/;
 
 function redirectToLogin(url: string) {
   return NextResponse.redirect(new URL('/login', url), { status: 302 });
 }
 
+// Auth.js always sets Expires on session cookies based on session.maxAge. We strip it here
+// so the cookie becomes a session cookie (expires when the browser closes), which is required
+// for Essential cookie classification compliance.
+function stripSessionCookieExpiry(response: Response): Response {
+  const setCookies = response.headers.getSetCookie();
+  const stripped = setCookies.map((cookie) =>
+    SESSION_TOKEN_COOKIE_RE.test(cookie)
+      ? cookie.replace(/;\s*(?:expires|max-age)=[^;]+/gi, '')
+      : cookie,
+  );
+
+  if (stripped.every((c, i) => c === setCookies[i])) return response;
+
+  const modified = new Response(response.body, response);
+
+  modified.headers.delete('set-cookie');
+  stripped.forEach((cookie) => modified.headers.append('set-cookie', cookie));
+
+  return modified;
+}
+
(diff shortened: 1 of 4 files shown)


core/proxies/with-auth.ts
```diff
@@ -6,14 +6,36 @@ import { type ProxyFactory } from './compose-proxies';
...
 const protectedPathPattern = new URLPattern({ pathname: `{/:locale}?/(account)/*` });
+const SESSION_TOKEN_COOKIE_RE = /^(__Secure-)?authjs\.session-token(\.\d+)?=/;
 
 function redirectToLogin(url: string) {
   return NextResponse.redirect(new URL('/login', url), { status: 302 });
 }
 
+// Auth.js always sets Expires on session cookies based on session.maxAge. We strip it here
+// so the cookie becomes a session cookie (expires when the browser closes), which is required
+// for Essential cookie classification compliance.
```
jorgemoya, line 17: There's no scenario for a consent selection that won't expire the cookie on browser close?

chanceaclark: Yep, consent selection cookies are handled separately and aren't affected here. The `stripSessionCookieExpiry` function only modifies cookies that match `SESSION_TOKEN_COOKIE_RE`, which targets `authjs.session-token` specifically. Consent cookies have their own names and expiry, so they'll persist across browser sessions as expected.

jorgemoya (approved):

PR #3051, opened 2026-06-16
LTRAC-908: feat(cli) - Add `catalyst channel link` command

jorgemoya:
Linear: [LTRAC-908](https://linear.app/commerce/issue/LTRAC-908)
Part of [LTRAC-138](https://linear.app/commerce/issue/LTRAC-138) — consolidating `create-catalyst` into the `catalyst` CLI.

## What/Why?
We decided **not** to port `create-catalyst`'s standalone `init` command. It was conceived for "download the Catalyst core repo → connect → run on Vercel", which doesn't apply to a native-hosting-only CLI, and `catalyst create` already connects a channel + writes `.env.local` for new projects.

The one residual need is real: **regenerating `.env.local` from a channel for an existing checkout** — e.g. a teammate who `git clone`d a Catalyst repo where `.env.local` (and `.bigcommerce/project.json`) are gitignored. That's a *channel* concern, distinct from `project link` (the native-hosting deploy target), so it lands as **`catalyst channel link`** — a sibling to `channel update`, and named to parallel the existing `project link` (the CLI's verb for linking a local checkout to a remote BigCommerce resource).

The command:
- Resolves credentials from flags/env → persisted `.bigcommerce/project.json` → interactive device-code `login()` (persisting on success), since a fresh clone has no stored creds — same pattern as `catalyst project create`.
- Selects a storefront channel (Catalyst-first ordering) or takes `--channel-id` to skip the picker.
- Fetches the channel's init data (`getChannelInit`) and writes `.env.local` to the **cwd** (where `dev`/`build`/`deploy` run), with repeatable `--env KEY=VALUE` overrides.

Reuses existing libs (`channels`, `login`, `project-config`, `env-config`, `telemetry`); no new dependencies. `program.ts` is untouched — `channel` is already registered.

## Testing
`pnpm test`, `pnpm typecheck`, and `pnpm lint` all pass in `packages/catalyst`. New `channel link` tests (in `channel.spec.ts`) cover: link by `--channel-id`, the interactive picker, `--env` overrides, the empty-channels exit, and the no-credentials login-and-persist path.

## Migration
None. Purely additive — adds a `channel link` subcommand. `create-catalyst init` is unaffected (the create-catalyst shim removal is tracked in LTRAC-910).

PR #3054, opened 2026-06-17
LTRAC-911: ref(cli) - Standardize interactive prompts on `@inquirer/prompts`

jorgemoya:
Linear: [LTRAC-911](https://linear.app/commerce/issue/LTRAC-911)
Part of [LTRAC-138](https://linear.app/commerce/issue/LTRAC-138) — consolidating `create-catalyst` into the `catalyst` CLI (Phase 3.1).

> **Stacked on #3051** (`catalyst channel link`). Base is that branch so the diff shows only the prompt-standardization changes; re-target to `alpha` once #3051 merges.

## What/Why?
The consolidated CLI mixed two prompt libraries — often in the same file (`login.ts`, `create.ts`, `commerce-hosting.ts` each used both). This standardizes **all interactive prompts on `@inquirer/prompts`**; `consola` is kept for logging only.

Why inquirer over consola: `consola.prompt` only supports `text`/`confirm`/`select`/`multiselect` — **no masked input**, but `login` collects the access token via `password({ mask: true })`. Standardizing on consola would have leaked the token in plaintext. inquirer also gives inline `validate` (now used for the ≤4 additional locales and project-name validation) and theming (create's locale picker).

## Changes
- Converted every `consola.prompt(...)` → inquirer: `confirm` / `input` / `select` / `checkbox`.
  - `confirm({ message, default })`, `input({ message })`, `select({ message, choices: [{ name, value, description }] })`, `checkbox(...)`.
  - Source: `login.ts`, `channel-site-flow.ts`, `commerce-hosting.ts`, `commands/{create,channel,project,deploy}.ts`.
- `create`'s additional-locales picker is now an inquirer `checkbox` with a `validate` (≤4) — removes the old `as unknown as string[]` cast + recursion-on-overflow hack. **UX improvement** (inline re-prompt instead of a warning + restart).
- Channel pickers now use the channel id directly as the choice `value` (number), dropping `String()`/`Number()` round-trips.
- Specs updated to mock `@inquirer/prompts` instead of spying on `consola.prompt`; assertions that inspected prompt args were retargeted to the inquirer calls (behavior assertions unchanged).

After: `grep -rn "consola.prompt" src/cli` → empty.

## Testing
`pnpm typecheck`, `pnpm lint`, and the full `pnpm test` (320 tests) all pass in `packages/catalyst`. Verified masked token entry in `login` is preserved.

## Notes
This standardizes the prompt **library** (the ticket's scope). Broader *visual* cohesion across prompts (inquirer) + logging (consola, ~340 calls) + spinners (yocto) is a separate, larger concern — `@clack/prompts` was evaluated and deemed not worth a CLI-wide migration (it can't replace consola's logging, so it wouldn't remove the mix). A scoped follow-up could revisit clack for the interactive create/onboarding flows only.

## Migration
None — purely an internal prompt-library swap; user-facing prompt behavior is unchanged (the locale picker is slightly improved).

### DIFF
--- a/packages/catalyst/src/cli/commands/auth.spec.ts
+++ b/packages/catalyst/src/cli/commands/auth.spec.ts
@@ -1,4 +1,4 @@
-import { password } from '@inquirer/prompts';
+import { confirm, input, password } from '@inquirer/prompts';
 import { Command } from 'commander';
 import { http, HttpResponse } from 'msw';
 import { realpath } from 'node:fs/promises';
@@ -27,9 +27,13 @@ import { auth } from './auth';
 vi.mock('yocto-spinner', () => import('../../../tests/mocks/spinner'));
 vi.mock('open', () => ({ default: vi.fn().mockResolvedValue(undefined) }));
 vi.mock('@inquirer/prompts', () => ({
+  confirm: vi.fn(),
+  input: vi.fn(),
   password: vi.fn(),
 }));
 
+const confirmMock = vi.mocked(confirm);
+const inputMock = vi.mocked(input);
 const passwordMock = vi.mocked(password);
 
 let exitMock: MockInstance;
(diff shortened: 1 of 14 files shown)


chanceaclark (approved):

chanceaclark:
🍹 Can you build the CLI locally and provide a proof of life here for all the CLI commands that changed?

jorgemoya:
Proof of life:

```
~/dev/catalyst/core on  jorgemoya/ltrac-911-standardize-prompt-library-across-the-consolidated-catalyst [?] took 4s
❯ node ../packages/catalyst/dist/cli.js channel link --api-host api.staging.zone --cli-api-origin https://cxm-stg.bigcommerceapp.com
◢ @bigcommerce/catalyst v1.0.0-alpha.5                                                                                                                 1:49:39 PM

? Which channel would you like to link?
❯ My Catalyst Storefront
  Test Store
Catalyst
```

PR #3062, opened 2026-06-24
feat(payment): PAYPAL-6499 add wallet buttons

bc-yaroslav-zhmutskyi:
## What/Why?

Add wallet button integration for cart page

## Testing

Manual

https://github.com/user-attachments/assets/646d1472-f31c-4600-aa9a-05b3dc5153b7

## Migration

No need

### DIFF
--- a/core/.env.example
+++ b/core/.env.example
@@ -44,3 +44,7 @@ DEFAULT_REVALIDATE_TARGET=3600
 
 # Disable automatic fetch instrumentation (not recommended unless using custom instrumentation)
 # NEXT_OTEL_FETCH_DISABLED=1
+
+# Checkout SDK loader URL used in local development for wallet buttons.
+# In non-development environments the production CDN loader is always used.
+NEXT_PUBLIC_CHECKOUT_SDK_DEV_URL=
--- a/core/app/[locale]/(default)/cart/page-data.ts
+++ b/core/app/[locale]/(default)/cart/page-data.ts
@@ -1,3 +1,4 @@
+import { removeEdgesAndNodes } from '@bigcommerce/catalyst-client';
 import { cache } from 'react';
 
 import { getSessionCustomerAccessToken } from '~/auth';
@@ -6,6 +7,8 @@ import { PricingFragment } from '~/client/fragments/pricing';
 import { graphql, VariablesOf } from '~/client/graphql';
 import { revalidate } from '~/client/revalidate-target';
 import { TAGS } from '~/client/tags';
+import { CurrencyCodeSchema } from '~/components/header/schema';
+import { getPreferredCurrencyCode } from '~/lib/currency';
 
 export const PhysicalItemFragment = graphql(
   `
@@ -252,6 +255,9 @@ const CartPageQuery = graphql(
           version
           currencyCode
           isTaxIncluded
+          amount {
+            value
+          }
           discountedAmount {
             ...MoneyFieldsFragment
           }
@@ -358,11 +364,109 @@ const SupportedShippingDestinationsQuery = graphql(`
   }
 `);
 
-export const getShippingCountries = cache(async () => {
+export const getShippingCountries = async () => {
   const { data } = await client.fetch({
     document: SupportedShippingDestinationsQuery,
     fetchOptions: { next: { revalidate } },
   });
 
   return data.site.settings?.shipping?.supportedShippingDestinations.countries ?? [];
+};
+
+const PaymentWalletsQuery = graphql(`
+  query PaymentWalletsQuery($filters: PaymentWalletsFilterInput) {
+    site {
+      paymentWallets(filter: $filters) {
+        edges {
+          node {
+            entityId
+            methodName
+            id
+          }
+        }
+      }
+    }
+  }
+`);
+
+type PaymentWalletsVariables = VariablesOf<typeof PaymentWalletsQuery>;
+
+export const getPaymentWallets = cache(async (variables: PaymentWalletsVariables) => {
+  const customerAccessToken = await getSessionCustomerAccessToken();
+
+  const { data } = await client.fetch({
+    document: PaymentWalletsQuery,
+    customerAccessToken,
+    fetchOptions: { cache: 'no-store' },
+    variables,
+  });
+
+  return removeEdgesAndNodes(data.site.paymentWallets).map(({ entityId }) => entityId);
+});
+
+const PaymentWalletWithInitializationDataQuery = graphql(`
+  query PaymentWalletWithInitializationDataQuery($entityId: String!, $cartId: String!) {
+    site {
+      paymentWalletWithInitializationData(
+        filter: { paymentWalletEntityId: $entityId, cartEntityId: $cartId }
+      ) {
+        clientToken
+        initializationData
+      }
+    }
+  }
+`);
+
+export const getPaymentWalletWithInitializationData = cache(
+  async (entityId: string, cartId: string) => {
+    const { data } = await client.fetch({
+      document: PaymentWalletWithInitializationDataQuery,
+      variables: {
+        entityId,
+        cartId,
+      },
+      customerAccessToken: await getSessionCustomerAccessToken(),
+      fetchOptions: { cache: 'no-store' },
+    });
+
+    return data.site.paymentWalletWithInitializationData;
+  },
+);
+
+const CurrencyQuery = graphql(`
+  query Currency($currencyCode: currencyCode!) {
+    site {
+      currency(currencyCode: $currencyCode) {
+        display {
+          decimalPlaces
+          symbol
+        }
+        name
+        code
+      }
+    }
+  }
+`);
+
+export const getCurrencyData = cache(async (currencyCode?: string) => {
+  const code = currencyCode ?? (await getPreferredCurrencyCode());
+
+  if (!code) {
+    throw new Error('Could not get currency code');
+  }
+
+  const parsedCode = CurrencyCodeSchema.parse(code);
+
+  const customerAccessToken = await getSessionCustomerAccessToken();
+
+  const { data } = await client.fetch({
+    document: CurrencyQuery,
+    fetchOptions: { cache: 'no-store' },
+    variables: {
+      currencyCode: parsedCode,
+    },
+    customerAccessToken,
+  });
+
+  return data.site.currency;
 });
--- a/core/app/[locale]/(default)/cart/page.tsx
+++ b/core/app/[locale]/(default)/cart/page.tsx
@@ -5,6 +5,7 @@ import { Streamable } from '@/vibes/soul/lib/streamable';
 import { Price } from '@/vibes/soul/primitives/price-label';
 import { Cart as CartComponent, CartEmptyState } from '@/vibes/soul/sections/cart';
 import { CartAnalyticsProvider } from '~/app/[locale]/(default)/cart/_components/cart-analytics-provider';
+import { ClientWalletButtons } from '~/components/wallet-buttons';
 import { pricesTransformer } from '~/data-transformers/prices-transformer';
 import { getCartId } from '~/lib/cart';
 import { getPreferredCurrencyCode } from '~/lib/currency';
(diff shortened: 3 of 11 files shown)


core/.env.example
```diff
@@ -44,3 +44,7 @@ DEFAULT_REVALIDATE_TARGET=3600
 
 # Disable automatic fetch instrumentation (not recommended unless using custom instrumentation)
 # NEXT_OTEL_FETCH_DISABLED=1
+
+# Checkout SDK loader URL used in local development for wallet buttons.
```
bc-nick, line 48: Mention that the related files should be places in the `core/public/v1` folder to be able use them via next server in dev mode

bc-nick:
Also can you please provide some information regarding how the release environment works here? How the release flow should be managed in our case? Probably it should be covered by exp or something before merging.

bc-yaroslav-zhmutskyi:
> Also can you please provide some information regarding how the release environment works here? How the release flow should be managed in our case? Probably it should be covered by exp or something before merging.

Our changes will be covered by experiment on backend side storefront service and will be disabled in prod for now. 
Canary branch containing the latest code in development and the most recent commit may not be stable. After merge of this PR we will need to create release

core/.env.example
bc-yaroslav-zhmutskyi: Added info

bc-nick (approved):
LGMT

core/lib/currency.ts
```diff
@@ -6,9 +6,9 @@ import type { CurrencyCode } from '~/components/header/fragment';
 import { CurrencyCodeSchema } from '~/components/header/schema';
 import { hasConsentFor } from '~/lib/consent-manager/has-consent-for';
 
-export async function getPreferredCurrencyCode(): Promise<CurrencyCode | undefined> {
+export async function getPreferredCurrencyCode(code?: string): Promise<CurrencyCode | undefined> {
   const cookieStore = await cookies();
-  const currencyCode = cookieStore.get('currencyCode')?.value;
+  const currencyCode = cookieStore.get('currencyCode')?.value || code;
```
chanceaclark: `cookie || code` means the cookie always overrides the `code` arg, so `getCurrencyData(cart.currencyCode)` can get silently overridden by a stale currency cookie. Cart currency's fixed at creation, so this can mismatch amount vs currency on the wallet button. I would say leave this helper's signature alone (~20 other call sites) and just do `(await getPreferredCurrencyCode()) ?? currencyCode` inline in `getCurrencyData`.

core/components/wallet-buttons/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+    if (!isMountedRef.current && initButtonProps.length) {
+      isMountedRef.current = true;
+
+      const initWalletButtons = async () => {
+        await new WalletButtonsInitializer().initialize(initButtonProps);
+      };
+
+      void initWalletButtons();
+    }
+  }, [cartId, initButtonProps]);
```
chanceaclark: `isMountedRef` makes this a one-shot init forever. Deps `[cartId, initButtonProps]` say "re-run on change," but the ref blocks that after the first run, so a quantity change won't update the PayPal amount. I would fix this before merge.

core/components/wallet-buttons/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  useEffect(() => {
+    if (!isMountedRef.current && initButtonProps.length) {
+      isMountedRef.current = true;
+
+      const initWalletButtons = async () => {
+        await new WalletButtonsInitializer().initialize(initButtonProps);
+      };
+
+      void initWalletButtons();
```
chanceaclark: `isMountedRef` gets set before `initialize()` resolves, and `void initWalletButtons()` has no `.catch`. If the SDK script fails to load, that rejection's swallowed and it never retries, wallet buttons just silently never show up. Worth a `.catch` that resets the ref.

core/components/wallet-buttons/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export const ClientWalletButtons = ({
+  walletButtonsInitOptions,
+  cartId,
+}: {
+  walletButtonsInitOptions: Streamable<InitializeButtonProps[]>;
+  cartId: string;
+}) => {
+  const isMountedRef = useRef(false);
+  const initButtonProps = useStreamable(walletButtonsInitOptions);
```
chanceaclark: This runs outside the `<Stream fallback={null}>` boundary below, so a rejected promise throws into render instead of degrading gracefully like the fallback implies. Might be worth moving the unwrap inside `Stream`'s render prop.

core/components/wallet-buttons/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+      void initWalletButtons();
+    }
+  }, [cartId, initButtonProps]);
+
+  return (
+    <Stream fallback={null} value={walletButtonsInitOptions}>
+      {(buttonOptions) => (
+        <div style={{ display: 'flex', alignItems: 'end', flexDirection: 'column' }}>
+          {buttonOptions.map((button) =>
+            button.containerId ? <div id={button.containerId} key={button.containerId} /> : null,
```
chanceaclark: Nit: `containerId` is required and always set, so `: null` here is dead. Could just be `buttonOptions.map((button) => <div id={button.containerId} key={button.containerId} />)`.

core/app/[locale]/(default)/cart/page.tsx
```diff
@@ -101,6 +141,16 @@ export default async function Cart({ params }: Props) {
 
   const taxIncluded = cart.isTaxIncluded;
 
+  const walletButtons = await getPaymentWallets({
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/cart/page.tsx]: This blocks every cart render (even zero-wallet carts) since it's a top-level await with `no-store`. Everything else wallet-related is streamed, would wrap this in `Streamable.from` too.

core/app/[locale]/(default)/cart/page-data.ts
```diff
@@ -366,3 +371,97 @@ export const getShippingCountries = cache(async () => {
...
+        }
+      }
+    }
+  }
+`);
+
+type PaymentWalletsVariables = VariablesOf<typeof PaymentWalletsQuery>;
+
+export const getPaymentWallets = async (variables: PaymentWalletsVariables) => {
+  const customerAccessToken = await getSessionCustomerAccessToken();
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/cart/page-data.ts:395]: `getSessionCustomerAccessToken` isn't `cache()`'d, so with N wallets it decodes the session N+2 times per render. Might be worth wrapping it in `cache()`.

core/proxy.ts
```diff
@@ -22,13 +22,14 @@ export const config = {
...
+     * - v1 (checkout SDK loader script)
      * - _vercel (vercel internals, eg: web vitals)
      * - favicon.ico (favicon file)
      * - admin (admin panel)
      * - sitemap.xml (sitemap route)
      * - xmlsitemap.php (legacy sitemap route)
      * - robots.txt (robots route)
      */
-    '/((?!api|admin|_next/static|_next/image|_vercel|favicon.ico|xmlsitemap.php|sitemap.xml|robots.txt).*)',
+    '/((?!api|admin|_next/static|_next/image|v1|_vercel|favicon.ico|xmlsitemap.php|sitemap.xml|robots.txt).*)',
```
chanceaclark: This `v1` isn't segment-anchored, so it'll match any slug starting with "v1" (e.g. `/v1-turbo-kit`), not just the loader path. Might want `v1/` instead.

core/vibes/soul/sections/cart/client.tsx
```diff
@@ -23,8 +24,10 @@ import {
   GiftCertificateCodeFormState,
 } from '@/vibes/soul/sections/cart/gift-certificate-code-form';
 import { StickySidebarLayout } from '@/vibes/soul/sections/sticky-sidebar-layout';
+import { ClientWalletButtons } from 'components/wallet-buttons';
```
chanceaclark: Two things: this import should be `~/components/wallet-buttons` like the rest of the file. Also, `vibes/soul`'s supposed to stay presentation-only, this pulls in app-level wallet logic plus a `cartId` prop. Could pass `ClientWalletButtons` in as a `ReactNode` slot instead (see `product-detail`'s `additionalActions`).

core/lib/wallet-buttons/index.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  private async initCheckoutButtonInitializer() {
+    if (!window.checkoutKitLoader) {
+      throw new InitializationError();
+    }
+
+    const checkoutButtonModule = await window.checkoutKitLoader.load('wallet-button');
+
+    return checkoutButtonModule.createWalletButtonInitializer({
+      graphQLEndpoint: '/graphql',
```
chanceaclark: We probably should react to the trailing slash environment variable and append it if it's enabled.

core/.env.example
```diff
@@ -44,3 +44,8 @@ DEFAULT_REVALIDATE_TARGET=3600
...
+# In non-development environments the production CDN loader is always used.
+# Related files can be placed in core/public/v1 so they are served by the Next server in dev mode
+NEXT_PUBLIC_CHECKOUT_SDK_DEV_URL=http://localhost:3000/v1/loader.js
```
chanceaclark: Can we avoid having to drop the file in the local Catalyst folder? My thoughts here is that since you are trying to make changes in both repos, you can have two separate running dev processes:
1. In Catalyst, `http://localhost:3000/...`
2. In `checkout-sdk-js`, `http://localhost:3001/v1/loader.js`

This way we can get rid of the proxy.ts change below, you can make changes in both repos without having to rebuild, we don't need to drop a file into `core/public/`.

core/app/[locale]/(default)/cart/page-data.ts
Comment (bc-yaroslav-zhmutskyi) [reply core/app/[locale]/(default)/cart/page-data.ts:395]: Added cached `getSessionCustomerAccessToken` on [line#13](https://github.com/bigcommerce/catalyst/pull/3062/files#diff-d5e5c3bd2d9602afaf35db11c6adea51a6a29268df218fe0e76e3035a31d57c5R13)

core/.env.example
bc-yaroslav-zhmutskyi: Deleted changes from proxy.ts. Will serve files from `checkout-sdk-js`

core/app/[locale]/(default)/cart/page-data.ts
Comment (chanceaclark) [reply core/app/[locale]/(default)/cart/page-data.ts:395]: The parent should be cached here, not the `getSessionCustomerAccessToken` call. 

```ts
export const getPaymentWallets = cache(async (variables: PaymentWalletsVariables) => {
```

Also apply this to the other functions as well.

core/lib/wallet-buttons/index.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  private async initCheckoutButtonInitializer() {
+    if (!window.checkoutKitLoader) {
+      throw new InitializationError();
+    }
+
+    const checkoutButtonModule = await window.checkoutKitLoader.load('wallet-button');
+
+    return checkoutButtonModule.createWalletButtonInitializer({
+      graphQLEndpoint: process.env.TRAILING_SLASH !== 'false' ? '/graphql/' : '/graphql',
```
chanceaclark: This class only ever runs in the browser, and `TRAILING_SLASH` isn't `NEXT_PUBLIC_` prefixed (and there's no `env` block in next.config), so it's always `undefined` in the client bundle and this always resolves to `/graphql/`. On stores deployed with `TRAILING_SLASH=false` it only works because Next 308s `/graphql/` back to `/graphql`, which adds a redirect round trip to every wallet call and leans on the SDK's fetch re-sending the body + requester header.

Mind passing the endpoint in from the server through the init options instead? That's where we can actually read `TRAILING_SLASH`.

core/app/[locale]/(default)/cart/page.tsx
```diff
@@ -101,6 +142,16 @@ export default async function Cart({ params }: Props) {
...
 
+  const walletButtonsInitOptions = Streamable.from(async () => {
+    const walletButtons = await getPaymentWallets({
+      filters: {
+        cartEntityId: cartId,
+      },
+    });
+
+    return createWalletButtonsInitOptions(walletButtons, cart);
+  });
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/cart/page.tsx:160]: `Stream` is just Suspense + `use()` with no error boundary, so if `paymentWallets` or any of the follow-up queries reject here, the whole cart page goes down with it. Wallet buttons are an enhancement, so I would say catch inside the thunk and return `[]` so the cart still renders when this fails.

chanceaclark (commented):
Looking so much better. Few comments before it looks ready to go.

chanceaclark (approved):

PR #3075, opened 2026-07-01
B2B-5056 Fix B2B company registration for catalyst stores

shan-pandian:
## What/Why?

While investigating an issue related to company registration from catalyst store front, I found that local catalyst (b2b-integration) setup doesn't work properly. Making these changes to fix local setup.

loader.tsx — reads BIGCOMMERCE_GRAPHQL_API_DOMAIN from env (defaults to mybigcommerce.com) and passes it down as bcGraphqlDomain to both script components
script-dev.tsx — injects bc_graphql_domain into window.B3.setting so the buyer portal can build the correct BC Storefront API URL; also fixes a pre-existing bug where data-channelid and data-storehash attributes were swapped
script-production.tsx — same bc_graphql_domain injection as the dev script

## Testing
Manually tested the company registration 

https://github.com/user-attachments/assets/49c43a42-fd86-450d-8586-999fdf1c9101

## Migration

N/A


shan-pandian:
Incorrect base branch. Closing

PR #3076, opened 2026-07-01
fix(core): support required custom customer fields in Account Settings

### LINKED ISSUE #3074: Update customer missing fields
Update customer data in Account Settings requires first name, last name and email, and an optional company:

https://github.com/bigcommerce/catalyst/blob/81b88df3e42ff06c86bd91628d7be570579af89d/core/vibes/soul/sections/account-settings/schema.ts#L11-L16

However, the update function only sends the first name and last name to actually update. The email, nor the company is updated.
https://github.com/bigcommerce/catalyst/blob/81b88df3e42ff06c86bd91628d7be570579af89d/core/app/%5Blocale%5D/(default)/account/settings/_actions/update-customer.ts#L19-L22

Additionally, if a new field was added later (e.g. Terms and Conditions) which is now required. Then older accounts cannot update their info because the field is required, but there is no option to change it:

```1:{"account":"$T0:0:account","lastResult":{"status":"error","intent":"$undefined","initialValue":{"company":"XXXXX","email":"xxxxx@xxx.com","firstName":"XXXX","intent":"update","lastName":"XXXXX"},"error":{"":["Required account form field ID 26 ('Terms and Conditions') is missing"]},"state":"$undefined","fields":["firstName","lastName","email","company","intent"]}}```

### PR DESCRIPTION
**Pull request by chanceaclark**
## What/Why?
Fixes #3074.

The issue reports two problems with Account Settings' update-customer flow:

1. Claimed the mutation only sends `firstName`/`lastName`, dropping `email`/`company`. This turned out to be a misread of the code — the lines cited are the mutation's *response selection*, not the input. `submission.value` (validated against `updateAccountSchema`) already includes `email`/`company`, and I confirmed via direct GraphQL calls against a test store that both persist correctly today. No fix needed for this part.
2. Claimed that once a merchant adds a new **required** custom customer field, existing accounts have no way to satisfy it. This is real, and worse than described: I confirmed against a live store that BigCommerce revalidates *every* required custom customer field on *every* `updateCustomer` call — even fields that already have a stored value. A customer with a `firstName`-only edit gets `ValidationError: Required account form field ID X is missing` if that field isn't resent, regardless of whether it was previously set. Account Settings only ever submitted `firstName`/`lastName`/`email`/`company`, so any store with a required custom customer field (e.g. a "Terms and Conditions" checkbox added after accounts already existed) locked customers out of saving *any* change at all.

This is a regression: custom/dynamic form-field support existed in Account Settings since #1257 (Aug 2024) but was dropped during the Vibes migration (`b14180c4`, Dec 2024) and never restored, while Register and Address kept it via the newer `DynamicForm`/`formFieldTransformer` pattern.

This PR restores that support in Account Settings by reusing the same `DynamicForm`/`formFieldTransformer` pattern already used by Register and Address:
- `page-data.tsx` now fetches the customer's existing custom form field values (`customer.formFields`) alongside the field *definitions* it already fetched but never used.
- `page.tsx` filters out fields already covered by the fixed inputs (email, password, name, company, phone, newsletter checkbox) and transforms the rest into `Field`s, pre-filled with the customer's current values.
- `update-account-form.tsx` renders those fields in the *same* form as the existing firstName/lastName/email/company inputs (via the shared `DynamicFormField` renderer), so a single save always resubmits the full required-field state — required given BigCommerce's per-call revalidation.
- `update-customer.ts` builds the mutation's `formFields` input from whatever fields are present, following the same per-type mapping already used by `registerCustomer`/`updateAddress`.
- Small supporting extraction in `dynamic-form/schema.ts` (`getFieldsShape`) so Account Settings' schema can merge the fixed fields with the dynamic ones without duplicating per-field-type zod logic.

## Testing
- Set up a test store with a required "Terms and Conditions" checkbox custom customer field, added *after* a test customer already existed.
- Verified directly against the Storefront GraphQL API (bypassing the app) that:
  - `email`/`company` already persist correctly via the existing mutation input.
  - Omitting the required custom field from `updateCustomer` fails with `ValidationError`, even when the customer already has a stored value for it — confirming the bug and that the fix must always resend it.
- Ran the app locally and rendered `/account/settings` with an authenticated session: confirmed the custom checkbox renders alongside the fixed fields, correctly prefilled from the customer's stored value (checked), and that email/company also prefill from live data.
- `pnpm run typecheck` and `pnpm run lint` pass.
- Not verified: an actual form submission through the browser (no browser automation available in this environment). The submit-side mutation shape was verified directly against the API in isolation and matches what the form now produces.

## Migration
N/A — no breaking changes. `updateAccountSchema` changed from a `ZodObject` to a factory function; only relevant to direct importers of `@/vibes/soul/sections/account-settings/schema`.

### DIFF
--- a/core/app/[locale]/(default)/account/settings/page.tsx
+++ b/core/app/[locale]/(default)/account/settings/page.tsx
@@ -3,12 +3,48 @@ import { Metadata } from 'next';
 import { notFound } from 'next/navigation';
 import { getTranslations, setRequestLocale } from 'next-intl/server';
 
+import { Field } from '@/vibes/soul/form/dynamic-form/schema';
 import { AccountSettingsSection } from '@/vibes/soul/sections/account-settings';
+import { formFieldTransformer } from '~/data-transformers/form-field-transformer';
+import {
+  ACCOUNT_SETTINGS_FIELDS_TO_EXCLUDE,
+  mapFormFieldValueToName,
+} from '~/data-transformers/form-field-transformer/utils';
+import { exists } from '~/lib/utils';
 
 import { changePassword } from './_actions/change-password';
 import { updateCustomer } from './_actions/update-customer';
 import { updateNewsletterSubscription } from './_actions/update-newsletter-subscription';
 import { getAccountSettingsQuery } from './page-data';
+
+function withDefaultValue(field: Field, value: unknown): Field {
+  if (field.type === 'checkbox-group') {
+    return { ...field, defaultValue: Array.isArray(value) ? value.map(String) : undefined };
+  }
+
+  if (typeof value !== 'string') {
+    return field;
+  }
+
+  switch (field.type) {
+    case 'radio-group':
+    case 'select':
+    case 'swatch-radio-group':
+    case 'card-radio-group':
+    case 'button-radio-group':
+    case 'checkbox':
+    case 'number':
+    case 'text':
+    case 'textarea':
+    case 'date':
+    case 'email':
+    case 'hidden':
+      return { ...field, defaultValue: value };
+
+    default:
+      return field;
+  }
+}
 
 interface Props {
   params: Promise<{ locale: string }>;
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

**Comment by chanceaclark** [comment]
Proof: 
Screenshot 2026-07-01 at 14 44 40

File: core/app/[locale]/(default)/account/settings/page.tsx [hunk 1]
```diff
@@ -3,13 +3,49 @@ import { Metadata } from 'next';
...
+    case 'number':
+    case 'text':
+    case 'textarea':
+    case 'date':
+    case 'email':
+    case 'hidden':
+      return { ...field, defaultValue: value };
+
+    default:
+      return field;
```
Comment (parthshahp) [inline core/app/[locale]/(default)/account/settings/page.tsx:45]: Just checking, do we want default to return the field?

parthshahp (approved):

core/app/[locale]/(default)/account/settings/page.tsx
Comment (chanceaclark) [reply core/app/[locale]/(default)/account/settings/page.tsx:45]: Yeah we do as we need to map the form submission with some field and we need the field regardless of the default value.

PR #3081, opened 2026-07-06
TRAC-291 hide State/Province when the selected country has no states

mfaris9:
Jira: [TRAC-291](https://bigcommercecloud.atlassian.net/browse/TRAC-291)

## What/Why?

The Account Registration form validated State/Province as required even when the selected country has no states (Algeria, Monaco, etc.).

The register page now queries per-country state data from BigCommerce. When the shopper picks a country with no states, the State/Province field is hidden entirely and validation skips it. When the country does have states, the field stays visible and required.

## Testing

Tested locally against an integration store:

- Registration with country = Algeria (no states configured), State/Province field is hidden → registration succeeds

Screenshot 2026-07-07 at 11 21 32 AM

- Registration with country = US, leave State/Province blank → error
- Registration with country = US, fill State/Province → registration succeeds

Screenshot 2026-07-07 at 11 22 16 AM

## Migration

None. No changes to server-side behavior or data model.

[TRAC-291]: https://bigcommercecloud.atlassian.net/browse/TRAC-291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


parthshahp:
Is there no way to trigger a request or something to obtain this information after we select a country? I don't know if I like putting "Optional" text when it's not really optional.

Also, the error message has `stateOrProvince` inside of it in the video. Let's make sure to clean that up.

mfaris9:
@parthshahp Yeah I don't love how its says optional when required for a country with states i'll update. I matched validation to cart. I can add a conditional render of the field, but it will require changes to the dynamic form and this would be the only field that would do this.

On your second point, I think thats out of scope since thats how all errors are being displayed. stateOrProvince comes from the backend and if we were to make this look better it would make sense to change all server error messages. If this is something you want I can make a separate ticket for it.

mfaris9:
@parthshahp Ready for another review

rtalvarez (commented):
any tests we can add for this?

mfaris9:
@rtalvarez yes I can add. Would E2E be overkill? Can do unit test

rtalvarez:
> @rtalvarez yes I can add. Would E2E be overkill? Can do unit test

Yeah id just do unit for this one

mfaris9:
@rtalvarez Looks like we don't have the infrastructure for unit in catalyst. Would you prefer I build this out, do E2E or leave as is?

rtalvarez (approved):

PR #3095, opened 2026-07-10
fix(cli): validate channel name and clarify invalid-name error

jorgemoya:
Linear: [LTRAC-1088](https://linear.app/commerce/issue/LTRAC-1088)

## What/Why?
Creating a channel with an unsupported name (e.g. an apostrophe in "Bob's Store") failed with an opaque API error, forcing users to guess. This validates the channel name in the shared `runCreateChannelFlow`, so both `catalyst create` and `catalyst channel create` benefit. The interactive prompt validates as you type, and an invalid `--name` flag fails fast with a `UserActionableError` before hitting the API. Both paths name the offending input and list the allowed characters.

Note/tradeoff: this is a client-side allowlist (Unicode letters/numbers, spaces, hyphens, underscores). Names the API might currently accept with other punctuation (`&`, `#`, `(`, `.`) are now rejected client-side. There's no documented character constraint for the channels API, so this follows a conservative set — easy to widen the regex + message if the API is more permissive.

## Testing
- New `create-channel-flow.spec.ts` (13 tests): validator matrix (accepts plain/hyphen/underscore/accented/padded; rejects apostrophe/ampersand/hash/parens/empty/whitespace) and flow wiring (invalid `--name` throws before the API call, prompt `validate` wired, valid `--name` passes through).

## Migration
None.

### DIFF
--- a/packages/catalyst/src/cli/lib/create-channel-flow.ts
+++ b/packages/catalyst/src/cli/lib/create-channel-flow.ts
@@ -7,8 +7,7 @@ import { getAvailableLocales } from './localization';
 
 // Human-readable summary of the allowed characters, reused in the prompt
 // validation and the flag-path error so both surface the same guidance.
-const ALLOWED_CHANNEL_NAME_CHARS =
-  'letters, numbers, spaces, hyphens (-), and underscores (_)';
+const ALLOWED_CHANNEL_NAME_CHARS = 'letters, numbers, spaces, hyphens (-), and underscores (_)';
 
 // The Catalyst channels API rejects names containing other punctuation (e.g. an
 // apostrophe in "Bob's Store") with an opaque server error. Validate up front so
--- a/packages/catalyst/src/cli/lib/create-channel-flow.ts
+++ b/packages/catalyst/src/cli/lib/create-channel-flow.ts
@@ -2,7 +2,33 @@ import { checkbox, input, select } from '@inquirer/prompts';
 import { colorize } from 'consola/utils';
 
 import { createChannel, type CreatedChannel } from './channels';
+import { UserActionableError } from './errors';
 import { getAvailableLocales } from './localization';
+
+// Human-readable summary of the allowed characters, reused in the prompt
+// validation and the flag-path error so both surface the same guidance.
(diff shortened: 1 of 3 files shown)


jorgemoya:
Testing

```
~/dev/catalyst/core on  jorgemoya/ltrac-1088-unclear-error-when-creating-a-channel-with-an-invalid-name [?] took 13s
❯ node ../packages/catalyst/dist/cli.js channel create
◢ @bigcommerce/catalyst v1.0.0

? What would you like to name your new channel? test' wrong channel 2 #@$#
> "test' wrong channel 2 #@$#" is not a valid channel name. Channel names may contain only letters, numbers, spaces, hyphens (-), and underscores (_).
```

packages/catalyst/src/cli/lib/create-channel-flow.ts
```diff
@@ -2,8 +2,33 @@ import { checkbox, input, select } from '@inquirer/prompts';
 import { colorize } from 'consola/utils';
 
 import { createChannel, type CreatedChannel } from './channels';
+import { UserActionableError } from './errors';
 import { getAvailableLocales } from './localization';
 
+// Human-readable summary of the allowed characters, reused in the prompt
+// validation and the flag-path error so both surface the same guidance.
+const ALLOWED_CHANNEL_NAME_CHARS = 'letters, numbers, spaces, hyphens (-), and underscores (_)';
```
mfaris9, line 10: Would we also want to add a max length check?

jorgemoya: I couldn't find quickly any character limits, and I tested this with `"name": "testingngksdnglksdnglkdsfnagklndsfgklndfslkgndsfklgndsfkgndsfkngsdngsdngflksdfngknsdfgkndfskgndfs",` and it works. I believe we should just have the server return an error if the limit is too long (or maybe there is no limit).

mfaris9: Sounds good, approving now. Maybe its 255.

mfaris9 (approved):

PR #3097, opened 2026-07-10
fix(cli): give a clear error on deploy when required env vars are missing

jorgemoya:
Linear: [LTRAC-1090](https://linear.app/commerce/issue/LTRAC-1090)

## What/Why?
Running `catalyst deploy` without the required environment variables produced an ugly raw build stack trace with no hint that missing env vars were the cause (the fix is `--env-path`, but nothing said so). This detects the missing-required-env condition before the build and fails with a `UserActionableError` that names the missing variables and points at `--env-path` and the likely `.env.local` location, instead of surfacing the raw build error.

See the PR diff for how "required" vars are determined; the check is intentionally minimal and targeted at the vars the core build asserts on.

## Testing
- New `required-build-env.spec.ts` + `build.spec.ts` — 9 tests pass, covering the missing-env path and the pass-through case.

## Migration
None.


jorgemoya:
Follow-up (cross-ref #3098): the missing-required-env message added here says the build auto-loads `.env.local` from the current directory. #3098 changes build/deploy env resolution to auto-load **both** `.env.local` and `.env` (with `.env.local` taking precedence). Whichever of these two PRs merges second should update this message to read "`.env.local` or `.env`" so the wording stays accurate.

jorgemoya:
Update: the missing-env message now references both `.env.local` and `.env` (with `.env.local` taking precedence), so no post-merge wording fix is needed regardless of order.

Caveat: the actual `.env` auto-loading lands in #3098. If #3097 merges to `canary` before #3098, there's a brief window where the message mentions `.env` but only `.env.local` is auto-loaded. Both PRs are headed to the same release, so this is intentional per discussion.

jorgemoya:
Testing:

```
◢ @bigcommerce/catalyst v1.0.0

ℹ Project is not set up for Commerce Hosting — running next build.
ℹ To deploy to Commerce Hosting, run catalyst deploy.

 ERROR  Missing required environment variables for the build:
  - BIGCOMMERCE_STORE_HASH
  - BIGCOMMERCE_STOREFRONT_TOKEN
  - BIGCOMMERCE_CHANNEL_ID
  - AUTH_SECRET

These are read by the build itself, so they must be set before it runs. The build automatically loads .env.local and .env from the current directory (/Users/jorge.moya/dev/catalyst/core) — add the variables above to one of those files (.env.local takes precedence).
To load an env file from a different location, pass --env-path <path> (for example --env-path ../.env.local when running from a subdirectory).
```

jairo-bc (approved):

PR #3098, opened 2026-07-10
feat(cli): require --env-path instead of auto-loading .env.local

jorgemoya:
Linear: [LTRAC-1091](https://linear.app/commerce/issue/LTRAC-1091)

## What/Why?
The CLI auto-loaded `.env.local` from cwd but not `.env`, an inconsistent bit of magic that surprised users. Per sign-off, this drops automatic env-file loading entirely: env files are now loaded **only** when passed explicitly via `--env-path`. This removes the `.env` vs `.env.local` asymmetry and makes env resolution explicit and predictable. Help text is updated to match.

## Testing
- `index.spec.ts` (7 tests) covers no-auto-load without `--env-path` and correct loading with `--env-path`.

## Migration
**Breaking behavior change.** Users who relied on the auto-loaded `.env.local` must now pass it explicitly:

```
catalyst deploy --env-path .env.local
```

(or the appropriate path, e.g. `--env-path ./core/.env.local`).

### DIFF
--- a/.changeset/require-env-path.md
+++ b/.changeset/require-env-path.md
@@ -2,11 +2,6 @@
 "@bigcommerce/catalyst": minor
 ---
 
-Env files are no longer auto-loaded by the CLI. Previously `.env.local` was
-automatically read from the current working directory (but `.env` was not),
-which was surprising and inconsistent. Now nothing is loaded automatically;
-pass `--env-path <path>` to load an env file explicitly.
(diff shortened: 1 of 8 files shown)


mfaris9 (approved):
Looks good to me!

jorgemoya:
Follow-up (cross-ref #3097): this PR makes build/deploy auto-load both `.env.local` and `.env`. #3097 adds a missing-required-env error message that currently only mentions `.env.local`. Whichever of these two PRs merges second should update that message (in `packages/catalyst/src/cli/lib/required-build-env.ts`) to read "`.env.local` or `.env`" so it matches the loading behavior.

PR #3103, opened 2026-07-13
fix: TRAC-284 Coalesce rapid cart quantity/delete clicks

chanceaclark:
Jira: [TRAC-284](https://bigcommercecloud.atlassian.net/browse/TRAC-284)

## What/Why?

Server actions run strictly serially with no abort/parallelism, so rapid clicks on the cart's quantity +/- or delete buttons queued up unboundedly. Combined with optimistic UI, users could click far ahead of the server, and navigating away while the queue drained had historically frozen client-side routing until a hard refresh. See [Linear LTRAC-284](https://linear.app/commerce/issue/LTRAC-284/cart-page-breaks-when-clicking-around-during-a-quantity-increase) for the full prior investigation and options considered.

**The fix:** a per-line-item pending-intent map + debounced, single-flight dispatcher in `CartClient`. Rapid clicks coalesce into one absolute-quantity `update` request (replacing the old per-click `increment`/`decrement` intents, which also collapsed ~150 duplicated lines in the server action into one branch). Deletes flush immediately and are serialized the same way. At most one action is ever in flight, so a failed request only rolls back one gesture instead of an entire queue.

**Progressive enhancement:** each +/-/delete control is its own `display: contents` micro-form carrying a real, pre-computed absolute quantity (or delete) straight to the server action. Without JS (or before hydration), clicking still works as a normal page round-trip. With JS, `onSubmit` intercepts and routes through the dispatcher instead. Quantity is clamped server-side only (`z.coerce.number().int().min(1)`) — no client-side quantity clamp, since the `disabled` attribute already prevents decrementing below 1 in every reachable path.

**Why the cart section is now keyed by `entityId` only** (previously `entityId`-`version`): keying on `version` remounted the whole section on every mutation. React drops actions dispatched into a hook that unmounts before they run, which could deadlock the old design, and clicks landing mid-DOM-swap hit handler-less nodes — this was the actual mechanism behind the "stuck forever" freeze a merchant reported on video. Dispatcher state now lives at the component instance level (not module scope), so if anything ever does remount mid-flight, the failure mode is a harmless snap-back to server truth, not a deadlock.

**Also fixed:** the checkout button spinning forever if its navigation was cancelled (pressing Esc, choosing "Stay" on the leave-page confirm, a dropped connection). The URL-string checkout action awaited a promise around `window.location.assign` that never resolved by design; a cancelled redirect left retry clicks queued behind that dead action forever. It now settles after 8s or on `pageshow` (bfcache restore), re-enabling the button as a retry.

Cart revalidation is also unified on `revalidateTag(TAGS.cart)` (previously a mix of that and `revalidatePath('/cart')`).

This branch went through a full review pass (code-review at high effort with two background verifier agents on timing edge cases, simplify, security-review) before this PR was opened; all findings were resolved and re-verified.

## Rollout/Rollback

Standard PR merge to `canary`; no feature flag, migration, or experiment involved. Rollback is a straight revert — no schema or data changes.

## Testing

Verified end-to-end against a live sandbox store with Playwright, covering:
- 10 rapid quantity-increase clicks coalesce into a single POST; final quantity and totals are correct.
- Reload mid-burst and after settling shows consistent server-truth quantities (the "phantom items" symptom from the original report).
- Clicking a nav link while a mutation is in flight completes navigation (confirm-dialog interceptors are still in place as a UX guard, kept intentionally).
- Rapid delete across multiple rows removes all rows correctly, no resurrected items.
- A net-zero burst (e.g. +3 then -3 within the debounce window) sends zero requests.
- No-JS fallback: with the app's JS bundle blocked (so hydration never completes, while leaving the browser's JS engine on, since fully disabling JS also breaks this page's pre-existing, unrelated `<Stream>`/Suspense fallback swap), quantity +/- and delete all work as real page round-trips against the server action.
- Checkout button recovers (re-enables, and a retry succeeds) after its navigation is cancelled.

Manual repro steps from the original ticket: add 10+ variants to cart, rapidly click delete/quantity buttons, then click a nav link — no freeze, no hard refresh required.

[TRAC-284]: https://bigcommercecloud.atlassian.net/browse/TRAC-284?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/tidy-hoops-refuse.md
+++ b/.changeset/tidy-hoops-refuse.md
@@ -1,0 +1,11 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Coalesce rapid cart line item quantity and delete clicks into at most one in-flight server action. Quantity buttons now update an optimistic pending intent that flushes a single absolute-quantity `update` intent after a short debounce (replacing the per-click `increment`/`decrement` intents), and deletes are serialized instead of queueing unboundedly. This prevents the serial server-action queue buildup that could lock up navigation on the cart page, and unifies cart revalidation on `revalidateTag`.
+
+The cart section is now keyed by cart `entityId` only (previously `entityId`-`version`) and renders line items from the revalidation-refreshed `cart` prop. Remounting on every mutation swallowed clicks landing during the DOM swap and could drop queued actions, leaving the summary skeletons and checkout button stuck in a pending state until hard refresh.
+
+Each quantity/delete control is now its own progressive-enhancement form that posts a real absolute quantity or delete request to the server action; JS intercepts the submit to route through the coalescing dispatcher, but the buttons keep working (as full page round-trips) before hydration or if the app bundle fails to load.
+
+Also fixes the checkout button getting stuck spinning forever if its navigation is cancelled (Esc, "Stay" on the leave-page prompt, a flaky connection). The URL-string checkout action previously awaited a promise that never resolved, leaving retry clicks queued behind a dead action; it now settles after a short timeout and on bfcache restore, re-enabling the button as a retry.
--- a/core/app/[locale]/(default)/cart/_actions/update-line-item.ts
+++ b/core/app/[locale]/(default)/cart/_actions/update-line-item.ts
@@ -53,7 +53,9 @@ export const updateLineItem = async (
   }
 
   switch (submission.value.intent) {
-    case 'increment': {
+    case 'update': {
+      const { quantity } = submission.value;
+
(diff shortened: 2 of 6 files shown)


chanceaclark:
## Proof

https://github.com/user-attachments/assets/8811d853-bffd-437d-b125-0e0d14ad2908

jairo-bc (approved):

PR #3105, opened 2026-07-14
LTRAC-1086: feat(cli) - Add --wrangler-version flag to build and deploy

jorgemoya:
Linear: [LTRAC-1086](https://linear.app/commerce/issue/LTRAC-1086/add-a-cli-flag-to-build-with-a-different-wrangler-version)

## What/Why?

`catalyst build` pins the Wrangler version it shells out to (`WRANGLER_VERSION = '4.90.0'`), so developers had no way to build against a different Wrangler version without editing CLI source. This adds a `--wrangler-version <version>` option to both `catalyst build` and `catalyst deploy`:

- When provided, it's used in place of the pinned default in the `wrangler@<version>` invocation.
- When omitted, the pinned default is used (unchanged behavior).
- On `deploy --prebuilt` the flag is ignored, since that path skips the build.

The value is validated (`^[a-zA-Z0-9][a-zA-Z0-9._+-]*$`) as a version or dist-tag before it's interpolated into the `wrangler@...` spec, so it can't smuggle extra args or shell metacharacters into the `pnpm dlx` call.

## Testing

`pnpm vitest run src/cli/commands/build.spec.ts src/cli/commands/deploy.spec.ts` — 40 tests pass. New unit tests cover the flag being threaded into the `wrangler@<version>` invocation, the pinned default being used when the flag is absent, and invalid values being rejected. `tsc --noEmit` and `eslint` are clean.

Manual:
- `catalyst build --wrangler-version 4.24.3` builds using the specified version.
- `catalyst build` (no flag) builds using the pinned default.
- `catalyst build --help` documents the flag.

## Migration

None.


packages/catalyst/src/cli/commands/build.ts
```diff
@@ -12,9 +12,28 @@ import { assertRequiredBuildEnv } from '../lib/required-build-env';
...
+
+export const parseWranglerVersion = (value: string): string => {
+  if (!WRANGLER_VERSION_PATTERN.test(value)) {
+    throw new InvalidArgumentError(
+      `"${value}" is not a valid Wrangler version or dist-tag (e.g. 4.90.0 or latest).`,
+    );
+  }
+
+  return value;
+};
```
chanceaclark: Do we just want to use https://www.npmjs.com/package/semver for this?

chanceaclark (approved):

packages/catalyst/src/cli/commands/build.ts
jorgemoya: Yeah, makes sense will update.

PR #3107, opened 2026-07-14
fix(cli): prevent dependency install from hanging on link/deploy setup

jorgemoya:
Linear: [LTRAC-1224](https://linear.app/commerce/issue/LTRAC-1224/fixcli-dependency-install-hangs-on-project-link-deploy-setup)

## What/Why?

`catalyst project link` (and the `catalyst deploy` first-run setup) could hang indefinitely on `Installing dependencies with pnpm. This could take a minute…` and never finish.

Root cause (not the corepack prompt — that's gated on `stdin.isTTY` and defaults off for the `corepack pnpm` entrypoint): nypm runs the install through **tinyexec**, which in silent mode drains the child's **stdout** to completion *before* it starts reading **stderr**. pnpm sends its progress to stdout but floods stderr with Node warnings — on Node 26 the `File descriptor N opened in unmanaged mode` spam alone is thousands of lines. That stderr fills its ~64KB pipe buffer while tinyexec is still busy on stdout, pnpm blocks on the stderr write, stdout never completes, and the install deadlocks.

Fix: keep the silent spinner UX and starve the flood instead. The install child now runs with `NODE_NO_WARNINGS=1` (drops pnpm/corepack's internal Node warnings, the bulk of the stderr traffic) and pins `COREPACK_ENABLE_DOWNLOAD_PROMPT=0` defensively. Both are saved/restored so they don't leak into later CLI operations. Verified that pnpm's reporter output goes to stdout (drained safely regardless of size) while only warnings hit stderr, so suppressing them keeps stderr small enough that the sequential drain can't deadlock.

Also: the `link`/`deploy` setup paths called `installDependencies(projectDir)` with no package manager, always defaulting to pnpm. They now use a new `detectProjectPackageManager()` that reads the project's lockfile / `packageManager` field, so npm/yarn/bun projects aren't forced through corepack pnpm.

## Testing

- `pnpm --filter @bigcommerce/catalyst test` — new `detect-package-manager.spec.ts` plus existing `project`/`deploy`/`create` specs pass (89 tests).
- `tsc --noEmit` and ESLint clean on the changed files.
- Manual: confirmed via captured streams that pnpm reporter output → stdout and Node warnings → stderr, and that the fd-warning spam disappears once the install runs the child under `NODE_NO_WARNINGS`.

## Migration

None.


packages/catalyst/src/cli/lib/detect-package-manager.ts
```diff
@@ -20,3 +22,20 @@ export const detectPackageManager = (): PackageManager => {
...
+// default) when detection is inconclusive or returns a manager we don't support.
+export const detectProjectPackageManager = async (projectDir: string): Promise<PackageManager> => {
+  const detected = await detectFromDir(projectDir);
+  const name = detected?.name;
+
+  if (name === 'pnpm' || name === 'yarn' || name === 'bun' || name === 'npm') {
+    return name;
+  }
+
+  return 'pnpm';
```
chanceaclark: Hmm I think we should default to npm as most people will have it rather than pnpm. What do you think?

chanceaclark (approved):

packages/catalyst/src/cli/lib/detect-package-manager.ts
jorgemoya: Agreed

PR #3109, opened 2026-07-15
feat(core): TRAC-282 Update cart locale when switching language

jordanarldt:
Jira: [TRAC-282](https://bigcommercecloud.atlassian.net/browse/TRAC-282)

## What/Why?

Switching the storefront locale left the cart on its original locale. Adds a `switchLocale` server action that calls the new `updateCartLocale` mutation to update the cart in place before navigating, mirroring the existing currency-switch behavior.

The mutation is flag-gated server-side and returns `null` when disabled — `switchLocale` no-ops in that case, and swallows/logs any other errors so a failure never blocks navigation.

## Testing

- `switchLocale` returns early with no cart, a `null` result (flag off), or mutation errors — navigation proceeds unaffected in all cases.

https://github.com/user-attachments/assets/ce4225b5-0067-49df-8596-55293ed20bf3

## Migration

N/A

[TRAC-282]: https://bigcommercecloud.atlassian.net/browse/TRAC-282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/sync-cart-locale-on-switch.md
+++ b/.changeset/sync-cart-locale-on-switch.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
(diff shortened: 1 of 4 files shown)


parthshahp (approved):
Looks good!

PR #3110, opened 2026-07-15
feat(cli): add `catalyst debug` command to gather diagnostics for bug reports

jorgemoya:
Linear: [LTRAC-1087](https://linear.app/commerce/issue/LTRAC-1087)

## What/Why?

When a developer hits a CLI problem we currently rely on them describing their setup by hand. This adds `catalyst debug`, a single command that prints a diagnostic report they can attach to a bug report / GitHub issue to speed up triage.

The report covers:
- **CLI**: package name + version
- **Runtime**: Node version, platform/arch, OS release, detected package manager
- **Project**: cwd, `projectUuid`, and the `getProjectState` flags (linked / transformed / fully-set-up, middleware/proxy/OpenNext presence)
- **Config** (resolved *without* secrets): store hash / access token / project UUID as `{ present, source }` following the `process.env > project.json` priority chain; `project.json` top-level key names; stored deployment env var key names; and each CLI-relevant env var as a presence boolean
- **Telemetry**: enabled flag + correlation ID
- **Files**: existence-only checks for `.env.local`, `.env`, `.bigcommerce/project.json`, `.bigcommerce/wrangler.jsonc`, `.open-next/`, `package.json`

Notable decisions:
- **Secrets/PII are never printed** — this was a hard requirement. Credentials, stored env vars, and environment variables are reported by presence only, never by value. There's a dedicated test asserting no seeded secret value appears anywhere in the output.
- The collector (`collect-diagnostics.ts`) reads `project.json` **raw** rather than via `getProjectConfig()`, which would instantiate `Conf` and create `.bigcommerce/` as a side effect (same reasoning as `project-state.ts`). There's a regression test for the no-side-effect guarantee.
- Logic lives in a pure, injectable collector so the secret-masking guarantee is unit-testable; the command is a thin formatter (human-readable by default, `--json` for copy/paste or piping).

## Testing

```bash
# Human-readable report
catalyst debug

# Machine-readable
catalyst debug --json
```

- `pnpm test` — new specs (`collect-diagnostics.spec.ts`, `debug.spec.ts`) pass; both new source files are at 100% coverage (the repo's threshold gate).
- `pnpm typecheck` and `pnpm lint` clean.
- Verified end-to-end against a project with a populated `.bigcommerce/project.json` and secret env vars set: config sources resolve correctly and **no secret value leaks** into either output mode.
- Confirmed `debug` is registered and appears in top-level `catalyst --help`.

## Migration

None — additive new command.

### DIFF
--- a/packages/catalyst/src/cli/commands/debug.spec.ts
+++ b/packages/catalyst/src/cli/commands/debug.spec.ts
@@ -1,0 +1,161 @@
+import { Command } from 'commander';
+import { afterEach, beforeAll, beforeEach, expect, test, vi } from 'vitest';
+
+import { collectDiagnostics, type Diagnostics } from '../lib/collect-diagnostics';
+import { consola } from '../lib/logger';
+import { program } from '../program';
+
+import { debug } from './debug';
+
+vi.mock('../lib/collect-diagnostics', () => ({
+  collectDiagnostics: vi.fn(),
+}));
+
+const fullDiagnostics: Diagnostics = {
+  cli: { name: '@bigcommerce/catalyst', version: '9.9.9' },
+  runtime: {
+    node: 'v20.0.0',
+    platform: 'darwin',
+    arch: 'arm64',
+    osRelease: '25.5.0',
+    packageManager: 'pnpm',
+  },
+  project: {
+    cwd: '/tmp/project',
+    projectUuid: 'uuid-123',
+    isLinked: true,
+    isTransformed: true,
+    isFullySetUp: true,
+    hasMiddleware: true,
+    hasProxy: false,
+    hasOpenNextDep: true,
+  },
+  config: {
+    storeHash: { present: true, source: 'project.json' },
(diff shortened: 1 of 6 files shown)


packages/catalyst/src/cli/commands/debug.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+  vi.mocked(collectDiagnostics).mockReturnValue(fullDiagnostics);
+
+  await program.parseAsync(['node', 'catalyst', 'debug']);
+
+  const output = lastLogOutput();
+
+  expect(output).toContain('Catalyst CLI Diagnostics');
+  expect(output).toContain('@bigcommerce/catalyst');
+  expect(output).toContain('pnpm');
+  expect(output).toContain('Catalyst core:      @bigcommerce/catalyst-core@1.8.0');
```
chanceaclark: Is "Catalyst core" the right label here?

chanceaclark (approved):

PR #3111, opened 2026-07-15
feat(cli): rename BIGCOMMERCE_API_HOST → CATALYST_API_HOST and resolve --api-host from project.json

jorgemoya:
Linear: [LTRAC-1249](https://linear.app/commerce/issue/LTRAC-1249)

## What/Why?

The CLI's `--api-host` was the odd one out among its config: bound to `BIGCOMMERCE_API_HOST` (not `CATALYST_`-prefixed) and resolved only as flag > env > hardcoded default via Commander — it couldn't be persisted in `.bigcommerce/project.json` like store hash / access token / project UUID.

This makes the API host a first-class `CATALYST_` config, resolved with the same precedence as the other credentials:

`--api-host` flag > `CATALYST_API_HOST` env > `.bigcommerce/project.json` `apiHost` > default `api.bigcommerce.com`

Key changes:
- `shared-options.ts`: rename the env binding to `CATALYST_API_HOST`, drop the Commander `.default('api.bigcommerce.com')`, and add a `resolveApiHost(options, config)` helper + `DEFAULT_API_HOST` constant (mirrors `resolveCredentials`/`resolveProjectUuid`).
- `project-config.ts`: add `apiHost?: string` to `ProjectConfigSchema` so it can be persisted/read from `project.json`.
- Route every command that accepts `--api-host` (`project`, `logs`, `domains`, `channel`, `deploy`, `auth`) through `resolveApiHost` instead of reading `options.apiHost` directly. Each action already has a `config` handle, so the resolver slots in at the top of the action. `create` is untouched (it builds its own host from `--bigcommerce-hostname`).

**Breaking change:** `BIGCOMMERCE_API_HOST` is no longer read — set `CATALYST_API_HOST` instead. It's a hidden/advanced knob (pre-GA), so blast radius is low; called out in the changeset.

Surfaced while reviewing the `catalyst debug` report (LTRAC-1087). The debug report's `CATALYST_API_HOST` label + `project.json` source will follow once #3110 merges (that branch is where the diagnostics collector lives).

## Testing

- New `shared-options.spec.ts` covers all four `resolveApiHost` precedence branches (flag / project.json / default / flag-beats-config).
- Updated `logs`/`deploy`/`project` command specs that asserted the old Commander default on `--api-host`.
- `pnpm typecheck` and `pnpm lint` clean; all touched command specs pass (199 tests across the 7 affected specs). The full-suite 100% coverage gate runs in CI (the `upgrade.*` integration specs are too slow/flaky to run locally).

## Migration

Replace `BIGCOMMERCE_API_HOST=...` with `CATALYST_API_HOST=...` in any environment or CI that overrides the API host. You can now also set `"apiHost"` in `.bigcommerce/project.json` or pass `--api-host`.

### DIFF
--- a/.changeset/catalyst-api-host-config.md
+++ b/.changeset/catalyst-api-host-config.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst": minor
+---
+
+`catalyst` now reads the API host from `CATALYST_API_HOST` (renamed from `BIGCOMMERCE_API_HOST`) and resolves it with the same precedence as other credentials: `--api-host` flag > `CATALYST_API_HOST` > `.bigcommerce/project.json` `apiHost` > default `api.bigcommerce.com`. **Breaking:** the `BIGCOMMERCE_API_HOST` environment variable is no longer read — set `CATALYST_API_HOST` instead.
--- a/packages/catalyst/src/cli/commands/auth.ts
+++ b/packages/catalyst/src/cli/commands/auth.ts
@@ -6,7 +6,7 @@ import { consola } from '../lib/logger';
 import { LoginAbortedError, login as runInteractiveLogin } from '../lib/login';
 import { fetchProjects } from '../lib/project';
 import { getProjectConfig } from '../lib/project-config';
-import { loginUrlOption } from '../lib/shared-options';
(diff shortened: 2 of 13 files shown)


parthshahp (approved):
Looks good

PR #3124, opened 2026-07-21
TRAC-934: feat - implement pagination for `catalyst logs query` command

parthshahp:
## What/Why?

Adding the ability to paginate for the `catalyst logs query` command.

## Testing

Tested via manually deploying a site and using the Catalyst CLI.

https://github.com/user-attachments/assets/a4dc57e1-1c13-4607-92c8-917029a50078

## Migration

No migration needed.


packages/catalyst/src/cli/lib/telemetry.ts
```diff
@@ -9,6 +9,12 @@ import { getUserConfig, UserConfigSchema } from './user-config';
 const TELEMETRY_KEY_ENABLED = 'telemetry.enabled';
 const TELEMETRY_KEY_ID = `telemetry.anonymousId`;
 
+// CLI telemetry is best-effort: command completion must never be delayed by a
```
parthshahp, line 12: This was causing problems when testing, let me know if I should remove this.

packages/catalyst/src/cli/lib/project.ts
```diff
@@ -29,9 +29,34 @@ export interface ProjectListItem {
 }
 
 function projectsUrl(storeHash: string, apiHost: string) {
+  console.log(`https://${apiHost}/stores/${storeHash}/v3/infrastructure/projects`);
```
jordanarldt: @parthshahp Should this be removed? Looks like it was for debugging?

parthshahp: Removed

chanceaclark (approved):

PR #3158, opened 2026-08-03
LTRAC-1421: Surface clear error for non-JWT storefront token

jorgemoya:
Linear: [LTRAC-1421](https://linear.app/commerce/issue/LTRAC-1421/unclear-401-when-storefront-token-isnt-a-storefront-jwt)

## What/Why?

Vortex IQ (closed beta partner) hit a bare 401 during Native Hosting setup because `BIGCOMMERCE_STOREFRONT_TOKEN` wasn't a storefront JWT (e.g. an OAuth access token was supplied instead). The response gave no indication of the cause, leaving them to reverse-engineer that the token type mattered.

`packages/client/src/client.ts` now checks, on any 401 response, whether the configured storefront token is a well-formed JWT (three base64url segments with a JSON-decodable payload — see `isWellFormedStorefrontToken` in `packages/client/src/lib/storefront-token.ts`). If it isn't, it throws a new `InvalidStorefrontTokenError` (extends `BigCommerceAPIError`, so existing `instanceof BigCommerceAPIError` handling still works) explaining that a storefront JWT is required and linking to how to generate one. A 401 with a well-formed token still falls through to the existing generic error, so this doesn't mask other causes of a 401 (expired/revoked JWT, wrong store hash, etc.).

Includes a changeset for `@bigcommerce/catalyst-client` (patch).

## Testing

No automated tests included yet. Manual verification:

1. Set `BIGCOMMERCE_STOREFRONT_TOKEN` to a non-JWT value (e.g. a BigCommerce OAuth access token) and hit any storefront page/GraphQL query that uses `client.fetch`. Confirm the thrown error is `InvalidStorefrontTokenError` with the actionable message, instead of a bare "BigCommerce API returned 401".
2. Confirm a valid storefront JWT continues to work normally (no regression on the happy path).
3. Confirm a 401 caused by something else (e.g. a well-formed but expired/invalid JWT) still throws the original generic `BigCommerceAPIError`, not the new error.

Follow-up: add unit tests for `isWellFormedStorefrontToken` and the `client.ts` 401 branch (deferred out of this PR).

## Migration

None.

### DIFF
--- a/packages/client/src/lib/storefront-token.ts
+++ b/packages/client/src/lib/storefront-token.ts
@@ -1,0 +1,23 @@
+const BASE64URL_SEGMENT = /^[A-Za-z0-9_-]+$/;
+
+// A BigCommerce storefront token is a JWT: three base64url segments
+// (header.payload.signature) whose payload decodes to a JSON object. Other
+// token types (e.g. OAuth access tokens) are opaque strings that do not match
+// this shape, so this lets us distinguish "wrong kind of token" from other
+// causes of a 401.
+export function isWellFormedStorefrontToken(token: string): boolean {
+  const segments = token.split('.');
+
+  if (segments.length !== 3 || !segments.every((segment) => BASE64URL_SEGMENT.test(segment))) {
+    return false;
+  }
+
+  try {
+    const payload = segments[1].replace(/-/g, '+').replace(/_/g, '/');
+    const decoded = JSON.parse(atob(payload)) as unknown;
+
+    return typeof decoded === 'object' && decoded !== null;
+  } catch {
+    return false;
+  }
+}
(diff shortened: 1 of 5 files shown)


packages/client/src/lib/storefront-token.ts
```diff
@@ -0,0 +1,23 @@
...
+// causes of a 401.
+export function isWellFormedStorefrontToken(token: string): boolean {
+  const segments = token.split('.');
+
+  if (segments.length !== 3 || !segments.every((segment) => BASE64URL_SEGMENT.test(segment))) {
+    return false;
+  }
+
+  try {
+    const payload = segments[1].replace(/-/g, '+').replace(/_/g, '/');
```
parthshahp, line 17: Is this really the best way to test if this is a valid storefront token?

jorgemoya: Not the best for sure. I will rename this function to make it clear that this is just a simple client check and not the real validation.

The other way to validate is by doing a network request but not sure that is worth it for the problem at hand. What do you think?

parthshahp: Honestly I think it's fine to do a network request to check, but fair point that it's not a huge problem. I'll leave it up to you then.

parthshahp (approved):

PR #3160, opened 2026-08-04
fix(login): B2B-5343 Fix runtime errors for local env

bc-lulu-yang:
## Summary

There are two errors when running Catalyst against a local Buyer Portal (`LOCAL_BUYER_PORTAL_HOST`) - _see the attached before video_  

- Encountered a script tag while rendering React component. Scripts inside React components are never executed when rendering on the client. Consider using template tag instead (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template).

- @vitejs/plugin-react can't detect preamble. Something is wrong.

Local B2B development loads the Buyer Portal from a Vite dev server (`http://localhost:3001`). That path depends on, in order:

1. Installing Vite's React refresh **preamble** before any Buyer Portal module evaluates
2. Loading `@vite/client`
3. Loading `/src/main.ts`
4. Having `window.B3` config available before the portal reads it

When the preamble is missing, the first Buyer Portal component module to evaluate throws:

```text
@vitejs/plugin-react can't detect preamble. Something is wrong.
```

(The reported location `Provider.tsx:20` is the preamble check that `@vitejs/plugin-react` injects into every transformed module, not meaningful source code, which made this confusing to debug.)

That blocks local login / Buyer Portal integration work on this branch. 

Not sure why it didn't work recently, as checked, seems React 19 is already there for a few months, so likely changed recently is Next.js (around `16.1` → `16.2` on this branch. But not too sure, just looks like the timing matches.

## What changed in this PR

| File | Change |
| --- | --- |
| `core/b2b/script-dev.tsx` | Now composes two small client components instead of rendering `next/script` tags |
| `core/b2b/b2b-dev-script-injector.tsx` (new) | Injects the config + one ordered Vite entry module via `document.createElement('script')` in `useEffect` |
| `core/b2b/b2b-client-effects.tsx` (new) | Pure extraction of the existing `useB2BAuth`/`useB2BCart` hooks — no behavior change |

Injection order is now guaranteed within a single module script:

1. Classic script setting `window.B3`
2. Module script that installs the React refresh preamble
3. `await import('${host}/@vite/client')`
4. `await import('${host}/src/main.ts')`

Scripts appended via the DOM API are always executed by the browser.

## Testing
Before

https://github.com/user-attachments/assets/382f6fa2-7b20-439b-b619-6bb5bcf430b2

After

https://github.com/user-attachments/assets/c0a401d9-24f0-4e36-951e-999217d00a62

## Migration
No needed, local env change only

### DIFF
--- a/core/b2b/b2b-client-effects.tsx
+++ b/core/b2b/b2b-client-effects.tsx
@@ -1,0 +1,16 @@
+'use client';
+
+import { useB2BAuth } from './use-b2b-auth';
+import { useB2BCart } from './use-b2b-cart';
+
+interface Props {
+  token?: string;
+  cartId?: string | null;
+}
+
+export function B2BClientEffects({ cartId, token }: Props) {
+  useB2BAuth(token);
+  useB2BCart(cartId);
+
+  return null;
+}
--- a/core/b2b/b2b-dev-script-injector.tsx
+++ b/core/b2b/b2b-dev-script-injector.tsx
@@ -1,0 +1,59 @@
+'use client';
+
+import { useEffect } from 'react';
+
+interface Props {
+  storeHash: string;
+  channelId: string;
+  hostname: string;
+  bcGraphqlDomain?: string;
+}
+
+export function B2BDevScriptInjector({ bcGraphqlDomain, channelId, hostname, storeHash }: Props) {
+  useEffect(() => {
+    if (document.getElementById('b2b-vite-entry')) {
+      return;
+    }
+
+    const mainSrc = `${hostname}/src/main.ts`;
+    const graphqlDomain = bcGraphqlDomain ?? 'mybigcommerce.com';
+
+    // Classic script so B3 config is available before the Vite modules evaluate.
+    const configScript = document.createElement('script');
+
+    configScript.id = 'b2b-config';
+    configScript.textContent = `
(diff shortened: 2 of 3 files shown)


parthshahp (approved):

CNanninga:
@bc-lulu-yang 

From the PR description, it sounds like this is an issue only with a locally running Buyer Portal, but I'd just recommend  scrutinizing whether any changes to the production loader are appropriate, for consistency.

Also please do make sure to review this official guide we have in the docs: https://docs.bigcommerce.com/developer/docs/storefront/catalyst/experiments/b2b-edition

There are instructions there for setting up a _separate_ loader for a deployed custom Buyer Portal.  So any time changes are being made to the way the BP scripts are injected, I just ask that there's awareness/guidance on whether/how that doc is affected.

I can't get into the weeds of the PR's changes, so I don't know whether there are actually implications for these areas, but just making sure there's visibility!

bc-lulu-yang:
Thanks @CNanninga for the context. That's a really good point.

Hi @parthshahp, thanks for the review. After thinking it over, I think it's better to leave verification (and any fixes, if needed) to your team. I'm going to close this PR since my current knowledge of the Catalyst platform isn't enough for me to confidently cover and test everything.

PR #3163, opened 2026-08-04
TRAC-1395: fix - treat consent as granted when the store has cookie consent disabled

parthshahp:
## What/Why?

When `cookieConsentEnabled` (setting in the control panel) is off, c15t grants every consent category client-side but only in its in-memory store, so no `c15t-consent` cookie is ever written. `hasConsentFor` read the cookie as the sole source of truth and so returned false for every category on those stores.

`hasConsentFor` now falls back to the store's cookie consent setting when no consent cookie is present. An explicit consent cookie still takes precedence.

Also seeds an explicit functionality-consent cookie in the two alternate currency e2e tests so they assert the consented path rather than depending on the test store's `cookieConsentEnabled` setting, and moves the consent cookie builder duplicated in analytics-session.spec.ts into a shared test helper.

## Testing

Tested using local tests as well as CI.

Before, you can see that I must refresh for the currency to take effect:

https://github.com/user-attachments/assets/8eb4eb8d-fc60-44b6-82a7-699942b2593a

After:

https://github.com/user-attachments/assets/28a08921-6f5d-41f1-bca4-33ea833c0ab1

## Migration

No migration needed.

### DIFF
--- a/core/components/consent-manager/persist-auto-granted-consent.tsx
+++ b/core/components/consent-manager/persist-auto-granted-consent.tsx
@@ -1,0 +1,30 @@
+'use client';
+
+import { useConsentManager } from '@c15t/nextjs/client';
+import { useEffect } from 'react';
+
+import { getConsentCookie } from '~/lib/consent-manager/cookies/client';
+
+interface Props {
+  isCookieConsentEnabled: boolean;
+}
+
+// With consent disabled, c15t's initial state grants every category but never persists
+// it, so `hasConsentFor` sees no cookie. saveConsents writes to storage regardless of
+// that setting, keeping the cookie the single source of truth on both sides.
+//
+// `StartVisitOnConsent` waits on the cookie written here to start the first visit of the
+// session, since the analytics proxy couldn't see consent when it handled this request.
+export function PersistAutoGrantedConsent({ isCookieConsentEnabled }: Props) {
+  const { saveConsents } = useConsentManager();
+
+  useEffect(() => {
+    if (isCookieConsentEnabled || getConsentCookie()) {
+      return;
+    }
+
+    saveConsents('all');
+  }, [isCookieConsentEnabled, saveConsents]);
+
+  return null;
+}
--- a/core/components/consent-manager/start-visit-on-consent.tsx
+++ b/core/components/consent-manager/start-visit-on-consent.tsx
@@ -10,22 +10,41 @@ import { startVisit } from './_actions/start-visit';
 export function StartVisitOnConsent() {
   const { consents, hasConsented } = useConsentManager();
   const isMeasurementGranted = hasConsented() && consents.measurement;
-  const wasMeasurementGranted = useRef(isMeasurementGranted);
+
+  // True once this session's visit is accounted for: either the analytics proxy already
+  // saw consent when it handled this request, or we dispatched startVisit ourselves.
+  // Seeded from the consent cookie rather than c15t's in-memory state, since the cookie is
+  // what the proxy read. On stores with cookie consent disabled, c15t grants every category
+  // while constructing its store — before the first render — so an in-memory seed would
+  // treat that grant as pre-existing and never dispatch.
+  const hasStartedVisit = useRef<boolean | null>(null);
+
+  if (hasStartedVisit.current === null) {
+    hasStartedVisit.current = getConsentCookie()?.['c.measurement'] ?? false;
+  }
 
   useEffect(() => {
-    const shouldStartVisit = isMeasurementGranted && !wasMeasurementGranted.current;
+    if (!isMeasurementGranted) {
+      // Consent withdrawn: the proxy clears the analytics cookies, so a later re-grant
+      // needs to start a fresh visit.
+      hasStartedVisit.current = false;
 
-    wasMeasurementGranted.current = isMeasurementGranted;
-
-    if (!shouldStartVisit) {
       return;
     }
 
-    // c15t updates its React state before persisting the consent cookie. The
-    // startVisit action validates consent from the cookie server-side, so poll
-    // document.cookie directly — only dispatch once the cookie is readable.
+    if (hasStartedVisit.current) {
+      return;
+    }
+
+    // c15t updates its React state before persisting the consent cookie. The startVisit
+    // action validates consent from the cookie server-side, so poll document.cookie
+    // directly — only dispatch once the cookie is readable. The flag flips on dispatch
+    // rather than up front, so a poll cancelled before it succeeds (StrictMode remounts
+    // effects in development) is retried rather than treated as already done.
     const tryStartVisit = () => {
       if (getConsentCookie()?.['c.measurement']) {
+        hasStartedVisit.current = true;
+
         // eslint-disable-next-line no-console
         void startVisit().catch(console.error);
 
--- a/core/lib/consent-manager/has-consent-for.ts
+++ b/core/lib/consent-manager/has-consent-for.ts
@@ -3,10 +3,8 @@ import { getConsentCookie } from './cookies/server';
 type ConsentCategory = 'functionality' | 'marketing' | 'measurement';
 
 // Server-side check for whether cookies of a given consent category may be stored.
-// The shopper's consent cookie is the source of truth. Absent a consent cookie,
-// treats consent as not given — the client-side consent manager will write the
-// cookie once the shopper decides (or c15t auto-grants when consent is disabled),
-// and startVisit handles recording the visit at that point.
+// The consent cookie is the source of truth: the consent manager writes it when the
+// shopper decides, or PersistAutoGrantedConsent writes it on stores that don't gate.
 export async function hasConsentFor(category: ConsentCategory) {
   const consent = await getConsentCookie();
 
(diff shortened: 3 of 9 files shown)


core/lib/consent-manager/has-consent-for.ts
```diff
@@ -1,18 +1,22 @@
...
+// so no cookie is ever written.
 export async function hasConsentFor(category: ConsentCategory) {
   const consent = await getConsentCookie();
 
   if (consent) {
     return consent[`c.${category}`];
   }
 
-  return false;
+  return !(await isCookieConsentEnabled());
```
parthshahp: This technically introduces another fetch in the middleware; I think the cost is very minimal since we already have other fetches but definitely something to point out.

core/tests/ui/e2e/compare.spec.ts
```diff
@@ -36,8 +37,18 @@ test('Validate compare page', async ({ page, catalog, currency }) => {
...
 
-test('Validate compare page with alternate currency', async ({ page, catalog, currency }) => {
+test('Validate compare page with alternate currency', async ({
+  page,
+  context,
+  catalog,
+  currency,
+}) => {
   const format = getFormatter();
+
```
parthshahp, line 47: This is the actual e2e test fix: we can just set the cookie explicitly for the test that we need

core/components/consent-manager/persist-auto-granted-consent.tsx
```diff
@@ -0,0 +1,36 @@
...
+ */
+export function PersistAutoGrantedConsent({ isCookieConsentEnabled }: Props) {
+  const { saveConsents } = useConsentManager();
+
+  useEffect(() => {
+    if (isCookieConsentEnabled || getConsentCookie()) {
+      return;
+    }
+
+    saveConsents('all');
```
parthshahp, line 26: If cookie consent is disabled in the control panel, act like the user "accepts all".

parthshahp: On second thought; this might not be such a good idea: https://c15t.com/docs/shared/react/hooks/use-consent-manager/setting-consent

From the documentation, it looks like `saveConsents` stores the value in localstorage. If a merchant decides to later enable cookie consent, this "fake consent" may still be stored in localstorage for the user.

(Edit: this is a rare case we needn't account for)

core/lib/consent-manager/auto-granted-consent.ts
parthshahp, line 1: This is what I came up with to resolve the issue of:

1. Consent cookies are disabled.
2. Shopper goes to site and consent is granted for `all` scope.
3. Consent cookies are enabled.
4. Shopper goes to site, but does not see banner because consent is granted for `all` scope still.

I'm not sure if this is a real scenario that we should consider or not.

Another solution here would be to just call `isCookieConsentEnabled` on the set currency code server action. We can also cache this request, and we need to remember to call it for any server action that may also depend on consent cookies (like analytics).

parthshahp: Thinking more about it I don't really like this `localstorage` solution, and we should just bite the bullet and make sure to check the consent cookie setting on relevant server actions, but I'll leave this up unchanged to get feedback first.

chanceaclark: I don't think that is a scenario we really need to worry about. Definitely an edge-case, because once a merchant enables it, the consent cookies will expire and will require consent after some time.

chanceaclark: We store cookie consent both in localstorage and in cookies. You'll need to dive into what c15t does, but we made sure we designed it that way to integrate better with Next.js.

Edit: technically, this method is not GDPR or CCPA compliant but we document this here: https://docs.bigcommerce.com/developer/docs/storefront/catalyst/features/cookie-consent

parthshahp: Ok, so then should I just go with 
1. the `saveConsents('all');` approach (if cookie consent is enabled)
2. make a request on server action specifically for set currency code?

core/components/consent-manager/persist-auto-granted-consent.tsx
```diff
@@ -0,0 +1,27 @@
...
+// that setting, keeping the cookie the single source of truth on both sides.
+export function PersistAutoGrantedConsent({ isCookieConsentEnabled }: Props) {
+  const { saveConsents } = useConsentManager();
+
+  useEffect(() => {
+    if (isCookieConsentEnabled || getConsentCookie()) {
+      return;
+    }
+
+    saveConsents('all');
```
chanceaclark, line 26: Heads up that `StartVisitOnConsent` won't call `startVisit` for this path. c15t grants synchronously when `enabled: false`, so `wasMeasurementGranted` over there initializes to `true` — there's no `false → true` transition to observe. The analytics recovery for first-load is deferred to the second request once this cookie is readable. I think that's fine, mind adding a note to the comment above so it's not mysterious later?

core/components/consent-manager/persist-auto-granted-consent.tsx
```diff
@@ -0,0 +1,27 @@
...
+  }, [isCookieConsentEnabled, saveConsents]);
```
chanceaclark, line 27: Is `saveConsents` referentially stable out of `useConsentManager`? If it's a new ref each render, this effect re-runs until the cookie exists and we could be calling `saveConsents` multiple times in a tight window before `getConsentCookie()` has a chance to return something. Probably fine if c15t handles idempotency here, but worth confirming.

parthshahp: https://github.com/c15t/c15t/blob/4e7ba5f4d1da27b9227cda645eedd0ce5bb9db8c/packages/core/src/store.ts#L473-L480

Seems like it's the same ref each render, it's in this zustand store.

core/components/consent-manager/persist-auto-granted-consent.tsx
parthshahp: Sure, updated the comment!

core/lib/consent-manager/auto-granted-consent.ts
chanceaclark: >make a request on server action specifically for set currency code?

What is this for?

core/components/consent-manager/persist-auto-granted-consent.tsx
chanceaclark: Well... no it actually shouldn't be a comment. We need to actually make sure it fires whenever cookie consent is turned off.

core/components/consent-manager/start-visit-on-consent.tsx
```diff
@@ -10,22 +10,41 @@ import { startVisit } from './_actions/start-visit';
...
+  // True once this session's visit is accounted for: either the analytics proxy already
+  // saw consent when it handled this request, or we dispatched startVisit ourselves.
+  // Seeded from the consent cookie rather than c15t's in-memory state, since the cookie is
+  // what the proxy read. On stores with cookie consent disabled, c15t grants every category
+  // while constructing its store — before the first render — so an in-memory seed would
+  // treat that grant as pre-existing and never dispatch.
+  const hasStartedVisit = useRef<boolean | null>(null);
+
+  if (hasStartedVisit.current === null) {
+    hasStartedVisit.current = getConsentCookie()?.['c.measurement'] ?? false;
```
parthshahp, line 23: `hasStartedVisit` depends on if the cookie exists. We've removed the `wasMeasurementGranted` ref since that only starts the visit on a `false -> true` change.

core/components/consent-manager/persist-auto-granted-consent.tsx
parthshahp: Answered with this change: https://github.com/bigcommerce/catalyst/pull/3163#discussion_r3724180988

core/lib/consent-manager/auto-granted-consent.ts
parthshahp: The specific issue regarding currency code is this:

On the set currency code server action, we check the consent cookie.

Before this PR, we weren't saving a consent cookie even when the consent cookies were disabled, meaning that the currency code wasn't able to be changed. https://github.com/bigcommerce/catalyst/blob/7848623eb5d8a30bd7308883fa8957422a5cdef7/core/lib/currency.ts#L25

The current PR's solution to this is to just set the cookie with 'all' scopes when we see that consent cookies are disabled in the control panel.

The alternative I was suggesting to my above solution is: we make an extra request inside of the set currency code server action to check the consent cookies control panel setting on our own (create a new graphql request). If it is set to false, we don't care about the consent cookie.

The issue with the alternative is that it doesn't fix the analytics case (doesn't start visit + deletes visits). https://github.com/bigcommerce/catalyst/blob/7848623eb5d8a30bd7308883fa8957422a5cdef7/core/proxies/with-analytics-cookies.ts#L21

core/components/consent-manager/persist-auto-granted-consent.tsx
jorgemoya: Are we planning to do something about this?

parthshahp: Nope, we probably don't really need to account for this. The consent cookie has expiration anyway

mfaris9:
LGTM, but should we run this by legal for this scenario? It might be overkill, but I would hate to cause a GDPR issue:

Day 1: consent disabled  -> shopper visits, cookie says "accepted everything"
Day 2: merchant enables consent in the control panel
Day 3: same shopper returns -> cookie already says "accepted", so
        no banner shows and they're tracked as if they opted in

parthshahp:
There's already a comment for this in the PR. TLDR this is an edge case and we have an expiration on the cookie anyway

PR #3174, opened 2026-08-13
TRAC-238: Update PDP metadata images to use correct image url

jordanarldt:
## What/Why?

`ProductPageMetadataQuery` requested `urlTemplate`, which returns a URL containing a literal `{:size}` placeholder. `<Image>` expands that through the CDN loader at render time, but `generateMetadata` uses the value verbatim — so `og:image` pointed at an unfetchable URL.

`url(width: 1200)` returns a concrete URL instead. Height is omitted deliberately: the stencil resizer fits the image inside the given box rather than cropping, so asking for a social aspect ratio like `1200x630` would just return a smaller square image for a square product photo.

## Testing

View source on a PDP and confirm `og:image` contains `/stencil/1200w/` rather than `/stencil/{:size}/`, and that the URL loads.

## Migration

In `core/app/[locale]/(default)/product/[slug]/page-data.ts`, update the `defaultImage` selection in `ProductPageMetadataQuery`:

```diff
  defaultImage {
    altText
-   url: urlTemplate(lossy: true)
+   url(width: 1200, lossy: true)
  }
```

Also included as a `## Migration` section in the changeset.

[TRAC-238]: https://bigcommercecloud.atlassian.net/browse/TRAC-238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core/app/[locale]/(default)/product/[slug]/page-data.ts
+++ b/core/app/[locale]/(default)/product/[slug]/page-data.ts
@@ -141,7 +141,7 @@ const ProductPageMetadataQuery = graphql(`
         path
         defaultImage {
           altText
-          url: urlTemplate(lossy: true)
+          url(width: 1200, lossy: true)
         }
         seo {
           pageTitle
--- a/.changeset/trac-238-pdp-metadata-image-url.md
+++ b/.changeset/trac-238-pdp-metadata-image-url.md
@@ -1,0 +1,19 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Fix the product page `og:image` tag pointing at an unfetchable URL. `ProductPageMetadataQuery` requested `urlTemplate`, which returns a URL containing a literal `{:size}` placeholder that the `<Image>` CDN loader substitutes at render time. `generateMetadata` has no such loader, so the placeholder was emitted verbatim into the Open Graph tag.
+
+## Migration
+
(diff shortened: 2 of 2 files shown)


core/app/[locale]/(default)/product/[slug]/page-data.ts
```diff
@@ -141,7 +141,7 @@ const ProductPageMetadataQuery = graphql(`
         path
         defaultImage {
           altText
-          url: urlTemplate(lossy: true)
+          url(width: 1200, lossy: true)
```
Comment (chanceaclark) [inline core/app/[locale]/(default)/product/[slug]/page-data.ts:144]: Is 1200 a little too big? Can we request it at a smaller url?

Also, won't this affect the page query as well? I can't remember if the product image carousel utilizes this image?

Comment (jordanarldt) [reply core/app/[locale]/(default)/product/[slug]/page-data.ts:144]: @chanceaclark 
> Also, won't this affect the page query as well? I can't remember if the product image carousel utilizes this image?

No, it's only used in the `getProductPageMetadata` query so it's strictly used for the metadata.

> Is 1200 a little too big? Can we request it at a smaller url?

The reason I went with 1200 is because twitter/X's `summary_large_image` metadata field suggests a width of 1200px, but I can reduce it if you think that's necessary

Comment (chanceaclark) [reply core/app/[locale]/(default)/product/[slug]/page-data.ts:144]: Sounds good!

chanceaclark (approved):

PR #3176, opened 2026-08-18
LTRAC-1583: Add a unit test harness to core and expire in-memory KV entries

jorgemoya:
Linear: [LTRAC-1583](https://linear.app/commerce/issue/LTRAC-1583/in-memory-kv-layer-never-expires-so-isolates-never-pick-up-each-others)

Two commits: the test harness `core` has never had, and the fix it exists to cover. Base for #3170.

## What/Why?

`lib/kv` keeps a per-process `MemoryKvAdapter` in front of whichever shared adapter is selected, and `KV.mget` skips the shared store whenever memory holds every requested key. Nothing ever expired those entries, so that short-circuit was permanent — once a process had seen a key it never consulted the shared store again, could not observe a refresh another process had written, and re-ran the refresh itself.

Caught on a deployed store:

- `01:19:01` — `storeStatus` written, window ends `01:24:01`
- `01:21:12` — a request reads the **previous** value and revalidates again

That read never reached the shared store despite a fresher value having sat there for two minutes. A later trace showed a route entry served from memory **14 hours** stale.

**Nothing was ever served incorrectly.** Staleness stayed bounded by the caller's own `expiryTime`, and stale-while-revalidate still served fast and refreshed behind the request. The cost was duplicated work — origin fetches and cache writes scaling with process count rather than being shared.

## The two values

**60s TTL** (was 30s) — matches Workers KV's floor for `cacheTtl` on a read, so the layers share one staleness window rather than one each. It sits inside the shortest window callers embed in their values (5 min for storefront status, 30 for routes).

**4096 entries** (was 500) — cache keys include the query string, so distinct keys accumulate much faster than the number of real paths suggests; a crawler walking `?utm_*` permutations alone can churn through the old limit.

## The trade, stated plainly

This adds a shared-store read per key per window that we previously didn't do, in exchange for origin fetches and cache writes no longer multiplying by process count. Storefront API calls are the scarce resource and KV reads are cheap, so it's favourable — but it isn't free.

It also pairs with a follow-up: `CloudflareKvAdapter` passing `cacheTtl` on the read would serve those extra reads from the colo cache, shared across every isolate in a data centre. That belongs in #3170, since this layer also fronts Upstash and Vercel. Worth noting the two only work together — `cacheTtl` is inert if memory never expires, because the read never happens.

## Considered and not adopted

**`allowStale` + `noDeleteOnStaleGet`** would serve an expired entry instantly and refresh it off the response path, which is nicer than a blocking read. It needs somewhere to run the refresh: `KV.mget` is called from `with-routes.ts:262` without the request's event and has no `waitUntil` handle. Returning stale values with no refresh would reinstate the exact bug this fixes.

**Single-flight on revalidation** — two concurrent requests observing the same stale entry currently fire two origin fetches and two KV writes, seen twice in traces including two writes to the same route key 207ms apart. That revalidation lives in `with-routes`, not here.

Also unchanged: `KV.mget` filters memory values with `.filter(Boolean)`, so a legitimately cached `null` — a path with no route — reads as a miss and re-hits the shared store.

## Testing

`vitest run` — 10 pass. Lint and `tsc --noEmit` clean.

`memory.spec.ts` covers the round trip, key ordering across a mixed hit/miss batch, `null` round-tripping, expiry at the default window, expiry inside the callers' shortest window, `ex` overriding the default, the window refreshing on rewrite, eviction at capacity, and that capacity now exceeds the old 500.

**Verified both changed values fail their specs when reverted** — removing the TTL breaks two specs, restoring `max: 500` breaks the capacity spec. Two things made that non-obvious, both noted at their call sites:

- `lru-cache` captures a reference to the `performance` object at module load. Vitest's fake timers replace the global with a new object that reference never sees, so the cache's clock has to be moved by stubbing the method in place.
- `lru-cache` treats a recorded start time of exactly `0` as "no TTL". The fake clock therefore starts at a non-zero baseline — starting at zero makes every expiry assertion pass whether or not the TTL works, which is how this nearly shipped green.

## Migration

None. No API change; the only behavioural difference is that a process re-reads the shared store for a key it has held for more than 60 seconds.

### DIFF
--- a/.changeset/ltrac-1583-memory-kv-ttl.md
+++ b/.changeset/ltrac-1583-memory-kv-ttl.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Expire entries in the in-memory KV layer after 60 seconds. `lib/kv` keeps a per-process `MemoryKvAdapter` in front of whichever shared adapter is selected (Cloudflare KV, Upstash, Vercel Runtime Cache) and skips the shared store whenever memory holds every requested key. Those entries never expired, so once a process had seen a key it stopped consulting the shared store for it entirely. Refreshes were then driven solely by the `expiryTime` each caller embeds in the cached value — and that path refetches from the **origin**, not from the shared store. So every process independently refetched on a clock starting from whenever it first cached the key, rather than picking up a value another process had already fetched and shared. Cached data was never wrong, but origin requests and cache writes scaled with process count. Capacity is also raised from 500 to 4096 entries, since cache keys include the query string and so accumulate faster than the number of real paths suggests.
--- a/core/lib/kv/adapters/memory.spec.ts
+++ b/core/lib/kv/adapters/memory.spec.ts
@@ -1,0 +1,128 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
+
+import { MemoryKvAdapter } from './memory';
+
+describe('MemoryKvAdapter', () => {
+  // `lru-cache` captures a reference to `performance` at module load. Vitest's
+  // fake timers replace the global with a new object that reference never sees,
+  // so the cache's clock has to be moved by stubbing the method in place.
+  // `lru-cache` records an entry's start time as `perf.now()` and treats a
+  // recorded start of exactly 0 as "no TTL set", so the clock must not start at
+  // zero — expiry would be silently disabled and every assertion below would
+  // pass regardless.
+  const CLOCK_BASE_MS = 1_000_000;
+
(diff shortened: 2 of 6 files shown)


chanceaclark (commented):
Before I review this any further, can you consult with our `platform-dispatch-router` (internal) and see if there is anything we can borrow from that in-memory cache, including TTLs and whatnot.

jorgemoya:
> Before I review this any further, can you consult with our `platform-dispatch-router` (internal) and see if there is anything we can borrow from that in-memory cache, including TTLs and whatnot.

Consulted and added what made sense.

chanceaclark (approved):

PR #3182, opened 2026-08-19
TRAC-1395: fix - treat disabled cookie consent as implicit consent server-side

parthshahp:
## What/Why?

Implicitly grant consent on server side requests when cookieConsent setting is disabled in the control panel.

## Testing

Revert this PR.

## Migration

This PR is just for sharing right now.


chanceaclark (approved):
I'm good with this @parthshahp

PR #3184, opened 2026-08-21
LTRAC-1462: Ship _headers so static assets are cached immutably

jorgemoya:
Linear: [LTRAC-1462](https://linear.app/commerce/issue/LTRAC-1462)

## What/Why?

`packages/catalyst/templates/public_headers` has always specified the right policy for `/_next/static/*`, but nothing ever copied it into the build output. `build.ts` wrote only `open-next.config.ts` and `wrangler.jsonc`, and no `_headers` file existed anywhere in the repo.

Without it, Workers Assets applies its own default. Measured on a deployed store — **all 32** hashed assets on a product page:

```
cache-control: public, max-age=0, must-revalidate
etag: "ea54572fc7599bec1529f18fae07fab9"
```

`max-age=0, must-revalidate` is wrong by construction for a **content-hashed** filename like `2080-_emu8td2.css`. The hash *is* the version, so a given URL can never return different bytes — yet browsers were forbidden from using a cached copy without asking first.

### Measured cost

| | |
| -- | -- |
| Hashed assets per product page | 32 (25 js, 4 css, 3 woff2) |
| One revalidation round trip | ~75-79ms → `304 Not Modified` |
| All 32 revalidated in parallel | 0.70-0.84s |

Every repeat page view issued 32 conditional requests, all answered "yes, use the copy you already had." HTTP/2 multiplexes them so it is not 32×75ms serially, but it is still several round trips plus 32 requests of overhead — and 4 of those CSS files and 3 fonts are **render-blocking**, so it delayed first paint on every repeat visit.

(The 0.70-0.84s figure is from a script that fetches each etag first, which a real browser already has cached — so treat the true repeat-visit cost as roughly half, ~350-400ms.)

With the intended `public,max-age=31536000,immutable`, all 32 requests disappear for a year.

### Why the header was missing at all

On Workers, `/_next/static/*` never reaches the Next.js server — Cloudflare's asset layer serves it directly from `.open-next/assets`, so the immutable header Next would normally set never applies. `_headers` is the supported override, and OpenNext's own `migrate` command generates a byte-identical `public/_headers`, treating it as the application's responsibility rather than something the adapter handles.

## Testing

`build.spec.ts` — 6/6 pass.

Added a regression test asserting the copy happens, and **verified it is a real guard**: removing the `copyFile` makes it fail, restoring it makes it pass. Worth having because the original bug was an *absent* value rather than a wrong one, with nothing asserting it — which is exactly how a correct template sat unused.

**Not yet verified on a deployed store.** After deploy the check is binary — re-request the same assets and confirm every one returns `public,max-age=31536000,immutable` instead of `max-age=0, must-revalidate`. At that point the revalidation benchmark above becomes moot, because a browser will not issue the requests at all.

## Migration

None. Build-output change only; no API surface, no file moves.

Placement note for reviewers: the copy runs **after** the OpenNext build, because that build regenerates `.open-next/assets` and would clobber anything written earlier, and **before** the Wrangler dry-run so Wrangler validates the file. The existing recursive copy into `.bigcommerce/dist/assets` carries it into the uploaded bundle.

Refs LTRAC-1462

### DIFF
--- a/.changeset/ltrac-1462-static-asset-immutable-headers.md
+++ b/.changeset/ltrac-1462-static-asset-immutable-headers.md
@@ -1,0 +1,22 @@
+---
+"@bigcommerce/catalyst": patch
+---
+
+Ship the `_headers` file so hashed static assets get an immutable `Cache-Control`, instead of being revalidated on every repeat page view.
+
+`packages/catalyst/templates/public_headers` has always specified the right policy for `/_next/static/*`, but nothing ever copied it into the build output — `build.ts` wrote only `open-next.config.ts` and `wrangler.jsonc`, and no `_headers` file existed anywhere in the repo.
+
+Without it, Workers Assets serves those files with its own default. Measured on a deployed store, all 32 hashed assets on a product page returned:
+
+```
+cache-control: public, max-age=0, must-revalidate
+```
+
+`max-age=0, must-revalidate` on a **content-hashed** filename is wrong by construction: the hash *is* the version, so a given URL can never return different bytes. The header forced browsers to issue a conditional request for all 32 assets on every repeat view (~75ms each, all answered `304 Not Modified`), including 4 render-blocking CSS files and 3 fonts. With the intended `public,max-age=31536000,immutable`, those requests disappear entirely.
+
+The reason the header was missing at all is that `/_next/static/*` never reaches the Next.js server on Workers — Cloudflare's asset layer serves it directly, so Next's own immutable header never applies. `_headers` is the supported override, and OpenNext's `migrate` command generates a byte-identical `public/_headers` for exactly this reason, treating it as the app's responsibility.
+
+## What changed
+
+- `build.ts` now copies `templates/public_headers` to `.open-next/assets/_headers`. It is written *after* the OpenNext build, because that build regenerates the assets directory, and *before* the Wrangler dry-run so Wrangler validates it. The existing recursive copy into `.bigcommerce/dist/assets` carries it into the uploaded bundle.
+- Added a regression test asserting the copy happens. The original bug was not a wrong value but an absent one, with nothing asserting it — verified the new test fails when the copy is removed.
--- a/packages/catalyst/src/cli/commands/build.spec.ts
+++ b/packages/catalyst/src/cli/commands/build.spec.ts
@@ -1,5 +1,7 @@
 import { Command } from 'commander';
 import { execa } from 'execa';
+import { copyFile } from 'node:fs/promises';
+import { join } from 'node:path';
 import { afterEach, beforeAll, beforeEach, expect, test, vi } from 'vitest';
 
 import { consola } from '../lib/logger';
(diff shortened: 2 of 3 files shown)


jorgemoya:
## Deployed and verified — this change is necessary but **not sufficient** on its own

Deployed to a native-hosting store on 2026-08-21. The `_headers` file ships correctly, but the header policy **did not take effect**. All 32 hashed assets still return the Workers Assets default:

```
cache-control: public, max-age=0, must-revalidate
```

### Why

The file is being served as a *static asset* rather than parsed as configuration:

```
GET /_headers
200  content-type: application/octet-stream  content-length: 66

/_next/static/*
  Cache-Control: public,max-age=31536000,immutable
```

Per [Cloudflare's docs](<https://developers.cloudflare.com/workers/static-assets/headers/>), `_headers` is never served as an asset when it is recognised — so it is not being recognised. Confirmed the deploy did include this change (that response body is the template verbatim), and confirmed the default it should be overriding is exactly `public, max-age=0, must-revalidate`.

The cause is the upload path. `_headers` is consumed during the **asset upload session** — Wrangler strips it from the manifest and submits it as configuration. Here Wrangler only runs `--dry-run`; **ignition performs the real upload** via the Cloudflare API, and its asset uploader (`pkg/cloudflare/upload/assets`) has no `_headers` handling, so the file is uploaded as an ordinary asset and no header config is ever registered.

The API does support it — `cloudflare-go v6.10.0`:

```go
type DispatchNamespaceScriptUpdateParamsMetadataAssetsConfig struct {
	// The contents of a \_headers file (used to attach custom headers on asset responses).
	Headers param.Field[string] `json:"_headers"`
	...
}
```

Ignition's `NewMetadata` currently sets only `Assets: { JWT: completionToken }`, with no `Config`, so there is nowhere for the policy to land.

### What this needs

| Half | Status |
| -- | -- |
| Catalyst — produce the `_headers` content | this PR |
| Ignition — read it, pass as `Assets.Config.Headers`, exclude from the uploaded manifest | separate change, in progress |

**Keeping this as a draft until the ignition half lands**, since merging it alone ships a file nothing reads — and makes `/_headers` publicly fetchable as a stray 66-byte asset, which the ignition change should strip.

### Worth noting for <issue id="LTRAC-1464" href="https://linear.app/commerce/issue/LTRAC-1464">LTRAC-1464</issue>

This is the **third** instance of the same drift pattern: Catalyst emits something the ignition deploy path does not consume. The others are the missing `WORKER_SELF_REFERENCE` service binding (<issue id="LTRAC-1457" href="https://linear.app/commerce/issue/LTRAC-1457">LTRAC-1457</issue>) and the missing `IMAGES` binding (15 `env.IMAGES binding is not defined` warnings in production). A contract test across the two repos would have caught all three.

jorgemoya:
## Verified working in production — resolves the blocker noted above

The ignition half ([ignition#331](https://github.com/bigcommerce/ignition/pull/331)) is now written and both halves were deployed to a native-hosting store on 2026-08-21. Re-running the benchmark captured before the change: **all 32 hashed assets on a product page flipped.**

| | before | after |
| -- | -- | -- |
| `cache-control` (all 32 assets) | `public, max-age=0, must-revalidate` | **`public,max-age=31536000,immutable`** |
| `GET /_headers` | `200`, 66 bytes served as a file | **`404`** |
| new warn/error log entries | — | **0** |

```
GET /_next/static/chunks/2080-_emu8td2.css
200
cache-control: public,max-age=31536000,immutable
```

The `/_headers` 404 confirms the earlier diagnosis was right: the file was being uploaded as an ordinary asset, and the missing piece was ignition consuming it as `assets.config._headers` and excluding it from the manifest. This PR's contribution — emitting the file — was correct; it simply had no consumer.

### Reading the timing numbers correctly

```
parallel revalidation of all 32 assets
  before: 0.863s / 0.841s / 0.893s
  after:  0.881s / 0.666s / 0.665s
```

**Not the win, and the similarity is expected.** The benchmark forces conditional requests via an explicit `If-None-Match`, measuring what revalidation costs *if performed*. Under `immutable` a browser issues none — the real after-value is **zero requests**.

The saving is the elimination of 32 conditional requests at ~75ms each (all previously `304 Not Modified`), including 4 render-blocking CSS files and 3 fonts: **~350-400ms off repeat page views**.

jordanarldt (approved):

PR #3193, opened 2026-08-25
LTRAC-1457: Revalidate via self-fetch instead of a Durable Object queue

jorgemoya:
Linear: [LTRAC-1457](https://linear.app/commerce/issue/LTRAC-1457)

Replaces the ignition-side approach, which was tried and proven impossible — see [ignition#334](https://github.com/bigcommerce/ignition/pull/334) (closed).

## What/Why?

OpenNext's `doQueue` routes ISR revalidation through a Durable Object whose constructor reads `env.WORKER_SELF_REFERENCE` and throws without it:

```js
this.service = env.WORKER_SELF_REFERENCE;
if (!this.service)
    throw new IgnorableError("No service binding for cache revalidation worker");
```

`queueCache.send()` catches that and only logs, so revalidation fails **silently** rather than surfacing.

### The binding cannot exist on native hosting

This was tested, not assumed. A Cloudflare `service` binding resolves against **account-level Workers**, and a Catalyst deployment is a script inside a dispatch namespace, which is not addressable that way. Emitting the binding from ignition was rejected at upload:

```
PUT .../dispatch/namespaces/ignition-integration-us/scripts/16693302-…
400 Bad Request

code 10143
"Service binding 'WORKER_SELF_REFERENCE' references Worker
 '16693302-5d818746-…' which was not found."
```

No value for `service` would work — it is a platform constraint.

**The `dispatch_namespace` binding often suggested as the alternative is worse.** Beyond the API mismatch (OpenNext calls `.fetch()` directly; that binding exposes `.get(name)`), `.get()` accepts *any* script in the namespace — binding it into a tenant Worker would let any deployment invoke any other deployment's Worker. Categorically worse than the problem.

### The fix

Revalidation never needed a binding. It is a `HEAD` request to the page's own public URL carrying the build-time preview secret — exactly what the Durable Object issues once it holds the service handle. This issues it with a plain `fetch`, which leaves and re-enters through `platform-dispatch-router` and arrives at the same Worker. `global_fetch_strictly_public` is already set, so the subrequest is not short-circuited internally.

Still wrapped in `queueCache`, so concurrent stale hits for one path still collapse into a single revalidation.

**Trade-off, stated plainly:** the Durable Object's retry and max-concurrency handling is lost. A failed revalidation is retried on the next stale hit rather than by the queue, and revalidations are no longer capped at a concurrency limit. Accepted because the alternative is not "keep the Durable Object" — it is "revalidation does not work at all."

## Rollout/Rollback

**Latent — no behaviour change for existing stores.** The build currently produces **zero routes with a revalidate window**, so the queue is never invoked. This makes revalidation work for when ISR is adopted rather than fixing an active failure.

`NEXT_CACHE_DO_QUEUE` is intentionally left bound on both sides. OpenNext's worker template exports all three Durable Object classes unconditionally, so the binding still resolves and **no Durable Object migration is needed** — deleting the class would have destroyed its storage. It simply goes inert.

Rollback is reverting this commit.

## Testing

Type-checked against the real build context (`core/.bigcommerce`, where the template is compiled), including a deliberate-error probe to confirm the check actually validates and that the queue object satisfies OpenNext's `Queue` interface structurally.

**Not verifiable through normal traffic**, because nothing exercises the queue while there are no ISR routes. Confirming it end-to-end needs a temporary route exporting `revalidate` to create a real ISR entry, then observing that a stale hit triggers a regeneration rather than the `No service binding…` error. Happy to add that probe if reviewers want it before merge.

Refs LTRAC-1457

### DIFF
--- a/.changeset/ltrac-1457-self-fetch-revalidation-queue.md
+++ b/.changeset/ltrac-1457-self-fetch-revalidation-queue.md
@@ -1,0 +1,37 @@
+---
+"@bigcommerce/catalyst": patch
+---
+
+Replace the Durable Object revalidation queue with a self-fetch queue, so ISR revalidation can work on native hosting at all.
+
+OpenNext's `doQueue` routes revalidation through a Durable Object whose constructor reads `env.WORKER_SELF_REFERENCE` and throws without it:
+
+```js
+this.service = env.WORKER_SELF_REFERENCE;
+if (!this.service)
+    throw new IgnorableError("No service binding for cache revalidation worker");
+```
+
+`queueCache.send()` catches that error and only logs, so revalidation would fail silently rather than surface.
+
+**That binding cannot exist on native hosting.** A Cloudflare `service` binding resolves against account-level Workers, and a Catalyst deployment is a script inside a dispatch namespace, which is not addressable that way. Adding it was attempted and rejected at upload:
+
+```
+400 Bad Request  code 10143
+Service binding 'WORKER_SELF_REFERENCE' references Worker
+'…' which was not found.
+```
+
+The `dispatch_namespace` binding sometimes suggested as the alternative is worse: OpenNext calls `.fetch()` directly on the value while that binding exposes `.get(name)`, and `.get()` accepts *any* script in the namespace — binding it into a tenant Worker would let any deployment invoke any other deployment's Worker.
+
+## What changed
+
+Revalidation does not require a binding. It is a `HEAD` request to the page's own public URL carrying the build-time preview secret — exactly what the Durable Object issues once it holds the service handle. `queue` now uses a small self-contained queue that issues that request with a plain `fetch`, which leaves and re-enters through the dispatch router and arrives at the same Worker. `global_fetch_strictly_public` is already set, so the subrequest is not short-circuited internally.
+
+It remains wrapped in `queueCache`, so concurrent stale hits for one path still collapse into a single revalidation.
+
+**Trade-off:** the Durable Object's retry and max-concurrency handling is lost. A failed revalidation is retried on the next stale hit rather than by the queue itself, and revalidations are no longer capped at a concurrency limit.
+
+The `NEXT_CACHE_DO_QUEUE` binding is intentionally left in place. OpenNext's worker template exports all three Durable Object classes unconditionally, so it still resolves and needs no Durable Object migration; it is simply inert.
+
+Note this is latent until ISR is adopted — the build currently produces no routes with a revalidate window, so the queue is never invoked.
--- a/.changeset/ltrac-1457-self-fetch-revalidation-queue.md
+++ b/.changeset/ltrac-1457-self-fetch-revalidation-queue.md
@@ -4,6 +4,10 @@
 
 Replace the Durable Object revalidation queue with a self-fetch queue, so ISR revalidation can work on native hosting at all.
 
+**This does not make anything faster, and changes no behavior today.** Catalyst currently ships no route with a revalidate window, so the queue is never invoked. From the built prerender manifest, every prerendered route is `initialRevalidateSeconds: false` and `dynamicRoutes` is empty. The `next: { revalidate }` options on the product and faceted-search queries are fetch-level data caching and do not feed this queue.
+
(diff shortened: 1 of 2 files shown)


jorgemoya:
## Correction: the alternatives were narrowed by security, not only by platform limits

The description above overstates the constraint. Correcting it, and recording the rejected options so they are not re-proposed as oversights.

**What was proven:** a `service` binding cannot target a **dispatch-namespace script**. Cloudflare's error is explicit — *"Verify the Worker exists in your account"* — so service bindings resolve against **account-level** Workers only. No value for `service` naming this deployment would work.

**What was not proven:** that no service binding can work at all. One would resolve if it pointed at an account-level Worker, and `platform-dispatch-router-*` is exactly that — it holds `dispatch_namespaces` bindings rather than living inside one. Since it routes purely on `new URL(c.req.url).hostname`, binding it would route straight back to the calling tenant:

```
env.WORKER_SELF_REFERENCE.fetch('https://<own-host>/page')
  -> dispatch router (account-level, resolves)
    -> routing KV lookup by hostname
      -> back to the same tenant Worker
```

That would have been **strictly better than this PR** — internal rather than a public round trip, and it keeps the Durable Object queue with its retry, max-concurrency, cross-region dedup and failed-route backoff.

**It was rejected deliberately.** It hands every tenant a handle able to reach any other tenant's Worker. Public fetches already reach those endpoints, so it is not new data access — but only through the Cloudflare edge. An internal handle bypasses WAF, caching and rate limiting, which makes it a usable amplification vector against another tenant. "No tenant holds a handle that can reach another tenant" is a cleaner invariant than re-deriving equivalence each time it comes up.

So the full picture is three options, two of which fail on tenant isolation rather than capability:

| Option | Resolves? | Rejected because |
| -- | -- | -- |
| `service` → own script | **no** — error 10143 | platform constraint |
| `dispatch_namespace` binding | yes | `.get()` reaches any script in the namespace; also wrong API shape |
| `service` → dispatch router | yes (untested) | tenant handle reaching any tenant, bypassing the edge |
| **self-fetch (this PR)** | n/a | — |

That is what makes the trade-off here acceptable: the comparison is not "this versus the Durable Object", it is "this versus no revalidation".

Also hardened since the first push: the fetch now carries a 10s timeout matching the Durable Object's default, and a non-ok response throws so a failed regeneration is logged by `queueCache` and retried on the next stale hit rather than silently recorded as done.

jorgemoya:
## Verified end to end on a deployed store

The concern raised against this approach was that OpenNext's docs say direct-style revalidation *"is intended for debugging purposes and is not recommended for use in production. It only works in preview mode (i.e. `wrangler dev`)."* That does not hold here, and it has now been demonstrated rather than argued.

Deployed 2026-08-26 with a temporary ISR route (`/api/isr-probe`, `revalidate = 60`) that returns its own generation timestamp. Catalyst produces no ISR routes of its own, so this path had **never once executed in production** — the probe exists to create a real ISR entry.

**Cycle 1**

| request | `x-nextjs-cache` | `generatedAt` |
| -- | -- | -- |
| 1 | `STALE` | 01:32:40 |
| 2 | `HIT` | **01:34:22** — regenerated |

**Cycle 2**, after waiting past the revalidate window

| request | `x-nextjs-cache` | `generatedAt` |
| -- | -- | -- |
| 1 | `STALE` | 01:34:22 |
| 2 | `STALE` | 01:34:22 — dedup held, no duplicate revalidation |
| 3 | `HIT` | **01:35:47** — regenerated |

Cycle 2 is the more informative one: a stale hit enqueues, `queueCache` suppresses a duplicate while the first revalidation is in flight, and the regenerated entry lands shortly after. That is the dedup behaviour the Durable Object provided, preserved through the wrapper.

**No swallowed failures.** Zero warn- or error-level log entries across the window. A failing queue would have surfaced `Error sending message to queue` from `queueCache`.

### Why it works here

`global_fetch_strictly_public` is set on these Workers, so the revalidation subrequest is forced out to the public edge and re-enters through `platform-dispatch-router`, arriving back at the same Worker. That is the mechanism the docs' preview-only caveat does not account for.

### Also established

* **ISR is viable on native hosting.** That was genuinely in doubt, since `doQueue` cannot work here (error 10143, see [ignition#334](https://github.com/bigcommerce/ignition/pull/334)) and the documented alternative is marked preview-only.
* This is the **first ISR entry ever created on this platform**, so the R2 incremental cache's ISR read/write path was exercised for the first time as well. It behaved correctly.

The probe lives on a separate branch (`jorgemoya/ltrac-1457-isr-probe-DO-NOT-MERGE`) and is not part of this PR. It should be deleted once this merges.

jairo-bc (approved):

PR #3194, opened 2026-08-26
TRAC-969: Create reusable preview deployment action

jordanarldt:
Jira: [TRAC-969](https://bigcommercecloud.atlassian.net/browse/TRAC-969)

## What/Why?

Adds `packages/deployment-preview-action`, a GitHub Action that deploys a preview of a pull request to BigCommerce native hosting and keeps the PR comments describing it accurate. `core/.github/` ships a ready-to-use workflow, so scaffolded storefronts get previews without writing one.

Three decisions worth knowing before reading the diff:

**All previews share one hosting project.** Native hosting assigns a hostname per *project*, not per deployment, so a second deploy replaces the first at the same URL. The project serves one PR at a time, which is why there is arbitration at all. The newest open PR deploys on push; any other takes the preview over by commenting `redeploy preview`. A PR that loses the preview gets its comment rewritten, so no thread advertises a URL that now serves someone else's code.

**Shipped as both an action and a reusable workflow.** A workflow can declare `concurrency`, `permissions` and `environment`; an action cannot. The concurrency group is load-bearing here — two deploys at once would race for the shared project — so the reusable wrapper is the recommended entry point and `core/` uses it.

**The comment path builds PR code with the store's deploy credentials.** Write-permission and same-repo checks run *before* checkout, so nothing untrusted is fetched, let alone executed by `pnpm install`. These live inside the action rather than in copy-pasteable YAML so they can't be omitted by mistake.

Unconfigured repositories skip silently rather than failing, so a freshly scaffolded storefront doesn't get a red X on its first PR.

## Testing

```bash
pnpm --filter @bigcommerce/deployment-preview-action test
```

43 tests against a stubbed BigCommerce and GitHub API — no credentials or network needed. They cover the deploy-eligibility rules, which comment updates notify vs. stay silent, and `action.yml` itself (every `run` block is syntax-checked, since nothing compiles it).

Verified end to end against a real store: auto-deploy on the newest PR, deferral on an older one, `redeploy preview` takeover, and the comment rewrite on the displaced PR.

## Migration

None for existing projects. Previews stay off until configured with a `PREVIEW_DEPLOYMENT_PROJECT_UUID` variable and the store credentials — see the package README.

[TRAC-969]: https://bigcommercecloud.atlassian.net/browse/TRAC-969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/deployment-preview-action/examples/with-action.yml
```diff
@@ -0,0 +1,49 @@
...
+      # Do not check out first. The action checks out the pull request itself,
+      # after the permission and fork checks have run.
+      - uses: bigcommerce/catalyst/packages/deployment-preview-action@preview-action-v1
+        with:
+          project-uuid: ${{ vars.PREVIEW_DEPLOYMENT_PROJECT_UUID }}
+          store-hash: ${{ vars.BIGCOMMERCE_STORE_HASH }}
+          access-token: ${{ secrets.BIGCOMMERCE_PREVIEW_DEPLOYMENT_ACCESS_TOKEN }}
+          storefront-token: ${{ secrets.BIGCOMMERCE_STOREFRONT_TOKEN }}
+          channel-id: ${{ vars.BIGCOMMERCE_CHANNEL_ID }}
+          auth-secret: ${{ secrets.AUTH_SECRET }}
```
chanceaclark: Normally, Github actions will ship with arguments you can pass like this and reserved vars and secrets. We can do both, but I would normalize the vars/secrets to be agnostic of sorts

packages/deployment-preview-action/scripts/catalyst-preview.mjs
```diff
@@ -0,0 +1,332 @@
...
+  }
+
+  if (!response.ok) {
+    fail(`GET ${path} failed: ${response.status} ${await response.text()}`);
+  }
+
+  return response.json();
+}
+
+async function github(path, init = {}) {
```
chanceaclark, line 98: Maybe use Octokit?

packages/deployment-preview-action/examples/with-action.yml
jordanarldt: @chanceaclark Can you clarify what you mean here? Do you mean like prefixing them to be more clear (e.g. `PREVIEW_CHANNEL_ID` etc)?

packages/deployment-preview-action/scripts/catalyst-preview.mjs
jordanarldt: It's a very simple github wrapper and we only use it for 3 endpoints really, so I'd like to avoid pulling in any dependencies on the action since it has 0 deps right now aside from the js-yaml dev dependency

chanceaclark: Pretty sure github does something where it makes oktokit very light to add, could be wrong though 🤷‍♂️

packages/deployment-preview-action/examples/with-action.yml
chanceaclark: Like this actually will probably run in a separate context so you could have env vars be `BIGCOMMERCE_PROJECT_UUID`, `BIGCOMMERCE_ACCESS_TOKEN`, etc....

packages/deployment-preview-action/scripts/catalyst-preview.mjs
jordanarldt: @chanceaclark It looks like if we use octokit we'll have to start committing a `dist/` folder in order for the action to work which might add some more complexity with the release flow.. If we keep it simple with 0 dependencies, it's much more straightforward and simple - less cognitive overhead

chanceaclark: Up to you 🤷

jorgemoya (approved):

PR #3203, opened 2026-09-02
LTRAC-1732: build(cli) - Bump @opennextjs/cloudflare to 1.20.6

jorgemoya:
Linear: [LTRAC-1732](https://linear.app/commerce/issue/LTRAC-1732/bump-opennextjscloudflare-1173-1205)

Follows #3202 (Next.js 16.3.4), which this required — 1.20.6's peer range is `next >=15.5.24 <16 || >=16.3.3`.

## What/Why?

Moves the Cloudflare adapter 1.17.3 → 1.20.6 and the Wrangler version the build shells out to 4.90.0 → 4.128.0.

**Both leave this repo's dependency graph.** The adapter and Wrangler only belong to a project transformed for Commerce Hosting, so the root `package.json` should never have carried the adapter as a dependency. It moves to a devDependency of `packages/catalyst`, which is where it is actually needed: `cloudflare-context-symbol.spec.ts` imports it, and tsc typechecks `templates/open-next.config.ts` (the tsconfig sets no `include`). Wrangler is now declared nowhere — the build invokes a pinned version via `dlx`, and in the monorepo pnpm resolves it from the adapter's own `^4.125.0` peer.

That is also what fixed the stale Wrangler pin, rather than declaring it. pnpm had auto-installed 4.31.0 and recorded it in the lockfile — short of even the old `^4.65.0` peer — and it treats auto-installed peer resolutions as sticky, so `pnpm update`, `pnpm install --force`, and stripping the lock entries all failed to move it. Removing the root declaration and letting it resolve fresh lands 4.128.0 with **no OpenNext peer warnings left**.

**API compatibility was verified by diffing the published tarballs**, not by reading the changelog. Unchanged: `defineCloudflareConfig`, the `OpenNextConfig` type (`index.d.ts` and `config.d.ts` byte-identical), the four override subpaths the generated `open-next.config.ts` imports, `ShardedDOTagCacheOptions`, and the `Queue`/`QueueMessage` interfaces the custom `selfFetchQueue` implements. `@opennextjs/aws` goes 3.9.16 → 4.1.3, but v4.0.0 was a mislabeled major — upstream says it "should really have been 3.10.5" with no breaking changes.

**No Ignition change is required**, checked against the Go source rather than assumed. Ignition pins neither the adapter nor Wrangler (no version constant, no `exec.Command`, no Node toolchain — it consumes a pre-built `worker.js` + `/assets/`). What it does mirror is unchanged: the compat flags in `pkg/cloudflare/upload/metadata.go` (1.20.5 needs the same single `nodejs_compat` flag as 1.17.3) and the DO migration tag + `DOQueueHandler`/`DOShardedTagCache`/`BucketCachePurge` class names in `pkg/cloudflare/upload/migrations/migrations.go`.

**Node 20 dropped from `engines`.** `WRANGLER_VERSION` is what `catalyst build`/`deploy` invoke, and every Wrangler satisfying the adapter's `^4.125.0` peer requires Node ≥22 (the last supporting Node 20 is 4.86.0, below the floor). It was already inaccurate: the previous `wrangler@4.90.0` pin also declares `node >=22.0.0`, so build/deploy could not work on Node 20 before this either. What genuinely regresses is `catalyst start`, which ran against the locally resolved 4.31.0 (Node ≥18).

**`catalyst build` and `catalyst deploy` now offer to move the adapter pin, and reinstall when it moves.** The pin lives in the project's own `package.json` and `deploy` skips `setupCommerceHosting` once `isTransformed`, so bumping the constant alone never reaches an existing Commerce Hosting project. `catalyst upgrade` cannot move it either — the adapter is injected into the project and absent from the upstream tree, so no 3-way merge ever sees it (`upgrade.ts:49-51` records the same constraint for the CLI's own pin, which is why `findStaleCli` checks it out-of-band).

The check lives in `buildCatalystProject`, the path both commands share, so it runs exactly once immediately before the adapter is invoked — early enough for the install to land first. Putting it in a single command would have missed the `catalyst build` → `deploy --prebuilt` flow that `setupCoreProject` scaffolds scripts for.

`reconcileOpenNextVersion` acts only on a project genuinely behind (`validRange` → `satisfies` → `minVersion` → `lt`), so one pinned ahead, one on a range that already admits the target, and one on a git/file specifier are left alone. Four gates stand before it writes anything:

| Condition | Behaviour | Why |
|---|---|---|
| Project's Next too old for the adapter | send to `catalyst upgrade`, don't offer | would hand them an unsupported peer set |
| `deploy --prebuilt` | warn only | skips the build, so the bundle would never be compiled against the new adapter |
| No TTY (CI, Docker, scripted) | warn only | pnpm treats the lockfile as frozen; the required install would fail the deploy |
| User declines | keep the pin | their call |

Every path that declines to act reports the exact command to run. `OPENNEXT_REQUIRED_NEXT_RANGE` backs the compatibility gate and is hand-synced with the version constant — read it off the new adapter's `peerDependencies.next` when bumping.

**The peer stays an optional `^1.17.3` range**, deliberately. An exact peer would make the upgraded CLI unresolvable for merchants still on the old adapter, so they could never install the version that offers the upgrade. Verified rather than assumed: npm fails with `ERESOLVE — Conflicting peer dependency` on a present-but-mismatched peer **even when it is marked `optional`**; `optional` excuses only an *absent* peer, which is what keeps it quiet for projects hosted somewhere that never installs the adapter.

## Areas worth careful review

- `reconcileOpenNextVersion` and its call site in `deploy.ts` — the only behavioural change; everything else is dependency and version bookkeeping.
- The `engines` narrowing, released as `minor`. `engines` is advisory (pnpm warns, npm doesn't block without `engine-strict`) and no working Node 20 configuration is being removed, but flagging it as a judgement call.

## Rollout/Rollback

Rollout is **gradual, not fleet-wide**: since Ignition doesn't pin the adapter, deployed stores keep running whatever version they last built with. The bump only takes effect per-store on next deploy.

On that first deploy, `DOShardedTagCache` runs an in-place SQLite migration adding `stale`/`expire` columns to its `revalidations` table, backing the SWR `revalidateTag` support added in 1.19.0. It is guarded and automatic. Rollback to 1.17.3 is tolerable — the older code ignores the extra columns.

`withRegionalCache` defaults were re-derived for Next 16 (`bypassTagCacheOnCacheHit` now defaults `false`, `shouldLazilyUpdateOnCacheHit` to `!bypassTagCacheOnCacheHit`). The generated config declares neither (see LTRAC-1458), so effective behaviour is unchanged and no config edit is needed.

Rollback is reverting this PR; no data migration blocks rolling forward.

## Testing

- CLI `tsc --noEmit`, `eslint . --max-warnings 0`, `tsup` build — all clean
- CLI tests **683/683** passing (677 pre-upgrade baseline + 6 new)
- `cloudflare-context-symbol.spec.ts` passes against the real 1.20.6 — confirming both that the `__cloudflare-context__` key `core/lib/kv/adapters/cloudflare-kv.ts` reads directly is unchanged, and that the adapter still resolves after moving to a devDependency
- `pnpm install --frozen-lockfile` clean after rebasing onto merged canary
- No OpenNext peer warnings remain; `wrangler@4.31.0` is gone from the lockfile
- Coverage: 91.24% lines vs a 91.21% baseline on canary — the `test:coverage` threshold already fails on canary and CI runs `pnpm run test`, so this is a marginal improvement, not a regression

**Not covered: a real `catalyst build` / OpenNext bundle**, which needs store credentials. This is the main residual risk, because 1.20.3 rewrote Turbopack wasm-helper patching for Next 16.3's emit shape, and it would also confirm the output still fits Ignition's bundle limits (`worker.js` ≤ 40 MiB, each asset ≤ 25 MiB, ≤ 1000 files per `pkg/cloudflare/bundle/limits.go`). Worth running before merge.

## Migration

None for this repo. Merchants on an older adapter pin are now prompted on deploy to move it, with the safe-to-take gates above.

Refs LTRAC-1732

### DIFF
--- a/packages/catalyst/src/cli/lib/commerce-hosting.ts
+++ b/packages/catalyst/src/cli/lib/commerce-hosting.ts
@@ -2,8 +2,10 @@ import { confirm, input, select, Separator } from '@inquirer/prompts';
 import { colorize } from 'consola/utils';
 import { existsSync, mkdirSync, readFileSync, unlinkSync, writeFileSync } from 'fs';
 import { dirname, join } from 'path';
+import { minVersion, satisfies, lt as semverLt, validRange } from 'semver';
 import { z } from 'zod';
 
+import type { PackageManager } from './detect-package-manager';
 import { consola } from './logger';
 import {
   createProject,
@@ -13,10 +15,18 @@ import {
 } from './project';
 import { sortPackageJsonFields } from './sort-package-json';
 
-// Exported so the Cloudflare context contract test can assert which version
-// its symbol-key assumption was verified against. See
-// `cloudflare-context-symbol.spec.ts`.
-export const OPENNEXT_CLOUDFLARE_VERSION = '1.17.3';
+// Exported so `cloudflare-context-symbol.spec.ts` can assert the version its
+// symbol-key assumption was verified against. Keep it exact: a range would
+// admit versions that key was never checked against, and that failure is silent
+// (null context, so every native-hosted store degrades to an in-process cache).
+//
+// This is what new projects are pinned to, not a floor existing ones are held
+// at -- see `reconcileOpenNextVersion`.
+export const OPENNEXT_CLOUDFLARE_VERSION = '1.20.6';
+
+// The `next` peer range OPENNEXT_CLOUDFLARE_VERSION declares. Update both
+// together, from the new adapter's `peerDependencies.next`.
+export const OPENNEXT_REQUIRED_NEXT_RANGE = '>=15.5.24 <16 || >=16.3.3';
 
 const corePackageJsonSchema = z.looseObject({
   dependencies: z.record(z.string(), z.string()).optional(),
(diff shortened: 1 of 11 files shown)


packages/catalyst/src/cli/lib/commerce-hosting.ts
parthshahp, line 1: Could we maybe make the comments more succinct if we need them? At this length they don't seem that useful

jorgemoya: Good call, cleaning them up.

jorgemoya:
Local test:

```
❯ node ../packages/catalyst/dist/cli.js build
◢ @bigcommerce/catalyst v1.3.0

Loaded environment variables from .env.local

✔ This project pins @opennextjs/cloudflare 1.17.3, but this CLI targets 1.20.6. Adapter fixes released since then are not applied. Update it to 1.20.6 and reinstall now?
 Yes
ℹ Updated @opennextjs/cloudflare to 1.20.6.
✔ Dependencies installed successfully.
◐ Copying templates...
✔ Templates copied
```

parthshahp (approved):

PR #3204, opened 2026-09-03
TRAC-1262: fix(cli) - run GraphQL codegen as part of the build

mfaris9:
## What/Why?

Running `pnpm catalyst deploy` on a fresh Native Hosting project failed during
the build with `response.data is of type 'unknown'`, unless you had already run
`pnpm build` (or `pnpm dev`) by hand.

Root cause: the build needs gql.tada codegen output (`bigcommerce-graphql.d.ts`,
imported by `client/graphql.ts`), but that step (`generate`) was only wired into
the scaffold's npm wrapper scripts, not into the `catalyst` CLI. `catalyst deploy`
and `catalyst build` ran the OpenNext/Next build without it, so on a fresh project
(where the gitignored codegen file does not exist yet) the typecheck failed.

Fix: run `generate` at the start of `buildCatalystProject`, before the OpenNext
build. Codegen stops being a silent prerequisite.

## Testing

Verified against a real Native Hosting project, `deploy --dry-run`, schema file
removed to simulate a fresh scaffold:

- Before (published 1.3.0): dies at `Running TypeScript` -> exit 1.
- After (this change): prints `Generating GraphQL types...`, passes TypeScript,
  builds the bundle -> exit 0.

Unit test: `build.spec.ts` "runs GraphQL codegen before the OpenNext build"
(RED without the change, GREEN with it). Full suite: build 7/7, deploy 35/35.

## Migration

None.

### DIFF
--- a/.changeset/trac-1262-run-generate-in-build.md
+++ b/.changeset/trac-1262-run-generate-in-build.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst": patch
+---
+
+`catalyst build` and `catalyst deploy` now run GraphQL codegen (`generate`) automatically, so a fresh project no longer needs a manual `pnpm build` before its first deploy.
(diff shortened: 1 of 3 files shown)


jorgemoya (approved):
Looks good, only need to fix the lint issues.

mfaris9:
@jorgemoya all fixed. Thanks for the review

PR #3205, opened 2026-09-04
LTRAC-1746: feat(cli) - Add channel checkout URL support

jorgemoya:
Linear: [LTRAC-1746](https://linear.app/commerce/issue/LTRAC-1746) — under [LTRAC-447](https://linear.app/commerce/issue/LTRAC-447)

**1 of 4 in a stack.** Review this one first; the other three build on it.

| | PR | Adds |
| --- | --- | --- |
| 1 | **this one** | The API layer and the channel checkout URL commands |
| 2 | #3210 | The cross-domain checkout warning |
| 3 | #3211 | `catalyst deploy --update-checkout-url` |
| 4 | #3212 | The interactive offer to set a checkout URL |

## What/Why?

Checkout is hosted by BigCommerce and the redirect target is resolved server-side from channel config — `core/app/[locale]/(default)/checkout/route.ts` just follows whatever `redirectedCheckoutUrl` the API returns. So pointing a channel's site URL at a Native Hosting domain leaves checkout on whatever domain the channel had before, and until now the CLI had no way to inspect or change it. On the sandbox store this left one channel still redirecting to a checkout domain last touched in May 2024.

Rather than a new command group, this folds into the channel surface that already exists:

```shell
catalyst channels info                                        # or bare `catalyst channels`
catalyst channels update --hostname <h> --checkout-url <url>
catalyst channels update --remove-checkout-url
```

`info` prints the storefront, canonical, and checkout URLs — three things that are easy to conflate when debugging a redirect — and says so when the channel has no checkout URL of its own. It's registered `isDefault`, so bare `catalyst channels` reports state instead of printing help, following the `logs tail` precedent already in the repo.

### Notes for review

**Why `--checkout-url` on `update` rather than its own command.** Every channel URL now moves through one place. The two checkout flags are mutually exclusive, enforced by commander's `Option.conflicts` rather than a hand-rolled guard — so it surfaces as a usage error, and there's no runtime check to keep in sync.

**Passing a checkout flag alone skips the site-URL flow entirely.** Going through it would prompt for a hostname the caller never asked to change. Two tests pin that: neither the site `PUT` nor the hostname prompt happens. `runChannelSiteUrlFlow` now returns the channel it resolved so combining flags doesn't resolve it twice.

**Removal reports after the fact rather than preflighting.** It's the one operation that *creates* the inherited-checkout state, and the domain it falls back to belongs to the store's *default* channel — no API field identifies that channel, so the result can't be known beforehand. It re-reads the site afterwards and prints where checkout landed. Judging that domain is the diagnostic's job and arrives in #3210.

**The same-main-domain rule is enforced by BigCommerce, not pre-checked locally.** Determining the registrable domain correctly needs the public suffix list, and a wrong local check would reject a valid setup. `httpError` already prefers server prose, so BC's own explanation surfaces verbatim:

> `Your checkout and storefront must be within the same main domain like 'main.com' and 'subdomain.main.com'`

Only the unambiguous parts are validated here — that the value parses and uses https, with a bare hostname getting `https://` prefixed the way `runChannelSiteUrlFlow` already treats site URLs.

**Two schema details come from hitting the live API, not the spec.** `ssl_status` is documented `dedicated | shared` but returns `null` for sites on shared SSL, and `urls[].type` is kept a plain string so an unrecognised role can't crash what is mostly a diagnostic read.

**Refactors:** the 401/403 re-auth branch moves out of `updateChannelSiteUrl` into a shared helper now that four call sites need it, and `resolveChannel` is exported from `channel-site-flow` so the new commands reuse the Catalyst-platform picker rather than re-deriving its filter and error copy.

**No `core/` change is needed or included** — this is entirely CLI plus channel configuration.

## Testing

`pnpm --filter @bigcommerce/catalyst test` — 718 pass. Lint and typecheck clean. Based on canary at `bb35f22d1`.

New coverage includes BigCommerce's real 422 body (asserting its message survives to the user), `ssl_status: null`, the removal reporting path, the response shape that omits the optional fields, each checkout flag leaving the storefront URL untouched, and the two flags being rejected together.

Also exercised against store `vcijajnodd`:

| Case | Result |
| --- | --- |
| `channels info`, channel 1825156 (vanity domain) | Shows `checkout.catalyst-demo.site` |
| `channels info`, channel 1799395 (auto-generated hostname) | Shows the stale inherited domain and notes the fallback |
| Bare `catalyst channel` | Resolves to `info` |
| `update --checkout-url` with the value already set | Idempotent; confirmed only `updated_at` moved |
| `update --checkout-url checkout.catalyst-demo.site` (bare hostname) | Normalised to `https://…`, accepted |
| `update --checkout-url https://checkout.example.com` (cross-domain) | BC's 422 surfaces verbatim, no bug-report framing |
| `update --checkout-url` alone | Storefront URL untouched, no hostname prompt |
| Both checkout flags together | Rejected by commander before any API call |
| `update --remove-checkout-url` on 1799395 (already uncustomized, so a state no-op) | Reports the inherited target; confirmed byte-identical state before/after |

Removal was not run against 1825156, which has a working custom checkout domain that DELETE would drop to the shared URL.

## Migration

None. Additive — one new subcommand and two new flags, no changed defaults. Nothing is published yet, so the earlier shapes this went through never shipped.

Bare `catalyst channels` now runs `info` instead of printing help. `resolveChannel` is newly exported from `channel-site-flow`, and `runChannelSiteUrlFlow` returns `{ channelId }` instead of `void` — both internal to the CLI package with no external callers.

Refs LTRAC-1746

### DIFF
--- a/packages/catalyst/src/cli/commands/channels.ts
+++ b/packages/catalyst/src/cli/commands/channels.ts
@@ -3,13 +3,18 @@ import { Command, InvalidArgumentError, Option } from 'commander';
 import type Conf from 'conf';
 import { colorize } from 'consola/utils';
 
-import { runChannelSiteUrlFlow } from '../lib/channel-site-flow';
+import { resolveChannel, runChannelSiteUrlFlow } from '../lib/channel-site-flow';
 import {
   channelPlatformLabel,
+  type ChannelSiteDetails,
   checkChannelEligibility,
+  deleteChannelCheckoutUrl,
   fetchAvailableChannels,
+  findChannelSiteUrl,
   getChannelInit,
+  getChannelSite,
   sortChannelsByPlatform,
+  updateChannelCheckoutUrl,
 } from '../lib/channels';
 import { NoLinkedProjectError } from '../lib/commerce-hosting';
 import { runCreateChannelFlow } from '../lib/create-channel-flow';
@@ -38,6 +43,42 @@ const parseChannelId = (value: string): number => {
 
   return parsed;
 };
+
+// Validates only the unambiguous parts: parses as a URL, uses https. The
+// same-main-domain rule is BigCommerce's to enforce (see
+// `updateChannelCheckoutUrl`). A bare hostname gets `https://` prefixed, as
+// `runChannelSiteUrlFlow` does for site URLs.
+const parseCheckoutUrl = (value: string): string => {
+  const withScheme = /^[a-z][a-z\d+.-]*:\/\//i.test(value) ? value : `https://${value}`;
+  let parsed: URL;
+
+  try {
+    parsed = new URL(withScheme);
+  } catch {
+    throw new InvalidArgumentError(
+      `"${value}" is not a valid URL. Pass a hostname or an https URL, e.g. https://checkout.example.com.`,
+    );
+  }
+
+  if (parsed.protocol !== 'https:') {
+    throw new InvalidArgumentError(
+      `The checkout URL must use https, but "${value}" uses ${parsed.protocol.replace(':', '')}.`,
+    );
+  }
+
+  // BigCommerce wants the origin; a pasted URL often carries more.
+  return parsed.origin;
+};
+
+const CHECKOUT_URL_NOTES = `
+Checkout is hosted by BigCommerce, so a checkout URL must be a domain you have
+pointed at BigCommerce with a certificate provisioned there — not one added with
+\`catalyst domains add\`, which only routes traffic to this project.
+
+BigCommerce also requires the checkout URL to share a main domain with the
+channel's storefront URL, so that sessions carry between the two. This means a
+custom checkout URL is only possible on a custom storefront domain; channels on
+an auto-generated deployment hostname use the shared checkout domain.`;
 
 // Resolve credentials from flags/env → persisted project config → interactive
 // login (persisting on success). Returns null when the user aborts login.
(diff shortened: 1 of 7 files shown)


jorgemoya:
Testing
```
◢ @bigcommerce/catalyst v1.3.0

◐ Fetching channels...
✔ Channels fetched.
✔ Select the channel to update. Canary Catalyst Demo Site
◐ Fetching channel site...
✔ Channel "Canary Catalyst Demo Site" (1543376):
  Storefront  https://canary.catalyst-demo.site
  Canonical   https://store-vcijajnodd-1543376.mybigcommerce.com
  Checkout    https://catalyst-demo-site.mybigcommerce.com
ℹ This channel has no checkout URL of its own, so BigCommerce falls back to the default channel's primary URL. That may be a different domain than the storefront above.

 WARN  Checkout for this channel is on catalyst-demo-site.mybigcommerce.com, a different domain than the storefront (canary.catalyst-demo.site). Shopper sessions and carts may not carry into checkout in browsers that restrict cross-domain cookies.

ℹ To fix it, point checkout.canary.catalyst-demo.site at BigCommerce and provision a certificate for it there.
✔ Set a checkout URL for this channel now? No
ℹ When you're ready, run catalyst channels checkout-url --channel-id 1543376 --url <url>.
```

packages/catalyst/src/cli/commands/channels.ts
```diff
@@ -361,12 +468,111 @@ Examples:
...
+    'after',
+    `${CHECKOUT_URL_NOTES}
+
+Examples:
+  # Pick a channel interactively; the same as bare \`catalyst channels\`
+  $ catalyst channels info
+
+  # Target a channel directly
+  $ catalyst channels info --channel-id 123`,
+  )
```
jordanarldt, line 544: Do we need the `--channel-id` flag? The linked project should be associated with a channel ID already, so I'm wondering if we should lean on that. Although maybe it is good to at least provide the option to get the data for other channels?

jordanarldt (approved):

packages/catalyst/src/cli/commands/channels.ts
jorgemoya: I think we can leave as is for now and remove if we don't find value for it. It may be important for some people to check channel info back and forth without having to swap channels.

PR #3208, opened 2026-09-08
TRAC-1266: fix(cli) - reconcile native hosting files on upgrade

mfaris9:
## What/Why?
catalyst upgrade could pull instrumentation.ts back into a native-hosted
project. It's incompatible with the Cloudflare Workers bundle (the deploy
fails at cold start), and nothing on the upgrade path removed it, so merchants
had to delete it by hand after every upgrade.

Run the same check catalyst deploy already uses, right after the merge, and
scrub the file from the merge result so it isn't reported as a leftover
conflict.

Gate on isLinked, not just isTransformed: an upgrade can leave package.json
with conflict markers, which makes isTransformed read false (it can't find the
opennext dep).

## Testing
```bash
# unit + static checks (CLI package)
npm test            # full CLI suite green
npm run lint && npm run typecheck

# end to end on a real store
node dist/cli.js create --hosting commerce --gh-ref @bigcommerce/catalyst-core@1.3.7
cd <project> && node ../dist/cli.js upgrade 1.4.0
#   without the change: instrumentation.ts persists, no prompt
#   with the change:    prompt fires, instrumentation.ts removed
```

## Migration
None.

### DIFF
--- a/.changeset/trac-1266-reconcile-instrumentation-on-upgrade.md
+++ b/.changeset/trac-1266-reconcile-instrumentation-on-upgrade.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst": patch
+---
+
+`catalyst upgrade` now removes native-hosting-incompatible `instrumentation.ts` that a merge reintroduces, instead of leaving it for the merchant to delete by hand.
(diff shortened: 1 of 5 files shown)


parthshahp:
Wasn't there an opennext change that didn't require this anymore or something? Or maybe it was related to `proxy`? @jorgemoya probably has more context.

jorgemoya (approved):

PR #3209, opened 2026-09-08
TRAC-295: fix(core) - respect date modifier range options in PDP date picker

mfaris9:
## What/Why?

On the product detail page, a Date modifier's range limits were ignored: the
calendar let shoppers pick any date, regardless of the control panel config.

Root cause: the API fields (`earliest`, `latest`, `limitDateBy`) are fetched,
but `productOptionsTransformer` dropped them, the `vibes` `DateField` type had
no min/max to hold them, and the PDP renderer passed no `disabledDays` to
`<DatePicker>`.

Fix: a small `date-field-limits` helper carries the bounds through.

## Testing

Live PDP, CP modifier set to RANGE, Sep 15-19 2026:
- Before: all 30 September days selectable (0 disabled).
- After: only Sep 15-19 selectable; Aug/Oct/Nov fully disabled.

Unit: `core/lib/date-field-limits.spec.ts`, 9 cases (all four modes + matcher
shapes), red against a no-bounds stub, green with the fix. Full vitest 74/74,
typecheck and lint clean.

## Migration

None.

## Screenshots

Before: 
image

After: 
image

### DIFF
--- a/.changeset/trac-295-pdp-date-range.md
+++ b/.changeset/trac-295-pdp-date-range.md
@@ -1,0 +1,5 @@
+---
+"@bigcommerce/catalyst-core": patch
+---
+
+Product detail page date modifiers now respect the date limits set in the control panel. A Date option's earliest date, latest date, and limit mode (earliest only, latest only, or a range) now disable out-of-range days in the PDP date picker, instead of letting shoppers pick any date.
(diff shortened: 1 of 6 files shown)


jorgemoya (approved):
Awesome, thanks for fixing this!

mfaris9:
> Awesome, thanks for fixing this!

@jorgemoya no problem! Thank you for reviewing

PR #3224, opened 2026-09-16
AI-309: Remove trailing slash before proxying UCP endpoints

davidchin:
## What/Why?
Requests to `/api/ucp/*` are proxied to the BigCommerce storefront. However, Catalyst normalizes URLs with a trailing slash before the proxy runs, changing `/api/ucp/checkout-sessions` to `/api/ucp/checkout-sessions/`. Because the platform serves UCP endpoints without trailing slashes, it responds with a `307` redirect to the bare path. The proxy now strips the trailing slash before forwarding the request, preventing the upstream redirect.

## Testing
Screenshot 2026-09-16 at 3 23 15 pm


parthshahp (approved):
😅

jorgemoya (approved):

PR #3231, opened 2026-09-18
TRAC-1555: fix - preserve Auth.js session cookie attributes

parthshahp:
## What/Why?

Unpartitioned duplicate cookeis would survive users attempting to lougout and cause redirect loops (see example below). This happened because the cookie rewrite on logout didn't use the `partitioned` property, and both cookies were retained by the browser.

https://github.com/user-attachments/assets/dd3c7e6c-013d-4475-a941-9d8f43126334

This PR instead preserves the cookie properties such as `partitioned` so that we can utilize the same cookie.

## Testing

Tested manually on a deployed native hosting site.

https://github.com/user-attachments/assets/3ac4ec65-32af-4095-af82-62ca473f6c9b

## Migration

No breaking changes.


mfaris9 (approved):
lgtm

jorgemoya (approved):

PR #3233, opened 2026-09-18
LTRAC-1328: ci - Automate Wrangler version bumps

jorgemoya:
Linear: [LTRAC-1328](https://linear.app/commerce/issue/LTRAC-1328/automate-wrangler-version-bumps)

## What/Why?

`catalyst build` interpolates its Wrangler version into `pnpm dlx wrangler@<version>`, so the pin is a plain string literal in the CLI source rather than a manifest entry. Dependabot can't see it and nothing else watched it, so it only moved when someone remembered to edit it by hand — it's currently six releases behind (`4.128.0` vs `4.134.0`).

A weekly job now reads the published versions from the npm registry, rewrites the pin, writes a changeset, and opens or force-pushes a single rolling PR.

Decisions worth flagging:

- **Tracks the pinned major only.** A Wrangler major can change bundling and compatibility behaviour for every native-hosted store, so a newer major is reported as a `> [!NOTE]` in the job summary and adopted deliberately rather than taken automatically. Prereleases are skipped, and a pin already ahead of the registry is left alone rather than downgraded.
- **One rolling branch, rebuilt from the canary tip each run,** so a superseded bump's changeset never lingers on it and the diff stays a single commit. If a PR is already open the job edits its title and body instead of opening a duplicate.
- **This PR deliberately does not bump the pin,** so the diff stays CI-only and reviewable. The actual `4.128.0 → 4.134.0` bump should arrive as the workflow's own PR — `workflow_dispatch` needs the workflow on the default branch, so dispatch it once after merge.
- **`test:scripts` now runs in CI.** It was defined in the root `package.json` but invoked by no workflow, so the 139 existing script tests gated nothing. Added as a step to `basic.yml`'s existing `lint-typecheck` job; the job ID and its display name are unchanged, since renaming a required check would break branch protection.
- **The generated PRs open without checks.** GitHub doesn't trigger workflows on pushes made with `GITHUB_TOKEN`, and this repo has no bot PAT. The generated body says to push an empty commit, which fires `synchronize` — a trigger all three PR workflows already subscribe to. Closing and reopening does *not* work here, because they subscribe to `opened`/`synchronize` and reopening only emits `reopened`. Widening the shared triggers for one automated PR didn't seem worth it.

`OPENNEXT_CLOUDFLARE_VERSION` in `commerce-hosting.ts` is the same shape of problem, and gets the same treatment in a follow-up PR stacked on this one. It needs more machinery than Wrangler, though: `OPENNEXT_REQUIRED_NEXT_RANGE` has to be re-derived from the new adapter's `peerDependencies.next`, the tripwire literal in `cloudflare-context-symbol.spec.ts` has to move with it, and the job has to commit a `pnpm-lock.yaml` update — otherwise the contract test in that spec keeps exercising the old *installed* version and passes for the wrong reason. The constant is currently up to date (`1.20.6` is latest), so there's no urgency.

Worth a conscious decision while reviewing: the lockfile already carries `wrangler@4.128.0`, installed as the adapter's peer (`wrangler: ^4.125.0` + `autoInstallPeers`) and coincidentally identical to `WRANGLER_VERSION`. This job moves only the literal, so the two will drift. They serve different purposes — `pnpm dlx` in the merchant's project vs. this monorepo's own dev copy — so I left them independent rather than coupling them.

## Testing

`pnpm test:scripts` — 157 pass (139 existing + 18 new).

Ran the script against the live npm registry:

- Bumped `4.128.0 → 4.134.0`, wrote `.changeset/bump-wrangler-4-134-0.md`, and emitted the expected step outputs. Reverted before committing.
- A second run correctly reported "already the newest 4.x release" and set `updated=false`.
- Held-back-major path, with the pin temporarily set to `3.114.0`: bumped within 3.x to `3.114.17` and flagged `4.134.0` in the job summary without taking it.

The workflow YAML parses and its embedded shell passes `bash -n`. The generated PR body was rendered with sample values to check the markdown and the release link.

The git/`gh` plumbing in the job itself can't run until the workflow is on `canary`, so the first `workflow_dispatch` after merge is what exercises it.

## Migration

None. CI only — no changeset, and no published package is affected.


jorgemoya:
Superseded by the unified job on `jorgemoya/ltrac-1328-automate-hosting-pin-bumps`.

This PR automated `WRANGLER_VERSION` on its own. A Codex review on the follow-up caught the gap that makes that insufficient: `@opennextjs/cloudflare` declares which Wrangler it supports (`wrangler: ^4.125.0`), so an adapter release that raises that floor leaves `catalyst build` running a Wrangler the adapter never supported — and two jobs on two rolling branches can't see each other's pending bump.

The replacement maintains both pins in one job and one PR, with Wrangler pinned to the newest release the pinned adapter allows rather than the newest that exists. That removes the drift by construction instead of gating against it, and reconciles the pins if they are already mismatched.

PR #3234, opened 2026-09-18
LTRAC-1328: ci - Automate the Commerce Hosting version pins

jorgemoya:
Linear: [LTRAC-1328](https://linear.app/commerce/issue/LTRAC-1328/automate-wrangler-version-bumps)

Supersedes #3233, which automated the Wrangler pin alone.

## What/Why?

Commerce Hosting builds rest on two versions that live as string literals in the CLI source rather than as manifest entries, so Dependabot can't see either and neither moved unless someone remembered:

- `WRANGLER_VERSION` in `commands/build.ts`, interpolated into `pnpm dlx wrangler@<version>`
- `OPENNEXT_CLOUDFLARE_VERSION` in `lib/commerce-hosting.ts`, what new projects get pinned to

One weekly job now maintains both, in one PR. They're deliberately not two jobs: `@opennextjs/cloudflare` declares which Wrangler it supports (`wrangler: ^4.125.0` today), so **Wrangler is pinned to the newest release the pinned adapter allows, not the newest that exists.** Bumping either in isolation is how they drift out of a supported pair — with separate jobs on separate branches, neither can see the other's pending bump, and an adapter release that raised its Wrangler floor would leave `catalyst build` running a Wrangler the adapter doesn't support.

### What holds a bump

The job is deliberately conservative — it holds *both* pins and reports why, rather than guessing:

| Condition | Why it can't be automated |
| --- | --- |
| Adapter moved its `peerDependencies.next` | Judging whether core's `next` satisfies a new range needs judgment. Shipping a pin core can't meet is worse than not bumping: `reconcileOpenNextVersion` refuses to apply it and tells the merchant to upgrade core first. |
| Adapter needs a Wrangler major this job doesn't track | A Wrangler major can change bundling for every native-hosted store. |
| Pin sits above a new adapter ceiling | Walking Wrangler backwards is a downgrade nobody asked for. |
| A newer adapter major exists | Same reasoning; reported in the job summary, never taken. |

Even a cosmetically reworded `next` range holds the bump, because deciding two ranges are equivalent needs range semantics the job doesn't apply to that question. There's a test asserting exactly that.

### The contract test, and the peer range

`cloudflare-context-symbol.spec.ts` is the reason this needs more care than a version string. `core` reads the Cloudflare context off `Symbol.for('__cloudflare-context__')` — an internal detail of the adapter — and the spec calls the real `getCloudflareContext()` to prove that key is still the one the adapter writes. If a release changed it, every native-hosted store would fall back to an in-process cache with no error and no signal.

That test exercises whichever adapter is **installed**, so it only says something about the pin while the two agree — and nothing made them agree. It now asserts that they do, replacing a hand-maintained `expect(OPENNEXT_CLOUDFLARE_VERSION).toBe('1.20.6')` literal that a lockfile bump could silently leave behind. Verified it bites: with the constant at `1.21.0` it fails with `expected '1.20.6' to be '1.21.0'`.

**Making that invariant hold means the peer range has to track the pin, so this raises it from `^1.17.3` to `^1.20.6`** — the one consumer-visible change here, and the piece most worth a second opinion. The old floor was fiction: nothing has ever targeted or verified `1.17.3`. A project deliberately held on an older adapter will now see an unmet-peer warning where it saw none. Nothing breaks — the peer is optional and `catalyst deploy` still offers the upgrade rather than requiring it — but it does cut against that tolerance, so it's called out in the changeset.

The job runs that spec against the new adapter *before* opening the PR. Unlike a pure version-string bump, running tests in-job earns its keep here: it's the only place the bump gets verified, since GitHub won't run checks on a PR pushed with `GITHUB_TOKEN`.

### Also

- `semver` is added at the root, used only to read the adapter's Wrangler peer range. Two earlier attempts at this were measured and rejected: `pnpm update` re-resolved 315 lines of unrelated lockfile tree, and a plain `pnpm install` after adding `semver` dragged in `eslint-plugin-import` and friends. `--lockfile-only` keeps it to the 24 lines here, all `semver` (7.7.4 → 7.8.5 across the tree).
- `test:scripts` now runs in CI. It was defined in the root `package.json` but invoked by no workflow, so 139 existing script tests gated nothing. Added as a step to `basic.yml`'s existing `lint-typecheck` job — the job ID and display name are untouched, since renaming a required check would break branch protection.
- This PR deliberately does **not** move either pin, so the diff stays reviewable. `workflow_dispatch` needs the workflow on the default branch, so dispatch it once after merge; as of writing that produces a Wrangler-only bump to `4.135.0` (the adapter is already current at `1.20.6`).

## Testing

`pnpm test:scripts` — 177 pass (139 existing + 38 new). Full CLI suite green. Lint and typecheck clean.

Ran the script against the live npm registry:

- Bumped Wrangler `4.128.0 → 4.135.0`, correctly gated on the adapter's `^4.125.0`, left the adapter at `1.20.6` (already latest), and wrote a changeset naming only the pin that moved. Reverted before committing.
- Adapter bump path, with the pin lowered to `1.19.0`: bumped to `1.20.6`, rewrote the constant and the peer range together, Next requirement recognised as unchanged.
- Hold path, with the pinned `next` range altered: refused the bump, wrote a `> [!WARNING]` job summary, touched no files.

Contract spec verified in both directions — passes when the installed adapter matches the pin, fails when it doesn't.

Workflow YAML parses and its embedded shell passes `bash -n`. The generated PR body was rendered with sample values.

The git/`gh` plumbing can't run until the workflow is on `canary`, so the first `workflow_dispatch` after merge is what exercises it.

## Migration

None for storefronts. Projects pinned below `@opennextjs/cloudflare@1.20.6` will see a new unmet-peer warning; `catalyst deploy` already prompts to move them to the pinned version.


jordanarldt:
@jorgemoya How can we verify that a wrangler version bump won't break anything?

jorgemoya:
> @jorgemoya How can we verify that a wrangler version bump won't break anything?

@jordanarldt Have included a build step to validate the build. Where you thinking of something else we need to validate?

jordanarldt:
@jorgemoya I'm mainly thinking of the deploy pipeline. Like if we update the wrangler version and that somehow breaks the deployment process on native hosting, or makes the storefront stop working. Is that possible to happen?

jorgemoya:
> @jorgemoya I'm mainly thinking of the deploy pipeline. Like if we update the wrangler version and that somehow breaks the deployment process on native hosting, or makes the storefront stop working. Is that possible to happen?

Anything is possible 😅 . I consider this rare/low risk, but yes it could happen. Maybe we manually verify the preview deployment you added succeeds before we merge it in?

jorgemoya:
@jordanarldt Added a step to preview a deployment so we can manually test as well.

